PR #16, opened 2020-11-30
Add tests for GuAutoScalingGroup

jamie-lynch:
## What does this change?

This PR adds unit tests for the `GuAutoScalingGroup` component. These tests each cover a specific piece of functionality and assert it directly, rather than making use of snapshots. 

This is different to how many of our tests work so far so I'm interested to hear thoughts on whether it's an approach we should or shouldn't take going forward (either instead or as well as snapshots of our constructs). Below are some thoughts on the pros and cons of each method. 

**Snapshots**

[+] Quick and easy to write
[+] Pick up any changes to the output cloudformation (particularly useful for unintended side effects)
[-] A change to the component may cause many unrelated tests to also fail
[-] When testing different permutations of a component, a number of snapshots will be created, each of which contains a 
      certain amount of information which is irrelevant to the particular test. This adds some extra effort to understand what 
      is relevant to the test which you're looking at
[-] Snapshots are easy to update which, especially when multiple are affected by a change, makes it easy to accidentally  
      update them incorrectly. Further to this, as the snapshot (which is essentially the assertion) is stored in a different file, 
      it's not immediately obvious if the assertions are valid

**Direct Assertions**

[+] Each test only contains the relevant assertions for the test, making it easier to understand the consequence of settings  
      certain props
[+] Changes will only fail tests that cover that particular area
[-]  More complex and time consuming to write

## Other

While writing the tests, I discovered that calling `this.attachToApplicationTargetGroup` more that once for the `AutoScalingGroup` causes an error. It appears that it is only possible to attach to one target group. Consequently, we should probably update the props and constructor to reflect that. I'll open a follow up PR to implement that.

No

## How to test

Run `npm run test` to see the result of the test run

All of the custom functionality implemented in the `GuAutoScalingGroup` is suitably tested.

n/a

## Images

n/a

### DIFF
--- a/src/constructs/autoscaling/asg.test.ts
+++ b/src/constructs/autoscaling/asg.test.ts
@@ -2,7 +2,7 @@ import "@aws-cdk/assert/jest";
 import { SynthUtils } from "@aws-cdk/assert/lib/synth-utils";
 import { Vpc } from "@aws-cdk/aws-ec2";
 import { ApplicationProtocol } from "@aws-cdk/aws-elasticloadbalancingv2";
-import { Stack } from "@aws-cdk/core";
+import { App, Stack } from "@aws-cdk/core";
 import { GuStack } from "../core/stack";
 import { GuSecurityGroup } from "../ec2";
 import { GuApplicationTargetGroup } from "../loadbalancing";
@@ -27,7 +27,8 @@ describe("The GuAutoScalingGroup", () => {
   };
 
   test("correctly sets the machine image using props and calling getImage", () => {
-    const stack = new GuStack();
+    const app = new App();
+    const stack = new GuStack(app);
 
     new GuAutoScalingGroup(stack, "AutoscalingGroup", { ...defaultProps, osType: 1 });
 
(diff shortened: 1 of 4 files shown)


stephengeller (approved):
Looks great! Tests are clear and legible.

src/constructs/autoscaling/asg.test.ts
```diff
@@ -0,0 +1,135 @@
...
+describe("The GuAutoScalingGroup", () => {
+  const vpc = Vpc.fromVpcAttributes(new Stack(), "VPC", {
+    vpcId: "test",
+    availabilityZones: [""],
+    publicSubnetIds: [""],
+  });
+  const defaultProps: GuAutoScalingGroupProps = {
+    vpc,
+    imageId: "123",
+    instanceType: "t3.macro",
```
akash1810, line 25: `macro`? Is this intentional or a typo of `micro`?

akash1810 (approved):
Looks great.

I like this approach of assertion testing for constructs as it seems more readable. Maybe its worth having a note in the readme along the lines of: for constructs, prefer assertion tests, for patterns prefer snapshot tests?

src/constructs/autoscaling/asg.test.ts
jamie-lynch: It was intentional just to guarantee that it had used the value from props rather than a default. Happy to change as it looks a bit like a typo though!

akash1810: Fair enough!

PR #19, opened 2020-12-01
Bump npm version to 0.1.1

stephengeller:
## What does this change?

This bumps the package version to 0.1.1, taking in changes made to the repository. cc @jorgeazevedo 

In the future, we should decide on how we would like to go about safely bumping the package.

No.


jamie-lynch (approved):
👍

PR #20, opened 2020-12-01
move aws-cdk/* libs to peerDependencies

akash1810:
## What does this change?

As I understand it, `peerDependencies` are a way of saying "@guardian/cdk is compatible with this version of @aws-cdk/*". This is important as the AWS CDK library has quite a [frequent release cycle](https://github.com/aws/aws-cdk/tags) with [breaking changes introduced in minor bumps](https://github.com/aws/aws-cdk/releases/tag/v1.75.0).

Another benefit of `peerDependencies` is it helps keep `node_modules` flat. [This](https://medium.com/angular-in-depth/npm-peer-dependencies-f843f3ac4e7f) is a good read.

No.

## How to test

n/a

More descriptive `package.json`?

n/a

## Images

n/a

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,9 +32,6 @@
     "typescript": "~4.1.2"
   },
   "peerDependencies": {
-    "@aws-cdk/core": "~1.74.0"
-  },
-  "dependencies": {
     "@aws-cdk/assert": "~1.74.0",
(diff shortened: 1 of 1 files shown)


akash1810:
Closing as peerDependencies don't get installed (?!) and so CI breaks.

PR #23, opened 2020-12-01
Use np to orchestrate release steps

akash1810:
## What does this change?

Uses [`np`](https://www.npmjs.com/package/np) to orchestrate the release steps, reducing the amount of manual work needed.

The biggest wins are that `np` will:
- ensure we're on the release branch (`main`)
- ensure we're up to date with release branch
- run tests
- push tags

It also:

> Warns about the possibility of extraneous files being published

This was yielding output similar to that below. Although this is related to [this issue](https://github.com/sindresorhus/np/issues/577), we can take this as an opportunity to simplify our `package.json` as the `files` setting was only ever temporary IIUC.

In the ideal world, we'd use something like [semantic-release](https://github.com/semantic-release/semantic-release) for a fully automated publishing process, however this requires an npm token to be added to GHA, which we've not fully solved. @guardian/client-side-infra have discussed this previously.

n/a

Fewer manual steps required to publish, thus a simpler process.

n/a

## Images

**"extraneous files being published"**
image


jamie-lynch (approved):
Looks good

PR #25, opened 2020-12-02
0.1.2

akash1810:
## What does this change?

Bumping version to 0.1.2.

After #23 went in, I've performed a release with `npm run release`. It has successfully published to npm and created a GitHub tag, however it failed when attempting to push this commit directly onto `main` due to our branch protection rules.

PR the commit to have GitHub reflect reality.

I don't think PRing version changes like this is the best solution as there's no guarantee it'll be merged at the correct point in the timeline.

The [`commercial-core` repo](https://github.com/guardian/commercial-core/blob/2d65abd8269ad4ddc785de686bb2e0281782115d/package.json#L3) solves this by not updating the version in `package.json`, instead relying on tags and release artefacts. I don't know if I like this approach or not...

Creating a seamless publishing process is probably something we want to think about as we continue to iterate on this library.

No.

## How to test

n/a

The head of main reflects reality.

Git history being out of sync.

## Images

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@guardian/cdk",
-  "version": "0.1.1",
+  "version": "0.1.2",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


jamie-lynch (approved):
Agreed that we want to look at a better release process. Guess that ties in to the wider discussions around how we manage projects which are published to npm.

PR #27, opened 2020-12-03
Move wazuh access to ec2/security-groups.ts

jamie-lynch:
## What does this change?

At the moment we have both `constructs/securitygroup` and `constructs/ec2/security-groups.ts` which both have classes related to security groups. This PR moves the `GuWazuhAccess` component into `constructs/ec2/security-groups.ts` file for consistency. 

Yes, existing stacks using the `GuWazuhAccess` component will have to be updated.

## How to test

n/a

`@guardian/cdk` only has one place which contains components related to security groups.

n/a

## Images

n/a


jorgeazevedo (approved):
Seems sensible to me!

akash1810 (approved):

PR #28, opened 2020-12-03
Add APIs prop to GuLambdaFunction

jamie-lynch:
## What does this change?

This PR adds a new `apis` prop to the `GuLambdaFunction` to which an array of config for `LambdaRestApis` can be passed. This means that lambdas with an API gateway can be instantiated in one go.

This is a new, non-breaking feature so existing stacks won't require changes. Currently no migrated stacks make use of this feature but soon to be migrated stacks will.

## How to test

A new snapshot has been added which includes the extra resources created when the `apis` prop is used. Further testing can be done by migrating a stack that already uses lambdas and API gateway and verifying that the output is the same.

Stacks which have api gateways along with lambda functions can specify them via the `apis` prop for simpler configuration.

n/a

## Images

n/a


stephengeller (commented):
Looks good! Won't approve but nice work getting tests on it

src/constructs/lambda/lambda.test.ts
```diff
@@ -27,12 +28,30 @@ describe("GuLambdaFunction", () => {
...
       ],
     });
 
     expect(SynthUtils.toCloudFormation(stack)).toMatchSnapshot();
   });
+
+  it("should create a lambda function with an api gateway", () => {
+    const stack = new Stack();
+
+    new GuLambdaFunction(stack, "lambda", {
```
akash1810, line 43: We probably want `GuLambdaFunction` to take in a `GuStack` as it's first argument, rather than a `Construct` as this will provide the stack and stage params and tags for free.

akash1810 (approved):

PR #29, opened 2020-12-04
Grant lambda read access to bucket containing code artifact

stephengeller:
## What does this change?

This PR ensures that the bucket that contains the code artefact can be read by the lambda.

Nope. This should simply add an extra permission to the stacks if not already there.

## How to test

`yarn test` from within the root of the repo.

No lambda stacks encounter issues with lambdas not being allowed to read from the code bucket.


akash1810 (commented):
This repo currently uses npm. Can we either replace the addition of yarn.lock with an update of package-lock.json or move the entire project to yarn?

On a related note, we should consider following [this pattern](https://github.com/github/scripts-to-rule-them-all) to abstract the implementation detail of which package manager we're using away.

src/constructs/lambda/lambda.ts
```diff
@@ -33,5 +33,7 @@ export class GuLambdaFunction extends Function {
         enabled: true,
       });
     });
+
+    bucket.grantRead(this);
```
akash1810, line 37: Is this bucket the same one accounts use for their EC2 based artefacts? If so, we should be able to re-use [this construct](https://github.com/guardian/cdk/blob/main/src/constructs/iam/policies.ts#L96) or even create a `GuGetDistributable` construct.

PR #30, opened 2020-12-04
Add an optional app prop to GuStack

jamie-lynch:
## What does this change?

This PR updates the GuStack to take a new `GuStackProps` interface which allows for an app value to be set. This app value is added as a tag to all resources on the stack and a getter is added to return either the string or throw an error if the app is not set.

This change means existing stacks can pass the app prop through instead of setting the tags themselves. It is a non-breaking change. 

## How to test

Create a new stack with the `app` prop set and check that the tag is correctly applied to all resources. 

Resources are all tagged with the relevant app with less effort for users. 

n/a

## Images

n/a

### DIFF
--- a/src/constructs/core/stack.test.ts
+++ b/src/constructs/core/stack.test.ts
@@ -53,4 +53,33 @@ describe("The GuStack construct", () => {
       ],
     });
   });
+
+  it("should apply the app tag to resources added to it if provided", () => {
+    const stack = new GuStack(new App(), "Test", { app: "MyApp" });
+
+    new Role(stack, "MyRole", {
+      assumedBy: new ServicePrincipal("ec2.amazonaws.com"),
+    });
+
+    expect(stack).toHaveResource("AWS::IAM::Role", {
+      Tags: [
+        {
+          Key: "App",
+          Value: "MyApp",
+        },
+        {
+          Key: "Stack",
+          Value: {
+            Ref: "Stack",
+          },
+        },
+        {
+          Key: "Stage",
+          Value: {
+            Ref: "Stage",
+          },
+        },
+      ],
+    });
+  });
 });
--- a/src/constructs/core/stack.test.ts
+++ b/src/constructs/core/stack.test.ts
@@ -88,59 +88,9 @@ describe("The GuStack construct", () => {
 
     expect(stack.app).toBe("MyApp");
   });
-
-  it("should create an app parameter and use that to tag the stack if the appParameter props is true", () => {
-    const stack = new GuStack(new App(), "Test", { appParameter: true });
-
-    new Role(stack, "MyRole", {
-      assumedBy: new ServicePrincipal("ec2.amazonaws.com"),
-    });
-
-    const json = SynthUtils.toCloudFormation(stack) as SynthedStack;
-    expect(json.Parameters.App).toEqual({
-      Description: "The name of the app",
-      Type: "String",
-    });
-
-    expect(stack).toHaveResource("AWS::IAM::Role", {
-      Tags: [
-        {
-          Key: "App",
-          Value: {
-            Ref: "App",
-          },
-        },
-        {
-          Key: "Stack",
-          Value: {
-            Ref: "Stack",
-          },
-        },
-        {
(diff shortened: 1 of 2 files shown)


src/constructs/core/stack.test.ts
```diff
@@ -15,4 +15,33 @@ describe("GuStack", () => {
...
   });
+
+  it("should apply the app tag to resources added to it if provided", () => {
+    const stack = new GuStack(new App(), "Test", { app: "MyApp" });
+
+    new Role(stack, "MyRole", {
+      assumedBy: new ServicePrincipal("ec2.amazonaws.com"),
+    });
+
+    expect(stack).toHaveResource("AWS::IAM::Role", {
```
akash1810, line 26: `toHaveResource` nice!

src/constructs/core/stack.ts
```diff
@@ -2,17 +2,23 @@ import type { App, StackProps } from "@aws-cdk/core";
...
+  constructor(app: App, id?: string, props?: GuStackProps) {
     super(app, id, props);
 
     this.stage = new GuStageParameter(this);
     this.stack = new GuStackParameter(this);
 
     Tags.of(this).add("Stack", this.stack.valueAsString);
     Tags.of(this).add("Stage", this.stage.valueAsString);
+
+    props?.app && Tags.of(this).add("App", props.app);
```
akash1810, line 22: I think we'd also want to create an `app` parameter too?

jamie-lynch: I wondered about that. Does app need to be a parameter or would you just hard code it in the app?

e.g. in `bin/cdk.ts`

```
const app = new App();
new CdkStack(app, "CdkStack", { app: "MyApp"} );
```

jamie-lynch: Or we can add another `appParameter?: boolean` prop which, if true, would create an app parameter and use that for the tag?

akash1810: Oh that's pretty neat!

akash1810: Ah, but we should attach the value to the class, maybe with a private field and a getter? That is, we'd want a way for other constructs to refer to it, for example in a `GuLaunchConfig` I think we'd want to do something like:

```ts
class GuLaunchConfig() {
  constructor() {
    ...
    this.userData = `
     #! /usr/bin/bash
     aws s3 cp ${distBucket}/${this.stage}/${this.stack}/${this.app}.deb
    `;
  }
}
```

Maybe one for later as we understand the use case a bit more.

jamie-lynch: Yep, I was just running into that issue. Maybe let's stick with this option only for now and I'll add a TODO which we ponder if there's a) a need to and b) a nice way to do this.

jamie-lynch: For now, in the above example `this.app` -> `props.app` and the tag would still be free. I think that's worth it but open to thoughts otherwise...

src/constructs/core/stack.ts
```diff
@@ -1,18 +1,46 @@
...
     super(app, id, props);
 
     this.stage = new GuStageParameter(this);
     this.stack = new GuStackParameter(this);
 
     Tags.of(this).add("Stack", this.stack.valueAsString);
     Tags.of(this).add("Stage", this.stage.valueAsString);
+
+    if (props?.app && props.appParameter) {
+      // TODO: How do we do this better with types?
```
akash1810, line 36: Could we infer `appParameter` from the presence of `props.app`?

jamie-lynch: Sorry, I don't quite follow..?

akash1810: No, no, I've misunderstood the logic. To help me: when would we ever want an app param?

jamie-lynch: I'm not sure. I can have a scan to see if there are any existing projects that do it.
Previously we were just taking app as an optional string so we can go back to that if there's no use case for it to be a parameter?

akash1810 (approved):

PR #37, opened 2020-12-07
Introduce ./script/task pattern

akash1810:
## What does this change?

Introduces a `./script/task` pattern as described [here](https://github.com/github/scripts-to-rule-them-all).

This pattern makes it easy to move between repositories (assuming all repositories follow this pattern) as it allows one to build up muscle memory: to set a repo up, run `./script/setup`, to start it `./script/start`. This is really useful when the technology differs between repositories, e.g. js, typescript, scala etc.

Adding this pattern to this repository now to abstract way the implementation detail of the package manager. We've gone from yarn to npm, back to yarn and back to npm again 😅 . Each time, we've updated the instructions. Being able to just run `./script/start` means we can change again later on but in a less dramatic way.

I've no idea what's going on with the diff in GitHub, it doesn't look particularly happy of file moves and edits in the same PR... Maybe easier to review changes on the branch itself?

No.

## How to test

Observe CI.

The repository is easier to contribute to.

n/a

## Images

n/a


script/release
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -e
+
+if npm whoami &>/dev/null; then
+  npm run release
+else
+  echo "You are not logged into npm!"
+  echo "  Run 'npm login' first!"
```
akash1810, line 9: We can test this by running:

```sh
docker run -it --rm --name test -v "$PWD":/usr/src/app -w /usr/src/app node:14.15.1 script/release
```

stephengeller (approved):
Tested all the scripts and works like a charm, nice one!

PR #38, opened 2020-12-08
Add tests for core constructs

jamie-lynch:
## What does this change?

This PR adds test for the constructs within the `core` directory. As part of that change it also:

- Modifies the tests on the `GuStack` construct to use direct assertions over snapshots
- Refactors some parameters to deal with default and override properties more consistently
- Creates a `SynthedStack` class in `test/utils/synthed-stack.ts` to be used by multiple test suites

No

## How to test

Run the test suite using `npm run test` 

The constructs within the `core` directory have good unit test coverage

n/a

## Images

n/a


stephengeller (approved):
Looks good! Nice one, tests pass locally

PR #39, opened 2020-12-08
Add GuArnParameter with ARN regex allowed values

stephengeller:
## What does this change?

This PR adds a `GuArnParameter`. This allows users of the library to add parameters to their stacks that are constrained by a regex enforcing them to pass ARNs compliant with [AWS's documentation](https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html).

The `AllowedValues` regex has been tested and appears to be working.

No.

## How to test

Paste the regex in `allowedValues` into a [regex tester](https://regex101.com/), then paste some valid (and invalid, if desired) ARNs against it and see it successfully validate the correct strings.

Future projects can create parameters that only allow ARN values.

## Screenshots

Valid ARN compositions:
image


src/constructs/core/parameters.ts
```diff
@@ -86,3 +86,13 @@ export class GuAmiParameter extends GuParameter {
     });
   }
 }
+
+export class GuArnParameter extends GuStringParameter {
+  constructor(scope: GuStack, id: string, props: GuNoTypeParameterProps) {
+    super(scope, id, {
+      ...props,
+      allowedPattern: "arn:aws:[a-z0-9]*:[a-z0-9\\-]*:[0-9]{12}:.*",
```
jamie-lynch, line 94: Nice!

akash1810, line 94: Could we add some unit tests against the regex? We could pull it out into a const somewhere and test it in it's own right as a regex pattern?

jamie-lynch (commented):
> No.

Although it doesn't require a change, looks like we have a couple of stacks that take ARN parameters. Happy to look through and PR once this is merged.

src/constructs/core/parameters.ts
stephengeller: Good shout! I've just added these, and discovered that we'll need a separate parameter regex for S3 ARNs, so I've added that!

src/constructs/core/parameters.ts
```diff
@@ -86,3 +86,29 @@ export class GuAmiParameter extends GuParameter {
...
+      allowedPattern: arnRegex,
+      constraintDescription: "Must be a valid ARN, eg: arn:partition:service:region:account-id:resource-id",
+    });
+  }
+}
+
+const s3BucketRegex = "(?!^(\\d{1,3}\\.){3}\\d{1,3}$)(^[a-z0-9]([a-z0-9-]*(\\.[a-z0-9])?)*$(?<!\\-))";
+export const s3ArnRegex = `arn:aws:s3:::${s3BucketRegex}*`;
+
+export class GuS3ArnParameter extends GuStringParameter {
```
akash1810, line 105: Looking at the tests, should we rename this to `GuS3ObjectArnParameter` as it looks like the regex is for an object path and not a bucket name?

stephengeller: Good shout, and done!

jamie-lynch (approved):
Looks good to me!

akash1810 (approved):
This is great!

We might want to put the regexes in `constants` to [reuse them](https://github.com/guardian/cdk/blob/61343ebf15b50dd3cbeb2dbf9add9b004d5ab80e/src/constructs/iam/policies.ts#L98), but we can do that in another PR later.

PR #40, opened 2020-12-09
Add tests for security group constructs

jamie-lynch:
## What does this change?

This PR adds tests for the security group constructs. The GuVpc static functions are _not_ tested as they can't be asserted directly and indirect/implicit testing predominantly tests the behaviour of the aws cdk library. 

No.

## How to test

Run the test suite and see the new tests pass.

Good coverage of the ec2 module.

n/a

## Images

n/a


stephengeller (approved):
Nice! Clear tests, look great and pass locally

PR #41, opened 2020-12-09
Prevent `yarn.lock` from being committed to the repo

stephengeller:
## What does this change?

This repository currently uses `npm` over `yarn` due to some complications in releasing the library with `yarn`. Now that we've settled on using `npm`, we should make it difficult and/or impossible to commit `yarn.lock` files accidentally.

This PR adds a script that is run in `script/build` (the first step in CI) to check for a `yarn.lock` file and fail if found.

Don't think so, no.

## How to test

In the root of the repo, to see it fail: 

```
$ touch yarn.lock
$ ./script/build # see script fail

$ rm yarn.lock
$ ./script/build # see script pass
```

We will never see `yarn.lock` in this repo! Until we later need to go back to yarn!!


akash1810 (commented):
Would a pre-commit/pre-push hook be more idiomatic here? [Husky](https://www.npmjs.com/package/husky) makes hooks super easy.

jamie-lynch:
> Would a pre-commit/pre-push hook be more idiomatic here? [Husky](https://www.npmjs.com/package/husky) makes hooks super easy.
> 
> Or even better, update the `.gitignore`.

My worry with the `.gitignore` option is that it would then be possible to make a change and not notice that the lock file was not updated accordingly which could lead to weird/inconsistent behaviour. Thinking about it again though, CI will probably fail in that case we're running `npm ci`.

twrichards:
just being nosy, but I would only add `yarn.lock` to `.gitignore` in **addition** to having the CI script fail if it detects it and a pre-commit hook to prevent a commit where one is present locally.

lock files are super important for consistency, and unless devs are completely blocked from using yarn, it could cause really pernicious and hard to spot issues.

script/check-yarn-lock
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+RED="\033[31m"
+PLAIN="\033[0m"
+
+DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+ROOT_DIR="${DIR}/.."
+
+if [[ -f "${ROOT_DIR}/yarn.lock" ]]; then
+  echo -e "${RED}There is a yarn.lock file when this is an npm directory!"
```
akash1810, line 10:
```suggestion
  echo -e "${RED}There is a yarn.lock file when this is an npm project!"
```

akash1810 (approved):

stephengeller:
Thanks for that @twrichards . That's a good point - it's definitely important to make sure we don't consider this change lightly. I think that, for now, we're okay with the possible pain of not having a pre-commit hook since it will get caught in CI and when running `script/release`. However, I think getting a pre-commit hook in this repository would be sensible in the future, and this check being present in it when it does.

PR #42, opened 2020-12-09
Add getters for stack and stage values

jamie-lynch:
## What does this change?

This PR adds getters for the `stack` and `stage` values in `GuStack` so that it is possible to use `this.stack` or `this.stage` directory rather than having to add `valueAsString` everywhere. This also means that once #30 is merged, the `GuStack` will have a consistent interface for accessing stack, stage and app.

Yes, everything that uses `this.stage.valueAsString` or `this.stack.valueAsString` at the moment will need to be updated.

## How to test

n/a

Slightly less typing is required to use the stack and stage values.

n/a

## Images

n/a

### DIFF
--- a/src/constructs/core/stack.ts
+++ b/src/constructs/core/stack.ts
@@ -7,8 +7,16 @@ export interface GuStackProps extends StackProps {
 }
 
 export class GuStack extends Stack {
-  protected stage: GuStageParameter;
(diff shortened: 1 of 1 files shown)


stephengeller (approved):
Great! Happy to help migrate existing usages of this to the new format.

akash1810 (approved):

PR #43, opened 2020-12-09
Add an identity interface

jamie-lynch:
## What does this change?

This PR creates a new identity interface which can be used Where there a places that take stack, stage and app params as input to save typing them out every time.

No.

## How to test

n/a

Reduced duplication of the stack, stage and app type definitions.

n/a

## Images

n/a

### DIFF
--- a/src/constructs/core/identity.ts
+++ b/src/constructs/core/identity.ts
@@ -1,0 +1,5 @@
+export interface Identity {
+  stack: string;
+  stage: string;
+  app: string;
+}


src/constructs/core/identity.ts
```diff
@@ -0,0 +1,5 @@
+export interface Identity {
+  stack: string;
+  stage: string;
+  app: string;
+}
```
jamie-lynch, line 5: Wasn't entirely sure where to put this so open to suggestions of where it could live.

src/constructs/core/identity.ts
```diff
@@ -0,0 +1,5 @@
+export interface Identity {
+  stack: string;
+  stage: string;
+  app: string;
```
akash1810, line 4: Should `app` be optional?

jamie-lynch: Hmm... I guess that might depend on the use case

akash1810: Good point. Let's get this in, we can always update it later if needed.

akash1810 (approved):

PR #44, opened 2020-12-09
Add tests for IAM constructs

jamie-lynch:
## What does this change?

This PR adds tests for the IAM constructs. 

No.

## How to test

Run the new test suite.

IAM constructs have good coverage.

n/a

## Images

n/a


akash1810 (approved):
Happy to get this in for now, we might want to consider reworking the `overrideId` prop; it would be good to have a list of immutable resources (load balancers, certificates, buckets, etc) and allow only those to keep pre-existing ids. For roles and permissions, I'm not sure it provides much value to use a custom logical ID.

PR #45, opened 2020-12-09
Add tests for RDS instance

jamie-lynch:
## What does this change?

This PR adds tests for the RDS constructs. 

No.

## How to test

Run the new test suite.

RDS constructs have good coverage.

n/a

## Images

n/a

### DIFF
--- a/src/constructs/rds/instance.test.ts
+++ b/src/constructs/rds/instance.test.ts
@@ -15,12 +15,15 @@ describe("The GuDatabaseInstance class", () => {
     privateSubnetIds: [""],
   });
 
-  it("removes the db prefix from the input instance type", () => {
+  it("removes the db prefix from the instanceType prop (before CDK adds it back again)", () => {
     const app = new App();
     const stack = new GuStack(app);
 
     new GuDatabaseInstance(stack, "DatabaseInstance", {
       vpc,
+      // When you pass an instanceType to the DatabaseInstance class, the db. prefix is added automagically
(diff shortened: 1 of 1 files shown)


src/constructs/rds/instance.test.ts
```diff
@@ -0,0 +1,135 @@
...
+
+describe("The GuDatabaseInstance class", () => {
+  const vpc = Vpc.fromVpcAttributes(new Stack(), "VPC", {
+    vpcId: "test",
+    availabilityZones: [""],
+    publicSubnetIds: [""],
+    privateSubnetIds: [""],
+  });
+
+  it("removes the db prefix from the input instance type", () => {
```
akash1810, line 18: Not sure I understand this test; `GuDatabaseInstance` gets instantiated with `instanceType: db.t3.small` and the `DBInstanceClass` is the same value? Should `DBInstanceClass` be `t3.small` in accordance to the test description?

jamie-lynch: The logic being tested here removes the `db.` prefix if it exists on the instanceType value. This is because there's some logic in the CDK which adds the prefix regardless of whether it exists already. Without this logic the `DBInstanceClass` ends up being `db.db.t3.small` given the inputs of this test. I'll update the test name and/or add some comments as this isn't particularly clear.

It might be that we want to enforce the lack of a `db.` prefix in some other way so maybe let's have a think if there's a better way of doing it.

stephengeller (approved):
Nice! Tests are also clear and pass locally.

PR #48, opened 2020-12-11
Update documentation and add ADRs

jamie-lynch:
## What does this change?

This PR adds a `docs` dir with separate files for general usage, starting a new CDK project and migrating existing cloudformation to CDK. Grouping the docs into their own folder makes it easier to split the files out logically as we expand the documentation about the library. 

It also adds a folder of [Architecture Design Records (ADRs)](https://github.com/joelparkerhenderson/architecture_decision_record). These files are where we can document the decisions we make around any form of structure, architecture or approach. By documenting them in this way, we can preserve the thought process behind all of the decisions whilst also laying out formally the preferences for all developers working on the library. 

The folder contains a template and some initial ADRs covering decisions we've discussed to date. 

No.

All of our design decisions are documented to remind future us why we thought something was a good idea at the time

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,20 +8,25 @@ Welcome to the Guardian CDK library! This library contains a number of reusable 
 
 You can read more about it in the [aws-cdk README](https://github.com/aws/aws-cdk).
 
-## Patterns
+## Architecture
+
+Read more about constructs, patterns and other architectural decisions in (docs)[docs]
+
+### Patterns
 
 Patterns are high level classes which compose a number of constructs to produce standard architectures. For example, you should be able to get all of the resources you need to deploy a new lambda function from one `GuLambdaStack` class. We're still working on these right now but hope to start bringing you
 some of the most common Guardian stacks soon!
 
 Patterns should be your default entry point to this library.
 
-## Constructs
+### Constructs
 
 Constructs are lower level classes which will create one or more resources to produce one element of a stack. For example, the `GuDatabaseInstance` will create an RDS instance as well as a parameter group, if required. This library defines a number of constructs which are combined to create the higher level patterns.
 
 If there is a pattern available for your use case, prefer to use that over composing constructs yourself. We welcome feedback and/or PRs to extend the functionality of patterns. Where you need to do something outside of currently available patterns, you can use the constructs to provide some level of abstraction. In this case, consider whether it's worth defining a pattern.
 
 ## Useful commands
+
 We follow the [`script/task`](https://github.com/github/scripts-to-rule-them-all) pattern,
 find useful scripts within the [`script`](./script) directory for common tasks.
 
@@ -33,6 +38,7 @@ find useful scripts within the [`script`](./script) directory for common tasks.
 - `./script/release` to release a new version to NPM
 
 There are also some other commands defined in `package.json`:
+
 - `npm run lint --fix` attempt to autofix any linter errors
 - `npm run format` format the code using Prettier
 - `npm run watch` watch for changes and compile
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -8,20 +8,25 @@ Welcome to the Guardian CDK library! This library contains a number of reusable
 
 You can read more about it in the [aws-cdk README](https://github.com/aws/aws-cdk).
 
-## Patterns
+## Architecture
+
+Read more about constructs, patterns and other architectural decisions in (docs)[docs]
```
stephengeller, line 13:
```suggestion
Read more about constructs, patterns and other architectural decisions in [docs](docs)
```

jamie-lynch:
I've now written up the ADRs so this is ready for review. Not sure if we want to finalise the ADRs here or merge the proposals and then open separate PRs to finalise and mark as agreed each one.

docs/001-general-usage.md
```diff
@@ -0,0 +1,35 @@
...
+
+If there is a pattern available for your use case, prefer to use that over composing constructs yourself. We welcome feedback and/or PRs to extend the functionality of patterns. Where you need to do something outside of currently available patterns, you can use the constructs to provide some level of abstraction. In this case, consider whether it's worth defining a pattern.
+
+## Using patterns and constructs
+
+Patterns can be imported from the top level of the library (e.g. `import { InstanceRole } from "@guardian/cdk";`) while constructs must be imported from their construct directory (e.g. `import { GuAutoScalingGroup } from "@guardian/cdk/lib/constructs/autoscaling";`)
+
+## AWS Library Versions
+
+Any versions of the `@aws-cdk` libraries that you have installed in your project must be the same version as those used in the `@guardian/cdk` library.
```
akash1810, line 35: I'd hope we can enforce this by listing our cdk deps in `peerDependencies` - https://github.com/guardian/cdk/pull/20 might need a bit more work to get right.

docs/002-starting-a-new-project.md
```diff
@@ -0,0 +1,27 @@
...
+
+This directory contains a file for each app that your cdk defines. For each app, there may be multiple stacks. For single apps, that file can be called `cdk.ts` whereas for multiple apps it is recommended to name them accordingly.
+
+**lib**
+
+The lib directory contains all of the stack defitions. For single apps, these file(s) will be at the top level of the directory and for a single stack, the file may be called `cdk-stack.ts`. Multiple stacks should be named accordingly and if multiple apps exist, nested in sub directories.
+
+**cdk.out**
+
+This directory is where the synthesised cloudformation will be written to.
```
akash1810, line 23: Do we need mention the default is json?

docs/architecture-decision-records/002-component-constuctors.md
```diff
@@ -0,0 +1,144 @@
...
+   :white_check_mark: Valid
+
+   ```ts
+   interface MyConstructProps {
+     prop1?: string;
+     prop2?: string
+   }
+
+   class MyConstruct {
+     constructor(scope: GuStack, id: string = "MyConstruct", props?: MyConstructProps) {
```
akash1810, line 122: Opinions wanted on this pattern that I've been playing around with locally:

```ts
class Foo {
  constructor(scope: GuStack, id: string, props?: FooProps) {
    // do stuff
  }

  static withProps(scope: GuStack, props?: FooProps) {
    return new Foo(scope, "MyFoo", props)
  }
}

class Bar {
  private foo: Foo;

  constructor(props: FooProps) {
    this.foo = Foo.withProps(props);
  }
}
```

The idea here is to avoid needing to specify the ID when you want to provide props; this is the closest thing I can find to an overloaded constructor in TypeScript.

docs/architecture-decision-records/003-package-manager.md
```diff
@@ -0,0 +1,48 @@
...
+
+1. Use `yarn` and fix the existing issues
+
+   `Yarn` is more widely used across the team's, and wider department's, projects so it would present a more consistent approach. There are more examples of common tasks written in `yarn` from our existing projects and developer's are more likely to default to running `yarn` commands.
+
+2. Use `npm`
+
+   Given that issues have been encountered using `yarn` and the fact that `npm` provides a working alternative with no obvious drawbacks (save the more common use of `yarn` in the department), it's not worth fixing the `yarn` issues.
+
+3. Use `npm` but use the script pattern to make it (more) transparent to the user
```
akash1810, line 30:
```suggestion
3. Use `npm` but use the script pattern to make it (more) opaque to the user
```

?

docs/architecture-decision-records/005-package-structure.md
```diff
@@ -0,0 +1,47 @@
...
+
+   While it does not provide the range of components, this library does provide multiple implementations of the same underlying resource (e.g. for the `Policy` construct, `GuLogShippingPolicy`, `GuSSMRunCommandPolicy`, `GuGetS3ObjectPolicy`).
+
+## Decision
+
+<!-- What is the change that we're proposing and/or doing? -->
+
+The top level directories with the `constructs` directory should mirror the AWS CDK library names.
+
+Each directory should contain an `index.ts` file which exports all of the classes within it.
```
akash1810, line 37: I think in the future we'd only export top level things at the module level; if you really want to import an "unusual" object, you need to work for it as a way of discouragement?

akash1810 (approved):
Looks great! Added a couple of comments, nothing blocking. Would be good to talk about some of these in the client side infra meetings.

docs/architecture-decision-records/002-component-constuctors.md
jamie-lynch: Interesting. I'd certainly favour this over removing the ID from the constructor. Maybe another one for discussions?

docs/architecture-decision-records/005-package-structure.md
jamie-lynch: Yeah, especially when we have most cases covered with patterns. I guess at that point we can take a view of what constructs are still commonly/somewhat used and export those at the top level whilst making people work to go even further off piste.

PR #52, opened 2020-12-15
Refactor lambda tests to direct assertions

jamie-lynch:
## What does this change?

This PR refactors the existing lambda tests to use directory assertions over snapshots to be in line with our agreed testing strategy.

The assertions for the API test case are far simpler than the snapshot as the APIs required the creation of many elements. Whilst this makes understanding the tests easier, it does raise the risk of an unintended change not being picked up by the test suite. The options here are:

1) Add a snapshot to this test
2) Add more direct assertions to check the most important requirements
3) Use snapshot testing on the pattern (tbd) to verify the overall stack is correct

No.

## How to test

Review the current tests and run them using `./script/test`

Lambda constructs have good unit test coverage following our agreed testing stategy.

The new tests don't cover a case which they should meaning that bugs could make it into production.


stephengeller (approved):
Looks good. I'm not 100% sure if I understand why we got rid of the snapshot test - is it because we strictly agreed to test resulting resources over snapshots? If so, all good! If not, happy to additionally have a snapshot test but I'm not 100% sure on the implications of that.

jamie-lynch:
> Looks good. I'm not 100% sure if I understand why we got rid of the snapshot test - is it because we strictly agreed to test resulting resources over snapshots? If so, all good! If not, happy to additionally have a snapshot test but I'm not 100% sure on the implications of that.

My understanding is that we agreed on direct assertions for constructs and snapshots for patterns but increasingly I wonder if that should be a preference rather than a strict requirement. In this case for example, I think adding the snapshot assertion back only for the `should add any apis passed in` test would be beneficial. 

Happy to do that here or raise the discussion as part of the ADRs and open a PR to update all the tests accordingly once we've had the discussion?

akash1810 (approved):

PR #53, opened 2020-12-15
Add overrideId prop to GuAutoScalingGroup and refactor for TODOs

stephengeller:
## What does this change?

This addresses a number of TODOs in the repository, specifically:

1. Adds an `overrideId` prop to `GuAutoScalingGroup`, rather than overriding by default
2. Extracts the SSM run command policy statement to its own variable so it can be used elsewhere in the future
3. Extracts the allowGetObjectPolicyStatement to its own variable for same reasons as above

Yes, it looks like [a few repositories](https://github.com/search?q=new+GuAutoScalingGroup+user%3Aguardian&type=code) will need to be updated in order to take the GuAutoScalingGroup change into account.

## How to test

`yarn test` from root of the repository.

The repository is cleaner (and littered with fewer TODOs) and our `GuAutoScalingGroup` construct is more consistent with the rest of the constructs.

We need to make sure we migrate the affected repositories successfully, as this could cause add/removes for the existing ASGs if not. However, it shouldn't be too hard as we'd just need to bump the version and add the `overrideId: true` prop to the usages.

### DIFF
--- a/src/constructs/autoscaling/asg.test.ts
+++ b/src/constructs/autoscaling/asg.test.ts
@@ -114,19 +114,29 @@ describe("The GuAutoScalingGroup", () => {
   test("does not include the UpdatePolicy property", () => {
     const app = new App();
     const stack = new GuStack(app);
-    new GuAutoScalingGroup(stack, "AutoscalingGroup", defaultProps);
+    new GuAutoScalingGroup(stack, "AutoscalingGroup", { ...defaultProps, overrideId: true });
 
     const json = SynthUtils.toCloudFormation(stack) as SynthedStack;
     expect(Object.keys(json.Resources.AutoscalingGroup)).not.toContain("UpdatePolicy");
   });
 
-  test("overrides the id to remove the 8 digit value added automagically", () => {
+  test("overrides the id with the overrideId prop set to true", () => {
+    const app = new App();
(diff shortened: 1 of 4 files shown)


src/constructs/autoscaling/asg.test.ts
```diff
@@ -114,19 +114,29 @@ describe("The GuAutoScalingGroup", () => {
   test("does not include the UpdatePolicy property", () => {
     const app = new App();
     const stack = new GuStack(app);
-    new GuAutoScalingGroup(stack, "AutoscalingGroup", defaultProps);
+    new GuAutoScalingGroup(stack, "AutoscalingGroup", { ...defaultProps, overrideId: true });
```
jamie-lynch, line 117: I think this test should pass regardless of the overrideId flag so maybe it's not needed here?

stephengeller: Unfortunately not, as `Object.keys(json.Resources.AutoscalingGroup)` fails because there is no `AutoscalingGroup` key in `Resources` since it'll be auto-generated.

jamie-lynch: Ah, good point!

jamie-lynch (approved):

PR #55, opened 2020-12-16
add dependabot.yml

akash1810:
## What does this change?

Hopefully this is an automated way of keeping deps up to date... aws-cdk's release cycle is pretty quick!

Config taken from [here](https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuration-options-for-dependency-updates).

## How to test

n/a

Dependencies are up to date.

`aws-cdk` has [breaking changes between minor versions](https://github.com/aws/aws-cdk/releases). This might result in a higher frequency of releases to this library as the APIs change, this will then impact consumers.

## Images

n/a

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+# See the documentation for all configuration options https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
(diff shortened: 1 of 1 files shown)


jamie-lynch (approved):
🤖

PR #56, opened 2020-12-16
split policies into individual files

akash1810:
## What does this change?

Reorganises `policies.ts` into individual files to make it easier to manage. This is a no-op: same class names, same tests, same exports. The second commit makes the return type explicit for simpler linting. There is also a net LOC increase as the `import`s are now replicated across the multiple files.

No.

## How to test

n/a


jamie-lynch (approved):
Makes sense to me. I think it would be good for us to define some clear rules / guidance on how to approach this across the project as there are a few places that have multiple classes per file. Maybe another one to add to ADRs..?

src/constructs/iam/policies/log-shipping.ts
```diff
@@ -0,0 +1,32 @@
...
+import { Effect, PolicyStatement } from "@aws-cdk/aws-iam";
+import type { GuStack } from "../../core";
+import type { GuPolicyProps } from "./base-policy";
+import { GuPolicy } from "./base-policy";
+
+export interface GuLogShippingPolicyProps extends GuPolicyProps {
+  kinesisStreamName: string;
+}
+
+// This interface is inconsistent. Should we add the ID here?
```
jamie-lynch, line 11: This comment can be removed as we clearly put the id back in at some point

src/constructs/iam/policies/base-policy.ts
```diff
@@ -0,0 +1,20 @@
+import type { CfnPolicy, PolicyProps } from "@aws-cdk/aws-iam";
+import { Policy } from "@aws-cdk/aws-iam";
+import type { GuStack } from "../../core";
+
+export interface GuPolicyProps extends PolicyProps {
+  overrideId?: boolean;
+}
+
+export class GuPolicy extends Policy {
+  static defaultProps: Partial<GuPolicyProps> = {};
```
jamie-lynch, line 10: Is it worth keeping this if it doesn't do anything at the moment?

src/constructs/iam/policies/log-shipping.test.ts
```diff
@@ -0,0 +1,94 @@
...
+import { Role, ServicePrincipal } from "@aws-cdk/aws-iam";
+import { App } from "@aws-cdk/core";
+import { GuStack } from "../../core";
+import { GuLogShippingPolicy } from "./log-shipping";
+
+describe("The GuLogShippingPolicy class", () => {
+  it("sets default props", () => {
+    const stack = new GuStack(new App());
+
+    const logShippingPolicy = new GuLogShippingPolicy(stack, "SSMRunCommandPolicy", { kinesisStreamName: "test" });
```
jamie-lynch, line 11: Tiny one but there's a copy and paste error in the id here

```suggestion
    const logShippingPolicy = new GuLogShippingPolicy(stack, "LogShippingPolicy", { kinesisStreamName: "test" });
```

src/constructs/iam/policies/log-shipping.test.ts
```diff
@@ -0,0 +1,94 @@
...
+          },
+        ],
+      },
+    });
+  });
+
+  it("merges defaults and passed in props", () => {
+    const stack = new GuStack(new App());
+
+    const logShippingPolicy = new GuLogShippingPolicy(stack, "SSMRunCommandPolicy", {
```
jamie-lynch, line 53:
```suggestion
    const logShippingPolicy = new GuLogShippingPolicy(stack, "LogShippingPolicy", {
```

jamie-lynch (commented):
Sorry, spotted some things when I looked back as I was reviewing #57

src/constructs/iam/policies/base-policy.ts
akash1810: Yes, but not for for this PR to keep it as no-op as possible. I've got another branch that makes a few changes to this area, so might make sense to add it there?

akash1810:
> I think it would be good for us to define some clear rules / guidance on how to approach this across the project as there are a few places that have multiple classes per file. Maybe another one to add to ADRs..?

Yep! Will have a think about what the ADR would be; recommend one file per class/concern and an `index.ts` that exports all the things?

jamie-lynch:
> > I think it would be good for us to define some clear rules / guidance on how to approach this across the project as there are a few places that have multiple classes per file. Maybe another one to add to ADRs..?
> 
> Yep! Will have a think about what the ADR would be; recommend one file per class/concern and an `index.ts` that exports all the things?

I actually added something about it in the last ADR this morning. Still in the `proposed` state so we can discuss more before we finalise.

https://github.com/guardian/cdk/pull/48/commits/5ad6610b687b02311e1b73c51fdafaee529e2c2d

PR #57, opened 2020-12-16
simplify interface InstanceRoleProps

akash1810:
## What does this change?

Extending multiple interfaces yields a simpler interface and consistent variable names.

No.

## How to test

Observe CI?

Simpler code?

n/a

## Images

n/a

### DIFF
--- a/src/constructs/iam/policies/ssm.ts
+++ b/src/constructs/iam/policies/ssm.ts
@@ -33,7 +33,7 @@ export class GuSSMRunCommandPolicy extends GuPolicy {
     };
   }
 
-  constructor(scope: GuStack, id: string = "SSMRunCommandPolicy", props: GuPolicyProps = { overrideId: false }) {
+  constructor(scope: GuStack, id: string = "SSMRunCommandPolicy", props?: GuPolicyProps) {
     super(scope, id, {
       ...GuSSMRunCommandPolicy.getDefaultProps(),
       ...props,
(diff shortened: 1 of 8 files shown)


src/constructs/iam/policies/ssm.ts
```diff
@@ -33,7 +33,7 @@ export class GuSSMRunCommandPolicy extends GuPolicy {
     };
   }
 
-  constructor(scope: GuStack, id: string, props: GuPolicyProps) {
+  constructor(scope: GuStack, id: string = "SSMRunCommandPolicy", props: GuPolicyProps = { overrideId: false }) {
```
jamie-lynch, line 36: I'm not sure that I like defining a default object for the prop and then also having a `getDefaultProps` function. It makes it a bit confusing to have two different places to define default props. If there are no required props, maybe the whole props parameter should be optional instead? 

```suggestion
  constructor(scope: GuStack, id: string = "SSMRunCommandPolicy", props?: GuPolicyProps) {
```

jamie-lynch:
> No.

jamie-lynch:
Clicked the wrong button there... 🤦

src/constructs/iam/policies/ssm.ts
akash1810: Yes, I think I agree with this, it gets us close to same shape as CDK classes too.

jamie-lynch (approved):
👍

PR #65, opened 2020-12-17
add helper function to add tags

akash1810:
## What does this change?

This provides a slightly nicer API, very slightly! Also applies tags to launched instances by default.

No.

## How to test

n/a

An ever so slightly nicer API.

n/a

## Images

n/a


jamie-lynch (approved):
Nice!

PR #66, opened 2020-12-18
add more permissions to InstanceRole as standard

akash1810:
## What does this change?

Updates the standard permissions given to `InstanceRole`, adding:
- Permissions to read instance and ASG state - applications read these at launch for service or config discovery
- Permissions to read from parameter store, for config

This is after observing the verbosity within the [new Typerigther stack](https://github.com/guardian/typerighter/compare/setup-rule-manager-infrastructure). cc @jonathonherbert @SHession 

**Opinions wanted**
Also adds a `Try` class that mimics Scala, namely the `getOrElse` - see the included tests. This is to handle the [possible error](https://github.com/guardian/cdk/blob/d1542ae73fb9e0c12a347dca7ff9d2e6f101c69d/src/constructs/core/stack.ts#L22-L30) when accessing `app` from `GuStack`. Alternatives:
- `GuStack` to have a `appForPolicy` property that returns `*` rather than throwing
- It could be an [`Either`](https://gcanti.github.io/fp-ts/modules/Either.ts.html#orelse)
- `Try` could be added to [`@guardian/types`](https://github.com/guardian/types)

Yes, the stacks can be thinned out as more things come as standard.

- [ ] [Prism](https://github.com/guardian/deploy-tools-platform/blob/c7d687d3e9a370cad04aaae7efc7bacd952ab97b/cdk/lib/prism/prism-stack.ts#L59-L68)
- [ ] [Grafana](https://github.com/guardian/deploy-tools-platform/blob/c7d687d3e9a370cad04aaae7efc7bacd952ab97b/cdk/lib/grafana/grafana-stack.ts#L147-L159)
- [ ] [AMIable](https://github.com/guardian/deploy-tools-platform/blob/c7d687d3e9a370cad04aaae7efc7bacd952ab97b/cdk/lib/amiable/amiable.ts#L93-L98)

## How to test

See unit tests.

Reducing the boilerplate needed to define a CDK stack.

n/a

## Images

n/a

### DIFF
--- a/script/test
+++ b/script/test
@@ -2,10 +2,18 @@
 
 set -e
 
+EXTRA_ARGS=""
+if [[ $# -gt 0 ]] ; then
+  # npm requires `--` for argument passing
+  # see https://docs.npmjs.com/cli/v6/commands/npm-run-script
+  EXTRA_ARGS="-- $*"
+fi
+
 # only install dependencies when running in CI (currently GitHub Actions)
 # see https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables#default-environment-variables
 if [ $CI ] ; then
   npm ci
 fi
 
-npm run test
+# shellcheck disable=SC2086
+npm run test $EXTRA_ARGS
--- a/src/constructs/iam/policies/describe-ec2.ts
+++ b/src/constructs/iam/policies/describe-ec2.ts
@@ -1,0 +1,29 @@
+import type { PolicyProps } from "@aws-cdk/aws-iam";
+import { Effect, PolicyStatement } from "@aws-cdk/aws-iam";
+import type { GuStack } from "../../core";
+import type { GuPolicyProps } from "./base-policy";
+import { GuPolicy } from "./base-policy";
+
+export class GuDescribeEC2Policy extends GuPolicy {
+  private static getDefaultProps(): PolicyProps {
+    return {
+      policyName: "describe-ec2-policy",
+      statements: [
+        new PolicyStatement({
+          effect: Effect.ALLOW,
+          actions: [
+            "autoscaling:DescribeAutoScalingInstances",
+            "autoscaling:DescribeAutoScalingGroups",
+            "ec2:DescribeTags",
+            "ec2:DescribeInstances",
+          ],
+          resources: ["*"],
+        }),
+      ],
+    };
+  }
+
+  constructor(scope: GuStack, id: string = "DescribeEC2Policy", props?: GuPolicyProps) {
+    super(scope, id, { ...GuDescribeEC2Policy.getDefaultProps(), ...props });
+  }
+}
--- a/src/constructs/iam/policies/parameter-store-read.ts
+++ b/src/constructs/iam/policies/parameter-store-read.ts
@@ -1,0 +1,27 @@
+import type { PolicyProps } from "@aws-cdk/aws-iam";
+import { Effect, PolicyStatement } from "@aws-cdk/aws-iam";
+import { Try } from "../../../utils/try";
+import type { GuStack } from "../../core";
+import type { GuPolicyProps } from "./base-policy";
+import { GuPolicy } from "./base-policy";
+
+export class GuParameterStoreReadPolicy extends GuPolicy {
+  private static getDefaultProps(scope: GuStack): PolicyProps {
+    const app: string = new Try<string>(() => scope.app).getOrElse("*");
+
+    return {
+      policyName: "parameter-store-read-policy",
+      statements: [
+        new PolicyStatement({
+          effect: Effect.ALLOW,
+          actions: ["ssm:GetParametersByPath"],
+          resources: [`arn:aws:ssm:${scope.region}:${scope.account}:parameter/${scope.stage}/${scope.stack}/${app}`],
+        }),
+      ],
+    };
+  }
+
+  constructor(scope: GuStack, id: string = "ParameterStoreRead", props?: GuPolicyProps) {
+    super(scope, id, { ...GuParameterStoreReadPolicy.getDefaultProps(scope), ...props });
+  }
+}
--- a/src/utils/try.test.ts
+++ b/src/utils/try.test.ts
@@ -1,0 +1,44 @@
+import { Try } from "./try";
+
+class Rectangle {
+  private readonly width: number;
+  private readonly height: number;
+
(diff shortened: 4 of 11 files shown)


script/test
```diff
@@ -8,4 +8,4 @@ if [ $CI ] ; then
   npm ci
 fi
 
-npm run test
+npm run test "$@"
```
akash1810, line 11: Pass arguments to the npm script - useful for updating tests as we can do `./script/test -- -u`.

src/utils/try.ts
```diff
@@ -0,0 +1,17 @@
+export class Try<T> {
```
jonathonherbert, line 1: Not against this by any means – I've done a similar thing in our telemetry service – but would it be worth standardising on an approach here w/ e.g. [fp-ts](https://gcanti.github.io/fp-ts/modules/Either.ts.html), or do we think the overhead (comprehension, opinions) isn't worth paying?

script/test
jonathonherbert: TIL! 🙏

src/utils/try.ts
akash1810: Wasn't entirely sure if `fp-ts` was too heavy right now, as it's for this single case; maybe adding `Try` to [`@guardian/types`](https://github.com/guardian/types) is an option?

jonathonherbert: Definitely an `Option` – `Either` would be good (not necessarily for now!), I wouldn't want to be seen as `Invariant` (although I suspect weight doesn't matter much in a node environment?), interested in the medium term in the `Group` consensus

script/test
akash1810: 😄 At second look, I'm not too keen on this, `./script/test --update` is a more obvious interface, will update.

jonathonherbert (approved):
LGTM, changes look sensible and this would remove twenty or so lines from our CFN.

src/utils/try.test.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  it("should set isSuccess to true on the happy path", () => {
+    const value = new Try<number>(() => goodRectangle.area);
+    expect(value.isSuccess).toBe(true);
+  });
+
+  it("should set isSuccess to false on the unhappy path", () => {
+    const value = new Try<number>(() => badRectangle.area);
+    expect(value.isSuccess).toBe(false);
+  });
```
jamie-lynch, line 43: Only a small one but this doesn't test the `get` method

script/test
akash1810: Slightly better now, we can now pass any [Jest flag](https://jestjs.io/docs/en/cli), for example `./script/test -u`.

src/utils/try.ts
akash1810: I think @jamie-lynch 's [comment](https://github.com/guardian/cdk/pull/66/files#r545678705) is key here, understanding the use-case for a multi-app stack will tell us if we need this.

src/constructs/iam/policies/parameter-store-read.ts
```diff
@@ -0,0 +1,27 @@
+import type { PolicyProps } from "@aws-cdk/aws-iam";
+import { Effect, PolicyStatement } from "@aws-cdk/aws-iam";
+import { Try } from "../../../utils/try";
+import type { GuStack } from "../../core";
+import type { GuPolicyProps } from "./base-policy";
+import { GuPolicy } from "./base-policy";
+
+export class GuParameterStoreReadPolicy extends GuPolicy {
+  private static getDefaultProps(scope: GuStack): PolicyProps {
+    const app: string = new Try<string>(() => scope.app).getOrElse("*");
```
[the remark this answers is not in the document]
akash1810: > I wonder if it's worth finding a stack that covers more than one app to better understand what the logic should be here.

Great point! How about we make `app` a standard prop in `GuStack` for now and refactor for a multi-app stack once we come across it.

[Grid](https://github.com/guardian/editorial-tools-platform/blob/main/cloudformation/media-service-account/grid/media-service.yaml) is the canonical example I use for this, however I think we're a bit far off from being able to tackle that stack.

src/constructs/iam/policies/describe-ec2.ts
```diff
@@ -0,0 +1,29 @@
+import type { PolicyProps } from "@aws-cdk/aws-iam";
+import { Effect, PolicyStatement } from "@aws-cdk/aws-iam";
+import type { GuStack } from "../../core";
+import type { GuPolicyProps } from "./base-policy";
+import { GuPolicy } from "./base-policy";
+
+export class GuDescribeEC2Policy extends GuPolicy {
+  private static getDefaultProps(): PolicyProps {
```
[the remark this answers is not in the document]
akash1810: Oh?! I was following the other `policies` we've defined. Definitely agree we should be consistent. Do you have examples of `DEFAULT_PROPS`?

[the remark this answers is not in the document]
jamie-lynch: The [GuApplicationListener](https://github.com/guardian/cdk/blob/6047cbcc12d55a33174dfc2cc7bf18fb4fd9588d/src/constructs/loadbalancing/elb.ts#L54) is the only example I can find now. I think we might have moved the others to spreading the defaults in the `super` call, which I guess is now a third way of doing it 🙃 .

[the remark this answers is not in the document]
jamie-lynch: I think spreading the defaults is fine for one or two values but gets a bit messy for anything more so having that and the `getDefaultProps` method are both valid use cases. Not a strongly held opinion though.

src/constructs/iam/policies/parameter-store-read.ts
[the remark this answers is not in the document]
akash1810: https://github.com/guardian/cdk/pull/69 to make `app` mandatory.

src/constructs/iam/policies/describe-ec2.ts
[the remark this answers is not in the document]
akash1810: Oh boy! Ideally we'd have a single pattern for doing this. I think it makes sense to address this in a separate PR.

akash1810:
Closing in favour of https://github.com/guardian/cdk/pull/76 since https://github.com/guardian/cdk/pull/69 has changed the shape of `GuStack`.

PR #69, opened 2020-12-18
make `app` mandatory in a GuStack

akash1810:
## What does this change?

For now, we do not support a multi-app stack so make `app` mandatory for simplicity.

Yes, as the signature of GuStack's constructor has changed.

## How to test

n/a

Simpler code, for now?

We'll need to support multi-app stacks at some point, so this might cause a massive diff later.

## Images

n/a

### DIFF
--- a/test/utils/simple-gu-stack.ts
+++ b/test/utils/simple-gu-stack.ts
@@ -1,4 +1,4 @@
 import { App } from "@aws-cdk/core";
 import { GuStack } from "../../src/constructs/core";
 
-export const simpleGuStackForTesting: () => GuStack = () => new GuStack(new App());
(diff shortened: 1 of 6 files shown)


test/utils/simple-gu-stack.ts
```diff
@@ -0,0 +1,4 @@
+import { App } from "@aws-cdk/core";
+import { GuStack } from "../../src/constructs/core";
+
+export const simpleGuStackForTesting = () => new GuStack(new App(), "Test", { app: "testing" });
```
jamie-lynch, line 4: This is a great idea! Maybe it should be it's own PR though as it's a pretty big diff?

akash1810: Good idea - https://github.com/guardian/cdk/pull/70

jamie-lynch (approved):

PR #72, opened 2020-12-18
Add a test helper for policy tests

jamie-lynch:
## What does this change?

This PR adds a helper function which can be used when testing policies to add the policy to a test role. This is required as policies that are not added to roles do not get included in the stacks and therefore can't be tested. Providing a test helper means that each test does not have to create the role itself, reducing the amount of duplicated code.

No.

## How to test

Run the test suite and ensure everything still passes

Less repeated code between tests


akash1810 (approved):
img

PR #73, opened 2020-12-18
update PR template

akash1810:
## What does this change?

Remove images section from PR template. There is little to no UI on this project (is terminal output UI?), so this section doesn't really apply to the project; remove it to reduce noise.

n/a

## How to test

n/a

It's faster to raise PRs?

n/a

## Images

img


jamie-lynch (approved):
👋

PR #77, opened 2020-12-18
make GuPolicy abstract to discourage usage

akash1810:
## What does this change?

Consumers of this library should primarily be using the concrete policy constructs. Making `GuPolicy` abstract is a way to signal this as it makes it harder to initialise a vanilla `GuPolicy`.

Hopefully not! Will update this section once the version of @guardian/cdk has been bumped in the other repos.

## How to test

n/a

Using the compiler to encourage the use of concrete constructs.

Stacks will need to be migrated away from using GuPolicy directly.

## Images

n/a


src/constructs/iam/policies/base-policy.ts
```diff
@@ -6,10 +6,10 @@ export interface GuPolicyProps extends PolicyProps {
   overrideId?: boolean;
 }
 
-export class GuPolicy extends Policy {
+export abstract class GuPolicy extends Policy {
```
jamie-lynch, line 9: My only concern with this is that there will be policies that aren't common enough for us to create dedicated constructs for where this would be useful. Having said that:

1) The only thing this class does over `Policy` is the `overrideId` functionality which we may or may not keep for resources where it is less important when the ID changes so it doesn't matter if they use `Policy` directly
2) I don't actually have examples of that so we can always change this back if we come across a genuine use for it.

jamie-lynch (approved):
One comment with a slight concern but I think I talked myself out of it being a problem as I was typing

src/constructs/iam/policies/base-policy.ts
akash1810: For anything truly uncommon, consumers would still be able to do:

```ts
class Foo extends GuPolicy {
  constructor(scope: GuStack, id: string, props?: FooProps) {
    super(scope, id, props);
  }
}

new Foo(...)
```

jamie-lynch: That's a good point! As I say, I don't have any specific examples in mind anyway so let's get this merged. 🚢

akash1810: Alas whilst updating our platform repo to the latest version of guardian/cdk it appears we make use of `GuPolicy` [quite a lot](https://github.com/guardian/deploy-tools-platform/search?q=%22new+GuPolicy%22) so the upgrade path is a bit more involved. I think this is ok though, we just need to provide more first party constructs.

PR #88, opened 2020-12-21
Clears the lib directory before build and releasing the latest

SHession:
## What does this change?

This PR ensures that old files are not released by clearing the lib directory before building and releasing the latest.


akash1810 (approved):

akash1810:
Automated deploys would have helped here as the chance of old artefacts being present between builds is reduced each build is run in a fresh environment.

PR #91, opened 2020-12-23
add an email sender policy

akash1810:
## What does this change?

Adds a new `GuSESSenderPolicy` construct which will:
- Add an `EmailSenderAddress` parameter to a `GuStack`
- Add an IAM policy to allow `ses:SendEmail` on that address

We have a few stacks that have this permission, each declare the policy in varying ways from [no restriction](https://github.com/guardian/anghammarad/blob/c919ebed337cb9be8653ac5ebf71ff2a6aa6990f/cloudformation/anghammarad.template.yaml#L40-L44) to [resource restricted](https://github.com/guardian/deploy-tools-platform/blob/1db44220323083ed89ef3fd03f8e1f27cfe2d390/cdk/lib/amiable/amiable.ts#L86-L92) to [condition restricted](https://github.com/guardian/editorial-tools-platform/blob/855ca76de244c9400c18a2ed9a34d428c99886fd/cloudformation/media-service-account/media-atom-maker/media-atom-maker.yaml#L786-L798).

After reading [this](https://docs.aws.amazon.com/ses/latest/DeveloperGuide/sending-authorization-policies.html), for our existing use cases it looks like a resource constraint and a condition amount to the same thing? So went with resource constraining as it was first in the docs.

Yes, it simplifies the AMIable stack.

## How to test

Run unit tests.

Providing more standard constructs and a standard way to declare this policy.

n/a

### DIFF
--- a/src/constructs/core/parameters.ts
+++ b/src/constructs/core/parameters.ts
@@ -105,3 +105,13 @@ export class GuS3ObjectArnParameter extends GuStringParameter {
     });
   }
 }
+
+export class GuGuardianEmailSenderParameter extends GuStringParameter {
+  constructor(scope: GuStack, id: string = "EmailSenderAddress", props?: GuNoTypeParameterProps) {
+    super(scope, id, {
+      ...props,
+      allowedPattern: RegexPattern.GUARDIAN_EMAIL,
+      constraintDescription: "Must be an @theguardian.com email address",
+    });
+  }
+}
--- a/src/constructs/core/parameters.ts
+++ b/src/constructs/core/parameters.ts
@@ -1,6 +1,6 @@
 import type { CfnParameterProps } from "@aws-cdk/core";
 import { CfnParameter } from "@aws-cdk/core";
-import { Stage, Stages } from "../../constants";
+import { RegexPattern, Stage, Stages } from "../../constants";
 import type { GuStack } from "./stack";
 
 export type GuParameterProps = CfnParameterProps;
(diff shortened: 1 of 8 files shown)


src/constructs/core/parameters.ts
```diff
@@ -1,6 +1,6 @@
 import type { CfnParameterProps } from "@aws-cdk/core";
 import { CfnParameter } from "@aws-cdk/core";
-import { Stage, Stages } from "../../constants";
+import { RegexPattern, Stage, Stages } from "../../constants";
```
jorgeazevedo, line 3: Encapsulation and unit tests ftw 👍

src/constructs/iam/policies/ses.ts
```diff
@@ -0,0 +1,22 @@
...
+  constructor(scope: GuStack, id: string = "GuSESSenderPolicy", props?: GuPolicyProps) {
+    super(scope, id, { ...props });
+
+    const emailSenderParam = new GuGuardianEmailSenderParameter(scope);
+
+    this.addStatements(
+      // see https://docs.aws.amazon.com/ses/latest/DeveloperGuide/sending-authorization-policies.html
+      new PolicyStatement({
+        effect: Effect.ALLOW,
+        actions: ["ses:SendEmail"],
```
jorgeazevedo, line 17: I remember there was one example from the datalake where they had both this and SendRawEmail. I was reading up on the difference in the [boto3 docs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ses.html) and SendRawEmail is what you need to use to send an email with an attachment, for example. Do you think it's overkill to include it here as well?

src/constructs/iam/policies/ses.test.ts
```diff
@@ -0,0 +1,48 @@
...
+    expect(stack).toHaveResource("AWS::IAM::Policy", {
+      PolicyDocument: {
+        Version: "2012-10-17",
+        Statement: [
+          {
+            Action: "ses:SendEmail",
+            Effect: "Allow",
+            Resource: {
+              "Fn::Join": [
+                "",
```
jorgeazevedo, line 20: This empty string is a cdk oddity isn't it? I was wondering if there was something to fix on our side, but it looks fine on our side

jorgeazevedo (approved):
I raise a question about SendRawEmail, but I don't think it's blocking and the rest of the implementation looks really solid

src/constructs/iam/policies/ses.ts
jorgeazevedo: Fair enough!

src/constructs/iam/policies/ses.test.ts
jorgeazevedo: Agreed, I don't think we should introduce a workaround for this.

akash1810: We might be able to use one of these functions https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_core.Arn.html.

PR #92, opened 2020-12-23
feat: add parameter helper functions on GuStack

akash1810:
## What does this change?

https://github.com/guardian/cdk/pull/65, but for parameters.

Helper functions to get/set parameters on the GuStack. An `Error` is thrown when getting a parameter that hasn't been set; this is to catch the error at compile time.

This is useful for when a construct reads from a parameter but isn't responsible for adding the parameter. For example, reading the `DistributionBucketName` param in the user data, where `DistributionBucketName` is added by [`GuGetDistributablePolicy`](https://github.com/guardian/cdk/blob/390a734dbfe228028581b172f1e920b3e223862e/src/constructs/iam/policies/s3-get-object.ts#L16) via [`GuInstanceRole`](https://github.com/guardian/cdk/blob/390a734dbfe228028581b172f1e920b3e223862e/src/constructs/iam/roles/instance-role.ts#L30). One such example can be seen [here](https://github.com/guardian/deploy-tools-platform/pull/375).

No.

## How to test

Observe CI.

It is simpler to share parameters across constructs.

n/a

### DIFF
--- a/src/constructs/core/stack.test.ts
+++ b/src/constructs/core/stack.test.ts
@@ -7,6 +7,7 @@ import { alphabeticalTags, simpleGuStackForTesting } from "../../../test/utils";
 import type { SynthedStack } from "../../../test/utils";
 import { Stage, Stages } from "../../constants";
 import { TrackingTag } from "../../constants/library-info";
+import { GuParameter } from "./parameters";
 import { GuStack } from "./stack";
 
 describe("The GuStack construct", () => {
@@ -62,5 +63,22 @@ describe("The GuStack construct", () => {
     const stack = new GuStack(new App(), "Test", { app: "MyApp", stack: "test" });
 
     expect(stack.app).toBe("MyApp");
+  });
+
+  it("should return a parameter that exists", () => {
+    const stack = new GuStack(new App(), "Test", { app: "MyApp", stack: "test" });
+    const testParam = new GuParameter(stack, "MyTestParam", {});
+    stack.setParam(testParam);
+
+    const actual = stack.getParam<GuParameter>("MyTestParam");
+    expect(actual).toBe(testParam);
+  });
+
+  it("should throw on attempt to get a parameter that doesn't exist", () => {
+    const stack = new GuStack(new App(), "Test", { app: "MyApp", stack: "test" });
+
+    expect(() => stack.getParam<GuParameter>("i-do-not-exist")).toThrowError(
+      "Attempting to read parameter i-do-not-exist which does not exist"
+    );
   });
 });
--- a/src/constructs/core/stack.ts
+++ b/src/constructs/core/stack.ts
@@ -40,7 +40,6 @@ export interface GuStackProps extends StackProps {
  * ```
  */
 export class GuStack extends Stack {
-  private readonly _stage: GuStageParameter;
   private readonly _stack: string;
   private readonly _app: string;
 
@@ -49,7 +48,7 @@ export class GuStack extends Stack {
   public readonly migratedFromCloudFormation: boolean;
 
   get stage(): string {
-    return this._stage.valueAsString;
+    return this.getParam(GuStageParameter.defaultId).valueAsString;
   }
 
   get stack(): string {
@@ -96,7 +95,7 @@ export class GuStack extends Stack {
 
     this.params = new Map<string, GuParameter>();
 
-    this._stage = new GuStageParameter(this);
+    new GuStageParameter(this);
     this._stack = props.stack;
     this._app = props.app;
 
--- a/src/constructs/core/stack.ts
+++ b/src/constructs/core/stack.ts
@@ -1,6 +1,7 @@
 import type { App, StackProps } from "@aws-cdk/core";
 import { Stack, Tags } from "@aws-cdk/core";
 import { TrackingTag } from "../../constants/library-info";
+import type { GuParameter } from "./parameters";
 import { GuStageParameter } from "./parameters";
 
 export interface GuStackProps extends StackProps {
@@ -43,6 +44,8 @@ export class GuStack extends Stack {
   private readonly _stack: string;
   private readonly _app: string;
 
+  private params: Map<string, GuParameter>;
+
   public readonly migratedFromCloudFormation: boolean;
 
   get stage(): string {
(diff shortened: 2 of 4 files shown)


src/constructs/core/stack.test.ts
```diff
@@ -63,4 +64,21 @@ describe("The GuStack construct", () => {
 
     expect(stack.app).toBe("MyApp");
   });
+
+  it("should return a parameter that exists", () => {
+    const stack = new GuStack(new App(), "Test", { app: "MyApp" });
+    const testParam = new GuParameter(stack, "MyTestParam", {});
+    stack.setParam(testParam);
```
jorgeazevedo, line 71: I'm a bit confused by the interface this is providing.

So is the idea for the client to call `stack.setParam` on parameters they want to make available via `stack.getParam`? I think I spot an opportunity to do the `setParam` in the constructor for `GuParameter` and make all of them available by default, but I suspect I'm actually misunderstanding the usecase

src/constructs/core/stack.test.ts
```diff
@@ -63,4 +64,21 @@ describe("The GuStack construct", () => {
...
+
+    const actual = stack.getParam<GuParameter>("MyTestParam");
```
jorgeazevedo, line 73: Probably a continuation of me misunderstanding the use case, but I'm seeing an opportunity to simplify the call to this getter by making it return GuParameter as a default, and leave the client to cast it if they need another type. Is getting a GuParameter sufficient for most cases?

```ts
const actual = stack.getParam("MyTestParam");
```

```ts
const actual = stack.getParam("MyTestParam") as GuSSMParameter;
```

jorgeazevedo (changes requested):

src/constructs/core/stack.test.ts
akash1810: > I think I spot an opportunity to do the setParam in the constructor for GuParameter and make all of them available by default

Yeah, that'd be brilliant actually!

> but I suspect I'm actually misunderstanding the usecase

I think I am too 😅 . This change was originally part of https://github.com/guardian/cdk/pull/91, but in that scenario it made more sense for the policy to bring its own parameter with it, rather than get it from the `GuStack`.

I think I misunderstood the API of parameters. I thought they were the same as tags where they had to be [explicitly added](https://github.com/guardian/cdk/blob/7d8fcb2bb0049f0bf3f92f0d36ce52ec6fd304f0/src/constructs/core/stack.ts#L29) to a stack, however it looks like simply declaring a parameter is enough to get it added to the stack.

I think this PR needs more thinking as it certainly feels useful, but not sure for what.

jorgeazevedo: Ok, cool! I've moved my review to just a comment rather than "request changes"

jorgeazevedo (commented):
Left some comments in the diff

jamie-lynch:
Definitely one for discussion. This is quite similar to something we had right at the start which we then moved away from. I like this general approach in theory but we didn't quite get it right the first time so resorted to just using `const parameters = { ... }` but if there's now another reason to go back to this then I'd be keen to explore it again.

akash1810:
I think I have a use case for this! 🎉 

A `UserData` construct will look something like this:

```
#!/bin/bash -ev
aws --region ${this.region} s3 cp s3://${distributablesBucket}/${this.stack}/${this.stage}/${this.app}/${this.app}.deb /tmp/
dpkg -i /tmp/${this.app}.deb
```

`distributablesBucket` is a parameter added to the stack from `GuGetDistributablePolicy`, and we'd need a way of accessing it from within the `UserData` construct.

akash1810:
Another use case is then you want to have multiple ASGs in a stack. Currently an ASG will bring an AMI parameter with it. This parameter is called "AMI" by default. If we try to add two ASGs to a stack, then we'll attempt to add the same parameter twice and an [exception is thrown](https://github.com/guardian/cdk/runs/1878351539?check_suite_focus=true). If we have a helper function to manage params, then we can do some validation.

I'll revive this branch today.

jacobwinch:
> Currently an ASG will bring an AMI parameter with it. This parameter is called "AMI" by default.

@akash1810 FYI I think it's possible to [override this behaviour](https://github.com/guardian/cdk/pull/213/files#diff-b258c1d87c022538e64b4609b0cef333e0f42ea12f0dff849a7bd644d69533fcR44-R63).

akash1810:
> @akash1810 FYI I think it's possible to [override this behaviour](https://github.com/guardian/cdk/pull/213/files#diff-b258c1d87c022538e64b4609b0cef333e0f42ea12f0dff849a7bd644d69533fcR44-R63).

Ah! I keep forgetting about this!

I don't think this'll allow a single AMI param to be used in multiple ASGs however as a [similar exception](https://github.com/guardian/cdk/runs/1878351539?check_suite_focus=true) noted above will be thrown?

Sharing AMIs across ASGs is fairly common and reasonable; here's an example from [Grid](https://github.com/guardian/editorial-tools-platform/blob/97bf1f94dfcd784f83defe97ab2b41bb793c5b7c/cloudformation/media-service-account/grid/media-service.yaml#L36-L46) the single param `AmiId` is used in 9 ASGs.

jacobwinch:
> I don't think this'll allow a single AMI param to be used in multiple ASGs however as a [similar exception](https://github.com/guardian/cdk/runs/1878351539?check_suite_focus=true) noted above will be thrown?

I haven't tested this, but I think this might work?

```
const mySharedAmi = new GuAmiParameter(stack, "MyAMI", {})

const myFirstAsg = new GuAutoScalingGroup(stack, "AutoscalingGroup", {
      // firstAsgProps
      imageId: mySharedAmi,
});

const mySecondAsg = new GuAutoScalingGroup(stack, "AutoscalingGroup", {
      // secondAsgProps
      imageId: mySharedAmi,
});
```

akash1810:
Oh lovely! Yeah that does work! The tests fail due to `InstanceType` being declared twice, but we can do a similar trick there too.

src/constructs/core/stack.ts
```diff
@@ -29,13 +32,27 @@ export class GuStack extends Stack {
...
+  }
+
   constructor(app: App, id: string, props: GuStackProps) {
     super(app, id, props);
 
     this._stage = new GuStageParameter(this);
     this._stack = new GuStackParameter(this);
     this._app = props.app;
 
+    this.params = new Map<string, GuParameter>();
```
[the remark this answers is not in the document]
akash1810: TypeScript has `private` and `public` keywords, which solves this in a way that doesn't rely on conventions. `this.params` is marked as `private`.

src/constructs/core/stack.test.ts
akash1810: This syntax is still possible. The default type of `getParam` is `GuParameter`. You can do both of these:

```typescript
const foo = scope.getParam("Foo") as GuAmiParameter;
const bar = scope.getParam<GuAmiParameter>("Bar");
```

Both `foo` and `bar` will be of type `GuAmiParameter`. Will rework the tests to make this more obvious.

akash1810: Hmm, the linter isn't happy with the assertion variant (see [this rule](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-unnecessary-type-assertion.md)), so the generics version is the only way to cast. I think?

jacobwinch (approved):

akash1810:
Merging w/out full approval as @jorgeazevedo is currently on leave.

PR #94, opened 2020-12-23
Make GuSESSenderPolicy available within the @guardian/cdk/lib/constru…

akash1810:
## What does this change?

Make GuSESSenderPolicy available within the @guardian/cdk/lib/constructs/iam module.

Just makes it easier to interact with.

This was missed in https://github.com/guardian/cdk/pull/91.

No.

## How to test

n/a

Importing goes from:

```ts
import { GuSESSenderPolicy } from "@guardian/cdk/lib/constructs/iam/policies/ses";
```

to:

```ts
import { GuSESSenderPolicy } from "@guardian/cdk/lib/constructs/iam
```

as standard with other constructs in the library.

n/a

### DIFF
--- a/src/constructs/iam/policies/index.ts
+++ b/src/constructs/iam/policies/index.ts
@@ -3,4 +3,5 @@ export * from "./describe-ec2";
 export * from "./log-shipping";
 export * from "./parameter-store-read";
 export * from "./s3-get-object";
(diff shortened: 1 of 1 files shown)


jorgeazevedo (approved):
Yep, agreed!

PR #96, opened 2020-12-24
Add DynamoDB read and write policies

akash1810:
## What does this change?

Adds two DynamoDB policies using a different (and possibly simpler) format from existing policy constructs.

In https://github.com/guardian/cdk/pull/77 we made `GuPolicy` abstract to discourage usage of it. In that PR we noted that if one wanted to create a custom policy they still could by extending `GuPolicy`. In this PR we make that process easier with the addition of `GuAllowPolicy` and `GuDenyPolicy` constructs.

These new constructs have a super simple interface and the dynamodb constructs added in this PR demonstrate their usage.

No, but it would mean a load of refactoring of existing constructs to extend from `GuAllowPolicy`.

## How to test

Observe CI. I've only added tests to the dynamo constructs as new policy constructs get tested implicitly as a result.

- A simpler interface to create policies
- A dynamodb read and write policy

n/a

### DIFF
--- a/src/constructs/iam/policies/dynamodb.ts
+++ b/src/constructs/iam/policies/dynamodb.ts
@@ -1,0 +1,32 @@
+import type { GuStack } from "../../core";
+import type { GuPolicyProps } from "./base-policy";
+import { GuAllowPolicy } from "./base-policy";
+
+interface GuDynamoDBPolicyProps {
+  tableName: string;
+}
+
+interface GuDynamoDBPolicyPropsWithActions extends GuPolicyProps, GuDynamoDBPolicyProps {
+  actions: string[];
+}
+
+abstract class GuDynamoDBPolicy extends GuAllowPolicy {
+  protected constructor(scope: GuStack, id: string, props: GuDynamoDBPolicyPropsWithActions) {
+    super(scope, id, {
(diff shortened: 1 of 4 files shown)


src/constructs/iam/policies/dynamodb.ts
```diff
@@ -0,0 +1,32 @@
...
+  protected constructor(scope: GuStack, id: string = "GuDynamoReadPolicy", props: GuDynamoDBPolicyPropsWithActions) {
+    super(scope, id, {
+      actions: props.actions.map((action) => `dynamodb:${action}`),
+      resources: [`arn:aws:dynamodb:${scope.region}:${scope.account}:table/${props.tableName}`],
+    });
+  }
+}
+
+export class GuDynamoDBReadPolicy extends GuDynamoDBPolicy {
+  constructor(scope: GuStack, id: string = "GuDynamoDBReadPolicy", props: GuDynamoDBPolicyProps) {
```
jamie-lynch, line 23: I think this will fail the linter check once it's rebased as the id can only have a default if the props are optional given that the default won't ever be used.

jamie-lynch (approved):
👍

PR #115, opened 2020-12-30
Add a custom eslint rule to check the constructors

jamie-lynch:
## What does this change?

This PR creates a custom ESLint rule which is used to check that the constructors of all constructs and patterns follow the best practice laid out in the [ADR](https://github.com/guardian/cdk/blob/main/docs/architecture-decision-records/002-component-constuctors.md).

Any constructs breaking those rules have been updated. In all cases except one, this change is non-breaking.

The InstanceRole pattern has had an `id` parameter added which means that any stacks implementing it will need to be updated.

Yes! The InstanceRole pattern has now had an `id` parameter added to conform to the style guide. This means that any stacks implementing it will need to be updated.

## How to test

Run the lint command to see if any errors are detected. Check a deliberate violation to see it catch things. 

The best practice agreed around component constructors is verified by an automated check rather than relying on peer reviews to catch any violations.

The rule will either fail to catch violations or detect numerous false positives.

### DIFF
--- a/eslint/rules/valid-constructors.js
+++ b/eslint/rules/valid-constructors.js
@@ -121,21 +121,21 @@ module.exports = {
             );
           }
 
-          // 10. Third parameter type should be custom
-          if (getStandardOrLeftProp(props, "typeAnnotation.typeAnnotation.type") !== "TSTypeReference") {
-            return context.report(
-              node,
-              getStandardOrLeftProp(props, "typeAnnotation.loc"),
-              `The third parameter in a construct or pattern contructor must be a custom type`
-            );
-          }
-
           // 8. If third parameter is present and non-optional then the second parameter should not be initialised
           if (props.type !== "AssignmentPattern" && !props.optional && id.type === "AssignmentPattern") {
             return context.report(
               node,
               id.loc,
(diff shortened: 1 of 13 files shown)


eslint/rules/valid-constructors.js
```diff
@@ -0,0 +1,147 @@
...
+          // 7. Third parameter (if exists) must called props
+          if (getStandardOrLeftProp(props, "name") !== "props") {
+            return context.report(
+              node,
+              props.loc,
+              `The third parameter in a construct or pattern contructor must be called props`
+            );
+          }
+
+          // 10. Third parameter type should be custom
```
jacobwinch, line 124: Minor: should we swap the order of 8 & 10?

jacobwinch (approved):
Disclaimer: I've never written a custom eslint rule before, but based on the docs and a quick local test this looks good to me 👍

eslint/rules/valid-constructors.js
jamie-lynch: Good point!

akash1810:
Oooh making the linter work for us, this is great!

PR #121, opened 2021-01-04
Add deletion protection by default

jamie-lynch:
## What does this change?

This PR sets the `deletionProtection` value to true by default for the `GuDatabaseInstance` and `GuApplicationLoadBalancer` constructs. 

This change updates a default value which will have an effect on existing projects (once the version of the library is bumped). For most projects, this new default value should be a sensible change however in some cases, it may be necessary to override it. 

## How to test

Two new unit tests cover the change. It could also be tested by applying it to a previously migrated stack.

Components which are risky to remove have deletion protection enabled by default.

There is a risk that this isn't a sensible default and either it is mistakenly set in many cases causing extra developer effort or it requires overriding in many cases.


stephengeller (approved):
Nice!

PR #126, opened 2021-01-05
Add GuClassicLoadBalancer

jamie-lynch:
## What does this change?

This PR adds a new `GuClassicLoadBalancer` component which extends the `LoadBalancer` construct provided by the `@aws-cdk/aws-elasticloadbalancing` library. 

It adds three main pieces of functionality:

* A boolean value can be passed to the `overrideId` prop to remove the 8 digit hash added to the end of the logical id
* An array of values can be provided to the `deletionOverrideProperties` prop which will be passed to the `addPropertyDeletionOverride` method on the underlying `CfnLoadBalancer`. An enum is defined containing the acceptable values. This allows for properties to be remove to match existing implementations more easily
* An object of property names and values can be passed to the `overrideProperties` prop which will each be passed to the `addPropertyOverride` method on the underlying `CfnLoadBalancer`. This allows properties to be overwritten more easily.

This is not a breaking change but any existing stacks using the `LoadBalancer` construct directly can be updated to make use of this new construct.

## How to test

Implement this construct in any stacks which are using the `LoadBalancer` construct directly and check that the snapshot tests still pass (indicating that nothing has changed)

Stacks using the `LoadBalancer` construct directly can now use this shared component and make use of the extra props provided.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "@aws-cdk/aws-autoscaling": "~1.74.0",
     "@aws-cdk/aws-ec2": "~1.74.0",
     "@aws-cdk/aws-apigateway": "~1.74.0",
+    "@aws-cdk/aws-elasticloadbalancing": "~1.74.0",
     "@aws-cdk/aws-elasticloadbalancingv2": "~1.74.0",
     "@aws-cdk/aws-events-targets": "~1.74.0",
     "@aws-cdk/aws-iam": "~1.74.0",
--- a/src/constructs/loadbalancing/clb.ts
+++ b/src/constructs/loadbalancing/clb.ts
@@ -1,0 +1,17 @@
+import type { CfnLoadBalancer, LoadBalancerProps } from "@aws-cdk/aws-elasticloadbalancing";
+import { LoadBalancer } from "@aws-cdk/aws-elasticloadbalancing";
+import type { GuStack } from "../core";
+
+interface GuClassicLoadBalancerProps extends LoadBalancerProps {
+  overrideId?: boolean;
+}
+
+export class GuClassicLoadBalancer extends LoadBalancer {
+  constructor(scope: GuStack, id: string, props: GuClassicLoadBalancerProps) {
+    super(scope, id, props);
+
+    const cfnLb = this.node.defaultChild as CfnLoadBalancer;
+
+    if (props.overrideId) cfnLb.overrideLogicalId(id);
+  }
+}
--- a/src/constructs/loadbalancing/clb.ts
+++ b/src/constructs/loadbalancing/clb.ts
@@ -2,16 +2,31 @@ import type { CfnLoadBalancer, LoadBalancerProps } from "@aws-cdk/aws-elasticloa
 import { LoadBalancer } from "@aws-cdk/aws-elasticloadbalancing";
 import type { GuStack } from "../core";
 
+enum DeletionOverrideProperties {
+  SCHEME = "Scheme",
+}
+
 interface GuClassicLoadBalancerProps extends LoadBalancerProps {
   overrideId?: boolean;
+  deletionOverrideProperties?: DeletionOverrideProperties[];
+  overrideProperties?: Record<string, unknown>;
 }
 
 export class GuClassicLoadBalancer extends LoadBalancer {
+  static DeletionOverrideProperties = DeletionOverrideProperties;
+
   constructor(scope: GuStack, id: string, props: GuClassicLoadBalancerProps) {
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@guardian/cdk",
-  "version": "0.10.0",
+  "version": "0.11.0",
```
jamie-lynch, line 3: Whoops! This is because I branched off of main locally but the release PR hasn't been merged yet. Will rebase once that's approved and merged.

src/constructs/loadbalancing/clb.ts
```diff
@@ -0,0 +1,32 @@
+import type { CfnLoadBalancer, LoadBalancerProps } from "@aws-cdk/aws-elasticloadbalancing";
+import { LoadBalancer } from "@aws-cdk/aws-elasticloadbalancing";
+import type { GuStack } from "../core";
+
+enum DeletionOverrideProperties {
```
jacobwinch, line 5: I'm not sure how much of the `cdk` jargon we're hoping to hide from users of this library, but `RemovableElbProperties` makes more sense to someone (like me!) who hasn't spent much time looking at `cdk` yet!

src/constructs/loadbalancing/clb.ts
```diff
@@ -0,0 +1,32 @@
...
+  SCHEME = "Scheme",
+}
+
+interface GuClassicLoadBalancerProps extends LoadBalancerProps {
+  overrideId?: boolean;
+  deletionOverrideProperties?: DeletionOverrideProperties[];
```
jacobwinch, line 11: Similarly, `propertiesToRemove` and `customProperties`/`bespokeProperties` would make a little more sense (to me) here.

src/constructs/loadbalancing/clb.ts
```diff
@@ -0,0 +1,32 @@
+import type { CfnLoadBalancer, LoadBalancerProps } from "@aws-cdk/aws-elasticloadbalancing";
+import { LoadBalancer } from "@aws-cdk/aws-elasticloadbalancing";
+import type { GuStack } from "../core";
+
+enum DeletionOverrideProperties {
+  SCHEME = "Scheme",
+}
+
+interface GuClassicLoadBalancerProps extends LoadBalancerProps {
+  overrideId?: boolean;
```
jacobwinch, line 10: I guess the reasons that we'd want to do this have been documented [here](https://github.com/guardian/cdk/discussions/111)? I wonder if we should add a link to that issue and/or some other short explanation in the code to help people who are migrating existing stacks?

jamie-lynch: Good point! Although this is something we've done in a lot of constructs so I'm not sure if we want to link from everywhere. There's also a doc about [migrating existing stacks](https://github.com/guardian/cdk/blob/main/docs/003-migrating-existing-stacks.md) that might be a good place for it to live. 

We should probably have a wider think about documentation as most of the constructs aren't that accessible right now to those who haven't worked on the library and/or don't have much prior CDK knowledge.

jacobwinch:
Presumably we are primarily adding support for classic load balancers in order to accommodate the migration of older stacks, which existed pre-ALB? Based on [this documentation](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/introduction.html#application-load-balancer-benefits), I assume that we would prefer new stacks to use an ALB. Do you think we need to add any docs to prevent developers from using an 'old' load balancer accidentally, or are we hoping to enforce best practices like this via patterns instead of via docs?

src/constructs/loadbalancing/clb.ts
jacobwinch: Ah great, thanks for sharing that - I agree that repeating this comment everywhere will be unhelpful! Feel free to disregard the original comment for now and we'll think about a more general solution for this later.

jacobwinch (approved):
Nice work 👍

jamie-lynch:
> this documentation

Yep! We did have a discussion about not providing constructs for 'old' components to encourage migration to the 'new' ones but (as fair a I remember) decided that this was more likely to just act as a blocker to people moving to CDK. Certainly our patterns should (by default at least) use an ALB over a CLB. We could add docs/a comment to express preference for using ALBs or even go as far as logging some sort of warning when synthesising a stack that implements a CLB?

Another discussion we've had which I'd like to explore more is adding a `GuLoadBalancer` construct which offers a consistent interface over both ALBs and CLBs and takes a prop to set which is used (defaulting to ALB). This would hopefully make migration from CLB to ALB easier!

PR #134, opened 2021-01-12
Add GuAssumeRolePolicy

jamie-lynch:
## What does this change?

This PR adds a new `GuAssumeRolePolicy` which takes `resources` as a prop. This reduces the amount of code required to provide assume role permissions. 

No changes are required but existing stacks where a Policy grants assume role permissions may wish to migrate to this. 

## How to test

Unit tests cover the new construct. Implementing it in place of a policy which grants this permissions in an existing stack shouldn't cause any changes when running the snapshot tests.

Fewer lines of code are required to define common policies.

### DIFF
--- a/src/constructs/iam/policies/assume-role.test.ts
+++ b/src/constructs/iam/policies/assume-role.test.ts
@@ -1,0 +1,51 @@
+import "@aws-cdk/assert/jest";
+import { attachPolicyToTestRole, simpleGuStackForTesting } from "../../../../test/utils";
+import { GuAssumeRolePolicy } from "./assume-role";
+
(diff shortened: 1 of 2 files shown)


stephengeller (approved):
Looks good! Sensible to have the option to override with `...props`, I'm guessing because this is an `AssumeRole` policy it'll always need to be `Effect.ALLOW` right?

jamie-lynch:
> I'm guessing because this is an `AssumeRole` policy it'll always need to be `Effect.ALLOW` right?

I guess in theory you _may_ want to allow a fairly wide pattern but explicitly deny one thing but I've only seen fairly simple cases so far. We can always add something later if we do find a case for it.

PR #135, opened 2021-01-12
Add migrated prop to GuStack

jamie-lynch:
## What does this change?

This PR adds a new `migrated`  prop to the `GuStack`. This prop can be set to true to denote that a stack has been migrated from a cloudformation template. Other constructs can then modify their behaviour where that is the case. 

Currently, the only implementation is in the `GuApplicationLoadBalancer`, `GuApplicationListener`, `GuApplicationTargetGroup`, `GuClassicLoadBalancer` and `GuDatabaseInstance` constructs. If the `migrated` value is true and the override id prop is _not_ false then the override id logic will be invoked. The overrideId prop has been left for cases outside of the default where the id still needs to be constant. Including the `&& props.overrideId !== false` condition means that it is also possible to explicitly override the default behaviour for a migrated stack if required. 

Credit to @jacobwinch for the idea! 👏 

No changes are required but migrated stacks can use the new prop in place of passing the `overrideId` prop to each construct.

## How to test

New units tests cover the new functionality. Trying adding the `migrated` prop to a migrated stack. Removing the relevant `overrideId` props should cause no change. 

Migrating existing stacks is simpler and requires fewer lines of code.

### DIFF
--- a/src/constructs/core/stack.ts
+++ b/src/constructs/core/stack.ts
@@ -6,7 +6,7 @@ export interface GuStackProps extends StackProps {
   // This limits GuStack to supporting a single app.
   // In the future, support for stacks with multiple apps may be required
   app: string;
-  migrated?: boolean;
+  migratedFromCloudFormation?: boolean;
 }
 
 export class GuStack extends Stack {
@@ -26,7 +26,7 @@ export class GuStack extends Stack {
     return this._app;
   }
 
-  migrated: boolean;
+  migratedFromCloudFormation: boolean;
 
   protected addTag(key: string, value: string, applyToLaunchedInstances: boolean = true): void {
     Tags.of(this).add(key, value, { applyToLaunchedInstances });
@@ -36,7 +36,7 @@ export class GuStack extends Stack {
   constructor(app: App, id: string, props: GuStackProps) {
     super(app, id, props);
 
-    this.migrated = !!props.migrated;
+    this.migratedFromCloudFormation = !!props.migratedFromCloudFormation;
 
     this._stage = new GuStageParameter(this);
     this._stack = new GuStackParameter(this);
--- a/src/constructs/core/stack.ts
+++ b/src/constructs/core/stack.ts
@@ -6,6 +6,7 @@ export interface GuStackProps extends StackProps {
   // This limits GuStack to supporting a single app.
   // In the future, support for stacks with multiple apps may be required
   app: string;
+  migrated?: boolean;
 }
 
 export class GuStack extends Stack {
@@ -25,6 +26,8 @@ export class GuStack extends Stack {
     return this._app;
   }
 
+  migrated: boolean;
+
   protected addTag(key: string, value: string, applyToLaunchedInstances: boolean = true): void {
     Tags.of(this).add(key, value, { applyToLaunchedInstances });
   }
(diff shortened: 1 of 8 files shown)


src/constructs/loadbalancing/elb.ts
```diff
@@ -38,7 +38,8 @@ export class GuApplicationTargetGroup extends ApplicationTargetGroup {
   constructor(scope: GuStack, id: string, props: GuApplicationTargetGroupProps) {
     super(scope, id, props);
 
-    if (props.overrideId) (this.node.defaultChild as CfnTargetGroup).overrideLogicalId(id);
+    if (props.overrideId || (scope.migrated && props.overrideId !== false))
+      (this.node.defaultChild as CfnTargetGroup).overrideLogicalId(id);
```
jamie-lynch, line 42: I'm not 100% whether the `GuApplicationTargetGroup` and `GuApplicationListenerProps` need to implement this or not.

jacobwinch: Neither am I! 

However, generally speaking when a team migrates I expect that they will preview the change set in the console. Lots of `Delete` lines are going to look pretty intimidating and make them question whether the migration is going to put their application at risk of suffering downtime.

Consequently I think it's best to keep the same resources during a migration wherever possible.

src/constructs/core/stack.ts
```diff
@@ -6,6 +6,7 @@ export interface GuStackProps extends StackProps {
   // This limits GuStack to supporting a single app.
   // In the future, support for stacks with multiple apps may be required
   app: string;
+  migrated?: boolean;
```
jacobwinch, line 9: Minor: perhaps we want to be more specific (`migratedFromCloudFormation`?) in case there are other migration scenarios in the future?

jacobwinch (approved):
> The overrideId prop has been left for cases outside of the default where the id still needs to be constant. Including the && props.overrideId !== false condition means that it is also possible to explicitly override the default behaviour for a migrated stack if required.

I wonder if we could simplify things by removing `overrideId` from `props` entirely. I can't think of a particular use-case for this other than a migration, but perhaps you had something in mind?

Overall though, this looks good to me 👍 🚢

src/constructs/loadbalancing/elb.ts
jamie-lynch: Good point. Let's leave this then.

src/constructs/core/stack.ts
jamie-lynch: 👍

jamie-lynch:
> > The overrideId prop has been left for cases outside of the default where the id still needs to be constant. Including the && props.overrideId !== false condition means that it is also possible to explicitly override the default behaviour for a migrated stack if required.
> 
> I wonder if we could simplify things by removing `overrideId` from `props` entirely. I can't think of a particular use-case for this other than a migration, but perhaps you had something in mind?
> 
> Overall though, this looks good to me 👍 🚢

We've had at least one case where something was referred to outside of the stack by its logicalId and needed to be kept the same. However, on the whole I think we should avoid using it wherever possible. I guess longer term we should maybe look at improving how that process works but for now I think the easiest thing is to keep the functionality.

jacobwinch:
> for now I think the easiest thing is to keep the functionality

Agreed, thanks for explaining this though!

PR #137, opened 2021-01-13
update GuLogShippingPolicy to bring it's own parameter

akash1810:
## What does this change?

We've established best practice for accounts/stacks to define the kinesis stream to send logs to to be a SSM Parameter called `/account/services/logging.stream.name`.

This updates `GuLogShippingPolicy` to add the parameter to the stack within that construct. The construct is bringing its own resources with it, rather than being passed in as props. This makes it more self contained.

This builds on the BYO* pattern introduced in https://github.com/guardian/cdk/pull/91.

Yes! The signature of `InstanceRole` has changed to no longer require a stream name, but simply a `includeLogShipping` boolean.

## How to test

- Observe CI?

Simpler interface to `InstanceRole` and more self contained constructs.

n/a

### DIFF
--- a/src/patterns/instance-role.ts
+++ b/src/patterns/instance-role.ts
@@ -11,7 +11,7 @@ import {
 } from "../constructs/iam";
 
 interface InstanceRoleProps extends GuGetS3ObjectPolicyProps {
-  includeLogShipping: boolean;
+  withoutLogShipping?: boolean; // optional to have log shipping added by default, you have to opt out
   additionalPolicies?: GuPolicy[];
 }
 
@@ -30,7 +30,7 @@ export class InstanceRole extends GuRole {
       new GuGetS3ObjectPolicy(scope, "GetDistributablesPolicy", props),
       new GuDescribeEC2Policy(scope),
       new GuParameterStoreReadPolicy(scope),
-      ...(props.includeLogShipping ? [new GuLogShippingPolicy(scope)] : []),
+      ...(props.withoutLogShipping ? [] : [new GuLogShippingPolicy(scope)]),
       ...(props.additionalPolicies ? props.additionalPolicies : []),
     ];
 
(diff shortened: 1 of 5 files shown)


src/patterns/instance-role.ts
```diff
@@ -10,7 +10,8 @@ import {
   GuSSMRunCommandPolicy,
 } from "../constructs/iam";
 
-interface InstanceRoleProps extends GuGetS3ObjectPolicyProps, Partial<GuLogShippingPolicyProps> {
+interface InstanceRoleProps extends GuGetS3ObjectPolicyProps {
+  includeLogShipping: boolean;
```
akash1810, line 14: Not wedded to this name; `withLogShipping` was also considered.

jamie-lynch, line 14: Would it be worth providing a default value for this?

jamie-lynch (approved):

src/patterns/instance-role.ts
akash1810: Yes! Not sure how to achieve this in a nice way though. Maybe relying on javascript's ~insane~ logic of `undefined` being falsy and inverting the prop name?

```typescript
interface InstanceRoleProps {
  withoutLogShipping? boolean
}
```

Then the ternary becomes `...(props.withoutLogShipping ? [] : [new GuLogShippingPolicy(scope)])`.

That is, make it opt out.

jamie-lynch: Hmm... yeah - I was thinking of spreading the props with a default `{withLogShipping: true, ...props}` although I'm not sure where you'd do that. Maybe it'd need to be a new `const` before the `this.policies` line which isn't great. 

Having it as `withoutLogShipping` makes it more obvious to reason about what the default behaviour is though, I like it!

src/patterns/instance-role.ts
```diff
@@ -10,7 +10,8 @@ import {
   GuSSMRunCommandPolicy,
 } from "../constructs/iam";
 
-interface InstanceRoleProps extends GuGetS3ObjectPolicyProps, Partial<GuLogShippingPolicyProps> {
+interface InstanceRoleProps extends GuGetS3ObjectPolicyProps {
+  withoutLogShipping?: boolean; // optional to have log shipping added by default, you have to opt out
```
akash1810, line 14: We might want to start using https://typedoc.org/ to generate a documentation site and also provide hints via IDEs IntelliSense features.

src/patterns/instance-role.ts
akash1810: Updated now.

PR #139, opened 2021-01-13
Create GuGetDistributablePolicy construct

akash1810:
## What does this change?

The changes in https://github.com/guardian/cdk/pull/137 were quite nice, this PR applies the same methodology to the policy that allows GETting a object from S3.

Creates a new `GuGetDistributablePolicy` construct to create an IAM policy allowing S3 GET from a distribution bucket. This construct also brings with it a SSM Parameter for the name of the bucket, defaulting to `/account/services/artifact.bucket`. This simplifies the props for `InstanceRole`, removing `bucketName`.

This PR also changes `GuGetS3ObjectPolicy` to extend from `GuAllowPolicy` which yields simpler code.

Yes. Now an `InstanceRole` is simpler to instantiate (`new InstanceRole(scope)` is sufficient).

## How to test

Observe CI.

Simpler interfaces.

n/a

### DIFF
--- a/src/constructs/iam/policies/s3-get-object.test.ts
+++ b/src/constructs/iam/policies/s3-get-object.test.ts
@@ -1,4 +1,6 @@
 import "@aws-cdk/assert/jest";
+import { SynthUtils } from "@aws-cdk/assert/lib/synth-utils";
+import type { SynthedStack } from "../../../../test/utils";
 import { attachPolicyToTestRole, simpleGuStackForTesting } from "../../../../test/utils";
 import { GuGetDistributablePolicy, GuGetS3ObjectPolicy } from "./s3-get-object";
 
@@ -51,6 +53,19 @@ describe("The GuGetDistributablePolicy construct", () => {
   it("creates the correct policy", () => {
     const stack = simpleGuStackForTesting();
     attachPolicyToTestRole(stack, new GuGetDistributablePolicy(stack));
+
+    const json = SynthUtils.toCloudFormation(stack) as SynthedStack;
+
+    const parameterKeys = Object.keys(json.Parameters);
+    const expectedKeys = ["Stage", "Stack", "DistributionBucketName"];
+    expect(parameterKeys).toEqual(expectedKeys);
+
+    expect(json.Parameters.DistributionBucketName).toEqual({
+      Default: "/account/services/artifact.bucket",
+      Description: "SSM parameter containing the S3 bucket name holding distribution artifacts",
+      NoEcho: true,
+      Type: "AWS::SSM::Parameter::Value<String>",
+    });
 
     expect(stack).toHaveResource("AWS::IAM::Policy", {
       PolicyName: "GetDistributablePolicyC6B4A871",
--- a/src/constructs/iam/policies/s3-get-object.test.ts
+++ b/src/constructs/iam/policies/s3-get-object.test.ts
@@ -1,6 +1,6 @@
 import "@aws-cdk/assert/jest";
 import { attachPolicyToTestRole, simpleGuStackForTesting } from "../../../../test/utils";
-import { GuGetS3ObjectPolicy } from "./s3-get-object";
+import { GuGetDistributablePolicy, GuGetS3ObjectPolicy } from "./s3-get-object";
 
 describe("The GuGetS3ObjectPolicy class", () => {
   it("sets default props", () => {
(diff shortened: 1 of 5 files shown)


src/constructs/iam/policies/s3-get-object.ts
```diff
@@ -1,26 +1,25 @@
...
-export class GuGetS3ObjectPolicy extends GuPolicy {
+export class GuGetS3ObjectPolicy extends GuAllowPolicy {
   constructor(scope: GuStack, id: string, props: GuGetS3ObjectPolicyProps) {
-    super(scope, id, { ...allowGetObjectPolicyStatement(props.bucketName), ...props });
+    super(scope, id, { ...props, actions: ["s3:GetObject"], resources: [`arn:aws:s3:::${props.bucketName}/*`] });
+  }
+}
+
+export class GuGetDistributablePolicy extends GuGetS3ObjectPolicy {
+  constructor(scope: GuStack, id: string = "GetDistributablePolicy", props?: GuPolicyProps) {
```
jamie-lynch, line 17: Maybe this should be `GuNoStatementsPolicyProps`? And should `GuGetS3ObjectPolicyProps` extend that instead too?

src/constructs/iam/policies/s3-get-object.test.ts
```diff
@@ -46,3 +46,35 @@ describe("The GuGetS3ObjectPolicy class", () => {
     });
   });
 });
+
+describe("The GuGetDistributablePolicy construct", () => {
```
jamie-lynch, line 50: Is it worth testing that the correct parameter is added too?

src/patterns/instance-role.ts
```diff
@@ -1,24 +1,24 @@
...
+interface InstanceRoleProps {
   withoutLogShipping?: boolean; // optional to have log shipping added by default, you have to opt out
   additionalPolicies?: GuPolicy[];
 }
 
 export class InstanceRole extends GuRole {
   private policies: GuPolicy[];
 
-  constructor(scope: GuStack, id: string, props: InstanceRoleProps) {
+  constructor(scope: GuStack, id: string = "InstanceRole", props?: InstanceRoleProps) {
```
jamie-lynch, line 21: This is a nice side effect!

src/constructs/iam/policies/s3-get-object.ts
akash1810: Yeah wasn't sure what to use here. I don't think `GuGetS3ObjectPolicyProps` would be correct as it has a `bucketName` property, which we don't want the user to set as it comes from the parameter.

src/constructs/iam/policies/s3-get-object.test.ts
akash1810: Yep! Looks like this can only be done via snapshot testing as there's no `toHaveParameter` assertion available.

akash1810: Done.

src/constructs/iam/policies/s3-get-object.ts
jamie-lynch: I've done something along the lines of this elsewhere which might work here too

```ts

expect(json.Parameters.DistributionBucketName).toEqual({
  Default: "/account/services/artifact.bucket",
  Description: "SSM parameter containing the S3 bucket name holding distribution artifacts",
  NoEcho: true,
  Type: "AWS::SSM::Parameter::Value<String>",
});
```

akash1810: We could also remove `props` entirely? The only thing we're currently passing through is [`policyName`](https://github.com/guardian/cdk/blob/d240cb1587ec02d0ab746f9c72403e185ba380b0/src/constructs/iam/policies/s3-get-object.test.ts#L30) and I think the autogenerated name would be fine here.

jamie-lynch: That's true... although if we do that, let's do it everywhere for consistency. I'm somewhat hesitant about removing it but I don't actually have any valid reasons. As you say, `policyName` is the only other prop that we've used so far and I agree that the auto-generate name should be fine.

src/constructs/iam/policies/s3-get-object.test.ts
jamie-lynch: Moving this comment here as I managed to post it on the wrong thread the first time 🤦 

I've done something along the lines of this elsewhere which might work here too

```ts

expect(json.Parameters.DistributionBucketName).toEqual({
  Default: "/account/services/artifact.bucket",
  Description: "SSM parameter containing the S3 bucket name holding distribution artifacts",
  NoEcho: true,
  Type: "AWS::SSM::Parameter::Value<String>",
});
```

akash1810: Have implemented this style of test now.

src/constructs/iam/policies/s3-get-object.ts
akash1810: Lets go with `GuNoStatementsPolicyProps` for now then.

jamie-lynch (approved):
🚢

PR #140, opened 2021-01-14
export the assume-role construct

akash1810:
## What does this change?

This just makes it simpler to import as it becomes available on `@guardian/cdk/lib/constructs/iam`. This is similar to https://github.com/guardian/cdk/pull/94 - not sure how to catch this "issue" early?

https://github.com/guardian/deploy-tools-platform/pull/322 would need updating.

## How to test

n/a

Simpler imports.

n/a


jamie-lynch (approved):
Whoops - good spot! Yeah, would be good to have a way of catching this.

PR #141, opened 2021-01-14
Allow all parameters to be from SSM

jamie-lynch:
## What does this change?

This PR updates the `GuParameter` to take a new `fromSSM` prop. If this prop is set to true then the `type` value will be wrapped with `AWS::SSM::Parameter::Value`. This means that all parameters extending the `GuParameter` can now be denoted as `fromSSM`. Due to the, the `GuSSMParameter` has been removed and the one case where it was used has been updated.

Yes. Anywhere that is currently using the `GuSSMParameter` will need to be updated to use another construct with the `fromSSM` value set to true.

## How to test

New unit tests have been added to cover the exact functionality. The `GuLogShippingPolicy` was using `GuSSMParameter` and has been refactored so you can also see that it still produces the same cloudformation.

GuParameter constructs present a more consistent interface with regard to getting values from SSM.

### DIFF
--- a/src/constructs/core/parameters.test.ts
+++ b/src/constructs/core/parameters.test.ts
@@ -8,8 +8,8 @@ import {
   GuAmiParameter,
   GuArnParameter,
   GuInstanceTypeParameter,
+  GuParameter,
   GuS3ObjectArnParameter,
-  GuSSMParameter,
   GuStackParameter,
   GuStageParameter,
   GuStringParameter,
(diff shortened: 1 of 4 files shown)


src/constructs/core/parameters.test.ts
```diff
@@ -18,6 +18,57 @@ import {
...
+
+    expect(json.Parameters.Parameter).toEqual({
+      Type: "AWS::SSM::Parameter::Value<String>",
+    });
+  });
+
+  it("passes through other values without modification", () => {
+    const stack = simpleGuStackForTesting();
+
+    new GuParameter(stack, "Parameter", { type: "Boolean", fromSSM: true, description: "This is a test" });
```
akash1810, line 61: We might want to make `type` an enum or validate it at some point, for a later PR though.

akash1810 (approved):

src/constructs/iam/policies/log-shipping.ts
```diff
@@ -1,16 +1,18 @@
...
 export class GuLogShippingPolicy extends GuPolicy {
   constructor(scope: GuStack, id: string = "GuLogShippingPolicy", props?: GuPolicyProps) {
     super(scope, id, { ...props });
 
-    const loggingStreamNameParam = new GuSSMParameter(scope, "LoggingStreamName", {
+    const loggingStreamNameParam = new GuStringParameter(scope, "LoggingStreamName", {
       description: "SSM parameter containing the Name (not ARN) on the kinesis stream",
       default: "/account/services/logging.stream.name",
+      fromSSM: true,
+      noEcho: true,
```
akash1810, line 15: Suggestion from https://github.com/guardian/cdk/pull/143.

```suggestion
```

PR #143, opened 2021-01-14
echo GuSSMParameter values by default

akash1810:
## What does this change?

The value of GuSSMParameter will be the path into parameter store. This isn't private, so setting noEcho should not be the default.

I observed this whilst applying https://github.com/guardian/deploy-tools-platform/pull/323. The value of `LoggingStreamName` is defaulted and with `noEcho` set to `true`, it caused for an interesting CFN update as had to triple check the redacted value.

It is still possible to set `noEcho` and have added a test to demonstrate this.

No.

## How to test

Tests have been added.

Better default behaviour of the `GuSSMParameter`.

n/a

### DIFF
--- a/src/constructs/core/parameters.test.ts
+++ b/src/constructs/core/parameters.test.ts
@@ -110,6 +110,19 @@ describe("The GuSSMParameter class", () => {
     const json = SynthUtils.toCloudFormation(stack) as SynthedStack;
 
     expect(json.Parameters.Parameter).toEqual({
+      Type: "AWS::SSM::Parameter::Value<String>",
+      Description: "This is a test",
+    });
+  });
+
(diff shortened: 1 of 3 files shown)


jamie-lynch:
The `GuSSMParameter` is removed in https://github.com/guardian/cdk/pull/141 so this problem goes away. In that PR I updated one place where it was used to set `noEcho` to true explicitly which maybe can be removed?

akash1810:
> The `GuSSMParameter` is removed in #141 so this problem goes away. In that PR I updated one place where it was used to set `noEcho` to true explicitly which maybe can be removed

Ah, good point! Yeah, removing the explicit `noEcho` would be preferred.

PR #149, opened 2021-01-15
Add BYO pattern to GuVpc

jamie-lynch:
## What does this change?

This PR adds a new `fromIdParameter` method to the `GuVpc` class. This method will create a new parameter for the VpcId and use this to return the reference to the VPC to be used throughout the stack.

No changes are required but existing stack may migrate to the new method to reduce the amount of code required.

## How to test

Migrate an existing stack to the method and run the snapshot tests.

Fewer parameters have to be defined manually.

### DIFF
--- a/src/constructs/ec2/vpc.ts
+++ b/src/constructs/ec2/vpc.ts
@@ -21,7 +21,7 @@ export class GuVpc {
     });
   }
 
-  static fromIdParameter(scope: GuStack, id: string, props: VpcFromIdParameterProps): IVpc {
+  static fromIdParameter(scope: GuStack, id: string, props?: VpcFromIdParameterProps): IVpc {
     const vpc = new GuVpcParameter(scope, "VpcId", {
       description: "Virtual Private Cloud to run EC2 instances within",
       default: "/account/services/default.vpc",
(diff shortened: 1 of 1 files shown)


akash1810 (approved):

src/constructs/ec2/vpc.ts
```diff
@@ -1,17 +1,33 @@
...
-      vpcId: vpcId,
       availabilityZones: scope.availabilityZones,
       ...props,
     });
   }
+
+  static fromIdParameter(scope: GuStack, id: string, props?: VpcFromIdParameterProps): IVpc {
+    const vpc = new GuVpcParameter(scope, "VpcId", {
+      description: "Virtual Private Cloud to run EC2 instances within",
+      default: "/account/services/default.vpc",
```
katebee, line 27: this might be a good opportunity to discourage use of the default VPC?

akash1810: Ah, good idea! We're intending the value of this param to be a non-default VPC - maybe a better name is `/account/services/primary.vpc`?

sihil: I'm guilty of using default here to indicate the main VPC rather than the default in the AWS sense. `primary` is much better.

I would argue that VPCs warrant their own sub key. How about something like:
```
/account/vpc/primary/id
/account/vpc/primary/subnets/public
/account/vpc/primary/subnets/private
/account/vpc/primary/egress/ips
...
```

jamie-lynch: Ah yes, that's nice. Then we can make add a similar method around subnets too.

PR #151, opened 2021-01-15
update the GH Action trigger

akash1810:
## What does this change?

Following https://github.com/guardian/cdk/pull/148, it looks like releases and tags are different concepts in GitHub, even though, confusingly, tags show up on https://github.com/guardian/cdk/releases.

It looks like our release process is currently only creating tags and NOT creating releases. As a result, this change tweaks the GH Action to update docs on _tag_ creation.

We'll likely add semantic-release to this library, which does in fact create releases, but this is a little far away.

No.

## How to test

n/a

Docs are updated automatically when a tag is created.

n/a

### DIFF
--- a/.github/workflows/cd-docs.yaml
+++ b/.github/workflows/cd-docs.yaml
@@ -1,7 +1,8 @@
-name: Deploy Documentation on new release
+name: Deploy documentation on new tag
 on:
-  release:
-    types: [created]
+  create:
+    tags:
+      - 'v*'
 jobs:
   validate:
     runs-on: ubuntu-latest


jamie-lynch (approved):
🏷️

.github/workflows/cd-docs.yaml
```diff
@@ -1,7 +1,8 @@
-name: Deploy Documentation on new release
+name: Deploy documentation on new tag
 on:
-  release:
-    types: [created]
+  create:
+    tags:
+      - v*
```
jacobwinch, line 5: I think this might need to be `- 'v*'` (based on https://github.community/t/how-to-run-github-actions-workflow-only-for-new-tags/16075/10).

akash1810: Updated.

PR #157, opened 2021-01-15
Update VPC default path

jamie-lynch:
## What does this change?

This PR updates the default path of the VPC parameter created by the `fromIdParameter` to be `/account/services/primary.vpc` to discourage the use of the default VPC. 

See thread here: https://github.com/guardian/cdk/pull/149#discussion_r558456087

Any stacks using this method already will have the ensure the value is set at the new path. Although, to my knowledge, nothing is using it yet.

## How to test

Implement this method in an existing stack and view that changes by running the snapshot test.

People are discouraged from using the default VPC.

### DIFF
--- a/src/constructs/ec2/vpc.ts
+++ b/src/constructs/ec2/vpc.ts
@@ -24,7 +24,7 @@ export class GuVpc {
   static fromIdParameter(scope: GuStack, id: string, props?: VpcFromIdParameterProps): IVpc {
     const vpc = new GuVpcParameter(scope, "VpcId", {
       description: "Virtual Private Cloud to run EC2 instances within",
-      default: "/account/services/default.vpc",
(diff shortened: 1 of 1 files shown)


akash1810 (approved):
Have been thinking it would be good to document what we're suggesting an account looks like and why. Maybe one to ADR or add to the documentation side soon?

jamie-lynch:
> Have been thinking it would be good to document what we're suggesting an account looks like and why. Maybe one to ADR or add to the documentation side soon?

Yeah, good point! Maybe one for the docs plus getting the pages plugin working so that we can include them on gh pages.

jamie-lynch:
Closing in favour of vpcs having their own path (see https://github.com/guardian/cdk/pull/149#discussion_r558843564)

PR #184, opened 2021-01-20
Add default health check values for CLB

jamie-lynch:
## What does this change?

This PR introduces a static `DefaultHealthCheck` object to the `GuClassicLoadBalancer` which is merged with the provided props. This allows either the entire object or individual values to be omitted (see new test cases for examples). 

No changes are required however stacks that implement the `GuClassicLoadBalancer` can reduce the number of props they pass in. 

## How to test

Upgrade an existing stack and remove an healthCheck values which are the same as the default value. The snapshot tests should still pass, confirming that nothing in the output has changed.

Fewer lines of code are required to define load balancers and best practice around CLB health checks comes for free.

### DIFF
--- a/src/constructs/loadbalancing/clb.ts
+++ b/src/constructs/loadbalancing/clb.ts
@@ -1,33 +1,50 @@
-import type { CfnLoadBalancer, LoadBalancerProps } from "@aws-cdk/aws-elasticloadbalancing";
-import { LoadBalancer } from "@aws-cdk/aws-elasticloadbalancing";
+import type { CfnLoadBalancer, HealthCheck, LoadBalancerProps } from "@aws-cdk/aws-elasticloadbalancing";
+import { LoadBalancer, LoadBalancingProtocol } from "@aws-cdk/aws-elasticloadbalancing";
+import { Duration } from "@aws-cdk/core";
 import type { GuStack } from "../core";
 
 enum RemoveableProperties {
   SCHEME = "Scheme",
 }
 
-interface GuClassicLoadBalancerProps extends LoadBalancerProps {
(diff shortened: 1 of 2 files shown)


src/constructs/loadbalancing/clb.ts
```diff
@@ -1,33 +1,50 @@
...
+    interval: Duration.seconds(30),
+    timeout: Duration.seconds(10),
+  };
+
   constructor(scope: GuStack, id: string, props: GuClassicLoadBalancerProps) {
-    super(scope, id, props);
+    const mergedProps = {
+      ...props,
+      healthCheck: { ...GuClassicLoadBalancer.DefaultHealthCheck, ...props.healthCheck },
+    };
```
jamie-lynch, line 34: Although we don't use the healthCheck props outside of the `super` call, doing this merge there means that we have a different props object passed to the parent constructor compared to the one available in this constructor which is a massive gotcha.

src/constructs/loadbalancing/clb.ts
```diff
@@ -1,33 +1,50 @@
...
 
+  static DefaultHealthCheck = {
+    port: 9000,
+    path: "/healthcheck",
+    protocol: LoadBalancingProtocol.HTTP,
+    healthyThreshold: 2,
+    unhealthyThreshold: 5,
+    interval: Duration.seconds(30),
+    timeout: Duration.seconds(10),
+  };
```
jamie-lynch, line 28: These seemed like sensible defaults at first pass but keen to hear any opinions on what these values should be.

akash1810 (approved):

PR #186, opened 2021-01-21
Add pattern for scheduled lambdas

jacobwinch:
## What does this change?

This PR is a first attempt at a scheduled lambda pattern. It also introduces a construct for CloudWatch alarms, and a pattern for a lambda-specific CloudWatch alarms, based on error %. 

I've introduced the CloudWatch-y stuff at the same time as the scheduled lambda pattern because a number of examples of this pattern 'in the wild' also include an alert. I think we should be encouraging developers to think about monitoring their software whenever they add new compute resources, so hopefully this will help to do that.

No changes are required, however stacks that use schedule-based lambdas can simplify their infrastructure definition slightly. 

## How to test

1. Upgrade an existing stack to use this new pattern. The snapshot tests should show 0 (or trivial) changes.
2. Create a new stack based on this pattern. You should end up with a functional scheduled lambda.
3. Configure monitoring and force the lambda to fail X% of the time to ensure that the new alert pattern works as expected.

I've tested 2 & 3 in the `developerPlayground` account.

* Projects already using this library (e.g. `tag-janitor` and `pagerduty-summary`) should be able to simplify their infrastructure definition slightly
* New projects should get sensible monitoring for a tiny amount of extra code/effort

* This is all new functionality; merging this PR should not have an impact on any projects currently using this library.


src/patterns/lambda-alarms.ts
```diff
@@ -0,0 +1,32 @@
+import { ComparisonOperator, MathExpression } from "@aws-cdk/aws-cloudwatch";
+import type { GuAlarmProps } from "../constructs/cloudwatch/alarm";
+import { GuAlarm } from "../constructs/cloudwatch/alarm";
+import type { GuStack } from "../constructs/core";
+import type { GuLambdaFunction } from "../constructs/lambda";
+
+export interface ErrorPercentageMonitoring
+  extends Omit<GuAlarmProps, "metric" | "threshold" | "comparisonOperator" | "evaluationPeriods"> {
+  toleratedErrorPercentage: number;
+  numberOfFiveMinutePeriodsToEvaluate?: number;
```
jacobwinch, line 10: Better suggestions for this name are welcome!

src/patterns/scheduled-lambda.ts
```diff
@@ -0,0 +1,27 @@
+import type { Schedule } from "@aws-cdk/aws-events";
+import type { GuStack } from "../constructs/core";
+import { GuLambdaFunction } from "../constructs/lambda";
+import type { GuFunctionProps } from "../constructs/lambda";
+import type { ErrorPercentageMonitoring } from "./lambda-alarms";
+import { GuLambdaErrorPercentageAlarm } from "./lambda-alarms";
+
+interface GuScheduledLambdaProps extends Omit<GuFunctionProps, "rules" | "apis"> {
+  schedule: Schedule;
+  monitoringConfiguration?: ErrorPercentageMonitoring;
```
jacobwinch, line 10: I wondered about making the opt-out for this more explicit. In Scala I would do something like:

```
sealed trait LambdaMonitoring
case object NoMonitoring extends LambdaMonitoring
case class ErrorPercentageMonitoring(toleratedErrorPercentage: Int, snsTopicArn: String, numberOfFiveMinutePeriodsToEvaluate: Option[Int]=Some(5)) extends LambdaMonitoring
```

The interface would then be:

```
monitoringConfiguration: LambdaMonitoring
```

So to opt-out, you'd have to state:

```monitoringConfiguration: NoMonitoring```

This would make it clearer (and highlight to reviewers) that you were intentionally omitting this (hopefully for a good reason e.g. you prefer some different alerting system to CloudWatch), rather than just being (another) optional property that can be easily forgotten.

Any opinions on whether we'd want to do this? If so, any suggestions on how we could go about this in TypeScript? I tried a couple of options using [Unions](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#discriminating-unions) but they were all pretty ugly!

File: src/patterns/lambda-alarms.ts [hunk 3]
```diff
@@ -0,0 +1,32 @@
+import { ComparisonOperator, MathExpression } from "@aws-cdk/aws-cloudwatch";
```
jacobwinch, line 1: I wasn't 100% sure whether this should be a `pattern` or a `construct` - any feedback on this would be helpful.

src/constructs/cloudwatch/alarm.ts
```diff
@@ -0,0 +1,18 @@
+import type { AlarmProps } from "@aws-cdk/aws-cloudwatch";
+import { Alarm } from "@aws-cdk/aws-cloudwatch";
+import { SnsAction } from "@aws-cdk/aws-cloudwatch-actions";
+import type { ITopic } from "@aws-cdk/aws-sns";
+import { Topic } from "@aws-cdk/aws-sns";
+import type { GuStack } from "../core";
+
+export interface GuAlarmProps extends AlarmProps {
+  snsTopicArn: string;
```
jacobwinch, line 9: Most teams I've worked on have 1 or 2 topics configured for all alerts and then pass the `arn` into various templates as parameters.

We could also consider taking the topic name (which is not private) instead of the full `arn` (this contains an AWS account number, which should be kept private), to simplify things slightly?

jamie-lynch: Yeah, it looks like that's what the [GuLogShippingPolicy](https://github.com/guardian/cdk/blob/964b73aa6a8f928a4d83de8427ae42fa7b825bd5/src/constructs/iam/policies/log-shipping.ts#L21) does. Then it gets the region and account from `scope`

src/patterns/lambda-alarms.ts
jamie-lynch: My first impression is that this should be in constructs as patterns should essentially encompass an entire stack but interested to hear other people's opinions too

src/patterns/scheduled-lambda.ts
jamie-lynch: Yeah, that would be nice. I don't know of a nice way of doing that in typescript although I often feel like there must be one that I'm missing.

src/constructs/cloudwatch/alarm.ts
akash1810: Is it worth having a single topic, set in param store, for use by multiple stacks? Or is the separation of concerns important?

src/constructs/lambda/lambda.ts
```diff
@@ -12,7 +12,7 @@ interface ApiProps extends Omit<LambdaRestApiProps, "handler"> {
   id: string;
 }
 
-interface GuFunctionProps extends Omit<FunctionProps, "code"> {
+export interface GuFunctionProps extends Omit<FunctionProps, "code"> {
   code: { bucket: string; key: string };
```
akash1810, line 16: Possibly not for this PR, but would be good for the `bucket` to come for free, [GuGetDistributablePolicy](https://github.com/guardian/cdk/blob/dc7c8999c1afaf5bc7b20d89ae4c7421d4c70ec9/src/constructs/iam/policies/s3-get-object.ts#L16-L26) is a vague demonstration of this.

src/patterns/lambda-alarms.ts
jacobwinch: Having read this comment and listened to @akash1810's excellent explanation during the presentations, I'm inclined to agree - I'll move this.

akash1810: Agree with @jamie-lynch and reminded that `InstanceRole` needs moving out of the patterns directory! https://docs.aws.amazon.com/cdk/latest/guide/constructs.html is a good read.

src/patterns/scheduled-lambda.ts
akash1810: Does [this](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgDJwLYCMAmcCyA9iMGIVKAObIDeyAvgFCOiSyIoByhRJZFIahAAekEDgDOaTLgLFS5KrQbNW0eEmQBRKFHIAFaEnBxKEXgoFDREcVPTY8F-kpqNkyMgBtocSDh09KEMoYzBTCAAuZBAAV2xoABpkd2QJEAkAFUIAB2AEAEEoEGiJMCtk1LiEqAB5GAAxYAA3c1BYyBDgQklsrWa4L1i-CAB+aOqsaEYmRhhYkAQwbpBkHEIAZUIMCDAACyoACmFoh1lnRUFkAF5aemQ4KW4LqwBKZVmEYjLkE+Rn+QuK63OhML4ZMDIACe0UCBiMtnCZheShBnkIPigIwCunhoUREWiAEZkuksrl8kUSsgAEQwQiEGkqRjrLY7fZHYSvFmbba7A6CQ5Q15AA) work? (I'm not too certain on the `as` though!)

src/constructs/cloudwatch/alarm.ts
jacobwinch: Good question. My initial thought is that configuring a single topic in parameter store would not work so well in this case because multiple teams sometimes share the same AWS account, but not the same topic for alerts (or the same mailing list). Similarly teams might want to use different topics for alerts of a different priority etc. 

I think it'd be worth thinking about how we might standardise this further in the future, but I'm inclined to leave it as is for now.

src/patterns/scheduled-lambda.ts
jacobwinch: Thanks for the suggestion @akash1810 - I've used this as the basis for https://github.com/guardian/cdk/pull/186/commits/dc3972c3d7f68386314c7e1644137cbc5e5f91de. To be honest, I'm not sure whether I like this more than my original implementation, but I've spent quite a bit of time experimenting and I can't find a cleaner way. Happy to revert that commit if people are not keen on this approach!

package.json
```diff
@@ -40,6 +40,7 @@
   "dependencies": {
     "@aws-cdk/assert": "~1.74.0",
     "@aws-cdk/aws-autoscaling": "~1.74.0",
+    "@aws-cdk/aws-cloudwatch-actions": "~1.74.0",
```
akash1810, line 43: This'll need rebasing from main since https://github.com/guardian/cdk/pull/191.

```suggestion
    "@aws-cdk/aws-cloudwatch-actions": "1.86.0",
```

src/constructs/cloudwatch/alarm.test.ts
```diff
@@ -0,0 +1,65 @@
...
+describe("The GuAlarm class", () => {
+  it("should create a CloudWatch alarm", () => {
+    const stack = simpleGuStackForTesting();
+    const lambda = new GuLambdaFunction(stack, "lambda", {
+      code: { bucket: "bucket1", key: "folder/to/key" },
+      handler: "handler.ts",
+      runtime: Runtime.NODEJS_12_X,
+    });
+    new GuAlarm(stack, "alarm", {
+      alarmName: `Alarm in ${stack.stage}`,
```
jamie-lynch, line 17: One for a future PR but I wonder if it would be useful to append stack, stage and app to the alarm name (or including the information in some other way) as part of the `GuAlarm`

src/constructs/cloudwatch/lambda-alarms.ts
```diff
@@ -0,0 +1,44 @@
...
+export interface ErrorPercentageMonitoring
+  extends Omit<GuAlarmProps, "metric" | "threshold" | "comparisonOperator" | "evaluationPeriods"> {
+  toleratedErrorPercentage: number;
+  numberOfFiveMinutePeriodsToEvaluate?: number;
+  noMonitoring?: false;
+}
+
+interface GuLambdaAlarmProps extends ErrorPercentageMonitoring {
+  lambda: GuLambdaFunction;
+}
```
jamie-lynch, line 22: Maybe one for a future PR but I wonder if the `noMonitoring` stuff is something that should live at a higher level as it's unlikely to be specific to Lambda monitoring and also it feels a bit odd to define a prop here which then isn't used. 

Maybe `NoMonitoring` should be exported from somewhere else and the `monitoringConfiguration` prop becomes `monitoringConfiguration: LambdaMonitoring | NoMonitoring`? Then I guess `ErrorPercentageMonitoring` could extend `GuMonitoringProps` to add the `noMonitoring?: false`.

src/constructs/cloudwatch/lambda-alarms.test.ts
```diff
@@ -0,0 +1,86 @@
...
+      toleratedErrorPercentage: 65,
+      numberOfFiveMinutePeriodsToEvaluate: 12,
+      snsTopicName: "alerts-topic",
+      lambda: lambda,
+    };
+    new GuLambdaErrorPercentageAlarm(stack, "my-lambda-function", props);
+    expect(SynthUtils.toCloudFormation(stack)).toMatchSnapshot();
+    expect(stack).toHaveResource("AWS::CloudWatch::Alarm", {
+      EvaluationPeriods: 12,
+    });
```
jamie-lynch, line 42: I think given this has been moved to constructs we should prefer direct assertions over snapshots wherever possible. For the first test, the snapshot is maybe still the better choice as it will pick up any unexpected side effects of changes we make but for these tests that also have a direct assertion, I'd vote that the snapshot assertions can be removed.

jamie-lynch (approved):
This looks great! I've left a couple of comments but either stuff for us to think about in the future or personal preference on how we test so happy for this to be merged.

src/constructs/cloudwatch/alarm.test.ts
jacobwinch: Good idea. Ideally we would add these as tags, but unfortunately [that's not currently possible](https://github.com/aws-cloudformation/aws-cloudformation-coverage-roadmap/issues/64), so perhaps we should put these into the name by default for now.

src/constructs/cloudwatch/lambda-alarms.test.ts
jacobwinch: Sounds sensible, I've updated the tests and removed obsolete snapshots in https://github.com/guardian/cdk/pull/186/commits/139d5241156b516aef5dc4c219874bddc8963687.

src/constructs/cloudwatch/lambda-alarms.ts
jacobwinch: Agreed - this part of the code needs a little more thought as we cover more use-cases. I think I will go ahead and merge this as is for now, but perhaps we can refactor this slightly as part of adding additional lambda patterns.

PR #187, opened 2021-01-21
ASG adds instanceType and AMI parameters

jamie-lynch:
## What does this change?

This PR updates the `GuAutoscalingGroup` class to follow the bring-your-own-parameter pattern by adding the `InstanceType` and `AMI` parameters within the constructor.

Yes. Existing stacks that use the `GuAutoscalingGroup` will be able to remove the manually configured parameters. Where parameter names differ, this will have to be updated.

## How to test

Update an existing stack to use the new `GuAutoscalingGroup` and run the snapshot tests. This should general no or minimal changes.

Less code is required to configure constructs that always require parameters.


jacobwinch (approved):
Nice work 👍

akash1810:
Excellent!

PR #188, opened 2021-01-21
Provide sensible defaults for GuLambda memorySize and timeout

jacobwinch:
## What does this change?

Whilst testing https://github.com/guardian/cdk/pull/186, I noticed that `memorySize` and `timeout` are optional parameters for a `GuLambda`. If these values are omitted, CloudFormation defaults to [128MB](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-function.html#cfn-lambda-function-memorysize) and [3 seconds](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-function.html#cfn-lambda-function-timeout), respectively.

I don't think that these default values are particularly helpful for many of the Guardian use-cases that I've seen (they are especially unhelpful for Scala lambdas!), so I've tried to suggest something more appropriate.

There are no required changes, but some projects may automatically increase the memory or timeout values for lambda functions. This change would show up in a snapshot diff though, so people could set these low values again manually if they felt strongly about keeping them.

## How to test

I think this is adequately covered by unit tests added in this PR.

When setting projects up from scratch developers should save time, as they are less likely to encounter timeouts or OOM errors immediately after deploying for the first time.

I suppose there is a small risk that this could incur unnecessary costs (due to the way that [lambda pricing](https://aws.amazon.com/lambda/pricing/) works).

There is also a risk that we will over or under provision lambdas due to my assumptions about what 'sensible' defaults are. An alternative approach would be to make these properties mandatory for a `GuLambda`. This would force developers to make these choices up-front, but at least it would avoid the annoying case where you end up with the super-low CloudFormation defaults!

### DIFF
--- a/src/constructs/lambda/lambda.ts
+++ b/src/constructs/lambda/lambda.ts
@@ -39,11 +39,10 @@ export class GuLambdaFunction extends Function {
   constructor(scope: GuStack, id: string, props: GuFunctionProps) {
     const bucket = Bucket.fromBucketName(scope, `${id}-bucket`, props.code.bucket);
     const code = Code.fromBucket(bucket, props.code.key);
-    const { memorySize, timeout, ...rest } = props;
     super(scope, id, {
-      ...rest,
-      memorySize: defaultMemorySize(props.runtime, memorySize),
-      timeout: props.timeout ? props.timeout : Duration.seconds(30),
+      ...props,
+      memorySize: defaultMemorySize(props.runtime, props.memorySize),
+      timeout: props.timeout ?? Duration.seconds(30),
       code,
     });
 
--- a/src/constructs/lambda/lambda.ts
+++ b/src/constructs/lambda/lambda.ts
@@ -3,8 +3,8 @@ import { LambdaRestApi } from "@aws-cdk/aws-apigateway";
 import type { Schedule } from "@aws-cdk/aws-events";
 import { Rule } from "@aws-cdk/aws-events";
 import { LambdaFunction } from "@aws-cdk/aws-events-targets";
-import { Code, Function } from "@aws-cdk/aws-lambda";
-import type { FunctionProps } from "@aws-cdk/aws-lambda";
+import { Code, Function, RuntimeFamily } from "@aws-cdk/aws-lambda";
+import type { FunctionProps, Runtime } from "@aws-cdk/aws-lambda";
 import { Bucket } from "@aws-cdk/aws-s3";
 import type { GuStack } from "../core";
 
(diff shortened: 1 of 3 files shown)


src/constructs/lambda/lambda.ts
```diff
@@ -21,12 +22,28 @@ interface GuFunctionProps extends Omit<FunctionProps, "code"> {
   apis?: ApiProps[];
 }
 
+function defaultMemorySize(runtime: Runtime, memorySize?: number): number {
+  if (memorySize) {
+    return memorySize;
+  } else {
+    switch (runtime.family) {
+      case RuntimeFamily.JAVA:
+        return 1024;
```
jacobwinch, line 31: I based this on https://github.com/guardian/lambda-scala-boilerplate.g8/blob/master/src/main/g8/cfn.yaml#L67.

src/constructs/lambda/lambda.ts
```diff
@@ -21,12 +22,28 @@ interface GuFunctionProps extends Omit<FunctionProps, "code"> {
...
+      default:
+        return 512;
```
jacobwinch, line 33: I think the other main family that we want to cater for is `NODE`, so I've based this on https://github.com/guardian/tag-janitor/blob/main/cdk/lib/cdk-stack.ts#L54 and https://github.com/guardian/google-chat-bots/blob/main/cdk/lib/constants.ts#L29.

src/constructs/lambda/lambda.ts
```diff
@@ -21,12 +22,28 @@ interface GuFunctionProps extends Omit<FunctionProps, "code"> {
...
+    }
+  }
+}
+
 export class GuLambdaFunction extends Function {
   constructor(scope: GuStack, id: string, props: GuFunctionProps) {
     const bucket = Bucket.fromBucketName(scope, `${id}-bucket`, props.code.bucket);
     const code = Code.fromBucket(bucket, props.code.key);
+    const { memorySize, timeout, ...rest } = props;
```
jamie-lynch, line 42: Do these still need to be extracted from props or would it be easier just to use `props.memorySize` on line 45?

src/constructs/lambda/lambda.ts
```diff
@@ -21,12 +22,28 @@ interface GuFunctionProps extends Omit<FunctionProps, "code"> {
...
     super(scope, id, {
-      ...props,
+      ...rest,
+      memorySize: defaultMemorySize(props.runtime, memorySize),
+      timeout: props.timeout ? props.timeout : Duration.seconds(30),
```
jamie-lynch, line 46: Minor but the following should achieve the same result with marginally less code
```suggestion
      timeout: props.timeout ?? Duration.seconds(30),
```

src/constructs/lambda/lambda.ts
```diff
@@ -21,12 +22,28 @@ interface GuFunctionProps extends Omit<FunctionProps, "code"> {
...
+function defaultMemorySize(runtime: Runtime, memorySize?: number): number {
+  if (memorySize) {
+    return memorySize;
+  } else {
+    switch (runtime.family) {
+      case RuntimeFamily.JAVA:
+        return 1024;
+      default:
+        return 512;
+    }
```
jamie-lynch, line 34: Nice 👍

src/constructs/lambda/lambda.ts
jacobwinch: Ah yes thanks, I didn't realise this simpler version would work (but now I've read the docs and re-tested I can see that my version was needlessly complicated 😀).

jamie-lynch (approved):

PR #191, opened 2021-01-21
bump aws-cdk to current latest version

akash1810:
## What does this change?

Updates AWS CDK to the [latest version](https://github.com/aws/aws-cdk/releases) (latest at time of writing!). This is to keep up to date.

I think so. As `peerDependencies` are not installed (unless on npm@7), we'd need to ensure dependant projects' version of AWS CDK match this.

## How to test

CI should suffice.

Keeping up to date.

n/a

### DIFF
--- a/src/constructs/ec2/vpc.ts
+++ b/src/constructs/ec2/vpc.ts
@@ -20,7 +20,21 @@ export interface GuSubnetProps {
 
 export class GuVpc {
   static subnets(scope: GuStack, subnets: string[]): ISubnet[] {
-    return subnets.map((subnetId) => Subnet.fromSubnetId(scope, `subnet-${subnetId}`, subnetId));
+    /**
+     *  As of ~1.77.0 of the aws/cdk, this line
+     * [ previosuly Subnet.fromSubnetId(scope, `subnet-${subnetId}`, subnetId)) ]
+     * resulted in the following error
+     * "Found an encoded list token string in a scalar string context. Use 'Fn.select(0, list)' (not 'list[0]') to extract elements from token lists."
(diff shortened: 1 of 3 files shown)


akash1810:
I don't really know why this is failing, help wanted!

jamie-lynch:
> I don't really know why this is failing, help wanted!

Wow! This looks like it's going to be a fun one 🙃

src/constructs/ec2/vpc.ts
```diff
@@ -20,7 +20,21 @@ export interface GuSubnetProps {
...
+     *  As of ~1.77.0 of the aws/cdk, this line
+     * [ previosuly Subnet.fromSubnetId(scope, `subnet-${subnetId}`, subnetId)) ]
+     * resulted in the following error
+     * "Found an encoded list token string in a scalar string context. Use 'Fn.select(0, list)' (not 'list[0]') to extract elements from token lists."
+     * See https://github.com/aws/aws-cdk/issues/11945 for more information
+     *
+     * As a hacky workaround, we have moved to using the fromSubnetAttributes method and hardcoded an empty value
+     * for the routeTableId prop. This prevents the error and, when tested on existings stacks, results in no change to the CFN output
+     *
+     * TODO: Understand VPCs and Subnets better and developer a better solution to this problem
```
stephengeller, line 33:
```suggestion
     * TODO: Understand VPCs and Subnets better and develop a better solution to this problem
```

stephengeller (approved):
Looks fine to me - I think had the snapshot altered, I'd be a bit more concerned, but it seems safe if the resulting yaml is fine?

PR #192, opened 2021-01-21
check age of aws-cdk/core during ci

akash1810:
## What does this change?

We introduced dependabot in #55 to keep our dependencies up to date. However dependabot doesn't report against the CDK libraries (related https://github.com/dependabot/dependabot-core/issues/1171).

This adds a fairly gnarly bash script to:
- List all versions of `@aws-cdk/core` (and assume it's ordered!)
- Use `jq` to find how out of date we are
- If we are 5* versions out of date, exit 1

This is all run in CI.

Also takes inspiration [from Source](https://theguardian.design/2a1e5182b/p/876251-getting-started/t/06dcf1) and uses [`npm-check-updates`](https://www.npmjs.com/package/npm-check-updates) to update the `@aws-cdk/*` libraries. All wrapped in a new `update-aws-cdk` script.

Related to https://github.com/guardian/cdk/pull/191.

*this was an arbitrary choice, we can tweak as needed

Assuming #191 goes in first, then no.

## How to test

Run the two new scripts locally:
- `./script/check-aws-cdk`
- `./script/update-aws-cdk`

We keep up to date with `@aws-cdk/*` more frequently, especially as they [release fairly often](https://github.com/aws/aws-cdk/releases).

Too much of a chore to keep up to date?

### DIFF
--- a/script/check-aws-cdk
+++ b/script/check-aws-cdk
@@ -13,7 +13,7 @@ TOTAL_VERSIONS=$(echo "$VERSIONS" | jq '. | length')
 
 INSTALLED_VERSION_NUMBER=$(jq -r ".dependencies.\"${LIBRARY_TO_CHECK}\"" < "$DIR/../package.json")
 
-INSTALLED_VERSION_POSITION=$(echo "$VERSIONS" | jq ". | index(\"${INSTALLED_VERSION_NUMBER:1}\")") # :1 to remove the first character (~)
+INSTALLED_VERSION_POSITION=$(echo "$VERSIONS" | jq ". | index(\"${INSTALLED_VERSION_NUMBER}\")")
 
 DIFFERENCE=$((TOTAL_VERSIONS - INSTALLED_VERSION_POSITION))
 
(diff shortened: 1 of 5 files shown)


jamie-lynch (approved):
Nice 👍

akash1810:
This isn't quite working as intended after rebasing with #191 as the diff should be 0 not 111! Will investigate.

jamie-lynch:
> This isn't quite working as intended after rebasing with #191 as the diff should be 0 not 111! Will investigate.

Looks like we went from `~1.74.0` to `1.86.0` in #191 but the script still removes the first char of the installed version to account for the `~`

script/check-aws-cdk
```diff
@@ -0,0 +1,25 @@
...
+TOTAL_VERSIONS=$(echo "$VERSIONS" | jq '. | length')
+
+INSTALLED_VERSION_NUMBER=$(jq -r ".dependencies.\"${LIBRARY_TO_CHECK}\"" < "$DIR/../package.json")
+
+INSTALLED_VERSION_POSITION=$(echo "$VERSIONS" | jq ". | index(\"${INSTALLED_VERSION_NUMBER}\")")
+
+DIFFERENCE=$((TOTAL_VERSIONS - INSTALLED_VERSION_POSITION))
+
+if [ $DIFFERENCE -gt $TOLERANCE ]; then
+  echo "⚠️  This repository is currently using version $LIBRARY_TO_CHECK at version $INSTALLED_VERSION_NUMBER."
```
jamie-lynch, line 21:
```suggestion
  echo "⚠️  This repository is currently using $LIBRARY_TO_CHECK at version $INSTALLED_VERSION_NUMBER."
```

akash1810: Ha, just noticed that too and force pushed.

jamie-lynch (approved):

PR #200, opened 2021-01-25
Move instance role to constructs

jamie-lynch:
## What does this change?

Now that we have our first proper pattern (🎉 ), this PR move the `InstanceRole` class into constructs as it more closely fits that definition.

Yes, existing stacks that use the `InstanceRole` class will need to update where they import it from.

## How to test

Run the tests to see that no actual changes have been made to any constructs or patterns. Update an existing stack's imports and see that the snapshot tests still pass.

Classes live in the location which best fits their definition.

This is a breaking change and so any projects making use of the `InstanceRole` will have to make changes before they can use the latest version of the library.


jacobwinch (approved):
Thanks for tidying this up 👍

PR #201, opened 2021-01-25
Add default health check values for GuTargetGroup

jamie-lynch:
## What does this change?

Following on from https://github.com/guardian/cdk/pull/184, this PR adds some default values for the `healthCheck` prop on the `GuTargetGroup` construct. For consistency, the default values are the same as defined for the `GuClassicLoadBalancer`. The default values are spread with any passed in so that you can do the following:

```ts
new GuTargetGroup(this, "TargetGroup", {
  ...
  healthCheck: {
    port: "3000"
  }
  ...
})
```

No changes are required but stacks the use the `GuTargetGroup` can reduce the number of props they have to pass in.

## How to test

Update an existing stack that implements the `GuTargetGroup` and remove any props that match the defaults. Run the snapshot test and confirm that it still passes.

Stacks implementing the `GuTargetGroup` get sensible default health check values for free.

The default values provided may not the optimum.

### DIFF
--- a/src/constructs/loadbalancing/elb.test.ts
+++ b/src/constructs/loadbalancing/elb.test.ts
@@ -107,6 +107,43 @@ describe("The GuApplicationTargetGroup class", () => {
     const json = SynthUtils.toCloudFormation(stack) as SynthedStack;
     expect(Object.keys(json.Resources)).not.toContain("ApplicationTargetGroup");
   });
+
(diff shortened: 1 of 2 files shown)


jacobwinch:
> Stacks implementing the `GuTargetGroup` get sensible default health check values for free.

🙌 🎉 💯

jacobwinch (approved):

PR #204, opened 2021-01-25
Add GuHttpsClassicLoadBalancer

jamie-lynch:
## What does this change?

This PR adds a new `GuHttpsClassicLoadBalancer` construct which sets defaults for the load balancer listener. It restricts the number of listeners to one with the following defaults:

```ts
InstancePort: "9000",
InstanceProtocol: "http",
LoadBalancerPort: "443",
Protocol: "https",
```

If a value for the `sslCertificateId` prop is not provided, it also creates a new parameter and sets the output of that as the prop value.

No changes are required but stacks which implement the `GuClassLoadBalancer` and have an HTTPS listener can reduce the number of props they provide.

## How to test

Migrate an existing stack to use the `GuHttpsClassLoadBalancer` and run the snapshot test to confirm that nothing has changed.

Stacks can define fewer props to get a standard load balancer which listens for HTTPS traffic

### DIFF
--- a/src/constructs/loadbalancing/clb.test.ts
+++ b/src/constructs/loadbalancing/clb.test.ts
@@ -5,6 +5,7 @@ import { Stack } from "@aws-cdk/core";
 import { simpleGuStackForTesting } from "../../../test/utils/simple-gu-stack";
 import type { SynthedStack } from "../../../test/utils/synthed-stack";
 import { GuClassicLoadBalancer } from "../loadbalancing";
+import { GuHttpsClassicLoadBalancer } from "./clb";
 
 describe("The GuClassicLoadBalancer class", () => {
   const vpc = Vpc.fromVpcAttributes(new Stack(), "VPC", {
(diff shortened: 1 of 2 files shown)


src/constructs/loadbalancing/clb.test.ts
```diff
@@ -116,3 +117,99 @@ describe("The GuClassicLoadBalancer class", () => {
...
+          Protocol: "https",
+          SSLCertificateId: {
+            Ref: "CertificateARN",
+          },
+        },
+      ],
+    });
+  });
+
+  test("adds the CertificateARN parameter if no value provided", () => {
```
jacobwinch, line 150: Nice!

jacobwinch (approved):
Looks good to me 👍

src/constructs/loadbalancing/clb.ts
```diff
@@ -48,3 +54,34 @@ export class GuClassicLoadBalancer extends LoadBalancer {
...
+    externalPort: 443,
+    internalProtocol: LoadBalancingProtocol.HTTP,
+    externalProtocol: LoadBalancingProtocol.HTTPS,
+  };
+
+  constructor(scope: GuStack, id: string, props: GuHttpsClassicLoadBalancerProps) {
+    const listenerProps = { ...GuHttpsClassicLoadBalancer.DefaultListener, ...props.listener };
+
+    if (!listenerProps.sslCertificateId) {
+      const certificateId = new GuArnParameter(scope, "CertificateARN", {
```
akash1810, line 74: ACM certificates can be created via CloudFormation. Historically, it's been painful as it's a blocking resource - the stack will remain in "updating" until the certificate has been validated.

Validation takes two forms:
- If email validated, an email will be sent with a link to be clicked. This email goes to a list of domain admins (around 5 people)
- If DNS validated, a CNAME needs created in DNS. Again, admins for the domain service is around 5 people.

However, since the creation of https://github.com/guardian/dns-validation-lambda, it's become more viable to create ACM certs via CFN if domain validated.

(almost at the end of this story!)

Rather than this construct bringing a param for the user to fill in at runtime, should/could it bring an ACM resource with it? Providing the benefit of:
- Less involvement for user (aka reduce human error)
- Applies best practice of domain validated certificates (best practice because they auto-renew vs. email validated which needs manual renewal once a year)

jamie-lynch: That would be cool (and thanks for providing the context)! I guess if we find that not all cases can move to this for whatever reason then we can always provide a prop to add a param instead. I'll add this as an idea to the discussions section for now. 💡

PR #208, opened 2021-01-26
update GuRole import in instance role

jamie-lynch:
## What does this change?

This PR updates the import of `GuRole` in the `instance-role.ts` file to import directly from the `roles.ts` file rather than the `index.ts` in the `roles` directory as this was causing an error when importing anything from `@guardian/cdk/lib/constructs/iam`. 

Any projects importing `GuInstanceRole` or `GuRole` from their specific file will now be able to get them from the `@guardian/cdk/lib/constructs/iam` path.

## How to test

Bump to the latest version and try importing the `GuInstanceRole` construct from `@guardian/cdk/lib/constructs/iam`

All imports work as expected.

### DIFF
--- a/src/constructs/iam/roles/instance-role.ts
+++ b/src/constructs/iam/roles/instance-role.ts
@@ -8,7 +8,7 @@ import {
   GuParameterStoreReadPolicy,
   GuSSMRunCommandPolicy,
 } from "../policies";
-import { GuRole } from "../roles";
+import { GuRole } from "./roles";
 
 interface GuInstanceRoleProps {
   withoutLogShipping?: boolean; // optional to have log shipping added by default, you have to opt out


jamie-lynch:
It looks like this issue didn't show up in our tests as the classes were only ever imported directly from their files. One thing that might be useful is to include some example stacks in this repository and try to build them during CI as an integration test. 

As a side-effect, keeping these examples up to date alongside feature work might make it easier for users to see what a current "best practice" stack looks like and what changes are needed to incorporate new features. It would also give us a good test bed to trial features on as we work on them, rather than having to deploy and bump the version used in external projects.

I've started https://github.com/guardian/cdk/discussions/209 to discuss this further.

akash1810 (approved):

PR #213, opened 2021-01-28
Allow imageId and instanceType to be passed into ASG

jamie-lynch:
## What does this change?

Currently the `GuAutoscalingGroup` creates parameters for the `imageId` (`AMI`) and `instanceType` values and omits the props. This PR adds the props back in as optional and only creates the parameters if no value is passed. This maintains the BYO pattern as the default behaviour but allows flexibility where required.

No

## How to test

New unit tests have been written to cover the change in behaviour.

The default behaviour of our constructs makes them easy to use but also allows for flexibility.

### DIFF
--- a/src/constructs/autoscaling/asg.test.ts
+++ b/src/constructs/autoscaling/asg.test.ts
@@ -5,6 +5,7 @@ import { ApplicationProtocol } from "@aws-cdk/aws-elasticloadbalancingv2";
 import { Stack } from "@aws-cdk/core";
 import { simpleGuStackForTesting } from "../../../test/utils/simple-gu-stack";
 import type { SynthedStack } from "../../../test/utils/synthed-stack";
+import { GuAmiParameter } from "../core";
 import { GuSecurityGroup } from "../ec2";
 import { GuApplicationTargetGroup } from "../loadbalancing";
 import type { GuAutoScalingGroupProps } from "./asg";
@@ -43,14 +44,21 @@ describe("The GuAutoScalingGroup", () => {
   test("does not add the AMI parameter if an imageId prop provided", () => {
     const stack = simpleGuStackForTesting();
 
-    new GuAutoScalingGroup(stack, "AutoscalingGroup", { ...defaultProps, osType: 1, imageId: "123" });
+    new GuAutoScalingGroup(stack, "AutoscalingGroup", {
+      ...defaultProps,
+      osType: 1,
+      imageId: new GuAmiParameter(stack, "CustomAMI", {}),
+    });
 
     const json = SynthUtils.toCloudFormation(stack) as SynthedStack;
 
     expect(Object.keys(json.Parameters)).not.toContain("AMI");
+    expect(Object.keys(json.Parameters)).toContain("CustomAMI");
 
     expect(stack).toHaveResource("AWS::AutoScaling::LaunchConfiguration", {
-      ImageId: "123",
+      ImageId: {
+        Ref: "CustomAMI",
+      },
     });
   });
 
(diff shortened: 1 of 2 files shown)


src/constructs/autoscaling/asg.test.ts
```diff
@@ -40,7 +40,21 @@ describe("The GuAutoScalingGroup", () => {
     });
   });
 
-  test("adds the instanceType parameter", () => {
+  test("does not add the AMI parameter if an imageId prop provided", () => {
```
jacobwinch, line 43: I believe that Riff-Raff [currently relies on AMI being configured as a parameter](https://riffraff.gutools.co.uk/docs/magenta-lib/types#amicloudformationparameter) in order to automatically update the underlying AMI during a deployment. That's a widely used feature in the department (and ideally we want every application to do this due to the security benefits it brings), so I think we might need to discourage passing it as a prop for now.

jacobwinch:
I wonder if we should go even further for `instanceType` and steer users away from using a parameter for this value entirely. @akash1810  described some of the advantages of using `Mappings` for min/max values [in this thread](https://chat.google.com/room/AAAAeK2MOq4/AQ3lNfNCY_4) and I think that the same logic applies to `instanceType` (which is, to my mind at least, not private). 

In addition to the benefits @akash1810 mentioned, using a `Mapping` is also neater if you deploy your CFN changes via Riff-Raff, because it allows you to update `instanceType` via merging a PR instead of via manually CloudForming in the console (which is required when using a parameter).

If we included a default `InstanceType` in a `Mappings` section in the (soon to exist!) `GuEc2App` pattern, I wonder if people really need the option to use a parameter for this at all?

src/constructs/autoscaling/asg.test.ts
jamie-lynch: That's a good point. The reason this was added back as an optional prop was to allow for stacks that have multiple ASGs. Currently, that would fail as we'd try to create the same parameter more than once. This would allow a stack to define multiple parameters and pass them into each ASG as required. If we want to go as far as requiring AMI values to be a parameter, we could change the type of `imageId` to `GuAMIParameter` to allow for this overriding but prevent the value being hardcoded in the stack.

jacobwinch: > The reason this was added back as an optional prop was to allow for stacks that have multiple ASGs. 

Hmm, I hadn't thought about that other use-case! Maybe we should still allow this then and use patterns/docs to push 'typical' single-ASG stacks towards using an AMI parameter.

jacobwinch (approved):
Although I've raised a couple of questions about the approach, I think this is a step in the right direction and the code looks great, so feel free to merge if you want to keep things moving 🚢

src/constructs/autoscaling/asg.ts
```diff
@@ -11,6 +11,8 @@ import { GuAmiParameter, GuInstanceTypeParameter } from "../core";
 // https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys
 export interface GuAutoScalingGroupProps
   extends Omit<AutoScalingGroupProps, "imageId" | "osType" | "machineImage" | "instanceType" | "userData"> {
+  instanceType?: InstanceType;
+  imageId?: string;
```
jamie-lynch, line 15: This would require the prop to be a parameter so we can cater for stacks with multiple ASGs but enforce using parameters for AMIs.

```suggestion
  imageId?: GuAMIParameter;
```

src/constructs/autoscaling/asg.test.ts
jamie-lynch: I've added a suggested below which might be the best of both worlds. Agreed that the standard EC2 pattern should then just create the parameter for you!

src/constructs/autoscaling/asg.ts
jacobwinch: That's a nice idea 👍

jacobwinch (approved):
Nice work 👍

PR #214, opened 2021-01-28
Add new GuHttpsApplicationListener construct

jamie-lynch:
## What does this change?

This PR introduces a new `GuHttpsApplicationListener` construct which abstracts away some of the complexity of the `GuApplicationListener`. The two key abstractions are as follows:

* It omits the `defaultAction` prop and instead takes the `targetGroup` prop. The `defaultAction` value is then set to `ListenerAction.forward([props.targetGroup])`
* It omits the `certificates` prop and instead takes a single `certificate` value. The `certificates` value is then set to `[{ certificateArn: props.certificates }]`. The certificate prop is optional and if not provided a parameter will be created.

No but projects that use the `GuApplicationListener` might be able to move to the `GuHttpsApplicationListener` to reduce the complexity of their stack slightly.

## How to test

New unit tests are added to cover the new construct. Existing stacks that use the `GuHttpsApplicationListener` might be able to migrate to the new construct without any change to the generated cloudformation.

Users can make use of components which abstract away more things making it easier to define their stack.


jacobwinch (approved):
Great work 👍

PR #215, opened 2021-01-28
Add pattern for SNS lambdas

jacobwinch:
## What does this change?

This PR adds a pattern for lambda's triggered by an SNS topic. It also refactors some of the alarm-related code added in https://github.com/guardian/cdk/pull/186 to make it easier to share that code with this new pattern.

Existing projects using the SNS -> lambda architecture should be able to reduce the amount of boilerplate in their templates. You can see an example of this here: https://github.com/guardian/google-chat-bots/pull/152.

Due to the aforementioned refactor, some projects might need to make minor changes (e.g. fixing import statements) for the alarm-related code.

## How to test

1. Create a new stack based on this pattern. You should end up with a functional lambda triggered by SNS messages.
2. Upgrade existing stacks to use this pattern:
    1. A stack which defines its own SNS topic
    2. A stack which uses an external SNS topic

* Projects already using this library (e.g. `sns-to-gchat`) should be able to simplify their infrastructure definition
* Projects using this architecture can configure sensible monitoring for a tiny amount of extra code/effort
* It should be faster to set up new projects correctly using this pattern

When migrating existing stacks, SNS Subscriptions and Invoke permissions will be replaced due to changing ids (i.e. we create new versions of these resources and delete the old ones). I am reasonably confident that the `Create` operations will run before the `Delete` operations, which should prevent any disruption to service. Although I have tested this, I'm struggling to find docs from AWS which explicitly state that this will always be the case, so there is a small risk that migrating to this pattern could introduce a very brief period of downtime.


src/constructs/cloudwatch/lambda-alarms.ts
```diff
@@ -4,20 +4,14 @@ import type { GuLambdaFunction } from "../lambda";
...
   extends Omit<GuAlarmProps, "metric" | "threshold" | "comparisonOperator" | "evaluationPeriods"> {
   toleratedErrorPercentage: number;
   numberOfFiveMinutePeriodsToEvaluate?: number;
   noMonitoring?: false;
 }
 
-interface GuLambdaAlarmProps extends ErrorPercentageMonitoring {
+interface GuLambdaAlarmProps extends LambdaErrorPercentageMonitoring {
   lambda: GuLambdaFunction;
 }
```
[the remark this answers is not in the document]
jacobwinch: I agree that your suggestion would be an improvement. At the moment it’s tricky to do it this way because it means that any pattern which wants to create an alarm has to pass a `GuLambdaFunction` in as a prop. Since all of the patterns which use this construct are also creating the `GuLambdaFunction` itself, we end up with a weird circular dependency!

To work around this, I think we’d have to:

a) Define a separate interface that the lambda patterns can populate e.g.

```
interface GuLambdaMonitoringProps
  extends Omit<GuAlarmProps, "metric" | "threshold" | "comparisonOperator" | "evaluationPeriods"> {
  noMonitoring?: false;
}

// Library users create this object and pass it to the pattern
export interface GuLambdaErrorPercentageMonitoringProps extends GuLambdaMonitoringProps {
  toleratedErrorPercentage: number;
  numberOfFiveMinutePeriodsToEvaluate?: number;
}

// Patterns supplement the user-defined object with an additional lambda prop
interface GuLambdaErrorPercentageAlarmProps extends GuLambdaErrorPercentageMonitoringProps {
  lambda: GuLambdaFunction;
}
```

b) Relax [this linting rule](https://github.com/guardian/cdk/blob/main/eslint/rules/valid-constructors.js#L63-L70) so the patterns can pass the `GuLambdaFunction` in separately from the other props.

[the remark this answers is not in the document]
jamie-lynch: Ah yes! I hadn't thought about that. In that case, I think how it is now is fine. Maybe it's something that can be refactored if and when we ever come to adding more lambda monitoring.

src/constructs/cloudwatch/no-monitoring.ts
```diff
@@ -0,0 +1,3 @@
+export interface NoMonitoring {
```
akash1810, line 1: Not too certain on the naming or placement of this interface. It looks like it is only ever used in conjunction with `GuLambdaErrorPercentageMonitoringProps`? Maybe we can colocate them?

src/constructs/lambda/lambda.test.ts
```diff
@@ -96,6 +96,22 @@ describe("The GuLambdaFunction class", () => {
...
+      code: { bucket: "bucket1", key: "folder/to/key" },
+      handler: "handler.ts",
+      runtime: Runtime.NODEJS_12_X,
+      errorPercentageMonitoring: {
+        toleratedErrorPercentage: 5,
+        snsTopicName: "test-topic",
+      },
+    });
+
+    expect(stack).toHaveResource("AWS::CloudWatch::Alarm");
```
akash1810, line 112: Do we want to test the shape of this resource too?

src/constructs/lambda/lambda.ts
```diff
@@ -63,6 +66,13 @@ export class GuLambdaFunction extends Function {
       });
     });
 
+    if (props.errorPercentageMonitoring) {
+      new GuLambdaErrorPercentageAlarm(scope, "error-percentage-alarm-for-lambda", {
```
akash1810, line 70: I think we've used `TitleCase` for ids so far. Any objection to continue with that pattern here? And/or should we have a linter rule for future?

src/patterns/sns-lambda.test.ts
```diff
@@ -0,0 +1,71 @@
...
+  it("should create the correct resources for a new stack with minimal config", () => {
+    const stack = simpleGuStackForTesting();
+    const noMonitoring: NoMonitoring = { noMonitoring: true };
+    const props = {
+      code: { bucket: "test-dist", key: "lambda.zip" },
+      functionName: "my-lambda-function",
+      handler: "my-lambda/handler",
+      runtime: Runtime.NODEJS_12_X,
+      monitoringConfiguration: noMonitoring,
+    };
```
akash1810, line 19: Possibly not for addressing in this PR, but I think these props are too big. I wonder if we could default `code`, `handler` and `runtime` with sensible defaults but still allow one to override them.

This will, for example, make it easier to keep to the latest node runtime environment. Version X of the CDK library will default the runtime to v12. When AWS support a newer runtime, version Y of CDK gets published and teams update their stacks by bumping a version number in package.json and get the updated runtime for free.

src/patterns/sns-lambda.ts
```diff
@@ -0,0 +1,40 @@
...
+  logicalIdFromCloudFormation?: string; // Allows us to inherit an SNS topic which has already been created via a CloudFormation stack
+  externalTopicName?: string; // Allows us to link the lambda to an SNS topic owned by a different stack (or created outside of CloudFormation)
+}
+
+interface GuSnsLambdaProps extends Omit<GuFunctionProps, "rules" | "apis" | "errorPercentageMonitoring"> {
+  monitoringConfiguration: NoMonitoring | GuLambdaErrorPercentageMonitoringProps;
+  existingSnsTopic?: ExistingSnsTopic;
+}
+
+export class GuSnsLambda extends GuLambdaFunction {
```
akash1810, line 21: As patterns are the primary entry point to the library, should we add documentation to them by default to make the typedoc site friendly?

src/constructs/cloudwatch/no-monitoring.ts
jacobwinch: These were previously colocated, but this PR has refactored based on this suggestion: https://github.com/guardian/cdk/pull/186#discussion_r563738361. It's slightly preemptive, but we are expecting to share this object when we add different types of monitoring.

src/constructs/lambda/lambda.test.ts
jacobwinch: We also have a few tests related to this alarm here: https://github.com/guardian/cdk/blob/main/src/constructs/cloudwatch/lambda-alarms.test.ts. I feel like it's probably better to check specifics about the alarm there?

src/constructs/lambda/lambda.ts
jacobwinch: No objections, just an apology that I also broke this convention in my previous PR! 🙈 

If we want consistency here then we probably do want a linter rule (or a type, if that's possible?!) to enforce this convention.

jacobwinch: Although thinking about this some more, teams will need to be able to bring their existing ids (regardless of casing) in order to preserve resources that have been created by CFN, so maybe we can't enforce this too strictly?

src/patterns/sns-lambda.ts
jacobwinch: Yes, definitely! Thanks for the reminder!

src/patterns/sns-lambda.test.ts
jacobwinch: > I wonder if we could default code, handler and runtime with sensible defaults but still allow one to override them.

I worry that with the current distribution of projects between Scala and TypeScript/JavaScript there is no single sensible default here. Perhaps we can take [RuntimeFamily](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-lambda.RuntimeFamily.html) as a prop, and produce sensible defaults based on that. We already do something similar for memory size [here](https://github.com/guardian/cdk/blob/main/src/constructs/lambda/lambda.ts#L25-L36).

> When AWS support a newer runtime, version Y of CDK gets published and teams update their stacks by bumping a version number in package.json and get the updated runtime for free.

I think this will be great technique for ensuring that new projects always use the latest runtime for their 'family'. For existing projects, I guess that a major Node upgrade may also require code changes / more careful testing, so we can probably do this as long as we treat it as a major version change?

Overall, I think it makes sense to implement this type of abstraction at the `GuLambdaFunction` (construct) level. So I'd be inclined to create a new card and tackle it in a separate PR, if that's OK with you?

src/constructs/lambda/lambda.ts
akash1810: IIUC linting will only be scoped to this repository - consumers of this library would need to perform linting for themselves, bringing their own rules. That is, I think linting is still good here.

src/constructs/lambda/lambda.test.ts
akash1810: Maybe we can add an inline comment to that effect?

src/constructs/cloudwatch/no-monitoring.ts
akash1810: Ah ok, happy to leave then.

src/patterns/sns-lambda.test.ts
akash1810: Another version of this just came to mind: make `GuSnsLambda` abstract and then publish a `GuSnsNodeLambda` and `GuSnsJVMLambda` that extend `GuSnsLambda`. If consumers want a different runtime, they can extend `GuSnsLambda` themselves. Making it similar to [`GuPolicy`, `GuAllowPolicy` and `GuDenyPolicy`](https://github.com/guardian/cdk/blob/404bd66d9c5c06a851b885f22490ecca397d5c18/src/constructs/iam/policies/base-policy.ts#L11-L42).

https://github.com/guardian/cdk/blob/404bd66d9c5c06a851b885f22490ecca397d5c18/src/constructs/iam/policies/base-policy.ts#L11-L42

> So I'd be inclined to create a new card and tackle it in a separate PR, if that's OK with you?

Absolutely! 👍🏽

src/constructs/lambda/lambda.ts
akash1810: Let's add a linter rule in a different PR to keep things simple.

src/patterns/sns-lambda.ts
jacobwinch: First pass at docs is here: https://github.com/guardian/cdk/pull/215/commits/80058bc3d1846b33dd6c46fa334f8bba3f204973

src/patterns/sns-lambda.test.ts
jacobwinch: I've created a [card](https://trello.com/c/vrpIdRIX/530-reduce-number-of-required-props-for-gulambdafunction) to remind us about this potential improvement.

akash1810 (approved):
👍🏽  looks great!

PR #223, opened 2021-02-03
feat: Add `X-Gu-CDK-Version` tag to GuStack

akash1810:
## What does this change?

Adds `X-Gu-CDK-Version` tag to `GuStack`. We can use this to track which stacks have been created using the library, and with which version. Adding the tag to `GuStack` means it'll get applied to all resources.

The absence of this tag suggests the stack hasn't been created with the cdk library. Using the version number allows us to track how up to date stacks are too.

We mock the library version in tests to remove having to update snapshots whenever we publish a new version of the library, which, among other things, would make automated publishing difficult! The mock is applied globally for simplicity.

Yes! The change is small though, it should be a matter of bumping the version of this library being used.

## How to test

Observe existing tests.

We know what version of this library was used to create a stack, if any.

~The version number is retrieved from the `package.json`. Does this resolve the the correct value once published? That is, if we use the library in a project, is the version number that of guardian/cdk or of the consuming project?~ Update: this is fine. The import path is relative so will always resolve to this package's `package.json` file.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -3,4 +3,5 @@ module.exports = {
   transform: {
     "^.+\\.tsx?$": "ts-jest",
   },
+  setupFilesAfterEnv: ["./jest.setup.js"],
 };
--- a/src/constants/__mocks__/library-info.ts
+++ b/src/constants/__mocks__/library-info.ts
@@ -1,0 +1,8 @@
+export const LibraryInfo = {
+  VERSION: "TEST",
+};
+
+export const TrackingTag = {
+  Key: "X-Gu-CDK-Version",
+  Value: LibraryInfo.VERSION,
+};
--- a/jest.setup.js
+++ b/jest.setup.js
@@ -1,0 +1,1 @@
+jest.mock("./src/constants/library-info");
--- a/src/constants/library-info.ts
+++ b/src/constants/library-info.ts
@@ -1,0 +1,10 @@
+import { version } from "../../package.json";
+
+export const LibraryInfo = {
+  VERSION: version,
+};
+
+export const TrackingTag = {
+  Key: "X-Gu-CDK-Version",
+  Value: LibraryInfo.VERSION,
+};
--- a/src/constructs/cloudwatch/__snapshots__/lambda-alarms.test.ts.snap
+++ b/src/constructs/cloudwatch/__snapshots__/lambda-alarms.test.ts.snap
@@ -55,6 +55,10 @@ Object {
               "Ref": "Stage",
             },
           },
+          Object {
+            "Key": "X-Gu-CDK-Version",
+            "Value": "0.31.0",
+          },
         ],
         "Timeout": 30,
       },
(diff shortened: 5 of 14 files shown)


src/constants/__mocks__/library-info.ts
```diff
@@ -0,0 +1,8 @@
+export const LibraryInfo = {
```
akash1810, line 1: see https://jestjs.io/docs/en/manual-mocks

jest.config.js
```diff
@@ -3,4 +3,5 @@ module.exports = {
   transform: {
     "^.+\\.tsx?$": "ts-jest",
   },
+  setupFilesAfterEnv: ["./jest.setup.js"],
```
akash1810, line 6: see https://jestjs.io/docs/en/configuration.html#setupfilesafterenv-array

jacobwinch:
> We know what version of this library was used to create a stack, if any.

Great idea 💡 I think this will be really useful in the future, especially if we can easily find stacks using specific versions via Prism!

jacobwinch (approved):
Great work 💯 

As far as I can tell, the only downside of this approach is that it will attempt to modify a large number of resources whenever users upgrade the library (because we'll need to perform tag updates, even if the underlying resources are unchanged)? 

I think it'll be really useful to have this info, but perhaps we should document this behaviour so that users are not concerned by these snapshot updates or change sets?

akash1810:
> As far as I can tell, the only downside of this approach is that it will attempt to modify a large number of resources whenever users upgrade the library (because we'll need to perform tag updates, even if the underlying resources are unchanged)?

Ah great point! Will have a look to see if a tag can be applied only to the stack and not spread across all its resources.

akash1810:
> Will have a look to see if a tag can be applied only to the stack and not spread across all its resources.

From the [docs](https://docs.aws.amazon.com/cdk/latest/guide/tagging.html#tagging_props) it is possible to choose which resources to add a tag to, however it looks like tags for the stack can only be set at [create/update time](https://docs.aws.amazon.com/cli/latest/reference/cloudformation/create-stack.html#id2:~:text=%2D%2Dtags%20(list))*, that is, they cannot be part of the template.

There is [this hack(?)](https://stackoverflow.com/a/37025655/3868241), but I'm not particularly keen in it as it is pretty obscure and unintuitive.

I see three options to progress:
1. Document this behaviour and merge
2. Make the tag more static, something like `X-Gu-Stack-Author => guardian/cdk`
3. Make it easier to apply updates through CD, which can hide the verbose change sets away

\* as we can't automate the application of tags against the stack, our tracking strategy might need to change.

jacobwinch (approved):
> I think I favour 1 ([3245a2c](https://github.com/guardian/cdk/commit/3245a2c174527551c91f3032968cc0e74ded2c86)).

Yep, I agree - thanks for looking into alternatives though! 

👍 🚢

PR #225, opened 2021-02-03
docs: Adds GitHub Action to lint PRs.

akash1810:
## What does this change?

In order for [machines](https://github.com/semantic-release/semantic-release) to automatically create releases for us, we need to structure the git history in standard way.

By default, [semantic-release](https://github.com/semantic-release/semantic-release) will enforce _individual commits_ are well structured.

The [prosemirror-typerighter](https://github.com/guardian/prosemirror-typerighter) repository uses semantic-release and after speaking with @jonathonherbert and @SHession about it, they recommend PR linting over commit linting as there is less friction.

Also removes CI from `main` as branch protection is enabled on `main`, so commits will only reach it via merging a PR. The repository is also configured for PRs to pass CI before they can be merged. These two combined, we can be certain commits to `main` pass CI.

I plan to create a `cd.yaml` workflow to run only on `main` which will perform the steps of CI and also create a release.

No.

## How to test

n/a

Our PR titles are structured.

n/a

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -2,9 +2,6 @@ name: CI
 on:
   pull_request:
   workflow_dispatch:
-  push:
-    branches:
-      - main
 jobs:
   check-aws-cdk:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -3,6 +3,13 @@ on:
   pull_request:
   workflow_dispatch:
 jobs:
+  # see https://github.com/marketplace/actions/semantic-pull-request
+  lint-PR:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: amannn/action-semantic-pull-request@v3.1.0
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
   check-aws-cdk:
     runs-on: ubuntu-latest
     strategy:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -7,7 +7,7 @@ jobs:
   lint-PR:
     runs-on: ubuntu-latest
     steps:
-      - uses: amannn/action-semantic-pull-request@v3.1.0
+      - uses: amannn/action-semantic-pull-request@v3.2.6
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
   check-aws-cdk:


jonathonherbert:
Great to see this approach used elsewhere – let us know how you get on!

jamie-lynch:
> Also removes CI from main as branch protection is enabled on main, so commits will only reach it via merging a PR. The repository is also configured for PRs to pass CI before they can be merged. These two combined, we can be certain commits to main pass CI.

Does this also rely on the `Require branches to be up to date before merging` setting to be enabled to be entirely sure? Having that set would would mean that checks will definitely be running on the merge of the branch with the latest code from main. I think this is what caused us a problem previously where there were no merge conflicts and the checks passed but once merged, the tests failed on main.

akash1810:
> Does this also rely on the `Require branches to be up to date before merging` setting to be enabled to be entirely sure?

Good point. Am tempted to enable this setting to see if it's painful or not. Any objections @guardian/developer-experience ?

jacobwinch:
> I think this is what caused us a problem previously where there were no merge conflicts and the checks passed but once merged, the tests failed on main.

We definitely expose ourselves to edge cases like this by removing:

```
  push:	
    branches:	
      - main
```

I presume that the removal of this trigger event is partly motivated by the fact that it doesn't make sense to run new the `lint-PR` job against the `main` branch? If that's the case then we could:

a) Move this new job into it's own workflow file (so that it can have different trigger events) 
or
b) Add a [conditional](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idif) to this job e.g. `if: ${{ github.ref != 'main' }}` to prevent it from running unnecessarily 

As @jamie-lynch mentions, we could also avoid these edge cases by enabling the `Require branches to be up to date before merging` setting, but having tried this out on previous projects (and found it very irritating) I'd prefer to avoid enabling this if possible!

jamie-lynch:
> As @jamie-lynch mentions, we could also avoid these edge cases by enabling the `Require branches to be up to date before merging` setting, but having tried this out on previous projects (and found it very irritating) I'd prefer to avoid enabling this if possible!

It's been on for the cdk-cli repo for a while and I agree that it is a bit of a pain if you have to rebase a few PRs every time you merge something. I do think it's the only way to be 100% sure that a release is going to work though. If the release process also runs the tests then you can avoid releasing broken code but you'll still have to raise a new PR to fix main which isn't ideal. 

Whilst we've been fairly lucky so far to not hit this, IIUC something as simple as one PR that renames a construct followed by a PR to add a new pattern which uses that construct (but the old name) could easily result in an automatically released version of the library where that pattern doesn't work even those the status checks passed. 

Given that using the PR title for versioning also relies on squash and merge, what about moving to merging main back to the branch rather than rebasing? Then the merge commit will be squashed anyway. From what I remember, when you have the `Require branches to be up to date before merging` on, GitHub gives you a button to do this from within the PR which might make it a bit less painful.

akash1810:
I'm keen to avoid conditionals in GHA files, they're already getting pretty big for what they do adding conditionals adds to the complexity for not much gain, imo.

What if we have a `ci.yaml` for CI on feature branches and `cd.yaml` that runs on updates to `main`.

CI would do:
- Checkout
- PR lint
- AWS version check
- Build
- TS lint
- Test

CD would do:
- Checkout
- Build
- Test
- Release

There is some overlap between the two, however as `build` is a simple call to `./script/build` we'll be relatively DRY. This is sort of analogous to how projects that uses TeamCity for CI and RiffRaff for CD. That is, they each have their own configuration.

I think this setup addresses the fears of enabling `Require branches to be up to date before merging`?

Something something, separation of concerns, something something.

jacobwinch (approved):
The suggestion [here](https://github.com/guardian/cdk/pull/225#issuecomment-772551834) sounds good to me.

> If the release process also runs the tests then you can avoid releasing broken code but you'll still have to raise a new PR to fix main which isn't ideal.

This is true. Ultimately it's a trade off between occasionally (hopefully!) having to do this and having to rebase/merge in `main` before you can get anything merged. If the "occasionally" part of my previous statement proves to be false then we should definitely consider turning the setting on again!

akash1810:
https://github.com/guardian/cdk/pull/226 implements [this](https://github.com/guardian/cdk/pull/225#issuecomment-772551834).

PR #226, opened 2021-02-03
chore: Add a GHA for continuous delivery

akash1810:
## What does this change?

Following #225, add a GHA workflow for continuous delivery. Where continuous deliver is:
- build
- test
- release (TODO!)

No.

## How to test

n/a

CD is on for the `main` branch.

n/a

### DIFF
--- a/.github/workflows/cd.yaml
+++ b/.github/workflows/cd.yaml
@@ -1,0 +1,44 @@
+name: CD
+on:
+  push:
+    branches:
+      - main
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [14.15.1]
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v2.1.4
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: ./script/build
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/cd.yaml
```diff
@@ -0,0 +1,43 @@
...
+      matrix:
+        node-version: [14.15.1]
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v2.1.4
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: ./script/test
+  release:
```
jacobwinch, line 31: I know this doesn't matter until we _actually_ release automatically, but by default these [jobs will run in parallel](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobs), which probably isn't what we want here. 

We could either refactor so these all run as steps in a single job (which will run sequentially), or add the [`needs` syntax](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idneeds) to the release job.

Happy for that to happen when we add the automated release step, as long as you think we'll remember!

jacobwinch (approved):

.github/workflows/cd.yaml
akash1810: Didn't know about `needs`, nice! I think it would be nice to have a single `job` which simply runs `./script/cd`??? However that's different from the approach taken in the CI step, so for consistency have added `needs`.

PR #233, opened 2021-02-05
refactor!: follow aws tag name convention of `org:something`

akash1810:
## What does this change?

Follow the AWS tag name convention of `org:something`: `X-Gu-CDK-Version` becomes `gu:cdk:version`.

Also adds a test helper `alphabeticalTags` to sort tags alphabetically by tag name as the test matcher checks two arrays are the same order.

The would library need bumping, but no real impact.

## How to test

See tests.

> convention (noun)
>   1. a way in which something is usually done.

n/a

### DIFF
--- a/src/constants/__mocks__/library-info.ts
+++ b/src/constants/__mocks__/library-info.ts
@@ -3,6 +3,6 @@ export const LibraryInfo = {
 };
 
 export const TrackingTag = {
-  Key: "X-Gu-CDK-Version",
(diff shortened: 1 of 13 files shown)


sihil (approved):
This seems super noisy for a relatively small change. I guess that's a symptom of the testing approach.

akash1810:
> I guess that's a symptom of the testing approach.

Correct - we test the output of synthesising a stack.

PR #243, opened 2021-02-11
feat(stack): GuStack accepts Stack value as prop

stephengeller:
## What does this change?

As we look to reduce the number of parameters requiring manual updating and maintaining in CFN, this allows library users to hard-code stack and stage variables so we can reason with their values in other parts of the codebase.

If a user would like to avoid having the stack and stage as tokens (which later gets resolved at deployment by writing the values in the CFN UI), you could now simply state these values in the stack declaration:

```typescript
const stack =  new GuStack(new App(), "Test", { app: "testing", stack: "deploy", stage: "CODE" });
```

Since we don't often (or ever) write stage-specific CDK declarations, I expect the `stage` prop will not be used often. However, being able to provide the `stack` as a hardcoded string I think provides value in a) having more configuration explicit in the repository, b) remove a parameter from CFN UI and c) remove a tiny bit of cognitive load in the process

No, this shouldn't affect existing projects.

## How to test

`./script/test` from the root of the repository.

More projects can be explicit about what stack (and perhaps rarely stage) the CDK stacks belong to.

### DIFF
--- a/src/constructs/core/stack.test.ts
+++ b/src/constructs/core/stack.test.ts
@@ -17,30 +17,21 @@ describe("The GuStack construct", () => {
     const json = SynthUtils.toCloudFormation(stack) as SynthedStack;
 
     expect(json.Parameters).toEqual({
-      Stack: {
+      Stage: {
         Type: "String",
-        Description: "Name of this stack",
-        Default: "deploy",
+        Description: "Stage name",
+        AllowedValues: Stages,
+        Default: Stage.CODE,
       },
     });
   });
 
-  it("can accept only one of stack or stage", function () {
-    const stack = simpleGuStackForTesting({ stage: "some-stage" });
-    expect(stack.stage).toEqual("some-stage");
-  });
-
-  it("should have stack and stage parameters", () => {
-    const stack = simpleGuStackForTesting();
+  it("should have a stage parameter", () => {
+    const stack = simpleGuStackForTesting({ stack: "test" });
 
     const json = SynthUtils.toCloudFormation(stack) as SynthedStack;
 
     expect(json.Parameters).toEqual({
-      Stack: {
-        Type: "String",
-        Description: "Name of this stack",
-        Default: "deploy",
-      },
       Stage: {
         Type: "String",
         Description: "Stage name",
@@ -51,7 +42,7 @@ describe("The GuStack construct", () => {
   });
 
   it("should apply the stack, stage and app tags to resources added to it", () => {
-    const stack = new GuStack(new App(), "Test", { app: "MyApp" });
+    const stack = new GuStack(new App(), "Test", { app: "MyApp", stack: "test" });
 
     new Role(stack, "MyRole", {
       assumedBy: new ServicePrincipal("ec2.amazonaws.com"),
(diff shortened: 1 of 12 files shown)


akash1810:
I think we'd always want `stage` to be a parameter as it can only be defined at runtime - otherwise we'd be using a different template per stage.

I also wonder if it would be simpler to remove the `stack` parameter entirely too in favour of a prop?

stephengeller:
> I think we'd always want `stage` to be a parameter as it can only be defined at runtime - otherwise we'd be using a different template per stage.

Agreed, but would it not be nice to make it possible to have it defined at declaration? I agree that for the Guardian, we would almost never use this, but I can envision certain circumstances where maybe an external user might want to create very specific stacks per stage, and have them hard-coded in the file.

> I also wonder if it would be simpler to remove the `stack` parameter entirely too in favour of a prop?

Yeah, I think it would be better to remove the parameter altogether. However, would this cause a breaking change for stacks that have it as a parameter currently? If so, I would maybe suggest allowing it for now then maybe removing it later, although it could be better to just make the necessary changes in the stacks and make this our best practice. I think it _should_ be fine to **remove** parameters, whereas **adding** one causes issues iirc.

akash1810:
> I agree that for the Guardian, we would almost never use this, but I can envision certain circumstances where maybe an external user might want to create very specific stacks per stage, and have them hard-coded in the file.

Default values can help alleviate this pain, I think?

Whilst we should consider external use, I'm not sure we should be overly concerned with those use cases at the moment. That is, I think we should build for our internal use to make adoption easy and once the library is mature, we can focus on external parties. Maybe we can add a TODO to remind us that this feature might be tricky for external users?

> However, would this cause a breaking change for stacks that have it as a parameter currently?

Semver releases solve this 😉. We also don't have too many stacks on CDK so should be easy to manage this change?

> I think it should be fine to remove parameters, whereas adding one causes issues iirc.

Great point! Maybe this warrants an ADR?

stephengeller:
> Whilst we should consider external use, I'm not sure we should be overly concerned with those use cases at the moment. That is, I think we should build for our internal use to make adoption easy and once the library is mature, we can focus on external parties. Maybe we can add a TODO to remind us that this feature might be tricky for external users?

You're right, that makes sense. I've made a change for the stage value to be **only** a parameter, while the stack one should **only** be passed in. It makes sense, and is in line with our best practices.

> > I think it should be fine to remove parameters, whereas adding one causes issues iirc.
> 
> Great point! Maybe this warrants an ADR?

Sure! How do we go about that again?

src/constructs/core/stack.test.ts
```diff
@@ -10,17 +10,28 @@ import { TrackingTag } from "../../constants/library-info";
...
 describe("The GuStack construct", () => {
-  it("should have stack and stage parameters", () => {
-    const stack = simpleGuStackForTesting();
+  it("requires passing the stack value as props", function () {
+    const stack = simpleGuStackForTesting({ stack: "some-stack" });
+    expect(stack.stack).toEqual("some-stack");
 
     const json = SynthUtils.toCloudFormation(stack) as SynthedStack;
 
     expect(json.Parameters).toEqual({
```
jacobwinch, line 19: I don't think we need this assertion here, as it's tested as part of `"should have a stage parameter"`

jacobwinch (approved):
This looks good to me. I've made one suggestion which allows us to simplify the tests slightly. Feel free to merge this once that's been tidied up 👍

PR #244, opened 2021-02-12
feat: create a GuCertificateArnParameter construct

akash1810:
## What does this change?

`GuCertificateArnParameter` has an allowed pattern of an ACM ARN string. Update `GuHttpsApplicationListener` to use it and validate the certificate prop too.

No.

## How to test

See updated tests and CI output.

Fewer errors as user input is restricted via the regex.

n/a

### DIFF
--- a/src/constructs/loadbalancing/elb.test.ts
+++ b/src/constructs/loadbalancing/elb.test.ts
@@ -5,6 +5,7 @@ import { ApplicationProtocol, ListenerAction } from "@aws-cdk/aws-elasticloadbal
 import { Stack } from "@aws-cdk/core";
 import { simpleGuStackForTesting } from "../../../test/utils/simple-gu-stack";
 import type { SynthedStack } from "../../../test/utils/synthed-stack";
+import { RegexPattern } from "../../constants";
 import {
   GuApplicationListener,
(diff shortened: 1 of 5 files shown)


akash1810:
I think it would be good to create the certificate ourselves. Imagine a construct/pattern with props that look like this:

```typescript
interface Props {
  withLoadBalancer: true,
  withDomain: "foo.theguardian.com"
}
```

Then the construct/pattern will create an ACM resource. We probably need mappings to solve this, so leaving that for now.

src/constructs/loadbalancing/elb.test.ts
```diff
@@ -331,24 +332,43 @@ describe("The GuHttpsApplicationListener class", () => {
...
           CertificateArn: {
-            Ref: "CertificateARN",
+            Ref: "TLSCertificate",
           },
         },
       ],
     });
   });
 
+  test("passing in an invalid ACM ARN", () => {
```
jacobwinch, line 353: Minor suggestion, it might be nice to describe the expected behaviour in the test name?

```suggestion
  test("throws an error if an invalid ACM ARN is passed in", () => {
```

jacobwinch (approved):

PR #251, opened 2021-02-16
feat(secrets): Add GuSecret construct

stephengeller:
## What does this change?

Context: https://trello.com/c/Hi2XN6Xm/545-create-a-gusecret-construct-to-facilitate-referencing-parameters-in-ssm

At the moment, we implement secrets in CFN stacks by passing them in as (mostly string) parameters at runtime. This makes it hard to track, maintain and version control secrets. It also creates more work for developers who need to manage and update multiple parameters in order to change configuration, as well as navigate to the CFN UI to do so.

This PR adds a `GuSecret` construct, allowing us to state secrets in CDK as keys that assume a standard SSM parameter store path of `$stage/$stack/$app/$secretName`:

```typescript
// Assuming Stack is "deploy", Stage is "CODE" and App is "grafana"

const oldMethodUsingParameters = new GuStringParameter(this, "databasePassword", {
        description: "pasword of the RDS DB",
        default: "foobarbaz",
      }),
 
const newGuSecretMethod = GuSecret(this, "databasePassword")
// evaluates to a string in SSM with the path `/CODE/deploy/grafana/databasePassword`
```

### Drawbacks

- This currently only works for string values in SSM (namely not `SecureString`), but that can be easily implemented in the future.
- At the moment, I couldn't get specifying versions to work. I decided to raise the PR anyways as I don't think that feature is crucial and that we would already get value out of this construct using the latest SSM value, but would be keen to hear thoughts.

Nope, this should be an "easy" switch from traditional parameters to SSM parameter paths using `GuSecret`.

## How to test

`./script/test`, from the root of the repo.

CDK stacks begin using `GuSecret` construct and have cleaner declarations of secret values.


package.json
```diff
@@ -26,6 +26,7 @@
     "@types/node": "14.14.25",
     "@typescript-eslint/eslint-plugin": "^4.14.2",
     "@typescript-eslint/parser": "^4.14.2",
+    "cz-conventional-changelog": "^3.3.0",
```
jacobwinch, line 29: Not sure if this is included intentionally, but it might be better to decouple adding this dependency from implementing the `GuSecret` construct?

jacobwinch:
> * This currently only works for string values in SSM (namely not `SecureString`), but that can be easily implemented in the future.

Perhaps we should add support for `SecureString` in this initial version? Based on the [AWS advice](https://docs.aws.amazon.com/systems-manager/latest/userguide/systems-manager-parameter-store.html#what-is-a-parameter), I think we want any secret to be stored as a `SecureString`.

jacobwinch:
Assuming that this PR and #252 are both merged, we should probably add some documentation which explains how/when to use a `GuSecret` vs how/when to use a `GuStageDependentValue`. Perhaps we can pair on that task and add those docs via a new PR?

stephengeller:
This has been updated to take @jacobwinch 's notes into account.

stephengeller:
This requires more thinking, so going to close this for now.

PR #252, opened 2021-02-17
feat!: Implement stage dependent ASG capacity via Mappings

jacobwinch:
## What does this change?

Since Guardian stacks typically use the same CFN template for different stages (e.g. `CODE` or `PROD`), we often need a way to provide different values to a template depending on which stage is being CloudFormed. For example, many applications use fewer ec2 instances in `CODE` (needs to handle minimal test traffic) than they do in `PROD` (needs to cope with real production workloads). Unfortunately, stage is passed into templates via a parameter which means that it's not available at build-time. Consequently, we cannot simply use TypeScript logic to conditionally change values. 

This PR adds functionality for defining (non-secret) values which differ depending on stage, without forcing library users to think about CFN concepts like `Parameters` or `Mappings` and without requiring them to pass in values via the CFN UI. Under the hood, this is achieved via CFN `Mappings`.

In order to illustrate usage of this technique, this PR also:

* Removes the `minCapacity` and `maxCapacity` props from `GuAutoScalingGroup`, in favour of using new stage-based `Mappings`.
* Removes the `desiredCapacity` prop, due to the advice [here](https://riffraff.gutools.co.uk/docs/magenta-lib/types#cloudformation):

```NOTE: It is strongly recommended you do NOT set a desired-capacity on auto-scaling groups, managed with CloudFormation templates deployed in this way [via Riff-Raff], as otherwise any deployment will reset the capacity to this number, even if scaling actions have triggered, changing the capacity, in the mean-time.```

Yes, once this version is released any stack which defines an AutoScaling Group will need to make use of the new `capacity` prop in order to upgrade.

## How to test

I've tested this by creating change sets for the ec2-based apps in `deploy-tools-platform`. I've also CloudFormed `AMIable-CODE` to confirm that removing [`desiredCapacity` from an existing stack](https://github.com/guardian/deploy-tools-platform/blob/c5a075d872198fda05cb08d84ca105728a0682fa/cdk/lib/amiable/amiable.ts#L76) is a safe operation.

* Users of the library can define values which differ between stages without understanding CFN concepts
* It should also be harder to accidentally set up ASGs which are incompatible with Riff-Raff deployments

* This is a breaking change so it will impact the upgrade path for a number of projects.

### DIFF
--- a/src/constructs/autoscaling/asg.ts
+++ b/src/constructs/autoscaling/asg.ts
@@ -3,14 +3,25 @@ import { AutoScalingGroup } from "@aws-cdk/aws-autoscaling";
 import type { ISecurityGroup, MachineImage, MachineImageConfig } from "@aws-cdk/aws-ec2";
 import { InstanceType, OperatingSystemType, UserData } from "@aws-cdk/aws-ec2";
 import type { ApplicationTargetGroup } from "@aws-cdk/aws-elasticloadbalancingv2";
-import type { GuStack } from "../core";
+import type { GuStack, GuStageVariable } from "../core";
 import { GuAmiParameter, GuInstanceTypeParameter } from "../core";
 
 // Since we want to override the types of what gets passed in for the below props,
 // we need to use Omit<T, U> to remove them from the interface this extends
 // https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys
 export interface GuAutoScalingGroupProps
-  extends Omit<AutoScalingGroupProps, "imageId" | "osType" | "machineImage" | "instanceType" | "userData"> {
+  extends Omit<
+    AutoScalingGroupProps,
+    | "imageId"
+    | "osType"
+    | "machineImage"
+    | "instanceType"
+    | "userData"
+    | "minCapacity"
+    | "maxCapacity"
+    | "desiredCapacity"
+  > {
+  capacity: GuAsgCapacityProps;
   instanceType?: InstanceType;
   imageId?: GuAmiParameter;
   osType?: OperatingSystemType;
@@ -19,6 +30,50 @@ export interface GuAutoScalingGroupProps
   securityGroups?: ISecurityGroup[];
   targetGroup?: ApplicationTargetGroup;
   overrideId?: boolean;
+}
+
+/**
+ * `minimumCodeInstances` and `minimumProdInstances` determine the number of ec2 instances running
+ * in each stage under normal circumstances (i.e. when there are no deployment or scaling events in progress).
+ *
+ * `maximumCodeInstances` and `maximumProdInstances` are both optional. If omitted, these
+ * will be set to `minimumCodeInstances * 2` and `minimumProdInstances * 2`, respectively.
+ * This allows us to support Riff-Raff's autoscaling deployment type by default.
+ *
+ * The maximum capacity values can be set manually if you need to scale beyond the default limit (e.g. due to heavy traffic)
+ * or restrict scaling for a specific reason.
+ */
+export interface GuAsgCapacityProps {
+  minimumCodeInstances: number;
+  minimumProdInstances: number;
+  maximumCodeInstances?: number;
+  maximumProdInstances?: number;
+}
+
+interface AwsAsgCapacityProps {
+  minCapacity: number;
+  maxCapacity: number;
+}
+
+function wireCapacityProps(scope: GuStack, capacity: GuAsgCapacityProps): AwsAsgCapacityProps {
+  const minInstancesKey = "minInstances";
+  const maxInstancesKey = "maxInstances";
+  const minInstances: GuStageVariable = {
+    variableName: minInstancesKey,
+    codeValue: capacity.minimumCodeInstances,
+    prodValue: capacity.minimumProdInstances,
+  };
+  const maxInstances: GuStageVariable = {
+    variableName: maxInstancesKey,
+    codeValue: capacity.maximumCodeInstances ?? capacity.minimumCodeInstances * 2,
+    prodValue: capacity.maximumProdInstances ?? capacity.minimumProdInstances * 2,
+  };
+  scope.mappings.addStageVariable(minInstances);
+  scope.mappings.addStageVariable(maxInstances);
+  return {
+    minCapacity: (scope.mappings.findInMap(scope.stage, minInstancesKey) as unknown) as number,
+    maxCapacity: (scope.mappings.findInMap(scope.stage, maxInstancesKey) as unknown) as number,
+  };
 }
 
 export class GuAutoScalingGroup extends AutoScalingGroup {
@@ -40,6 +95,7 @@ export class GuAutoScalingGroup extends AutoScalingGroup {
 
     const mergedProps = {
       ...props,
+      ...wireCapacityProps(scope, props.capacity),
       machineImage: { getImage: getImage },
       instanceType: props.instanceType ?? new InstanceType(new GuInstanceTypeParameter(scope).valueAsString),
       userData: UserData.custom(props.userData),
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: src/constructs/autoscaling/asg.ts [hunk 1]
```diff
@@ -21,6 +32,50 @@ export interface GuAutoScalingGroupProps
...
+ *
+ * The maximum capacity values can be set manually if you need to scale beyond the default limit (e.g. due to heavy traffic)
+ * or restrict scaling for a specific reason.
+ */
+export interface GuAsgCapacityProps {
+  minimumCodeInstances: number;
+  minimumProdInstances: number;
+  maximumCodeInstances?: number;
+  maximumProdInstances?: number;
+}
```
akash1810, line 51: This interface feels a bit verbose, I wonder if TS allows us to do something like this:

```typescript
interface ASGProps {
  min: number,
  max: number
}

interface GuASGProps {
  Stage.PROD: ASGProps
  [stage: Stage]?: ASGProps
}
```

Where usage then becomes:

```typescript
const myAsgProps = {
  Stage.PROD: {
    min: 3,
    max: 6
  },
  Stage.CODE: {
    min: 1,
    max: 2
  }
}
````

Comment (akash1810) [reply src/constructs/autoscaling/asg.ts:51]: Alas, I don't think it's possible to make `PROD` mandatory and other stages optional in `GuASGProps` as suggested above (nor do I think it's correct, as if we didn't define CODE values, but deployed to CODE, what do we do?).

Here's a [working example](https://www.typescriptlang.org/play?#code/KYOwrgtgBAygLgQwObCgbwFBSgYQPIAiAolALxQBE+xFANFlAAoBKhZlLhdGAvhhgEsQcYACcAZggDGqAIIwA4o1EB7AA4BndA2wQhALijgIAIzH1suhAA9Dxs6N784ATzWoFYeUtWb2mbABtDUQUKCFYUOAAXUNvZXUNAG5eFIwpFRAQqHEVFUNPeN8tcgDLTgJDND0QQwBmWigIG0MANh4LS2oiKprDAEZG5tsoACY+HiA) of what I'm talking about.

Comment (jacobwinch) [reply src/constructs/autoscaling/asg.ts:51]: The assumption with my implementation is that most people will omit the optional props anyway, giving them:

```
capacity: {
      minimumCodeInstances: 1,
      minimumProdInstances: 3,
    }
```

As we calculate sensible default max values here: https://github.com/guardian/cdk/pull/252/files#diff-b219fec113d015feaadba12ad710f805f4ba8fc1fcf3c270be34c8fd31b67c0fR66-R70

Comment (akash1810) [reply src/constructs/autoscaling/asg.ts:51]: What about something like this?

```typescript
enum Stage {
  CODE = "CODE",
  PROD = "PROD",
}

interface ASGProps {
  min: number;
  max: number;
}

type GuASGProps = Record<Stage, ASGProps>;

class GuASG {
  public static defaultProps: GuASGProps = {
    [Stage.PROD]: {
      min: 3,
      max: 6,
    },
    [Stage.CODE]: {
      min: 1,
      max: 2,
    },
  };

  constructor(props: GuASGProps = GuASG.defaultProps) {
    console.log(props);
  }
}

const vanillaASG = new GuASG();

const myCustomProps: GuASGProps = {
  ...GuASG.defaultProps,
  [Stage.PROD]: {
    min: 3,
    max: 6,
  },
};

const myCustomGuASG = new GuASG(myCustomProps);
```

My main concern is the verbosity of adding more mappings over time - I suspect we'd want to move the instance type to a mapping soon, for example. In the above proposal, this can be achieved by:

```typescript
interface ASGProps {
  min: number;
  max: number;
  instanceType: string;
}
```

As in, one extra line. The benefit here is a strong relation between the `Stage` enum and the ASG props.

(The downside of both is it becomes difficult to cater for CAPI's `zebra` stage!)

src/constructs/autoscaling/asg.ts
```diff
@@ -21,6 +32,50 @@ export interface GuAutoScalingGroupProps
...
+    codeValue: capacity.maximumCodeInstances ?? capacity.minimumCodeInstances * 2,
+    prodValue: capacity.maximumProdInstances ?? capacity.minimumProdInstances * 2,
+  };
+  stack.setStageDependentValue(minInstances);
+  stack.setStageDependentValue(maxInstances);
+  return {
+    minCapacity: stack.getStageDependentValue(minInstancesKey),
+    maxCapacity: stack.getStageDependentValue(maxInstancesKey),
+  };
+}
```
stephengeller, line 77: Will we need to write code like this for every situation where we'll have stage-dependent non-private values to handle? Is there any way to make this easier going forward with helper functions, or should we just accept the responsibility to implement these on a case-by-case basis so that the library users can benefit from the result?

stephengeller (approved):
Looks great. I like how the interface looks for users and also like how you can only pass minimum values for ASG instances and have it multiply appropriately for max. I added one question regarding needing to write some code for each case where we have stage-dependant values, but happy to approve this and solve at a later date when we add more usages of this.

src/constructs/autoscaling/asg.ts
jacobwinch: Thanks for taking the time to code up this suggestion 💯  

I've tried to incorporate this (plus a few other small improvements) here: https://github.com/guardian/cdk/pull/252/commits/b3a4d9599875763995a986a60800d3e2132327b3. Please let me know if I've misunderstood anything!

> (The downside of both is it becomes difficult to cater for CAPI's zebra stage!)

Our [current `Stage` enum](https://github.com/guardian/cdk/blob/main/src/constants/stage.ts#L1-L7) already precludes this, but this PR certainly exacerbates the problem. 

I did have a go at implementing a version of this where an `AllowedStages` type could be passed in as a type parameter, but I got into a complete tangle and decided not to pursue this further yet! Hopefully we can cover the majority of stacks with `CODE` and `PROD` stages.

src/constructs/autoscaling/asg.ts
jacobwinch: I guess the main opportunity for cutting down on boilerplate here would be to provide a single function which sets up the mapping and returns the [`FindInMap`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-findinmap.html) which references it.

src/constructs/autoscaling/asg.ts
akash1810: > I did have a go at implementing a version of this where an `AllowedStages` type could be passed in as a type parameter, but I got into a complete tangle and decided not to pursue this further yet! 

Sounds promising! Happy to pair on this.

src/constructs/core/stack.ts
```diff
@@ -43,6 +46,7 @@ export class GuStack extends Stack {
   private readonly _stack: string;
   private readonly _app: string;
 
+  private _mappings: undefined | GuStageMapping;
```
akash1810, line 49: Could this be `private _mappings?: GuStageMapping;`?

src/constructs/autoscaling/asg.ts
```diff
@@ -21,6 +33,53 @@ export interface GuAutoScalingGroupProps
...
+    variableName: minInstancesKey,
+    stageValues: {
+      [Stage.CODE]: stageDependentProps.CODE.minimumInstances,
+      [Stage.PROD]: stageDependentProps.PROD.minimumInstances,
+    },
+  };
+  const maxInstances: GuStageDependentValue<number> = {
+    variableName: maxInstancesKey,
+    stageValues: {
+      [Stage.CODE]: stageDependentProps.CODE.maximumInstances ?? stageDependentProps.CODE.minimumInstances * 2,
```
akash1810, line 71: minor: would be good to do some validation here to assert min > max and throw otherwise. Maybe for a new PR.

akash1810 (approved):
Looks great!

src/constructs/autoscaling/asg.ts
jacobwinch: I will add this in separate PR (I don't want to hold up the queue of PRs which edit this file!)

PR #256, opened 2021-02-18
fix: always use linux in GuASG

akash1810:
## What does this change?

We don't use anything other than Linux, so we can hard code the osType.

Actually, are more accurate statement is we shouldn't be using anything other than Linux as AMIgo doesn't support it so we cannot guarantee non-Linux instances are being kept up to date.

Nope. It's was an optional prop and we don't appear to be explicitly setting it anywhere.

## How to test

n/a

Simpler signature.

n/a


stephengeller (approved):
Seems sensible, nice!

akash1810:
This was an effort to simplify the instantiation of an ASG. I think I've worked out an even better version of this, so closing for now. The diff on the next version might be a little big 😞 .

PR #267, opened 2021-02-22
feat: tweak type of GuASG's userData

akash1810:
## What does this change?

Update the type of `userData` in `GuAutoScalingGroup` to be `UserData | string`. This allows it to be more strongly typed.

A side-effect of this change is hard coding the `osType` to linux. Currently all our stacks run a flavour of linux, so this shouldn't be too controversial. There are ambitions from InfoSec and Enterprise to run Windows machines in AWS, however I think we can cross that bridge when we come to it as it'll likely mean other tools, such as AMIgo, need updating too.

This change also paves the way auto generating the user data for the most basic launch configurations of:
- Download config
- Download artefact
- Run artefact

This will be added in a subsequent PR - https://github.com/guardian/cdk/pull/269.

No mandatory, but we can change the type of `userData`.

## How to test

Existing tests have been updated, so observe CI?

Stricter types.

n/a

### DIFF
--- a/src/constructs/autoscaling/asg.ts
+++ b/src/constructs/autoscaling/asg.ts
@@ -1,6 +1,6 @@
 import type { AutoScalingGroupProps, CfnAutoScalingGroup } from "@aws-cdk/aws-autoscaling";
 import { AutoScalingGroup } from "@aws-cdk/aws-autoscaling";
-import type { ISecurityGroup, MachineImage, MachineImageConfig, S3DownloadOptions } from "@aws-cdk/aws-ec2";
+import type { ISecurityGroup, MachineImage, MachineImageConfig } from "@aws-cdk/aws-ec2";
 import { InstanceType, OperatingSystemType, UserData } from "@aws-cdk/aws-ec2";
 import type { ApplicationTargetGroup } from "@aws-cdk/aws-elasticloadbalancingv2";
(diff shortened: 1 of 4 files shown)


src/constructs/autoscaling/asg.ts
```diff
@@ -13,23 +13,24 @@ export interface GuAutoScalingGroupProps
...
+    const userData = props.userData instanceof UserData ? props.userData : UserData.custom(props.userData);
+
     // We need to override getImage() so that we can pass in the AMI as a parameter
     // Otherwise, MachineImage.lookup({ name: 'some str' }) would work as long
     // as the name is hard-coded
     function getImage(): MachineImageConfig {
       return {
-        osType: props.osType ?? OperatingSystemType.LINUX,
-        userData: UserData.custom(props.userData),
+        osType: OperatingSystemType.LINUX,
```
jacobwinch, line 32: Minor point: it might be worth mentioning this change in the PR description so that it's documented somewhere?

jacobwinch (approved):
This looks great - thanks for breaking this work up into small PRs 👍

src/constructs/autoscaling/asg.ts
akash1810: Done.

PR #268, opened 2021-02-22
feat!: refactor security groups in GuAutoScalingGroup

akash1810:
## What does this change?

Simplify how to define security groups in `GuAutoScalingGroup`. Currently, we can specify `securityGroup` (singular) _and_ `securityGroups` (multiple), where `securityGroup` overrides the [default one](https://github.com/aws/aws-cdk/blob/296a10d76a9f6fc2a374d1a6461c460bcc3eeb79/packages/%40aws-cdk/aws-autoscaling/lib/auto-scaling-group.ts#L913-L916) provided by AWS.

In this PR, we set the default security group to allow outgoing traffic on port 443 (https); this is tighter than AWS's default.

By setting the default security group, the props for `GuAutoScalingGroup` no longer needs the `securityGroup` field 🎉 .

Lastly, we're renaming `securityGroups` to `additionalSecurityGroups` to be more explicit; this is inspired by [`GuInstanceRole`](https://github.com/guardian/cdk/blob/974d1a864252e8d2344c512305266ea925276e0b/src/constructs/iam/roles/instance-role.ts).

Yes! The signature of `GuAutoScalingGroup` has changed.

## How to test

Observe CI.

Simpler API.

n/a

### DIFF
--- a/src/constructs/ec2/security-groups.ts
+++ b/src/constructs/ec2/security-groups.ts
@@ -70,3 +70,18 @@ export class GuPublicInternetAccessSecurityGroup extends GuSecurityGroup {
     });
   }
 }
+
+export class GuHttpsEgressSecurityGroup extends GuSecurityGroup {
+  constructor(scope: GuStack, id: string, props: SecurityGroupProps) {
+    super(scope, id, {
+      vpc: props.vpc,
+      allowAllOutbound: false,
+      description: "Allow all outbound traffic on port 443",
+      egresses: [{ range: Peer.anyIpv4(), port: Port.tcp(443) }],
+    });
+  }
+
+  public static forVpc(scope: GuStack, props: SecurityGroupProps): GuHttpsEgressSecurityGroup {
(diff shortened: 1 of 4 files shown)


src/constructs/ec2/security-groups.ts
```diff
@@ -70,3 +70,18 @@ export class GuPublicInternetAccessSecurityGroup extends GuSecurityGroup {
     });
   }
 }
+
+export class GuHTTPSEgressSecurityGroup extends GuSecurityGroup {
```
jacobwinch, line 74: Not sure if there is a proper convention for cases like this, but I find `GuHttpsEgressSecurityGroup` slightly more readable. Feel free to ignore without comment if you disagree!

jacobwinch (approved):
Thanks for updating this - I find the new API much more intuitive 👍

src/constructs/ec2/security-groups.ts
akash1810: Done!

PR #269, opened 2021-02-23
feat: auto generating UserData

akash1810:
## What does this change?

Following the change in #267 , introduce a way to autogenerate user data for the ideal scenario of:
- download config
- download distributable
- run distributable

This is part of a wider task of creating an EC2 based pattern.

No.

## How to test

See added tests.

For the simplest use case, the library has an easier API.

n/a

### DIFF
--- a/src/constructs/autoscaling/user-data.ts
+++ b/src/constructs/autoscaling/user-data.ts
@@ -11,6 +11,7 @@ import type { GuStack } from "../core";
 export interface GuUserDataS3DistributableProps {
   bucketName: string;
   fileName: string;
+  executionStatement: string; // TODO can we detect this and auto generate it? Maybe from the file extension?
 }
 
 /**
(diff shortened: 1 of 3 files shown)


src/constructs/autoscaling/user-data.ts
```diff
@@ -1,13 +1,77 @@
 import type { S3DownloadOptions } from "@aws-cdk/aws-ec2";
 import { UserData } from "@aws-cdk/aws-ec2";
+import { Bucket } from "@aws-cdk/aws-s3";
+import type { GuStack } from "../core";
+
+export interface GuUserDataS3DistributableProps {
```
jacobwinch, line 6: I think it's worth adding some basic docs for these interfaces now so that we don't forget?

jacobwinch (approved):
Great work 👍

src/constructs/autoscaling/user-data.ts
akash1810: Done.

PR #271, opened 2021-02-24
chore: enable auto publishing 🎉 

akash1810:
## What does this change?

Enables auto publishing of this library to NPM 🎉 .

We're moving from [`np`](https://github.com/sindresorhus/np) to [`semantic-release`](https://github.com/semantic-release/semantic-release) as `semantic-release` has a bit more automation, it will:
- Craft structured release notes, with emoji ❗ (`np`'s release notes appears to be the git history)
- Comment on PRs to say it's been shipped (similar to PRout)

As we have branch protection on `main`, we cannot use [`semantic-release/git`](https://github.com/semantic-release/git) to commit the new version number in `package.json` - we still need to create a PR, approve it and merge it. This can be automated with [`guardian/post-release-action`](https://github.com/guardian/post-release-action)  🎉 .

Note: we need to add two GitHub secrets to the repo for this:
- `NPM_TOKEN` to enable publishing to npm
- `POST_RELEASE_TOKEN` to raise a PR that triggers CI workflows

## Does this change require changes to existing projects or CDK CLI?

No.

## How to test

Merge to main 😱 .

## How can we measure success?

Releases are fully automated, no more "Can someone +1 this version bump PR quickly?".

## Have we considered potential risks?

n/a

PR #276, opened 2021-02-25
chore(docs): update readme to reflect new release process

akash1810:
## What does this change?

Removes instructions for manual release steps from README. Also updates code snippets to use an actual pattern that exists in the library.

See https://github.com/guardian/cdk/pull/271.

No.

## How to test

Look at the changes - https://github.com/guardian/cdk/blob/aa-readme/README.md?

Keeping up to date.

n/a

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ TL;DR We release new versions of the library to NPM automagically.
 We use [semantic-release](https://github.com/semantic-release/semantic-release) and [guardian/post-release-action](https://github.com/guardian/post-release-action) to automate releases.
 
 To release a new version:
-1. Raise a PR following using [Angular](https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#commits) / [Karma](http://karma-runner.github.io/6.1/dev/git-commit-msg.html) commits
+1. Raise a PR using [Angular](https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#commits) / [Karma](http://karma-runner.github.io/6.1/dev/git-commit-msg.html) commits
 1. Get your PR reviewed and approved
 1. Merge your PR
 1. Wait... The robots will parse your commits to determine the next version number (following [semantic versioning](https://semver.org/)), release to npm and update `package.json`. Automatically.
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # Guardian CDK Library
 
+![npm](https://img.shields.io/npm/v/@guardian/cdk?style=flat-square)
+
 Welcome to the Guardian CDK library! This library contains a number of reusable patterns and constructs which can be used to build up your AWS Cloudformation stacks.
 
 📖 View the available components in the [API documentation](https://guardian.github.io/cdk/).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -68,23 +70,32 @@ or
...
-> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.
+We use [semantic-release](https://github.com/semantic-release/semantic-release) and [guardian/post-release-action](https://github.com/guardian/post-release-action) to automate releases.
 
-For simplicity, this library only has minor version releases whilst still on a 0 major number.
+To release a new version:
+1. Raise a PR following using [Angular](https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#commits) / [Karma](http://karma-runner.github.io/6.1/dev/git-commit-msg.html) commits
+1. Get your PR reviewed and approved
+1. Merge your PR
+1. Wait... The robots will parse your commits to determine the next version number (following [semantic versioning](https://semver.org/)), release to npm and update `package.json`. Automatically.
+1. Enjoy a comment on your PR to inform you that your change has been released.
```
jamie-lynch, line 101: AFAIK PR titles can be used if you squash and merge as that's the commit you get on the main branch. Alternatively, you can ensure the commits conform and then merge as usual. At the moment the PR title is validated in CI and the commits don't have to conform but if you were to merge (rather than squash + merge) with a conforming PR and non-conforming commits, you wouldn't trigger a release.

I don't know if you want to enforce one style over the other or allow both. If it's both, maybe this should be a bit clearer on the two options and if it's one over the other, then the checks may want to change accordingly?

README.md
```diff
@@ -68,23 +70,32 @@ or
 yarn add @guardian/cdk
 ```
 
-Patterns can be imported from the top level of the library (e.g. `import { InstanceRole } from "@guardian/cdk";`) while constructs must be imported from their construct directory (e.g. `import { GuAutoScalingGroup } from "@guardian/cdk/lib/constructs/autoscaling";`)
+Patterns can be imported from the top level of the library:
 
-There are more details on using the CDK library in [docs](docs)
+```typescript
+import { GuScheduledLambda } from "@guardian/cdk";
```
jamie-lynch, line 76: 🎉  Yay for a real example!

README.md
akash1810: Ah, great point. Will discuss merge styles with team next week and update accordingly.

README.md
```diff
@@ -68,23 +70,32 @@ or
...
 
-This library follows [semantic versioning](https://semver.org/). Part of the semantic versioning specification [states](https://semver.org/#spec-item-4):
+TL;DR We release new versions of the library to NPM automagically.
 
-> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.
+We use [semantic-release](https://github.com/semantic-release/semantic-release) and [guardian/post-release-action](https://github.com/guardian/post-release-action) to automate releases.
 
-For simplicity, this library only has minor version releases whilst still on a 0 major number.
+To release a new version:
+1. Raise a PR following using [Angular](https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#commits) / [Karma](http://karma-runner.github.io/6.1/dev/git-commit-msg.html) commits
```
jacobwinch, line 97:
```suggestion
1. Raise a PR using [Angular](https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#commits) / [Karma](http://karma-runner.github.io/6.1/dev/git-commit-msg.html) commits
```

README.md
jacobwinch: I think you can [configure the repo so that you can _only_ use squash and merge](https://docs.github.com/en/github/administering-a-repository/configuring-commit-squashing-for-pull-requests)? This would probably get my vote as it means we are only enforcing conventions where it is strictly necessary. It also means that if someone creates a load of commit messages which don't conform, they can easily fix things by updating their PR title and hitting merge, which reduces friction for contributors.

README.md
```diff
@@ -68,23 +70,32 @@ or
...
-Patterns can be imported from the top level of the library (e.g. `import { InstanceRole } from "@guardian/cdk";`) while constructs must be imported from their construct directory (e.g. `import { GuAutoScalingGroup } from "@guardian/cdk/lib/constructs/autoscaling";`)
+Patterns can be imported from the top level of the library:
 
-There are more details on using the CDK library in [docs](docs)
+```typescript
+import { GuScheduledLambda } from "@guardian/cdk";
+```
 
-## Releasing
+While constructs must be imported from their construct directory:
```
jacobwinch, line 79: Perhaps we should repeat the point about using patterns over constructs here (in case people skip straight to the `Usage` section?

```suggestion
We encourage you to use patterns rather than constructs whenever possible. If you need to use a construct directly, they must be imported from their construct directory:
```

README.md
```diff
@@ -68,23 +70,32 @@ or
...
 
-We use [`np`](https://www.npmjs.com/package/np) to help orchestrate the release process.
-To release a new version, run `./script/release`. You will need to be logged in to your `npm` account (`npm login`) which must be part of the Guardian organisation. If you have 2fa enabled, you will be prompted for an OTP during the release process.
+```typescript
+import { GuAutoScalingGroup } from "@guardian/cdk/lib/constructs/autoscaling";
+```
 
-The script automatically prepares a [GitHub Release](https://docs.github.com/en/github/administering-a-repository/about-releases) and opens a pre-populated form in your browser. Please complete the GitHub Release creation manually.
+This is intentional as the patterns ideally solve the majority of use-cases.
+if they don't, PRs are welcome!
```
jacobwinch, line 86: Perhaps we want to encourage users who are working with constructs directly to tell us why they're having to do this too:

```suggestion
If they don't, please let us know about your use-case so that we can consider supporting it via a pattern. Alternatively, PRs are always welcome!
```

README.md
akash1810: Happy to change this setting. IIUC (for all intents and purposes) the only impact is the merge button on a PR is worded differently - there is no impact to our git workflow otherwise.

@stephengeller any objections to changing the repo settings to only allow squash and merge?

akash1810: Have changed the repo settings now to not allow merge commits; I think this resolves this thread?

jamie-lynch: Maybe the first point should be more explicit that it's only the PR title which needs to conform to the style. And similar with the penultimate point about `parse you commits`?

```
1. Raise a PR with a title that conforms to [Angular](https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#commits) / [Karma](http://karma-runner.github.io/6.1/dev/git-commit-msg.html) standards
...
1. Wait... The robots will parse your merge commit to determine the next version number (following [semantic versioning](https://semver.org/)), release to npm and update `package.json`. Automatically.
```

akash1810: Have reworded the entire section as it was pretty verbose.

jacobwinch (approved):
Great work 👍

PR #282, opened 2021-02-26
chore: auto deployment of docs 🙏

akash1810:
## What does this change?

Hopefully this works!

In #271 we enabled the automation of releasing to npm. Part of this process will create a git tag.

We had a GH Action workflow to run `./script/cd-docs` when a new tag was created. However, it appears that the default GitHub token (`secrets.GITHUB_TOKEN`) will not trigger another workflow. That is, the workflow `cd-docs.yaml` never gets triggered and thus the docs are not updated.

The [suggested solution](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#triggering-new-workflows-using-a-personal-access-token) to this is to use a custom token. We did this in #277 however it still doesn't seem to work, and so we can go back to using the standard token.

I can't think why we were updating docs in a seperate Action in the first place. Maybe to guarentee the need to update the docs?

Doing the update in the main `cd` script should be safe as:
  - If semantic-release determines a release should be made, it will publish to npm. We'll then run `npm run release:docs` to generate docs and push to the `gh-pages` branch.
  - If semantic-release determines a release should not be made, it'll do nothing. We'll then run `npm run release:docs` to generate docs and push to the `gh-pages` branch. However, as there will be no diff there will be nothing to push. That is, it's a no-op.

In both scenarios, the exit code from `npm run release:docs` will be 0 and so the Action will continue to the `post-release` step.

This is actually preferable as it reduces magic and also means we're more CI agnostic, should we ever want to move off GitHub Actions.

No.

## How to test

- Merge to `main`.
- Raise a PR that'll create a new version.
- Merge said PR.
- Witness a new version published to npm and the docs getting updated.

Docs stay up to date with releases.

n/a


jamie-lynch (approved):
Good thinking! This is much simpler than trying to get it work in a separate workflow file.

akash1810:
Reverting this as the rebase went wrong! The cd-docs.yaml file should have been deleted rather than emptied out!

PR #293, opened 2021-03-02
refactor: make SSM policy extend GuAllowPolicy for DRYness

akash1810:
## What does this change?

`GuAllowPolicy`'s signature makes it really simple to create policies, let's use it to DRY out `GuSSMRunCommandPolicy`.

No.

## How to test

Existing tests should continue to pass.

DRY.

n/a


jacobwinch (approved):
Nice work 👍

PR #294, opened 2021-03-02
feat(secrets): add constructs for accessing SSM parameter store parameters

stephengeller:
## What does this change?

This PR adds two new constructs:

### `GuSSMStringParameter`

This can be used to access string values in SSM. It can be used like so:

```ts
// Creates CFN parameter, uses latest version
GuSSMStringParameter(stack, "/path/to/some-secret-name");

// Can optionally specify a version, uses `resolve:ssm`
GuSSMStringParameter(stack, "/path/to/some-secret-name", 4);
```

Currently, there are limitations on how we can use SSM parameters in CDK, namely that **[you can't dynamically point to the latest version of a parameter unless it's created as a CFN parameter (docs)](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/dynamic-references.html#dynamic-references-ssm-pattern)**. This means that you can't make use of deploy-time variables such as the `stage`. So, for now, we can simply have a construct that requires a full SSM path rather than just the secret name, and improve the construct at a later date if & when AWS make it possible for SSM parameters to refer to CFN parameters (like stage) as well as not need to specify a version. It would look like this:

```ts
// This would assume the full path to be `/${stack.stage}/${stack.stack}/${stack.app}/secret-name`
GuSSMStringParameter(stack, "secret-name") 
```

However, there is still value in the above construct as some parameters are global to the account (eg vpcs) and don't rely on stage values.

### `GuSSMSecureStringParamter`

This can be used to access **secure string** values in SSM. It can be used like so:

```ts
// Requires specifying a version, uses `resolve:ssm-secure`
 GuSSMSecureStringParameter(stack, "/path/to/some-secret-name", 4);
```

This isn't super practical _yet_, but hopefully will become more useful as time progresses as we'd hope for AWS to allow us to dynamically fetch the latest version of an SSM parameter value at some point in the future, at which point you can drop the version argument and be confident in getting the latest version, every time.

No. 

## How to test

`./script/test` from the root of the repo.

Projects can implement these constructs instead of creating CFN parameters. For now, this would be mostly in the form of replacing global SSM values (eg vpcs) that are either manually entered in CFN parameters or otherwise specified in CDK, but these constructs will hopefully provide more value as AWS provide more functionality for querying Parameter Store.


src/constructs/core/ssm.ts
```diff
@@ -0,0 +1,69 @@
...
+interface GuSSMParameterProps {
+  version?: number;
+  path?: string; // In practice, some stacks will want to share usage of an account-wide SSM parameter (such as a vpc ID)
+  type: SSMParameterType;
+}
+
+type GuSSMParameterNoTypeProps = Omit<GuSSMParameterProps, "type">;
+type GuSSMSecureStringParameterProps = Omit<GuSSMParameterNoTypeProps, "version">;
+
+function GuSSMParameter(
```
akash1810, line 20: Does this mean the [`fromSSM` prop for `GuParameter`](https://github.com/guardian/cdk/blob/cd14aefa11b327bb686a5a379676d2cf4af3da0d/src/constructs/core/parameters.ts#L7) is redundant now?

src/constructs/core/ssm.ts
```diff
@@ -0,0 +1,69 @@
+import type { IStringListParameter, IStringParameter } from "@aws-cdk/aws-ssm";
```
akash1810, line 1: Let's export this file in the [index](https://github.com/guardian/cdk/blob/main/src/constructs/core/index.ts) too?

src/constructs/core/ssm.ts
akash1810: Should these be co-located with the other parameters? Or should the file be called `ssm-parameters` for consistency?

akash1810: Any reason these aren't classes?

akash1810 (commented):
I think this makes sense. Have added a few questions/comments.

stephengeller:
I've created a document to discuss how to move forward with this, as more thought is needed as to how to approach this work. Once that is done, we can re-explore how to proceed with this.

PR #296, opened 2021-03-03
feat: Add pattern for Lambda which processes Kinesis Stream

jacobwinch:
## What does this change?

This PR adds a pattern for lambdas which process Kinesis streams.

Example usage:

```typescript
new GuKinesisLambda(stack, "my-lambda-function", {
  code: { bucket: "test-dist", key: "lambda.zip" },
  functionName: "my-lambda-function",
  handler: "my-lambda/handler",
  runtime: Runtime.NODEJS_14_X,
  errorHandlingConfiguration: {
    bisectBatchOnError: true,
    retryBehaviour: StreamRetry.maxAttempts(3),
  },
  monitoringConfiguration: {
    toleratedErrorPercentage: 1,
    snsTopicName: "alerts-topic",
  },
});
```

## How to test

1. Create a new stack based on this pattern. You should end up with a functional lambda triggered whenever new records are put onto a Kinesis stream.
1. Migrate existing stacks to use this pattern:
    1. A stack which defines its own Kinesis stream
    1. A stack which uses an external Kinesis stream

* It should be faster to set up new projects correctly using this pattern
* When migrating existing projects, users can configure sensible monitoring for a tiny amount of extra code/effort
* When migrating existing projects, users will be encouraged to configure appropriate error handling. Currently this is optional, which means that we often use [AWS's (potentially unhelpful) defaults](https://aws.amazon.com/blogs/compute/new-aws-lambda-controls-for-stream-processing-and-asynchronous-invocations/).

It's tricky to test the migration scenarios (outside of unit tests) as Developer Experience don't seem to have any stacks which use this architecture. We will need to find some volunteers to work on trying this pattern out with us.


src/patterns/kinesis-lambda.ts
```diff
@@ -0,0 +1,111 @@
...
+    bisectBatchOnError: errorHandlingProps.bisectBatchOnError,
+    onFailure: errorHandlingProps.deadLetterQueueForSkippedRecords,
+    ...errorHandlingProps.retryBehaviour.toAwsProp(),
+  };
+}
+
+type AwsRetryProp = Pick<AwsErrorHandlingProps, "maxRecordAge" | "retryAttempts">;
+type RetryType = "attempts" | "recordAge";
+
+export class Retry {
```
akash1810, line 41: I think it'd be nice for patterns to be super simple to make it easy to follow, both in terms of their orchestration of constructs but also in terms of their source code - can we move this class into it's own file? I think some of the interfaces could also be moved too.

akash1810 (commented):
Looks good so far, added a comment on readability/style.

src/patterns/kinesis-lambda.ts
jacobwinch: Good idea. I've moved a lot of the 'helper' code into [its own file](https://github.com/guardian/cdk/pull/296/files#diff-c4842322ffd7111821d75fbe3cce9aecb7f630bc22e877934ae644a947f3c9d1) at the construct level.

src/constructs/lambda/event-sources.ts
```diff
@@ -0,0 +1,93 @@
...
+ * }
+ * ```
+ *
+ * Records which could not be processed successfully can (optionally) be sent to a dead letter queue
+ * via `deadLetterQueueForSkippedRecords`.
+ */
+export interface StreamErrorHandlingProps {
+  retryBehaviour: StreamRetry;
+  bisectBatchOnError: boolean;
+  deadLetterQueueForSkippedRecords?: IEventSourceDlq;
```
jacobwinch, line 44: We could think about providing a queue for users if they select this option, but this PR is already getting quite big!

src/constructs/kinesis/kinesis-stream.ts
```diff
@@ -0,0 +1,15 @@
+import type { CfnStream, StreamProps } from "@aws-cdk/aws-kinesis";
+import { Stream } from "@aws-cdk/aws-kinesis";
+import type { GuStack } from "../core";
+
+export interface GuKinesisStreamProps extends StreamProps {
```
jacobwinch, line 5: I'm not that knowledgeable about Kinesis... I wonder if we are happy with the [CFN default property values](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-kinesis-stream.html#aws-resource-kinesis-stream-properties)?

src/constructs/lambda/event-sources.ts
```diff
@@ -0,0 +1,93 @@
...
+/**
+ * Use with caution! This will cause your lambda to continuously retry if it encounters errors
+ * whilst processing a batch. If the errors are permanent (e.g. due to a bad record) your lambda
+ * will fall behind with stream processing i.e. it will only move onto new batches when the problematic
+ * records expire. Consider using [[`StreamErrorHandlingProps`]] instead.
+ *
+ * When using this approach to error handling, ensure that an alarm is configured to monitor
+ * lambda failures.
+ */
+export interface BlockProcessingAndRetryIndefinitely {
```
jacobwinch, line 19: Hopefully this is ugly enough to put users off!

src/patterns/kinesis-lambda.ts
```diff
@@ -0,0 +1,135 @@
...
+ * Other alarm properties (e.g. alarm name and description) will be pre-populated with sensible defaults.
+ * For a full list of optional properties, see [[`GuLambdaErrorPercentageMonitoringProps`]].
+ *
+ * If your team do not use CloudWatch, it's possible to opt-out with the following configuration:
+ * ```typescript
+ *  monitoringConfiguration: { noMonitoring: true } as NoMonitoring
+ * ```
+ */
+export interface GuKinesisLambdaProps extends Omit<GuFunctionProps, "rules" | "apis" | "errorPercentageMonitoring"> {
+  monitoringConfiguration: NoMonitoring | GuLambdaErrorPercentageMonitoringProps;
```
jacobwinch, line 90: We could consider different options for alarms here (e.g. alarms based on Kinesis metrics, or on the number of dead letters if a queue is configured), but as this PR is already pretty big we might want to tackle that as a second card/PR?

akash1810:
Can we add a code snippet for how one could use this pattern, for future travellers?

src/constructs/lambda/event-sources.ts
```diff
@@ -0,0 +1,93 @@
...
+  }
+  public toAwsProp(): AwsRetryProp {
+    return this.retryType === "attempts"
+      ? { retryAttempts: this.amount }
+      : { maxRecordAge: Duration.seconds(this.amount) };
+  }
+  private readonly amount: number;
+  readonly retryType: RetryType;
+  // eslint-disable-next-line custom-rules/valid-constructors -- TODO only lint for things that extend IConstruct
+  private constructor(amount: number, type: RetryType) {
```
akash1810, line 89: TIL private constructors!

src/constructs/lambda/event-sources.ts
```diff
@@ -0,0 +1,93 @@
+import type { IEventSourceDlq } from "@aws-cdk/aws-lambda";
```
akash1810, line 1: This file is confusing me a little, can we discuss IRL?

src/constructs/lambda/event-sources.ts
jacobwinch: I also learnt about private constructors via this PR! I wanted to provide something similar to `Duration.minutes` so I borrowed heavily from that [AWS code](https://github.com/aws/aws-cdk/blob/b96efd8aaa143845b9fe315a9ee1e8398c4d83c2/packages/%40aws-cdk/core/lib/duration.ts).

src/constructs/kinesis/kinesis-stream.ts
jacobwinch: > I wonder if we can abstract these settings at the pattern level?

I like this idea; it could be a nice future improvement for this pattern.

src/utils/lambda/event-source.ts
```diff
@@ -0,0 +1,80 @@
+import type { IEventSourceDlq } from "@aws-cdk/aws-lambda";
```
jacobwinch, line 1: I'll need to move some [existing code](https://github.com/guardian/cdk/tree/d1002d06d8d7f8bcae963e0485033d654cb3e094/src/utils) into more specific subdirectories in order to match this new structure. I will do that in its own PR as it's not directly related to this pattern.

src/constructs/lambda/event-sources.ts
jacobwinch: I think I've made all of the changes that we discussed now @akash1810 - let me know if I missed anything!

src/utils/lambda/event-source.ts
```diff
@@ -0,0 +1,80 @@
...
+ * const errorHandlingProps: ErrorHandlingProps = {
+ *   bisectBatchOnError: true,
+ *   retryBehaviour: StreamRetry.maxAttempts(5),
+ * }
+ * ```
+ *
+ * Records which could not be processed successfully can (optionally) be sent to a dead letter queue
+ * via `deadLetterQueueForSkippedRecords`.
+ */
+export interface StreamErrorHandlingProps {
```
akash1810, line 28: Great, clear, comments!

akash1810 (approved):
Brilliant! Thanks for your patience with my feedback!

LGTM 👍🏽

src/utils/lambda/event-source.ts
jacobwinch: https://github.com/guardian/cdk/pull/314

PR #298, opened 2021-03-04
feat: GuUserData supports additional statements

akash1810:
## What does this change?

Update `GuUserData` to allow additional statements to be specified. These will get run _after_ configuration is downloaded and _before_ the distributable is downloaded and executed.

There are two ways to interact with this prop:

1. An array of strings:

```typescript
const props: GuUserDataProps = {
  distributable: {
    bucket: new GuDistributionBucketParameter(stack),
    fileName: "my-app.deb",
    executionStatement: `dpkg -i /${stack.app}/my-app.deb`,
  },
  configuration: {
    bucketName: "test-app-config",
    files: ["secrets.json", "application.conf"],
  },
  additionalStatements: [
    "mkdir /etc/gu",
    "cat > /etc/gu/my-application.conf <<-'EOF'",
    `  include "application"`,
    "EOF",
  ],
};
```

2. With `GuUserData.stripMargin`:

```typescript
const props: GuUserDataProps = {
  distributable: {
    bucket: new GuDistributionBucketParameter(stack),
    fileName: "my-app.deb",
    executionStatement: `dpkg -i /${stack.app}/my-app.deb`,
  },
  configuration: {
    bucketName: "test-app-config",
    files: ["secrets.json", "application.conf"],
  },
  additionalStatements: [
    GuUserData.stripMargin`
    |mkdir /etc/gu
    |cat > /etc/gu/my-application.conf <<-'EOF'
    |  include "application"
    |EOF`,
  ],
};
```

`GuUserData.stripMargin` is an approximation of [Scala's stripMargin](https://docs.scala-lang.org/overviews/scala-book/two-notes-about-strings.html#multiline-strings); the hope is it makes it easier to reason about multiline bash statements.

The test have also been refactored to make them a little DRY.

This is inspired by recent pairing to update [Typerighter](https://github.com/guardian/typerighter/blob/beffe0582e9b0776d5d39e7d1098d281c42fb5a8/cdk/lib/rule-manager/rule-manager.ts#L118-L126) - cc @jonathonherbert @SHession.

No - the new prop is optional.

## How to test

See tests/CI.

User data is even simpler to write. I think we can default the value of `distributable` to something sensible to make user data even simpler. This is for another PR though.

The API might be a bit confusing, especially for the `stripMargin` function.

### DIFF
--- a/src/constructs/autoscaling/user-data.test.ts
+++ b/src/constructs/autoscaling/user-data.test.ts
@@ -3,6 +3,7 @@ import { Vpc } from "@aws-cdk/aws-ec2";
 import { Stack } from "@aws-cdk/core";
 import { simpleGuStackForTesting } from "../../../test/utils";
 import { Stage } from "../../constants";
+import type { GuStack } from "../core";
 import { GuDistributionBucketParameter } from "../core";
 import { GuAutoScalingGroup } from "./asg";
 import type { GuUserDataProps } from "./user-data";
@@ -13,6 +14,33 @@ describe("GuUserData", () => {
     vpcId: "test",
     availabilityZones: [""],
     publicSubnetIds: [""],
+  });
+
+  const createAsg = (stack: GuStack, props: GuUserDataProps) => {
+    return new GuAutoScalingGroup(stack, "AutoscalingGroup", {
+      vpc,
+      userData: new GuUserData(stack, props).userData,
+      stageDependentProps: {
+        [Stage.CODE]: {
+          minimumInstances: 1,
+        },
+        [Stage.PROD]: {
+          minimumInstances: 3,
+        },
+      },
+    });
+  };
+
+  test("GuUserData.stripMargin", () => {
+    const actual = GuUserData.stripMargin`
+      |mkdir /etc/gu
+      |cat > /etc/gu/my-application.conf <<-'EOF'
+      |  include "application"
+      |EOF
+      |`;
+
+    const expected = `\nmkdir /etc/gu\ncat > /etc/gu/my-application.conf <<-'EOF'\n  include "application"\nEOF\n`;
+    expect(actual).toEqual(expected);
   });
 
   test("Distributable should be downloaded from a standard path in S3 (bucket/stack/stage/app/filename)", () => {
@@ -26,20 +54,7 @@ describe("GuUserData", () => {
       },
     };
 
-    const { userData } = new GuUserData(stack, props);
-
-    new GuAutoScalingGroup(stack, "AutoscalingGroup", {
-      vpc,
-      userData,
-      stageDependentProps: {
-        [Stage.CODE]: {
-          minimumInstances: 1,
-        },
-        [Stage.PROD]: {
-          minimumInstances: 3,
-        },
-      },
-    });
+    createAsg(stack, props);
 
     expect(stack).toHaveResource("AWS::AutoScaling::LaunchConfiguration", {
       UserData: {
@@ -78,20 +93,7 @@ describe("GuUserData", () => {
       },
     };
 
-    const { userData } = new GuUserData(stack, props);
-
-    new GuAutoScalingGroup(stack, "AutoscalingGroup", {
-      vpc,
-      userData,
-      stageDependentProps: {
-        [Stage.CODE]: {
-          minimumInstances: 1,
-        },
-        [Stage.PROD]: {
-          minimumInstances: 3,
-        },
-      },
-    });
+    createAsg(stack, props);
 
     expect(stack).toHaveResource("AWS::AutoScaling::LaunchConfiguration", {
       UserData: {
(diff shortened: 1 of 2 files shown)


src/constructs/autoscaling/user-data.test.ts
```diff
@@ -15,6 +16,33 @@ describe("GuUserData", () => {
...
+  };
+
+  test("GuUserData.stripMargin", () => {
+    const actual = GuUserData.stripMargin`
+      |mkdir /etc/gu
+      |cat > /etc/gu/my-application.conf <<-'EOF'
+      |  include "application"
+      |EOF
+      |`;
+
```
stephengeller, line 41: I'm a bit confused about how `stripMargin` works, especially the ``` stripMargin`some-string` ``` syntax as I can see that it takes a `template`, and an `args` parameter. Do we need to test when it takes two params, or is it doing this already in this test?

src/constructs/autoscaling/user-data.ts
```diff
@@ -39,6 +44,27 @@ export interface GuUserDataProps {
...
+   * A helper function to make user data strings easier to write.
+   *
+   * Example usage:
+   * ```typescript
+   * const additionalStatements = GuUserData.stripMargin`
+   *   |mkdir /etc/gu
+   *   |cat > /etc/gu/my-application.conf <<-'EOF'
+   *   |  include "application"
+   *   |EOF
+   *   |`;
```
stephengeller, line 57: I was initially confused about how `stripMargin` worked, but I think this description is good enough to allow library users to use it without confusion. I'm still confused about how the ``` stripMargin`string foo bar baz` ``` syntax works but covered that in an earlier comment.

jonathonherbert:
Discussed out-of-band w/ @akash1810 and we think it'd be better to remove our need for this API rather than implement, as removal should be fairly straightforward. 

The question of adding config in this way for other projects is undecided, but perhaps we could avoid merging this until the need arises elsewhere, to avoid shipping and maintaining code that's (at the moment) not needed.

That said, PR looks lovely ✨ 😁

src/constructs/autoscaling/user-data.test.ts
akash1810: Yeah its pretty confusing and magical isn't it!! i'll confess to copying it from [here](https://gist.github.com/jimschubert/06fea56a6d2a1e7fdbc2).

[Here's](https://www.typescriptlang.org/play?#code/PTAEDMCcHsFtQBYBckAcDOAuEBzAlukgHT5IICuARkQMZzABWes6NFlAppEsAAwBs4DgEMArP2H8AJgCZhARg4B2cFMo0ZAWABQO8OQB2NJHmgHQhSHlQBZYZHzmAFEg6xUAG2GvMoACpunt4cAMpIVgY46ACCkJDCAJ4ANKBEafZRvoYA1gbQAO4GANoAugCUvpZ4kaAA3jqgoHQG6NAeHEQe0DgugV6uZQ1NZq3tnd1OGeiDutqNzYSgkBzo5B5IoAC8oK7u-R3LUuQ0HE6TNDQpqPZIKXhlWwB8oEXCFylTRXigALSg8iUrjcSkQGNBqk4AESQsplADcQ2WSHIkHMy1W6yIyyCJycwAAOpAAPxOfEGMr49AAKnxAB9gDgUpCACTyGEI7QAXw5OgWG1QXFa5m2kOWwikXEhOl5Iw2sBW6GEOA4Wws4Wsdgc1QABkNaQARESQUDM2oCyBCzl6vUATWg5FA9hVwnyKzgHAAhLrZgs2h0uj15ehFcqZjoQGqrLYMhDtYb7CbamKJZBOSkEvbHctHa7WvLtSlSjMgA) an attempt to reason about it, but it doesn't help much!

akash1810:
> better to remove our need for this API rather than implement, as removal should be fairly straightforward.

👍🏽 

I'm increasingly of the opinion that this library should make it really easy to walk down an opinionated road. It will never prevent you from taking a detour, but won't provide the tools to do so.

That is, this library should provide a default user data of:
- get config
- get distributable
- run distributable

If you want to do something different, you have to put the work in. The type system can help you here. This should result in a simpler API and a simpler library codebase too.

src/constructs/autoscaling/user-data.test.ts
```diff
@@ -76,30 +130,69 @@ describe("GuUserData", () => {
...
+      configuration: {
+        bucketName: "test-app-config",
+        files: ["secrets.json", "application.conf"],
+      },
+      additionalStatements: [
+        "mkdir /etc/gu",
+        "cat > /etc/gu/my-application.conf <<-'EOF'",
+        `  include "application"`,
+        "EOF",
+      ],
```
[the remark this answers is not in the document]
akash1810: `GuUserData.stripMargin` is a utility function that returns a `string` rather than a custom type. I think your suggestion would result in `additionalStatements` being `string | string[]` which might be confusing?

src/constructs/autoscaling/user-data.test.ts
akash1810: Ah, this is helpful https://www.tektutorialshub.com/typescript/typescript-tagged-templates.

> The Tagged Template is a template string, which is tagged by a function. The Tag is a function which has the ability to process the template string. We use it to Tag a template string by placing it before it. We do not use the `()` as in the function call.

akash1810: Here are the MDN docs https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#tagged_templates.

jonathonherbert:
> I'm increasingly of the opinion that this library should make it really easy to walk down an opinionated road. It will never prevent you from taking a detour, but won't provide the tools to do so.

This is a neatly expressed principle that might be worth documenting in some high-level project goals/non-goals ✨

stephengeller:
+1 to the above conclusions, I think that makes sense and should be what we're striving towards.

PR #304, opened 2021-03-08
refactor: split parameters into multiple files

akash1810:
## What does this change?

Whilst the diff is fairly sizeable, this should be a no-op.

Here we split the big `parameters.ts` file into multiple files. This is for a couple of reasons:
- Improve maintainability
- Identify which parameters do not have tests (not sure if we need them?)

The parameters are grouped by concern, which is hopefully easier to reason about. For example `ec2` contains parameters concerning an EC2 based stack:
- instance type
- ami

We're creating a `parameters` module/package (?) with its `index.ts` exporting everything, so for consumers, nothing has really changed. This can be demonstrated by the diff only affecting the parameters and none of the constructs that use them.

### Notes
It's worth noting this is a part of some wider thinking around parameters. I don't think any "concrete" parameters should accept an `id` or `props` as it changes their behaviour.

For example `GuInstanceTypeParameter` currently lets you override the `type` of the param; we have a test that overrides it to a `Number` 😱 .

I think we can enforce this via the linter; indeed I've started thinking about this [here](https://github.com/guardian/cdk/blob/ecf9adb55e8f02c832052554a952a23ed012e406/eslint/logic/gu-parameter-constructors.js#L1-L15).

No.

## How to test

Observe (pre-existing) tests and read the code to judge if maintainability is indeed improved.

Maintainability is improved.

n/a


jacobwinch (approved):
> The parameters are grouped by concern, which is hopefully easier to reason about.

Yep, I agree with this 👍

PR #310, opened 2021-03-09
fix!: add a default description to parameters

akash1810:
This is a multipart PR, each commit is a single change - feedback wanted regarding if it makes sense to bundle them together or as separate PRs.

## What does this change?

This PR makes it easier to reason about GuParameters by:
- Adding a default description which can be overridden if desired
- Not letting the `type` to be overridden to avoid breaking behaviour

It also removes `GuS3ObjectArnParameter` which [doesn't seem to be used](https://github.com/search?q=org%3Aguardian+GuS3ObjectArnParameter&type=code) and I'm not sure where it would be used too?

No, unless we're making use of the incorrect behaviour of overriding the `type` of the parameter.

## How to test

CI should pass.

Parameters are easier to understand both within the repo and also in a deployed stack as they have descriptions against them.

Are the default descriptions clear enough?


jacobwinch (approved):
Looks good to me 👍 

> feedback wanted regarding if it makes sense to bundle them together or as separate PRs.

Personally I'm happy for small, low-risk changes like this to be bundled together - thanks for doing this 👍

PR #312, opened 2021-03-10
refactor!: Share private config interface so it can be used by EC2 Pattern

jacobwinch:
## What does this change?

This refactor allows us to specify information about private configuration at the pattern-level. It seems necessary to do this because two separate constructs (`GuUserData` and `GuInstanceRole`) both need this same information. Similarly, we want users to pass the private config bucket name in via a parameter (we don't want bucket names in public GitHub repos) and we must only define this parameter once.

### How will this work in the context of the pattern?

The EC2 app pattern props will take `privateS3Configuration` as an optional prop:

```typescript
export interface GuEc2AppProps {
    privateS3Configuration?: GuPrivateS3ConfigurationProps
    // other props
}
```

If this prop is defined, we'll be able to wire it through into `GuUserData`. We'll also be able to create the appropriate IAM policies via `GuInstanceRole`'s `additionalPolicies` prop e.g.

```typescript
const additionalPolicies = privateS3Configuration ? [new GuGetPrivateConfigPolicy(this, "GetPrivateConfigFromS3Policy", privateS3Configuration)] : undefined 
new GuInstanceRole(stack, "GuInstanceRole", { additionalPolicies: additionalPolicies });
```

A single CFN parameter will be created and accessed via both constructs.

No, I don't think so.

## How to test

I've tested this by updating a stack to use my local version of this change - it seems to build a sensible looking template! We will need to test this more thoroughly as part of the EC2 Pattern work.

Users (will) only need to specify information about private config in one place. After this refactor, we should be able to easily wire up permissions and user-data, if necessary (see examples above).

I can't think of any in this context.

### DIFF
--- a/src/constructs/autoscaling/user-data.test.ts
+++ b/src/constructs/autoscaling/user-data.test.ts
@@ -73,7 +73,7 @@ describe("GuUserData", () => {
         executionStatement: `dpkg -i /${stack.app}/my-app.deb`,
       },
       configuration: {
-        bucketName: new GuPrivateConfigBucketParameter(stack),
+        bucket: new GuPrivateConfigBucketParameter(stack),
         files: ["secrets.json", "application.conf"],
       },
     };
--- a/src/constructs/autoscaling/user-data.test.ts
+++ b/src/constructs/autoscaling/user-data.test.ts
@@ -3,7 +3,7 @@ import { Vpc } from "@aws-cdk/aws-ec2";
 import { Stack } from "@aws-cdk/core";
 import { simpleGuStackForTesting } from "../../../test/utils";
 import { Stage } from "../../constants";
-import { GuDistributionBucketParameter } from "../core";
+import { GuDistributionBucketParameter, GuPrivateConfigBucketParameter } from "../core";
 import { GuAutoScalingGroup } from "./asg";
 import type { GuUserDataProps } from "./user-data";
 import { GuUserData } from "./user-data";
@@ -73,7 +73,7 @@ describe("GuUserData", () => {
         executionStatement: `dpkg -i /${stack.app}/my-app.deb`,
       },
       configuration: {
-        bucketName: "test-app-config",
+        bucketName: new GuPrivateConfigBucketParameter(stack),
         files: ["secrets.json", "application.conf"],
       },
     };
@@ -99,7 +99,15 @@ describe("GuUserData", () => {
           "Fn::Join": [
             "",
             [
-              "#!/bin/bash\nmkdir -p $(dirname '/etc/testing/secrets.json')\naws s3 cp 's3://test-app-config/secrets.json' '/etc/testing/secrets.json'\nmkdir -p $(dirname '/etc/testing/application.conf')\naws s3 cp 's3://test-app-config/application.conf' '/etc/testing/application.conf'\nmkdir -p $(dirname '/testing/my-app.deb')\naws s3 cp 's3://",
+              "#!/bin/bash\nmkdir -p $(dirname '/etc/testing/secrets.json')\naws s3 cp 's3://",
+              {
+                Ref: "PrivateConfigBucketName",
+              },
+              "/secrets.json' '/etc/testing/secrets.json'\nmkdir -p $(dirname '/etc/testing/application.conf')\naws s3 cp 's3://",
+              {
+                Ref: "PrivateConfigBucketName",
+              },
+              "/application.conf' '/etc/testing/application.conf'\nmkdir -p $(dirname '/testing/my-app.deb')\naws s3 cp 's3://",
               {
                 Ref: "DistributionBucketName",
               },
--- a/src/constructs/core/parameters/s3.ts
+++ b/src/constructs/core/parameters/s3.ts
@@ -12,3 +12,15 @@ export class GuDistributionBucketParameter extends GuStringParameter {
     });
   }
 }
+
+export class GuPrivateConfigBucketParameter extends GuStringParameter {
+  public static parameterName = "PrivateConfigBucketName";
+
+  constructor(scope: GuStack, id: string = GuPrivateConfigBucketParameter.parameterName) {
+    super(scope, id, {
+      description: "SSM parameter containing the S3 bucket name holding the app's private configuration",
+      default: "/account/services/private.config.bucket",
+      fromSSM: true,
+    });
+  }
+}
--- a/src/constructs/iam/policies/s3-get-object.test.ts
+++ b/src/constructs/iam/policies/s3-get-object.test.ts
@@ -3,13 +3,13 @@ import { SynthUtils } from "@aws-cdk/assert/lib/synth-utils";
 import type { SynthedStack } from "../../../../test/utils";
 import { attachPolicyToTestRole, simpleGuStackForTesting } from "../../../../test/utils";
 import { GuDistributionBucketParameter } from "../../core";
-import { GuGetDistributablePolicy, GuGetS3ObjectPolicy } from "./s3-get-object";
+import { GuGetDistributablePolicy, GuGetS3ObjectsPolicy } from "./s3-get-object";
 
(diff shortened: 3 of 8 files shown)


jacobwinch:
> The EC2 app pattern props will take `privateS3Configuration` as an optional prop

Lots of apps don't seem to need this, and I don't think we are pushing `always download your private config from s3` as a current best-practice, so I think this should be optional.

jacobwinch:
> It seems necessary to do this because two separate constructs (`GuUserData` and `GuInstanceRole`) both need this same information.

@akash1810 - you made a similar observation in the `Dev Notes` on [this card](https://trello.com/c/UjVkM1fc/591-create-a-pattern-for-an-ec2-based-app). Does this approach roughly line up with your thinking on this?

src/constructs/core/parameters/s3.ts
```diff
@@ -12,3 +12,15 @@ export class GuDistributionBucketParameter extends GuStringParameter {
...
   }
 }
+
+export class GuPrivateConfigBucketParameter extends GuStringParameter {
+  public static parameterName = "PrivateConfigBucketName";
+
+  constructor(scope: GuStack, id: string = GuPrivateConfigBucketParameter.parameterName) {
+    super(scope, id, {
+      description: "SSM parameter containing the S3 bucket name holding the app's private configuration",
+      default: "/account/services/private.config.bucket",
```
jacobwinch, line 22: Unfortunately there doesn't seem to be a strong convention for this across the department. Accounts which use a single bucket across all apps can make use of this default; others may need to add a few custom parameters in SSM.

akash1810:
> > It seems necessary to do this because two separate constructs (`GuUserData` and `GuInstanceRole`) both need this same information.
> 
> @akash1810 - you made a similar observation in the `Dev Notes` on [this card](https://trello.com/c/UjVkM1fc/591-create-a-pattern-for-an-ec2-based-app). Does this approach roughly line up with your thinking on this?

Yeah, I think it makes sense to move to the pattern creating the variable and passing it down, DI style. It should make it easier to reference values between constructs and we can remove the `getParameter` stuff from `GuStack`.

src/constructs/core/parameters/s3.ts
akash1810: Hmm, this parameter type forces one to create a parameter store entry for the location of their bucket. I wonder if we should allow either a param store path or a plain string?

Maybe one for a later PR as we understand the use case a bit more?

src/constructs/iam/policies/s3-get-object.ts
```diff
@@ -1,23 +1,31 @@
...
 import { GuAllowPolicy } from "./base-policy";
 
 export interface GuGetS3ObjectPolicyProps extends GuNoStatementsPolicyProps {
   bucketName: string;
-  path?: string;
+  paths?: string[];
 }
 
-export class GuGetS3ObjectPolicy extends GuAllowPolicy {
+export class GuGetS3ObjectsPolicy extends GuAllowPolicy {
```
akash1810, line 12: I think this rename warrants a minor version bump? The current PR title (`refactor: `) will be a no-op IIUC.

src/constructs/iam/policies/s3-get-object.test.ts
```diff
@@ -48,6 +48,30 @@ describe("The GuGetS3ObjectPolicy class", () => {
...
+
+    attachPolicyToTestRole(stack, s3Policy);
+
+    expect(stack).toHaveResource("AWS::IAM::Policy", {
+      PolicyDocument: {
+        Version: "2012-10-17",
+        Statement: [
+          {
+            Effect: "Allow",
+            Resource: ["arn:aws:s3:::test/file1.txt", "arn:aws:s3:::test/file2.txt"],
```
akash1810, line 68: This tight scoping is fantastic!

src/constructs/autoscaling/user-data.test.ts
```diff
@@ -99,7 +99,15 @@ describe("GuUserData", () => {
           "Fn::Join": [
             "",
             [
-              "#!/bin/bash\nmkdir -p $(dirname '/etc/testing/secrets.json')\naws s3 cp 's3://test-app-config/secrets.json' '/etc/testing/secrets.json'\nmkdir -p $(dirname '/etc/testing/application.conf')\naws s3 cp 's3://test-app-config/application.conf' '/etc/testing/application.conf'\nmkdir -p $(dirname '/testing/my-app.deb')\naws s3 cp 's3://",
+              "#!/bin/bash\nmkdir -p $(dirname '/etc/testing/secrets.json')\naws s3 cp 's3://",
```
akash1810, line 102: Gosh this generated output is hard to visualise 🤕

akash1810 (approved):
One comment on the PR title, otherwise this is great!

src/constructs/core/parameters/s3.ts
jacobwinch: You're right, this could be a bit of a pain... The advantage of doing it this way is that the parameter value is not private (as far as I know!), so if you do need to change the default, it could at least be set via your `riff-raff.yaml` (which is preferable if you use CD for your infrastructure).

src/constructs/iam/policies/s3-get-object.ts
jacobwinch: Good point. Is it best to add `!`? I think that would be a major version bump, is that what we want here?

src/constructs/autoscaling/user-data.test.ts
jacobwinch: Agreed - this is probably another good argument in support of your conclusion [here](https://github.com/guardian/cdk/pull/298#issuecomment-796525586)!

src/constructs/core/parameters/s3.ts
akash1810: Makes sense.

I was thinking we ultimately only need a string, which can come from:
- A parameter store parameter
- A string parameter
- Hard coded

Thinking about it more, making it a parameter store parameter makes sense as it follows the convention of the distribution bucket (I want to rename artifact bucket to distribution bucket!)

src/constructs/iam/policies/s3-get-object.ts
akash1810: Reading the [docs](https://github.com/semantic-release/commit-analyzer) again, it suggests `refactor: ` is actually ok and what we'd need to do is add `BREAKING CHANGE` to the footer?

From https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#-git-commit-guidelines:
> Footer
> The footer should contain any information about Breaking Changes and is also the place to reference GitHub issues that this commit closes.
>
> Breaking Changes should start with the word BREAKING CHANGE: with a space or two newlines. The rest of the commit message is then used for this.

A major version bump sounds correct. From https://semver.org/:

> MAJOR version when you make incompatible API changes,
> MINOR version when you add functionality in a backwards compatible manner, and
> PATCH version when you make backwards compatible bug fixes.

I think this'll amount to v0.41.0 after reading [this item](https://semver.org/#spec-item-4) in the spec, rather than v1.0.0?

I find this all a bit confusing, but this is likely because I've never written structured commits!

akash1810: More confusion: I think adding `!` is valid for the [PR title linter](https://github.com/amannn/action-semantic-pull-request#validation) to describe a breaking change.

I think I'm going to end up referring to something like [this guide](https://www.conventionalcommits.org/en/v1.0.0/) quite often!

jacobwinch: > I find this all a bit confusing

I'm glad it's not just me 😅 

> I think this'll amount to v0.41.0 after reading [this item](https://semver.org/#spec-item-4) in the spec, rather than v1.0.0?

Ah okay cool, I'd missed that part. I've merged now (with `!`) so let's hope we've figured this out correctly 🤞

PR #316, opened 2021-03-11
chore: switch node versions in script and move to node 14.16.0

akash1810:
### Problem

If we want to update to node 14.16.0 (the latest LTS version) we need to update:
- `.nvmrc`
- `.github/workflows/ci.yaml`
- `.github/workflows/cd.yaml`

Ew!

### Solution

Switch node versions directly in the scripts that are run in ci or cd.

This should make it easier to keep the node version used in dev and ci in-sync as they'll both use the `.nvmrc` file.

### Notes
- `nvm` is available as standard through the [environments that GitHub build](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu1804-README.md)
- The CI job gets updated in this PR so we'd need to update the branch protection rules if we agree this is a good change
- This is an alternative to [this fork](https://github.com/guardian/actions-setup-node#about-this-fork)

No.

## How to test

The `script/ci` script should switch to the node version defined in `.nvmrc`.

image

Changing node versions is simpler as only one file (`.nvmrc`) needs to be updated.

The second commit on this PR moves from node 14.15.1 to 14.16.0 (the latest LTS), by simply changing `.nvmrc`.

n/a


stephengeller (approved):
Nice, looks like a good and straightforward improvement.

sndrs:
i'm curious about why you prefer this approach to using `guardian/actions-setup-node`? 

if GHA actions offered the functionality of the fork by default, would you still go for this approach?

akash1810:
> i'm curious about why you prefer this approach to using `guardian/actions-setup-node`?
> 
> if GHA actions offered the functionality of the fork by default, would you still go for this approach?

The main reason is to encapsulate everything in one place, which should make it easy to move CI providers (one can correctly argue this is a premature optimisation!). This is a pattern employed by projects being build on TeamCity.

Having looked at Source's implementation of `guardian/actions-setup-node` it looks like you can benefit from [caching](https://github.com/guardian/source/runs/2085562449?check_suite_focus=true#step:3:6), which I hadn't considered. Doing things fresh every time doesn't lengthen the build significantly however.

All this being said, I've not got strong opinions on either solution and am happy to be swayed either way. Maybe we should do this for now and wait to hear from the PR as there could be reasons why the upstream version of `actions-setup-node` doesn't sniff the node version files today?

akash1810:
Closing as stale.

PR #319, opened 2021-03-12
refactor: create a simpler interface for defining security group rules

akash1810:
## What does this change?

This hopefully creates a simpler interface to defining security group ingress and egress rules. Now, to define an ingress or egress rule we provide a `SecurityGroupAccessRule`.

`SecurityGroupAccessRule` has three mandatory fields:
1. `range` which is the CIDR range
2. `port` which is the port number and protocol (tcp by default)
3. `description` to provide some useful information to the reader viewing the security group in the AWS console

Example usage:

```typescript
new GuSecurityGroup(stack, "FtpAccessSecurityGroup", {
  vpc,
  ingresses: [
    { range: Peer.anyIpv4(), port: 21, description: "Allow FTP access from the world" },
  ],
});
```

BREAKING CHANGES:
`CidrIngress` and `CidrEgress` have been removed in favour of a single `SecurityGroupAccessRule`. This change forces a description to be provided for every rule as it makes it easier to reason about in the AWS console. 

It also changes the type of `port` to `number | Port` - if you provide a `number` the protocol will default to `TCP` as this is the most common use-case. This should provide a simpler API.

`transformToCidrIngress` is renamed to `transformToSecurityGroupAccessRule` as it's not longer tied to ingress rules.

Port 22 is completely forbidden as SSH over SSM is preferred.

Yes - some types and function names have changed.

## How to test

See existing tests.

A simpler API.

n/a

### DIFF
--- a/src/constructs/ec2/security-group.test.ts
+++ b/src/constructs/ec2/security-group.test.ts
@@ -188,11 +188,11 @@ describe("The GuPublicInternetAccessSecurityGroup class", () => {
     });
 
     expect(stack).toHaveResource("AWS::EC2::SecurityGroup", {
-      GroupDescription: "Allows internet access on 443",
+      GroupDescription: "Allow all inbound traffic via HTTPS",
       SecurityGroupIngress: [
         {
           CidrIp: "0.0.0.0/0",
-          Description: "Allows internet access on 443",
+          Description: "Allow all inbound traffic via HTTPS",
           FromPort: 443,
           IpProtocol: "tcp",
           ToPort: 443,
@@ -215,11 +215,11 @@ describe("The GuHttpsEgressSecurityGroup class", () => {
     GuHttpsEgressSecurityGroup.forVpc(stack, { vpc });
 
     expect(stack).toHaveResource("AWS::EC2::SecurityGroup", {
-      GroupDescription: "Allow all outbound traffic on port 443",
+      GroupDescription: "Allow all outbound HTTPS traffic",
       SecurityGroupEgress: [
         {
           CidrIp: "0.0.0.0/0",
-          Description: "Allow all outbound traffic on port 443",
+          Description: "Allow all outbound HTTPS traffic",
           FromPort: 443,
           IpProtocol: "tcp",
           ToPort: 443,
--- a/src/constructs/ec2/security-group.test.ts
+++ b/src/constructs/ec2/security-group.test.ts
@@ -42,8 +42,8 @@ describe("The GuSecurityGroup class", () => {
     new GuSecurityGroup(stack, "TestSecurityGroup", {
       vpc,
       ingresses: [
-        { range: Peer.ipv4("127.0.0.1/24"), description: "ingress1" },
-        { range: Peer.ipv4("127.0.0.2/8"), description: "ingress2" },
+        { range: Peer.ipv4("127.0.0.1/24"), description: "ingress1", port: 443 },
+        { range: Peer.ipv4("127.0.0.2/8"), description: "ingress2", port: 443 },
       ],
     });
 
@@ -181,7 +181,7 @@ describe("The GuPublicInternetAccessSecurityGroup class", () => {
       SecurityGroupIngress: [
         {
           CidrIp: "0.0.0.0/0",
-          Description: "GLOBAL_ACCESS",
+          Description: "Allows internet access on 443",
           FromPort: 443,
           IpProtocol: "tcp",
           ToPort: 443,
@@ -208,7 +208,7 @@ describe("The GuHttpsEgressSecurityGroup class", () => {
       SecurityGroupEgress: [
         {
           CidrIp: "0.0.0.0/0",
-          Description: "from 0.0.0.0/0:443",
+          Description: "Allow all outbound traffic on port 443",
           FromPort: 443,
           IpProtocol: "tcp",
           ToPort: 443,
--- a/src/constructs/ec2/security-group.test.ts
+++ b/src/constructs/ec2/security-group.test.ts
@@ -98,6 +98,17 @@ describe("The GuSecurityGroup class", () => {
       ],
     });
   });
+
+  it("should not allow an ingress rule for port 22", () => {
+    const stack = simpleGuStackForTesting();
+
+    expect(() => {
+      new GuSecurityGroup(stack, "TestSecurityGroup", {
+        vpc,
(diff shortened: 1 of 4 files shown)


src/constructs/ec2/security-groups.ts
```diff
@@ -1,42 +1,60 @@
...
-
   constructor(scope: GuStack, id: string, props: GuSecurityGroupProps) {
     super(scope, id, props);
 
     if (props.overrideId) {
       (this.node.defaultChild as CfnSecurityGroup).overrideLogicalId(id);
     }
 
-    props.ingresses?.forEach(({ range, description }) =>
-      this.addIngressRule(range, GuSecurityGroup.defaultIngressPort, description)
```
akash1810, line 34: The ingress rule port was always set to port 443, which is a bug, I think?!

src/constructs/ec2/security-group.test.ts
```diff
@@ -98,6 +98,17 @@ describe("The GuSecurityGroup class", () => {
...
+
+  it("should not allow an ingress rule for port 22", () => {
+    const stack = simpleGuStackForTesting();
+
+    expect(() => {
+      new GuSecurityGroup(stack, "TestSecurityGroup", {
+        vpc,
+        ingresses: [{ range: Peer.anyIpv4(), description: "SSH access", port: 22 }],
+      });
+    }).toThrow(new Error("An ingress rule on port 22 is not allowed. Prefer to setup SSH via SSM."));
```
jacobwinch, line 110: This is great 💯

src/constructs/ec2/security-groups.ts
```diff
@@ -1,42 +1,60 @@
...
-export interface CidrEgress {
-  range: IPeer;
-  port: Port;
-  description?: string;
+  /**
+   * The port to open in a security group.
+   * The default protocol is TCP.
+   * Use `Port.udp(port)` for the UDP protocol.
+   */
+  port: number | Port;
```
jacobwinch, line 22: 👍

src/constructs/ec2/security-groups.ts
jacobwinch: As we are already providing the HTTPS default via `GuPublicInternetAccessSecurityGroup`, I think allowing more flexibility at the `GuSecurityGroup` level is fine.

If we do want to be stricter, we could follow the example you've set for port 22 and explicitly throw if users attempt to allow incoming traffic on port 80?

src/constructs/ec2/security-groups.ts
```diff
@@ -1,42 +1,60 @@
 import type { CfnSecurityGroup, IPeer, SecurityGroupProps } from "@aws-cdk/aws-ec2";
 import { Peer, Port, SecurityGroup } from "@aws-cdk/aws-ec2";
-import { transformToCidrIngress } from "../../utils/security-groups";
 import type { GuStack } from "../core";
 
-export interface CidrIngress {
+/**
+ * A way to describe an ingress or egress rule for a security group.
+ *
+ * See [[transformToSecurityGroupAccessRule]] for a handy helper function.
```
jacobwinch, line 8: Maybe we should remind users to make use of `GuWazuhAccess`, `GuPublicInternetAccessSecurityGroup`, `GuHttpsEgressSecurityGroup` (if possible), in case people stumble across these docs first and start unnecessarily reimplementing common cases?

src/constructs/ec2/security-groups.ts
```diff
@@ -47,8 +65,8 @@ export class GuWazuhAccess extends GuSecurityGroup {
       overrideId: true,
       allowAllOutbound: false,
       egresses: [
-        { range: Peer.anyIpv4(), port: Port.tcp(1514), description: "wazuh event logging" },
-        { range: Peer.anyIpv4(), port: Port.tcp(1515), description: "wazuh agent registration" },
+        { range: Peer.anyIpv4(), port: 1514, description: "wazuh event logging" },
+        { range: Peer.anyIpv4(), port: 1515, description: "wazuh agent registration" },
```
jacobwinch, line 69:
```suggestion
        { range: Peer.anyIpv4(), port: 1514, description: "Wazuh event logging" },
        { range: Peer.anyIpv4(), port: 1515, description: "Wazuh agent registration" },
```

Since we are defining defaults for the department, the pedant in me would like to see consistent capitalisation in the AWS console. 😅 

TIL: [multi-line suggestions](https://github.blog/changelog/2020-02-26-multi-line-code-suggestions-beta/)

src/constructs/ec2/security-groups.ts
```diff
@@ -65,8 +83,8 @@ export class GuPublicInternetAccessSecurityGroup extends GuSecurityGroup {
   constructor(scope: GuStack, id: string, props: SecurityGroupProps) {
     super(scope, id, {
       ...props,
-      ingresses: transformToCidrIngress(Object.entries({ GLOBAL_ACCESS: "0.0.0.0/0" })),
-      description: `Allows internet access on ${GuPublicInternetAccessSecurityGroup.defaultIngressPort.toString()}`,
+      ingresses: [{ range: Peer.anyIpv4(), port: 443, description: "Allows internet access on 443" }],
+      description: "Allows internet access on 443",
```
jacobwinch, line 87: Minor: there are a couple of descriptions where we mention allowing access over 443, but I think that's already reasonably clear from the port definition. Perhaps mentioning allowing traffic over HTTPS would add more helpful information for anyone who doesn't know what this port is used for?

jacobwinch (approved):
I've left a few _very_ minor comments, feel free to merge this as is if you don't agree with them 👍

src/constructs/ec2/security-groups.ts
akash1810: There might be legit use cases to allow incoming on port 80 - if your server is running on EC2 on 80, you'd want the load balancer to connect, but nothing else. Not sure if can encode that cleanly yet? Maybe it can be a warning that's printed out?

src/constructs/ec2/security-groups.ts
akash1810: Great suggestion. Done.

src/constructs/ec2/security-groups.ts
jacobwinch: Good point, let's leave it for now?

jacobwinch:
> This PR is included in version 1.0.0

`BREAKING CHANGE:` - @akash1810 - it's good to know that this works!

PR #326, opened 2021-03-16
feat!: Tag support for a stack with multiple apps

akash1810:
## What does this change?

Currently, as noted by [this comment](https://github.com/guardian/cdk/blob/27543797cc5c2f0dbfcf5c262c5456c13e730a10/src/constructs/core/stack.ts#L11-L12), our constructs are tailored for single app stacks. We don't support a stack that has, for example, a frontend web app and an API app. If we were to create two ASGs in the same stack, they'd both get incorrectly tagged as the same "app". That is, we don't support micro-services very well.

This change moves the app string into an interface, this interface is then extended by any construct's props if they need to be app-aware.

As "app" is not in `GuStack` and we want to continue adding the "app" tag to resources, we're having to call `Tags.of(this).add("App", props.app);` a few more times, maybe there's a better solution here that keeps things DRY?

Lastly, this is quite a big diff - a lot of the changes are within test files. I've tried to make each commit standalone, so it might be easier to review commit by commit?

BREAKING CHANGE:
- `GuStack` no longer takes knows about an "app"
- Constructors for `GuAutoScalingGroup`, `GuUserData`, `GuParameterStoreReadPolicy`,  `GuGetDistributablePolicy` and `GuInstanceRole` now require `props`
- Props for `GuLambdaFunction`, `GuDatabaseInstance`, `GuInstanceTypeParameter` and `GuAmiParameter` now take an "app" prop

Yes! There are numerous breaking changes in this PR!

## How to test

See updated tests, aka observe CI.

It is possible to have a stack with multiple apps that are tagged correctly.

### DIFF
--- a/src/constructs/core/identity.ts
+++ b/src/constructs/core/identity.ts
@@ -1,5 +1,10 @@
-export interface Identity {
+export interface StackStageIdentity {
   stack: string;
   stage: string;
+}
+
+export interface AppIdentity {
   app: string;
 }
+
+export interface Identity extends StackStageIdentity, AppIdentity {}
--- a/src/constructs/core/identity.ts
+++ b/src/constructs/core/identity.ts
@@ -17,7 +17,7 @@ export const AppIdentity = {
     const titleCaseApp = appIdentity.app.charAt(0).toUpperCase() + appIdentity.app.slice(1);
     return `${text}${titleCaseApp}`;
   },
-  addTag<T extends IConstruct>(appIdentity: AppIdentity, construct: T): T {
+  taggedConstruct<T extends IConstruct>(appIdentity: AppIdentity, construct: T): T {
     Tags.of(construct).add("App", appIdentity.app);
     return construct;
   },
--- a/src/constructs/core/identity.ts
+++ b/src/constructs/core/identity.ts
@@ -1,3 +1,6 @@
+import type { IConstruct } from "@aws-cdk/core";
+import { Tags } from "@aws-cdk/core";
+
 export interface StackStageIdentity {
   stack: string;
   stage: string;
@@ -13,5 +16,9 @@ export const AppIdentity = {
   suffixText(appIdentity: AppIdentity, text: string): string {
     const titleCaseApp = appIdentity.app.charAt(0).toUpperCase() + appIdentity.app.slice(1);
     return `${text}${titleCaseApp}`;
+  },
+  addTag<T extends IConstruct>(appIdentity: AppIdentity, construct: T): T {
+    Tags.of(construct).add("App", appIdentity.app);
+    return construct;
   },
 };
--- a/src/constructs/core/identity.ts
+++ b/src/constructs/core/identity.ts
@@ -8,3 +8,10 @@ export interface AppIdentity {
 }
 
 export interface Identity extends StackStageIdentity, AppIdentity {}
+
+export const AppIdentity = {
+  suffixText(appIdentity: AppIdentity, text: string): string {
+    const titleCaseApp = appIdentity.app.charAt(0).toUpperCase() + appIdentity.app.slice(1);
+    return `${text}${titleCaseApp}`;
+  },
+};
--- a/src/constructs/core/parameters/ec2.ts
+++ b/src/constructs/core/parameters/ec2.ts
@@ -1,12 +1,14 @@
+import { AppIdentity } from "../identity";
 import type { GuStack } from "../stack";
-import type { GuNoTypeParameterProps } from "./base";
+import type { GuNoTypeParameterPropsWithAppIdentity } from "./base";
 import { GuParameter } from "./base";
 
 export class GuInstanceTypeParameter extends GuParameter {
-  constructor(scope: GuStack, id: string = "InstanceType", props?: GuNoTypeParameterProps) {
-    super(scope, id, {
+  // eslint-disable-next-line custom-rules/valid-constructors -- TODO be better
+  constructor(scope: GuStack, props: GuNoTypeParameterPropsWithAppIdentity) {
+    super(scope, AppIdentity.suffixText(props, "InstanceType"), {
       type: "String",
-      description: "EC2 Instance Type",
+      description: `EC2 Instance Type for the app ${props.app}`,
       default: "t3.small",
       ...props,
     });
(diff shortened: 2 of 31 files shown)


akash1810:
PR to fix the build - https://github.com/guardian/cdk/pull/327.

src/constructs/iam/policies/s3-get-object.ts
```diff
@@ -18,9 +19,14 @@ export class GuGetS3ObjectsPolicy extends GuAllowPolicy {
 }
 
 export class GuGetDistributablePolicy extends GuGetS3ObjectsPolicy {
-  constructor(scope: GuStack, id: string = "GetDistributablePolicy", props?: GuNoStatementsPolicyProps) {
-    const path = [scope.stack, scope.stage, scope.app, "*"].join("/");
-    super(scope, id, { ...props, bucketName: new GuDistributionBucketParameter(scope).valueAsString, paths: [path] });
+  // eslint-disable-next-line custom-rules/valid-constructors -- TODO be better
```
jacobwinch, line 22: 😀

jacobwinch (approved):
Great work 💯 

👍 once you've rebased / merged in https://github.com/guardian/cdk/pull/327!

jacobwinch:
> I've tried to make each commit standalone, so it might be easier to review commit by commit?

Thanks for this suggestion, it was helpful!

src/constructs/iam/policies/s3-get-object.ts
akash1810: Yeah, I'm not sure if this is the best change as it means the constructor for the constructs take different forms and goes against [this decision](https://github.com/guardian/cdk/blob/main/docs/architecture-decision-records/002-component-constuctors.md).

src/constructs/iam/roles/instance-role.ts
```diff
@@ -10,30 +11,33 @@ import {
...
   withoutLogShipping?: boolean; // optional to have log shipping added by default, you have to opt out
   additionalPolicies?: GuPolicy[];
 }
 
 export class GuInstanceRole extends GuRole {
   private policies: GuPolicy[];
 
-  constructor(scope: GuStack, id: string = "InstanceRole", props?: GuInstanceRoleProps) {
-    super(scope, id, {
+  // eslint-disable-next-line custom-rules/valid-constructors -- TODO be better
```
jacobwinch, line 22: As we are starting to regularly break this convention perhaps we should consider relaxing the linting rule?

src/constructs/core/parameters/ec2.ts
```diff
@@ -1,23 +1,26 @@
...
 
 export class GuInstanceTypeParameter extends GuParameter {
-  constructor(scope: GuStack, id: string = "InstanceType", props?: GuNoTypeParameterProps) {
-    super(scope, id, {
+  // eslint-disable-next-line custom-rules/valid-constructors -- TODO be better
+  constructor(scope: GuStack, props: GuNoTypeParameterPropsWithAppIdentity) {
+    super(scope, AppIdentity.suffixText(props, "InstanceType"), {
       type: "String",
-      description: "EC2 Instance Type",
+      description: `EC2 Instance Type for the app ${props.app}`,
```
akash1810, line 11: For the app "janus", the ID/name of this parameter becomes "InstanceTypeJanus". Hopefully the description is helpful?

src/constructs/core/identity.ts
```diff
@@ -1,5 +1,24 @@
...
+export interface Identity extends StackStageIdentity, AppIdentity {}
+
+export const AppIdentity = {
+  suffixText(appIdentity: AppIdentity, text: string): string {
+    const titleCaseApp = appIdentity.app.charAt(0).toUpperCase() + appIdentity.app.slice(1);
+    return `${text}${titleCaseApp}`;
+  },
+  addTag<T extends IConstruct>(appIdentity: AppIdentity, construct: T): T {
+    Tags.of(construct).add("App", appIdentity.app);
+    return construct;
```
jacobwinch, line 22: I find this return type a little surprising for a function called `addTag`

src/constructs/iam/roles/instance-role.ts
akash1810: Definitely! I started on this a couple of weeks ago, but looks like I deleted my branch! 😱 

I _think_ the rules should be:
- 1, 2 or 3 arguments
- 1st argument is always GuStack
- If only 2 arguments, 2nd argument is required and are `props` w/no default
- If 3 arguments, 2nd is ID w/no default, 3rd are `props` w/no default and optional

Will raise this in a separate PR to see how it pans out.

src/constructs/core/identity.ts
akash1810: Agreed. It's mainly out of laziness as we can do this:

```typescript
const snsTopic = props.existingSnsTopic?.externalTopicName
  ? Topic.fromTopicArn(
      scope,
      topicId,
      `arn:aws:sns:${scope.region}:${scope.account}:${props.existingSnsTopic.externalTopicName}`
    )
  : AppIdentity.addTag(
      props,
      new GuSnsTopic(scope, topicId, {
        overrideId: !!props.existingSnsTopic?.logicalIdFromCloudFormation,
      })
    );
```

The alternative is:

```typescript
const newTopic = () => {
  const topic = new GuSnsTopic(scope, topicId, {
    overrideId: !!props.existingSnsTopic?.logicalIdFromCloudFormation,
  });

  AppIdentity.addTag(props, topic);
  return topic;
};

const snsTopic = props.existingSnsTopic?.externalTopicName
  ? Topic.fromTopicArn(
      scope,
      topicId,
      `arn:aws:sns:${scope.region}:${scope.account}:${props.existingSnsTopic.externalTopicName}`
    )
  : newTopic();
```

Maybe `withTag` is a better name?

jacobwinch: Ah yes, I see the benefit now. We could go for `withTag` or `taggedResource`, but it's not a big deal so feel free to ignore this!

jacobwinch (approved):

src/constructs/core/identity.ts
akash1810: Renamed to `taggedConstruct` to match the name of the parameter.

src/constructs/iam/roles/instance-role.ts
stephengeller: Agree with the rules stated above!

akash1810: https://github.com/guardian/cdk/pull/352 revises the linting rules

PR #332, opened 2021-03-18
feat: support multiple GuInstanceRoles in a stack

akash1810:
## What does this change?

A `GuInstanceRole` provides the basic permissions needed by EC2, it grants access to:
- SSM
- Describing tags
- Writing to kinesis for log shipping

#326 started the work to support the declaration of multiple apps in a stack, where an app is a collection of ASG, LB etc. However, when actually attempting to use the updated constructs in reality, the synth step failed due to multiple constructs using the same ID.

This change follows #341 and #337 and converts `GuSSMRunCommandPolicy` and `GuDescribeEC2Policy` to singletons. This means these policies will only get created once regardless of how many roles use them - there will be one policy with multiple roles attached to it. This helps keep the stack DRY and also reduces the chance of reaching the max file size limits of cloudformation.

The commits have been crafted to be standalone, reviewing them one by one might make the overall review easier.

Requires #341.

BREAKING CHANGE:
* `GuSSMRunCommandPolicy` and `GuDescribeEC2Policy` can no longer be directly instantiated

Ultimately no, as `GuInstanceRole` hasn't fundamentally changed.

## How to test

See new test!

It's even more possible to declare multiple apps within a stack.

n/a

### DIFF
--- a/src/constructs/iam/policies/describe-ec2.test.ts
+++ b/src/constructs/iam/policies/describe-ec2.test.ts
@@ -3,15 +3,15 @@ import { attachPolicyToTestRole, simpleGuStackForTesting } from "../../../../tes
 import { GuDescribeEC2Policy } from "./describe-ec2";
 
 describe("DescribeEC2Policy", () => {
-  it("can accept a custom policy name", () => {
+  it("creates the correct policy", () => {
     const stack = simpleGuStackForTesting();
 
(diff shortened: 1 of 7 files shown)


philmcmahon (approved):
This sounds like a good change to me 👍🏻 

best stack ever

PR #335, opened 2021-03-18
chore: make the repo commitizen-friendly

stephengeller:
## What does this change?

According to [these docs](https://github.com/commitizen/cz-cli#making-your-repo-commitizen-friendly), we haven't made our repository commitizen-friendly. This PR merges in the result of running `commitizen init cz-conventional-changelog --save-dev --save-exact` on the repo and setting up basic config.

This will enable everyone to use the proper commitizen cli in the CDK repo with `cz`, `git cz` or `npm cz`.

## Does this change require changes to existing projects or CDK CLI?

Nope.

## How to test

In the root of the repo, run `npx cz` and see the interactive menu work!

## How can we measure success?

We can use `npx cz` with ease!

PR #327, opened 2021-03-16
chore(deps): bump @aws-cdk/* to v1.94.1

akash1810:
## What does this change?

We're more than five releases behind https://github.com/aws/aws-cdk/releases, this PR bumps the version of the AWS CDK libraries being used using [the suggested](https://github.com/guardian/cdk/runs/2126107533?check_suite_focus=true#step:4:6) [script](https://github.com/guardian/cdk/blob/27543797cc5c2f0dbfcf5c262c5456c13e730a10/script/update-aws-cdk).

## Does this change require changes to existing projects or CDK CLI?

Yes - they'd need to bump their AWS CDK libraries to match the version used here. Not sure if this makes this PR a breaking change?

## How to test

CI should pass.

## How can we measure success?

Keeping up to date.

## Have we considered potential risks?

n/a

PR #336, opened 2021-03-18
chore: Create GuSSMParameter using AWS Custom Resources

stephengeller:
### Problem 

At the moment, the general practice for injecting secrets into CloudFormation templates is through CFN parameters. In CDK, this looks like this:

```ts
DatabaseUser: new GuStringParameter(this, "DatabaseUser", {
        description: "Username of the RDS DB",
        default: "grafana",
}),
```

This is a bit gross, as it:

1. Doesn't have a sensible default (and it wouldn't make sense to have one here), meaning you have to input it yourself every time you create the stack and somehow keep it safe in an unspecified place
2. Requires you to go into the UI every time you want to change it
3. Doesn't make it clear that it is a secure/sensitive value
4. Makes the CFN more unwieldy as more parameters == more things to remember and update and maintain

### Solution

This PR creates a `GuSSMParameter` construct, making use of the SSM Parameter Store to fetch values dynamically.

```ts
DatabaseUser: GuSSMParameter(this, "/PATH/TO/database-user")
```

Using AWS Custom Resources, the `GuSSMParameter` construct can dynamically fetch the latest version of an SSM parameter, [something that AWS' own CDK construct cannot do yet](https://github.com/aws/aws-cdk/issues/9793). 

This also creates `GuSSMDefaultParam` (would prefer a new name, suggestions welcome!) that assumes the default path of `/$STAGE/$STACK/$APP/$PARAM_NAME` as a prefix to the parameter name, for added simplicity.

All three of the below implementations work:

```ts
import type { App } from "@aws-cdk/core";
import { CfnOutput } from "@aws-cdk/core";
import type { GuStackProps } from "@guardian/cdk/lib/constructs/core";
import { GuStack } from "@guardian/cdk/lib/constructs/core";
import { GuSSMDefaultParam, GuSSMParameter } from "./ssmParams";

export class CdkStack extends GuStack {
    super(scope, id, props);

    const secrets = {
      fullPath: new GuSSMParameter(this, "/CODE/test/tag-janitor/test-output").getValue(),
      usingDefaultValues: GuSSMDefaultParam(this, "test-output").getValue(),
      usingTokensInPath: new GuSSMParameter(this, `/${this.stage}/${this.stack}/${this.app}/test-output`).getValue(),
    };

    new CfnOutput(this, "fullPath", { value: secrets.fullPath });
    new CfnOutput(this, "usingDefaultValues", { value: secrets.usingDefaultValues });
    new CfnOutput(this, "usingTokensInPath", { value: secrets.usingTokensInPath });
  }
}
```

This offers numerous advantages:

1. It's less code (yay!) and easier to declare
2. It requires **no user input** to deploy
3. Updating the value is easier, as you can navigate to SSM on the CLI as well as the UI and then redeploy the app in RiffRaff
4. You can optionally and easily make use of the `/$STAGE/$STACK/$APP/$PARAM_NAME` path prefix with ease and have it dynamically resolve based on the value of those parameters

## How does it work?

This uses [AWS Custom Resouces](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/template-custom-resources.html), which creates a few things alongside your normal resources. Namely, this creates a lambda which will essentially execute anything you want, but we've restricted ours to just call `SSM getParameter` for our purposes. When you create/update a stack, the lambda will run, fetch the latest version of the SSM parameters you need and populate the relevant resources with them.

Not yet, as this would be something to begin replacing traditional `GuParameter`s with.

## How to test

Tests will be added before moving this from draft to live.

 1. We pretty much eliminate all manually filled-in CFN parameters from CDK stacks
 2. We can always fetch the latest version of an SSM parameter for use in CDK stacks
 3. We can dymically fetch parameters for appropriate `$STAGE/$STACK/$APP` paths

Not risks as such, but AWS Custom Resources will somewhat unavoidably create additional resources per stack. This isn't necessarily a problem (and should be fine cost-wise due to the low number of executions), but is a bit of a gotcha that's worth knowing about. However, users will not need to touch these resources and can happily ignore them.


src/constructs/core/ssm.ts
```diff
@@ -0,0 +1,90 @@
...
+  defaultPath: boolean;
+}
+
+const stripped = (str: string) => str.replace(/[-/]/g, "");
+
+export class GuSSMParameter extends Construct implements IGrantable {
+  private readonly customResource: CustomResource;
+  readonly grantPrincipal: IPrincipal;
+
+  // eslint-disable-next-line custom-rules/valid-constructors -- I think stating an ID would be overkill for this, but happy to discuss
```
jacobwinch, line 26: I think there are plans to change this; see [similar a discussion](https://github.com/guardian/cdk/pull/326#discussion_r596057941) here. However perhaps `param` should be part of props, for consistency?

src/constructs/core/ssm.ts
```diff
@@ -0,0 +1,90 @@
...
+  constructor(scope: GuStack, param: string, props?: GuSSMParameterProps) {
+    // TODO: Establish if this is an accepted, safe way of creating IDs
+    const id = (id: string) =>
+      param.toUpperCase().includes("TOKEN") ? `${id}-token-${Date.now()}` : `${id}-${stripped(param)}`;
+
+    super(scope, id("GuSSMParameter"));
+
+    const provider = new SingletonFunction(scope, id("Provider"), {
+      code: Code.fromInline(readFileSync(join(__dirname, "/custom-resources/runtime/lambda.js")).toString()),
+      // runtime: new Runtime("nodejs14.x", RuntimeFamily.NODEJS, { supportsInlineCode: true }), -- we can use Node14 once we bump the version of @aws-cdk to v1.94 https://github.com/aws/aws-cdk/releases/tag/v1.94.0
```
jacobwinch, line 36: I think you're [in luck](https://github.com/guardian/cdk/pull/327)!

src/constructs/core/ssm.ts
```diff
@@ -0,0 +1,90 @@
...
+    return this.customResource.getAtt("Parameter.Value");
+  }
+
+  public getValue(): string {
+    console.log(this.customResource.toString());
+    return this.customResource.getAttString("Parameter.Value");
+  }
+}
+
+export function GuSSMDefaultParam(scope: GuStack, param: string): GuSSMParameter {
```
jacobwinch, line 88: If `GuSSMParameterProps` was:

```typescript
export interface GuSSMParameterProps {
  secure?: boolean;
  /*
   * If no path is provided we will assume it is`/${STAGE}/${STACK}/${APP}/${param}`
   * */
  overridePath?: string; // not sure about the name, but hopefully you get the idea
}
```

Then we could consider dropping `GuSSMDefaultParam` entirely?

Although it would still be possible to stray from the well-trodden path (no pun intended), it would be a little harder to do so accidentally.

package.json
```diff
@@ -22,10 +22,12 @@
   },
   "devDependencies": {
     "@guardian/eslint-config-typescript": "^0.5.0",
+    "@types/aws-lambda": "^8.10.72",
     "@types/jest": "^26.0.20",
     "@types/node": "14.14.32",
     "@typescript-eslint/eslint-plugin": "^4.16.1",
     "@typescript-eslint/parser": "^4.16.1",
+    "aws-sdk": "^2.866.0",
```
jacobwinch, line 30: I think that you could use [`v3` of this sdk](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/index.html). I had a few problems migrating a project to it recently, but unless you are aware of any known issues with the functionality you're using it could be worth a try?

jacobwinch:
Great PR description, thanks for taking the time to write this up! I've left a few comments but it's looking good so far!

src/constructs/core/ssm.ts
stephengeller: Good shout, and done. It's actually becoming apparent that passing in an ID might solve a couple of issues down the line but will investigate.

The problem is essentially that it looks like we need to ensure we create new custom resources every time (not the lambda, but possibly everything else) in order to trigger the fetching of an SSM parameter every time. If you have a fixed ID, it _might_ be the case that it won't update past creation unless you somehow inform it that there's been a change.

Will investigate though.

src/constructs/core/ssm.ts
stephengeller: Yay! Updated to use Node 14.

src/constructs/core/ssm.ts
stephengeller: Yeah, maybe if we called the prop `fullPath` instead, that might be clearer? eg:

```ts
const withFullPath = new GuSSMParameter(this, { parameter: "/path/to/param", fullPath: true })
const assumedPath = new GuSSMParameter(this, { parameter: "param" }) // assumes /stage/stack/app/param
```

package.json
stephengeller: Good shout! I've tried it and looks fine, thanks for the heads up about this.

stephengeller:
I've got a couple of things to work out before this is good to go (getting CI to pass included), so will work those out before requesting another review but all comments so far should be addressed

package.json
stephengeller: Sadly, I've had to revert to v2 of the library since an AWS Node Lambda [only has v2 of the library available](https://docs.aws.amazon.com/lambda/latest/dg/lambda-nodejs.html). If and when this changes we can definitely upgrade though, but not sure we can make this work with v3 at the moment.

stephengeller:
This is now ready for review! Good to get eyes on it from @jacobwinch @akash1810 and/or @sihil if possible.

src/constructs/core/custom-resources/runtime/lambda.test.ts
```diff
@@ -0,0 +1,17 @@
+/*
+ * You can't specify the file extension (eg .ts) of the file to import,
+ * and lambda.js exists as a dummy file, so we'd always import that over lambda.ts (the actual code to be tested) if
+ * we imported `lambda`
+ * lambda.symlink is a symlink to lambda.ts, so we can test it without importing lambda.js
+ * */
+import { flatten } from "./lambda.symlink";
```
akash1810, line 7: Oh wow! I'd be curious to know the output of `npm pack` now? IIUC [`npm pack`](https://docs.npmjs.com/cli/v6/commands/npm-pack) will create a local bundle that's representative of what get's published to NPM - a sort of publish dry run.

I'm not sure what JS file will be added - the empty lambda.js or the one that tsc will create?

src/constructs/core/ssm.test.ts
```diff
@@ -0,0 +1,33 @@
+import "@aws-cdk/assert/jest";
+import { simpleGuStackForTesting } from "../../../test/utils";
+import { GuSSMIdentityParameter, GuSSMParameter } from "./ssm";
+
+describe("SSM:", () => {
+  describe("The GuSSMIdentityParameter construct", () => {
+    beforeAll(() => {
+      // TODO: Experiment with compiling `custom-resources/runtime/lambda.ts` to JS so that tests run
+    });
```
akash1810, line 9: Don't think this is needed?

src/constructs/core/ssm.ts
```diff
@@ -0,0 +1,97 @@
...
+      // We need to create UIDs for the resources in this construct, as otherwise CFN will not trigger the lambda on updates for resources that appear to be the same
+      const uid = now.toString().substr(now.toString().length - 4);
+      return parameter.toUpperCase().includes("TOKEN") ? `${id}-token-${uid}` : `${id}-${stripped(parameter)}-${uid}`;
+    };
+
+    super(scope, id("GuSSMParameter"));
+
+    const provider = new SingletonFunction(scope, id("Provider"), {
+      code: Code.fromInline(readFileSync(join(__dirname, "/custom-resources/runtime/lambda.js")).toString()),
+      runtime: Runtime.NODEJS_12_X, // TODO: Ensure that the TS -> JS compile creates a NODE12-compliant file
```
akash1810, line 41: Node 12 is in [maintenance LTS](https://nodejs.org/en/about/releases/). AWS recently announced [support for node 14 in AWS Lambda](https://aws.amazon.com/blogs/compute/node-js-14-x-runtime-now-available-in-aws-lambda/).

How does the output for a node 14 runtime break? There might be some TS config tweaks we can make? Or raise it with our AWS TAM?

src/constructs/core/ssm.ts
```diff
@@ -0,0 +1,97 @@
...
+  // eslint-disable-next-line custom-rules/valid-constructors -- TODO: Remove once linting rules have been relaxed for this
+  constructor(scope: GuStack, props: GuSSMParameterProps) {
+    const { parameter } = props;
+
+    const id = (id: string) => {
+      const now = Date.now();
+      // We need to create UIDs for the resources in this construct, as otherwise CFN will not trigger the lambda on updates for resources that appear to be the same
+      const uid = now.toString().substr(now.toString().length - 4);
+      return parameter.toUpperCase().includes("TOKEN") ? `${id}-token-${uid}` : `${id}-${stripped(parameter)}-${uid}`;
+    };
```
akash1810, line 35: The comment suggests this is pretty crucial, should we explicitly test it?

akash1810 (commented):
Looks good! Added a few comments.

Could we add some usage guidance to the PR description? The example use case is a database user - when should this be normal application config vs a `GuSSMParameter` in the CFN template?

Lastly, should the PR's prefix be "feat" as we're adding a feature and a new minor version should be released?

package-lock.json
```diff
@@ -1,8 +1,19472 @@
 {
   "name": "@guardian/cdk",
   "version": "4.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
```
akash1810, line 4: I think you're on NPM v7? You'll want to downgrade to v6, I think . See https://github.com/guardian/cdk/pull/340 and https://github.com/guardian/cdk/pull/335#issuecomment-802678168.

src/constructs/core/ssm.ts
akash1810: https://github.com/guardian/cdk/pull/352 revises the linting rules.

src/constructs/core/ssm.test.ts
stephengeller: Good find, thanks!

package-lock.json
stephengeller: D'oh, thanks! Resolved in a coming push.

src/constructs/core/ssm.ts
stephengeller: I think this TODO is outdated, as the TS configuration in the app creates Node 12 compliant JS code. The problem with upgrading to Node 14 is that it for some reason prevents us from using inline code (ie `Code.fromInline`), which is necessary here. I can't find anything online yet that specifically explains why, but it means that we're somewhat restricted to 12 until we can inline the code or find a better solution.

src/constructs/core/ssm.ts
stephengeller: Yeah, we should. I've added two tests that hopefully capture this behaviour!

src/constructs/core/custom-resources/runtime/lambda.test.ts
stephengeller: Just ran it and it looks like it packs fine! By that I mean the resulting `lambda.js` file is the expected compiled TS lambda, not the dummy file. I assume it simply overwrites it no matter what, as the dummy file could equally be seen as a stale JS file in need of updating

stephengeller:
Thanks for the comments @akash1810 , I've now addressed them all as well as updated the PR name and added a use case, hope that's what you were after?

src/constructs/core/custom-resources/runtime/lambda.ts
```diff
@@ -0,0 +1,102 @@
...
+      Data: data,
+    });
+
+    console.log("Responding", responseBody);
+    const parsedUrl = parse(event.ResponseURL);
+    const requestOptions = {
+      hostname: parsedUrl.hostname,
+      path: parsedUrl.path,
+      method: "PUT",
+      headers: { "content-type": "", "content-length": responseBody.length },
```
akash1810, line 84: What does `content-type` of empty string do/mean? Would omitting it yield the same behaviour?

src/constructs/core/ssm.ts
akash1810: Would be good, but not a blocker, to isolate this function to make it super obvious that given this input, we expect this output. That is, test the `id` function outside of the constructor. We can do this later as needed though.

src/constructs/core/ssm.ts
```diff
@@ -0,0 +1,97 @@
...
+export interface GuSSMIdentityParameterProps extends GuSSMParameterProps, AppIdentity {}
+
+const stripped = (str: string) => str.replace(/[-/]/g, "");
+
+export class GuSSMParameter extends Construct implements IGrantable {
+  private readonly customResource: CustomResource;
+  readonly grantPrincipal: IPrincipal;
+
+  // eslint-disable-next-line custom-rules/valid-constructors -- TODO: Remove once linting rules have been relaxed for this
+  constructor(scope: GuStack, props: GuSSMParameterProps) {
```
akash1810, line 27: https://github.com/guardian/cdk/pull/352 allows for this constructor 🎉  a rebase of `main` should allow you to remove this line.

src/constructs/core/ssm.ts
```diff
@@ -0,0 +1,97 @@
...
+  }
+}
+
+/*
+ * Assumes the path of `/${STAGE}/${STACK}/${APP}/${parameter}`
+ *
+ * */
+export class GuSSMIdentityParameter extends GuSSMParameter {
+  // eslint-disable-next-line custom-rules/valid-constructors -- this may not be necessary going forward
+  constructor(scope: GuStack, props: GuSSMIdentityParameterProps) {
```
akash1810, line 94: https://github.com/guardian/cdk/pull/352 allows for this constructor 🎉  a rebase of `main` should allow you to remove this line.

akash1810 (approved):
Looks great. A couple of minor comments to consider, but feel free to address them in future PRs.

src/constructs/core/custom-resources/runtime/lambda.ts
stephengeller: I assume it's a bad thing, if anything. However, it was copied straight from [aws's source code](https://github.com/aws/aws-cdk/blob/37a5502ced1bf1b451ac4bd921752746277461bf/packages/%40aws-cdk/custom-resources/lib/aws-custom-resource/runtime/index.ts#L171), so I'm tempted to leave it for now and only change if it causes issues?

src/constructs/core/ssm.ts
stephengeller: Nice one, done.

stephengeller:
> Would be good, but not a blocker, to isolate this function to make it super obvious that given this input, we expect this output. That is, test the id function outside of the constructor. We can do this later as needed though.

I isolated the function and added some tests, hopefully that should do the trick @akash1810 .

PR #337, opened 2021-03-19
refactor(GuLogShippingPolicy)!: make GuLogShippingPolicy a singleton

akash1810:
## What does this change?

This is ultimately part of https://github.com/guardian/cdk/pull/332.

This policy is now a [singleton](https://en.wikipedia.org/wiki/Singleton_pattern) as we only ever want one per stack. If there are multiple roles defined in this stack that require log shipping permissions, a single policy will be defined and attached to the roles.

This DRYness helps keep within the file size limits of cloudformation too.

Also included in this change is a new `GuLoggingStreamNameParameter`, also written as a singleton for the same reasons.

To make the review easier, each commit can be read in sequence as they have been crafted to tell the story.

Usage hasn't really changed as this policy will never really be used by consumer stacks as they'l be using `GuInstanceRole` instead, which adds this policy by default. However, usage looks roughly like this:

```typescript
const stack = new GuStack();
const myRole = new GuRole(stack);

const policy = GuLogShippingPolicy.getInstance(stack);
policy.attachToRole(myRole);
```

It's worth noting that https://github.com/guardian/cdk/pull/92 was a naive implementation of singleton parameters and this new style removes the need to have the `getParam` helper function on the `GuStack`.

BREAKING CHANGE:
* The API for `GuLogShippingPolicy` has changed. It is now a singleton and direct instantiation is not possible.

Yes. They'll need a version bump by as `GuLogShippingPolicy` is provided [through the `GuInstanceRole`](https://github.com/guardian/cdk/blob/fa6a0d3d99953c8f0f3009f03525161bdc060864/src/constructs/iam/roles/instance-role.ts#L35), this should be a no-op.

## How to test

See updated and new tests. Either by running them or observing CI.

A much simpler API and ground work for supporting multiple apps (for example a backend api and frontend web app microservice) within a single stack.

n/a

### DIFF
--- a/test/utils/attach-policy-to-test-role.ts
+++ b/test/utils/attach-policy-to-test-role.ts
@@ -4,10 +4,10 @@ import type { Stack } from "@aws-cdk/core";
 import { Tags } from "@aws-cdk/core";
 
 // IAM Policies need to be attached to a role, group or user to be created in a stack
-export const attachPolicyToTestRole = (stack: Stack, policy: Policy, app: string = "testing"): void => {
-  const role = new Role(stack, "TestRole", {
+export const attachPolicyToTestRole = (stack: Stack, policy: Policy, id: string = "TestRole"): void => {
+  const role = new Role(stack, id, {
     assumedBy: new ServicePrincipal("ec2.amazonaws.com"),
   });
-  Tags.of(role).add("App", app);
+  Tags.of(role).add("App", "testing");
   policy.attachToRole(role);
 };
--- a/src/constructs/core/parameters/log-shipping.ts
+++ b/src/constructs/core/parameters/log-shipping.ts
@@ -1,0 +1,27 @@
+import type { GuStack } from "../stack";
+import { GuStringParameter } from "./base";
+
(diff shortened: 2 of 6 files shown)


test/utils/attach-policy-to-test-role.ts
```diff
@@ -4,10 +4,10 @@ import type { Stack } from "@aws-cdk/core";
 import { Tags } from "@aws-cdk/core";
 
 // IAM Policies need to be attached to a role, group or user to be created in a stack
-export const attachPolicyToTestRole = (stack: Stack, policy: Policy, app: string = "testing"): void => {
-  const role = new Role(stack, "TestRole", {
+export const attachPolicyToTestRole = (stack: Stack, policy: Policy, id: string = "TestRole"): void => {
```
akash1810, line 7: The default value of the `app` param was never overridden, so (safely) removed it.

stephengeller (approved):
Great work! Clear changes that make sense and thanks for the PR making it clear how to understand and review this!

PR #352, opened 2021-03-25
feat: revise custom lint rule

akash1810:
## What does this change?

We have a lint rule for constructors [because](https://github.com/guardian/cdk/blob/main/docs/architecture-decision-records/002-component-constuctors.md):

> This project contains a large number of classes, making up the various constructs and patterns. The project is intended to be used as a component library and, therefore, used by a number of people who don't have extensive knowledge of either the CDK or this library. It is therefore important to make the experience of using these classes as intuitive as possible.

We have a few too many instances of `// eslint-disable-next-line custom-rules/valid-constructors` 😅 .

It looks like this is mainly because the `id` field is static (for example the [`Stage` parameter](https://github.com/guardian/cdk/blob/0c32dbe8a73333b6ebf39d24c3e11d935845cfc0/src/constructs/core/parameters/identity.ts#L5-L14)) or generated by the `props` passed in (for example [`InstanceType`](https://github.com/guardian/cdk/blob/0c32dbe8a73333b6ebf39d24c3e11d935845cfc0/src/constructs/core/parameters/ec2.ts#L6-L16)).

This PR revises the rules to be:

1. Private constructors don't get linted
2. Must be 1, 2 or 3 parameters
3. First parameter must be called scope
4. First parameter must be of type GuStack
5. If 2 parameters:
   - The second parameter must be called props
   - The second parameter must be a custom type
6. If 3 parameters:
   - The second parameter must be called id
   - The second parameter must be of type string
   - The third parameter must be called props
   - The third parameter must be a custom type
7. Only the third parameter can be optional or have a default value

The rules can be seen in practice in the file `eslint/rules/valid-constructors.test.ts`.

It might be easier to review this change commit by commit.

No. A few constructors have been changed, but the way in which they're being used hasn't; that is default parameters are now removed and inlined in the constructor body. I think this makes it ok for this to be a minor release (via feat: prefix in PR title)?

## How to test

See tests.

The custom lint rule reflects reality.

n/a

### DIFF
--- a/docs/architecture-decision-records/002-component-constuctors.md
+++ b/docs/architecture-decision-records/002-component-constuctors.md
@@ -20,7 +20,7 @@ This project contains a large number of classes, making up the various construct
 
    By having a consistent style for constructors, users will know what to expect every time they use a class from the library. This makes the developer experience easier and faster as users do not have to check what is required for each class.
 
-2. Constrctors should be the best fit for each class
+2. Constructors should be the best fit for each class
 
    In the case of this library, this mainly means either missing out the id or changing the order of the id and props inputs. In scenarios where a sensible id can either be hardcoded or defaulted, this would prevent users from having to add ids everywhere, reducing the amount of code they have to write.
 
@@ -30,13 +30,13 @@ This project contains a large number of classes, making up the various construct
 
 Constructors should follow the following rules for consistency.
 
-1. Constructors should all accept a `scope` of type `GuStack` and an `id` of type `string`
+1. The first parameter should be a `scope` of type `GuStack`:
 
    :white_check_mark: Valid
 
    ```ts
    class MyConstruct {
-     constructor(scope: GuStack, id: string) {
+     constructor(scope: GuStack) {
        ...
      }
    }
@@ -46,13 +46,39 @@ Constructors should follow the following rules for consistency.
 
    ```ts
    class MyConstruct {
-     constructor(scope: Stack, id: string) {
+     constructor(scope: Stack) {
        ...
      }
    }
    ```
 
-2. They can also take a `props` object which should be correctly typed
+   The construct/pattern will then have a static `id` as it will never change, for example the `Stage` parameter.
+
+2. They can also take a `props` object which should be correctly typed:
+
+   :white_check_mark: Valid
+
+   ```ts
+   class MyConstruct {
+     constructor(scope: GuStack, props: MyConstructProps) {
+       ...
+     }
+   }
+   ```
+
+   :x: Invalid
+
+   ```ts
+   class MyConstruct {
+     constructor(scope: Stack, props: object) {
+       ...
+     }
+   }
+   ```
+
+   The construct/pattern will then derive `id` from `props` as it will never change, for example `InstanceTypeFor${props.app}`.
+
+3. They can also take an `id` of type string and a `props` object which should be correctly typed
 
    :white_check_mark: Valid
 
@@ -70,13 +96,13 @@ Constructors should follow the following rules for consistency.
 
    ```ts
    class MyConstruct {
-     constructor(scope: GuStack, id: string, props: object) {
+     constructor(scope: GuStack, id: any, props: object) {
        ...
      }
    }
    ```
 
-3. Where all `props` are optional, the `props` object should be optional as a whole
+4. Where all `props` are optional, the `props` object should be optional as a whole
 
    :white_check_mark: Valid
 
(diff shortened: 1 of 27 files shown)


eslint/rules/valid-constructors.js
```diff
@@ -1,38 +1,102 @@
-// Rules
```
akash1810, line 1: The rendering of the diff on this file isn't amazingly helpful - reviewing the [end result](https://github.com/guardian/cdk/blob/aa-update-custom-lint-rule/eslint/rules/valid-constructors.js) might be easier?

src/constructs/core/parameters/acm.ts
```diff
@@ -1,15 +1,14 @@
 import { RegexPattern } from "../../../constants";
+import { AppIdentity } from "../identity";
 import type { GuStack } from "../stack";
-import type { GuNoTypeParameterProps } from "./base";
 import { GuStringParameter } from "./base";
 
 export class GuCertificateArnParameter extends GuStringParameter {
-  constructor(scope: GuStack, id: string = "TLSCertificate", props?: GuNoTypeParameterProps) {
-    super(scope, id, {
+  constructor(scope: GuStack, props: AppIdentity) {
```
akash1810, line 7: This gives the ability to have a `GuCertificateArnParameter` for each `app` defined in a  stack 🎉 .

src/constructs/iam/policies/ses.ts
```diff
@@ -1,22 +1,14 @@
...
 import { GuGuardianEmailSenderParameter } from "../../core";
-import type { GuPolicyProps } from "./base-policy";
-import { GuPolicy } from "./base-policy";
-
-export class GuSESSenderPolicy extends GuPolicy {
-  constructor(scope: GuStack, id: string = "GuSESSenderPolicy", props?: GuPolicyProps) {
-    super(scope, id, { ...props });
+import { GuAllowPolicy } from "./base-policy";
 
+export class GuSESSenderPolicy extends GuAllowPolicy {
```
akash1810, line 5: Extending `GuAllowPolicy` yields a simpler implementation of this policy 🎉 .

eslint/rules/valid-constructors.js
```diff
@@ -1,38 +1,102 @@
...
-const getPath = (node, path) => {
-  const keys = path.split(".");
-  let value = node;
-
-  for (const key of keys) {
-    if (!value) break;
-    value = value[key];
+/*
+RULES
+1. Private constructors don't get linted
```
jacobwinch, line 3: Good thinking 💡

src/constructs/iam/policies/ses.ts
jacobwinch: 💯

docs/architecture-decision-records/002-component-constuctors.md
```diff
@@ -46,21 +46,21 @@ Constructors should follow the following rules for consistency.
...
+     constructor(scope: Stack) {
        ...
      }
    }
    ```
 
-2. They can also take a `props` object which should be correctly typed
+   The construct/pattern will then have a static `id` as it will never change, for example the `Stage` parameter.
+
+2. They should all accept a `scope` of type `GuStack` and `props` of a custom type:
```
jacobwinch, line 57: I don't think that "should _all_" is quite true? I think the original phrasing ("They can also take a `props` object which should be correctly typed"), is probably still fine here?

jacobwinch:
https://github.com/guardian/cdk/pull/352/commits/a9b5a0695c5a2fa2cde032de8f6f674a931b1d52 was a very pleasant commit to review 🏅

eslint/rules/valid-constructors.js
```diff
@@ -1,38 +1,102 @@
...
+7. Only the third parameter can be optional or have a default value
+
+See `valid-constructors.test.ts` and `npm run test:custom-lint-rule`
+
+TODO: If all values in third type are optional then parameter should be optional
+ */
+
+const NO_LINT_ERRORS = null;
+
+const lintParameter = (param, node, context, { name, type, allowOptional, allowDefault, position }) => {
```
jacobwinch, line 24: This is neat!

jacobwinch (approved):
This looks great; I've just made [one small suggestion](https://github.com/guardian/cdk/pull/352#discussion_r603216251) related to the docs that you might want to address before merging.

Feel free to merge once conflicts have been resolved 🚢

docs/architecture-decision-records/002-component-constuctors.md
akash1810: 👍🏽  have reverted that bit and force pushed.

PR #362, opened 2021-03-30
fix: simplify props for `GuClassicLoadBalancer`

akash1810:
## What does this change?

Removes the `propertiesToRemove` prop. It looks like we were using this to set the scheme on the load balancer, however this is already being driven by the `internetFacing` prop provided by AWS.

For a scheme of "internal", set `internetFacing` to `false`. For a scheme of "internet-facing", set `internetFacing` to `true`.

See:
- https://github.com/aws/aws-cdk/blob/fb6512314db1b11fc608cd62753090684ad0d3c4/packages/%40aws-cdk/aws-elasticloadbalancing/lib/load-balancer.ts#L249
- https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-elb.html#cfn-ec2-elb-scheme

Yes - the props to `GuClassicLoadBalancer` is changing, with `propertiesToRemove` disappearing, so stacks would need to migrate. The migration is just setting a boolean however, so not too bad.

## How to test

n/a

A simpler API to the construct.

n/a

### DIFF
--- a/src/constructs/loadbalancing/elb.test.ts
+++ b/src/constructs/loadbalancing/elb.test.ts
@@ -46,18 +46,6 @@ describe("The GuClassicLoadBalancer class", () => {
     expect(Object.keys(json.Resources)).not.toContain("ClassicLoadBalancer");
   });
 
-  test("deletes any provided properties", () => {
-    const stack = simpleGuStackForTesting();
-    new GuClassicLoadBalancer(stack, "ClassicLoadBalancer", {
-      vpc,
-      overrideId: true,
-      propertiesToRemove: [GuClassicLoadBalancer.RemoveableProperties.SCHEME],
-    });
-
-    const json = SynthUtils.toCloudFormation(stack) as SynthedStack;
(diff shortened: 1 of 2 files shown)


jacobwinch (approved):
Seems sensible to me 👍 

If `@aws/cdk` is adding this property by default, perhaps we should mention it [here](https://github.com/guardian/cdk/blob/main/docs/003-migrating-existing-stacks.md#changes) (I think it's optional in CFN, so many stacks may not have this value explicitly set).

akash1810:
> If @aws/cdk adds this property by default, perhaps we should mention it here (I think it's optional in CFN, so many stacks may not have this value explicitly set).

This change is part of my bigger plans to simplify the load balancing constructs. That is, 👍🏽  to documentation, however I'd be tempted to delay it once the other load balancing changes have shipped too.

The theme of the change is to create a set of custom props, rather than extending AWS's. This should make it easier to set defaults and also to understand what those defaults are.

PR #364, opened 2021-03-30
refactor: Encapsulate overriding of logicalId in a single place

akash1810:
## What does this change?

A resource's logicalId maps to the name of the created resource.

For example, take this CloudFormation YAML for an ASG:

```yaml
DataNodeAutoScalingGroup:
  Type: AWS::AutoScaling::AutoScalingGroup
  Properties:
    AvailabilityZones: !GetAZs ""
    LaunchConfigurationName: !Ref DataNodeLaunchConfig
    MinSize: 3
    MaxSize: 4
    Cooldown: 180
    HealthCheckType: EC2
    HealthCheckGracePeriod: 300
    VPCZoneIdentifier: !Ref PrivateVpcSubnets
    Tags:
    - { Key: Stack, Value: !Ref Stack, PropagateAtLaunch: true }
    - { Key: App, Value: "elk-es-data", PropagateAtLaunch: true }
    - { Key: Stage, Value: !Ref Stage, PropagateAtLaunch: true }
    - { Key: ElasticSearchCluster, Value: !Ref ClusterName, PropagateAtLaunch: true }
```

This creates an ASG called `DataNodeAutoScalingGroup`. This pattern is widely used in YAML definitions of CloudFormation as YAML files are typically hand-written.

CDK changes this as it auto-generates the logicalIds, as there one will gain more understanding about the resource from looking at the CDK template.

In order to enable a no-op migration between a YAML stack a CDK stack, we allow the logicalId to be overridden.
Our constructs seem to follow different rules regarding when to override the logicalId of a CloudFormation resource.

For example, we have [this](https://github.com/guardian/cdk/blob/60cbf21698b3ddbb8cfd23556fd2af886e9551c7/src/constructs/rds/instance.ts#L38-L40):

```typescript
if (props.overrideId || (scope.migratedFromCloudFormation && props.overrideId !== false))
```

And [this](https://github.com/guardian/cdk/blob/60cbf21698b3ddbb8cfd23556fd2af886e9551c7/src/constructs/iam/policies/base-policy.ts#L15-L18):

```typescript
if (props.overrideId) {
```

This change aims to bring consistency and simplicity across our constructs by applying a universal rule that,
`migratedFromCloudFormation` on the `GuStack` must be `true` and `existingLogicalId` to be set for the logicalId to be overridden.

If these conditions are not met, we let CDK auto-generate the logicalId.

It's worth noting that this doesn't implement the change on the constructs, it's simply to define the behaviour.
This is mainly to keep the diff down - we'll migrate the constructs in follow-up PRs.

Usage will roughly look like this:

```typescript
const MyComponentProps extends GuMigratingResource { }

class MyComponent extends SomeGuConstruct {
  constructor(scope: GuStack, id: string, props: MyComponentProps) {
    super(scope, id, props);
    GuMigratingResource.setLogicalId(this, scope, props);
  }
}
```

No.

## How to test

See added tests.

Consistency in behaviour across constructs.

n/a

### DIFF
--- a/src/constructs/core/migrating.test.ts
+++ b/src/constructs/core/migrating.test.ts
@@ -1,0 +1,122 @@
+import "@aws-cdk/assert/jest";
+import { SynthUtils } from "@aws-cdk/assert";
+import type { BucketProps } from "@aws-cdk/aws-s3";
+import { Bucket } from "@aws-cdk/aws-s3";
+import type { SynthedStack } from "../../utils/test";
+import { simpleGuStackForTesting } from "../../utils/test";
+import type { GuStatefulConstruct } from "./migrating";
+import { GuMigratingResource } from "./migrating";
+import type { GuStack } from "./stack";
+
+class TestGuStatefulConstruct extends Bucket implements GuStatefulConstruct {
+  isStatefulConstruct: true;
+
+  constructor(scope: GuStack, id: string, props?: BucketProps) {
+    super(scope, id, props);
+    this.isStatefulConstruct = true;
+  }
+}
+
+/*
+NOTE: In reality, we'd never directly call `GuMigratingResource.setLogicalId` as the GuConstruct will do that.
+We're calling it here to test the function in isolation.
+ */
+
+describe("GuMigratingResource", () => {
+  // eslint-disable-next-line @typescript-eslint/no-empty-function -- we are testing `console.info` being called, we don't need to see the message
+  const info = jest.spyOn(console, "info").mockImplementation(() => {});
+
+  // eslint-disable-next-line @typescript-eslint/no-empty-function -- we are testing `console.warn` being called, we don't need to see the message
+  const warn = jest.spyOn(console, "warn").mockImplementation(() => {});
+
+  afterEach(() => {
+    warn.mockReset();
+    info.mockReset();
+  });
+
+  test("Creating a new resource in a new stack produces a new auto-generated ID", () => {
+    const stack = simpleGuStackForTesting({ migratedFromCloudFormation: false });
+    const construct = new Bucket(stack, "MyBucket");
+
+    GuMigratingResource.setLogicalId(construct, stack, {});
+
+    expect(warn).toHaveBeenCalledTimes(0);
+    expect(info).toHaveBeenCalledTimes(0);
+
+    const json = SynthUtils.toCloudFormation(stack) as SynthedStack;
+    const resourceKeys = Object.keys(json.Resources);
+
+    expect(resourceKeys).toHaveLength(1);
+    expect(resourceKeys[0]).toMatch(/^MyBucket[A-Z0-9]+$/);
+  });
+
+  test("Keeping a resource's logicalId when migrating a stack", () => {
+    const stack = simpleGuStackForTesting({ migratedFromCloudFormation: true });
+    const construct = new Bucket(stack, "MyBucket");
+
+    GuMigratingResource.setLogicalId(construct, stack, { existingLogicalId: "my-pre-existing-bucket" });
+
+    expect(warn).toHaveBeenCalledTimes(0);
+    expect(info).toHaveBeenCalledTimes(0);
+
+    const json = SynthUtils.toCloudFormation(stack) as SynthedStack;
+    expect(Object.keys(json.Resources)).toContain("my-pre-existing-bucket");
+  });
+
+  test("Creating a construct in a migrating stack, w/out setting existingLogicalId", () => {
+    const stack = simpleGuStackForTesting({ migratedFromCloudFormation: true });
+    const construct = new Bucket(stack, "MyBucket");
+
+    GuMigratingResource.setLogicalId(construct, stack, {});
+
+    expect(info).toHaveBeenCalledTimes(0);
+    expect(warn).toHaveBeenCalledTimes(0);
+
+    const json = SynthUtils.toCloudFormation(stack) as SynthedStack;
+    const resourceKeys = Object.keys(json.Resources);
+
+    expect(resourceKeys).toHaveLength(1);
+    expect(resourceKeys[0]).toMatch(/^MyBucket[A-Z0-9]+$/);
+  });
+
+  test("Specifying a construct's existingLogicalId in a new stack", () => {
+    const stack = simpleGuStackForTesting({ migratedFromCloudFormation: false });
+    const construct = new Bucket(stack, "MyBucket");
+
+    GuMigratingResource.setLogicalId(construct, stack, { existingLogicalId: "my-pre-existing-bucket" });
+
+    expect(info).toHaveBeenCalledTimes(0);
+    expect(warn).toHaveBeenCalledTimes(1);
+    expect(warn).toHaveBeenCalledWith(
+      "GuStack has 'migratedFromCloudFormation' set to false. MyBucket has an 'existingLogicalId' set to my-pre-existing-bucket. This will have no effect - the logicalId will be auto-generated. Set 'migratedFromCloudFormation' to true for 'existingLogicalId' to be observed."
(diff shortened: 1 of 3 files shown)


src/constructs/core/migrating.ts
```diff
@@ -0,0 +1,41 @@
...
+    { existingLogicalId }: GuMigratingResource
+  ): void {
+    const overrideLogicalId = (logicalId: string) => {
+      const defaultChild = construct.node.defaultChild as CfnElement;
+      defaultChild.overrideLogicalId(logicalId);
+    };
+
+    const id = construct.node.id;
+
+    if (!migratedFromCloudFormation && existingLogicalId) {
```
jacobwinch, line 21: I hadn't really thought about it before, but I guess this could be a valid thing to do (in the future) if the user has created a resource via one cdk template and is now trying to move it into another? It's a bit of an edge case though, so feel free to ignore!

jacobwinch: Actually I'm not sure this makes sense at all - definitely ignore it!

akash1810: It's odd, right?! This behaviour is implicitly in place in things like [GuPolicy](https://github.com/guardian/cdk/blob/60cbf21698b3ddbb8cfd23556fd2af886e9551c7/src/constructs/iam/policies/base-policy.ts#L15-L18) as it's not checking the value of `migratedFromCloudFormation`.

It sounds like you're suggesting we only ever observe `existingLogicalId` if `migratedFromCloudFormation` is explicitly true? That makes perfect sense to me!

jacobwinch (approved):
Looks good to me 👍

src/constructs/core/migrating.test.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  // eslint-disable-next-line @typescript-eslint/no-empty-function -- we are testing `console.error` being called, we don't need to see the message
+  const error = jest.spyOn(console, "error").mockImplementation(() => {});
+
+  afterEach(() => {
+    warn.mockReset();
+    error.mockReset();
+  });
+
+  test("creating a new resource in a new stack", () => {
```
stephengeller, line 28: Being annoying here, but could we capture the expected behaviour for when we create a new resource in a new stack? eg

```suggestion
  test("creating a new resource in a new stack produces a new auto-generated ID", () => {
```

src/constructs/core/migrating.test.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  test("Migrating a stack and falling back to the construct's ID when existingLogicalId is not set", () => {
+    const stack = simpleGuStackForTesting({ migratedFromCloudFormation: true });
+    const construct = new Bucket(stack, "MyBucket");
+
+    GuMigratingResource.setLogicalId(construct, stack, {});
+
+    expect(warn).toHaveBeenCalledTimes(1);
+    expect(warn).toHaveBeenCalledWith(
+      "The migratedFromCloudFormation prop on the GuStack is true, however existingLogicalId has not been set on the construct MyBucket. This will create a resource with a new, unique, logicalId. Is this correct?"
```
stephengeller, line 65: Is this warning right? I'm likely misreading it, but the test name suggests that it'll fall back to the construct's ID, but the warning says it'll create a resource with a new, unique logicalId. I'm not sure if I'm reading this correctly though.

src/constructs/core/migrating.test.ts
```diff
@@ -0,0 +1,105 @@
...
+    );
+
+    const json = SynthUtils.toCloudFormation(stack) as SynthedStack;
+    expect(Object.keys(json.Resources)).toContain("MyBucket");
```
stephengeller, line 69: If the above warning was true, I would expect this to be a unique ID, not the ID of the construct itself, no?

src/constructs/core/migrating.ts
stephengeller: Yeah, I guess we really want to account for _any_ migration, not just from a CFN YAML but also from CDK. 

I'm wondering if there's ever a situation in which you're _not_ migrating the stack, but you _are_ migrating the resource. If there isn't, then we can definitely follow the suggestion above to only observe `existingLogicalId` if `migratedFromCloudFormation` is true.

stephengeller: Yeah, I guess we really want to account for _any_ migration, not just from a CFN YAML but also from CDK. 

I'm wondering if there's ever a situation in which you're _not_ migrating the stack, but you _are_ migrating the resource. If there isn't, then we can definitely follow the suggestion above to only observe `existingLogicalId` if `migratedFromCloudFormation` is true.

akash1810: > Yeah, I guess we really want to account for any migration, not just from a CFN YAML but also from CDK.

I suspect a CDK to Gu-CDK migration would generate the same ID if the output of the construct looks the same. For example, if migrating a policy and the policy documents matched.

If we do migrate between CDK and Gu-CDK, it would be intuitive to set `migratedFromCloudFormation` to true as it _is_ a migration. That said, I'm increasingly thinking we should mandate `migratedFromCloudFormation` be true for `existingLogicalId` to be observed.

akash1810:
@stephengeller @jacobwinch I've re-worked and re-worded this to make the logic more assertive, principally `migratedFromCloudFormation` has to be set - the function is greatly simplified and is more readable as a result. A review would be welcomed.

src/constructs/core/migrating.ts
```diff
@@ -0,0 +1,55 @@
...
+    const isStateful = isGuStatefulConstruct(construct);
+
+    if (migratedFromCloudFormation) {
+      if (existingLogicalId) {
+        return overrideLogicalId(existingLogicalId);
+      }
+
+      if (isStateful) {
+        console.warn(
+          `GuStack has 'migratedFromCloudFormation' set to true. ${id} is a stateful construct and 'existingLogicalId' has not been set. ${id}'s logicalId will be auto-generated. This is not advised as downstream services, such as DNS, will likely need updating.`
```
jacobwinch, line 38: Perhaps we can be explicit here about the impact of auto-generated logical ids?

```suggestion
          `GuStack has 'migratedFromCloudFormation' set to true. ${id} is a stateful construct and 'existingLogicalId' has not been set. ${id}'s logicalId will be auto-generated and consequently AWS will create a new resource rather than inheriting an existing one. This is not advised as downstream services, such as DNS, will likely need updating.`
```

src/constructs/core/migrating.ts
```diff
@@ -0,0 +1,55 @@
...
+        );
+      }
+    } else {
+      if (existingLogicalId) {
+        console.warn(
+          `GuStack has 'migratedFromCloudFormation' set to false. ${id} has an 'existingLogicalId' set to ${existingLogicalId}. This will have no effect - the logicalId will be auto-generated. Set 'migratedFromCloudFormation' to true for 'existingLogicalId' to be observed.`
```
jacobwinch, line 44: Much clearer 👍

src/constructs/core/migrating.ts
```diff
@@ -0,0 +1,55 @@
...
+        );
+      }
+
+      if (isStateful) {
+        console.info(
+          `GuStack has 'migratedFromCloudFormation' set to false. ${id} is a stateful construct, it's logicalId will be new and auto-generated.`
```
jacobwinch, line 50: Minor suggestion:

```suggestion
          `GuStack has 'migratedFromCloudFormation' set to false. ${id} is a stateful construct, it's logicalId will be auto-generated and AWS will create a new resource.`
```

jacobwinch (approved):
Great work - thanks for taking the time to document and simplify this 💯

akash1810:
Now requires https://github.com/guardian/cdk/pull/388.

PR #365, opened 2021-03-30
fix: change to output as ES modules

akash1810:
## What does this change?

This library is broken since v6.0.1 after PR #356. Users face this stack trace:

```
Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /Users/philip_mcmahon/code/editions/projects/aws/node_modules/read-pkg-up/index.js
require() of ES modules is not supported.
require() of /Users/philip_mcmahon/code/editions/projects/aws/node_modules/read-pkg-up/index.js from /Users/philip_mcmahon/code/editions/projects/aws/node_modules/@guardian/cdk/lib/constants/library-info.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename index.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from /Users/philip_mcmahon/code/editions/projects/aws/node_modules/read-pkg-up/package.json.
```

This change updates tsconfig to output as ES modules rather than commonjs. This slightly goes against the recommendations (https://github.com/guardian/recommendations/blob/master/npm-packages.md#compiling), however as the library only gets used in controlled environments, this should be fine.

No.

## How to test

- Checkout the branch
- npm link the project
- Update an existing project to use the linked version of `@guardian/cdk`
- Synth a stack
- Observe no errors

The library can be used again!

n/a

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,9 +1,10 @@
 {
   "compilerOptions": {
     "target": "ES2020",
-    "module": "commonjs",
+    "module": "ES2020",
     "moduleResolution": "node",
     "declaration": true,
+    "declarationMap": true,
     "esModuleInterop": true,
     "strict": true,
     "noUnusedLocals": true,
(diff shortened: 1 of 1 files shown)


philmcmahon (approved):
🚀

PR #367, opened 2021-03-30
chore: add an integration test project

akash1810:
## What does this change?

This PR creates an integration test 🎉 and requires https://github.com/guardian/cdk/pull/368 (I can't quite get the commit hashes between this and #368 to line up, so the diff here looks massive. Will rebase once #368 is merged).

A new project is added to the "integration-test" directory. To ensure it's using the latest version of `@guardian/cdk`, this project installs it from a relative path. The integration-test project doesn't do too much, it will:
- Create an empty `GuStack`
- Perform a snapshot based test
- Synth the stack

This is in reaction to #356 and #365 which broke the package. The integration test helps to ensure any changes we make to the package in the future work, _before_ publishing them. That is, the feedback loop is shortened.

No.

## How to test

The CI step has been updated, so see that?

We have, basic, integration tests on the project.

n/a


jacobwinch (approved):
Nice idea 💯

PR #369, opened 2021-03-31
fix: compile to commonjs

akash1810:
## What does this change?

**TL;DR** Publishing `@guardian/cdk` in ESM format is tricky, it is easiest to stay on CommonJS.

In https://github.com/guardian/cdk/pull/365, we changed the output format to ES Modules. This was after https://github.com/guardian/cdk/pull/356 which bumped `read-pkg-up` to v8.0.0.

Looking at `read-pkg-up`'s [release notes for v8.0.0](https://github.com/sindresorhus/read-pkg-up/releases/tag/v8.0.0) the package has changed to ESM only and consuming projects [need to](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c):

> 1. Use ESM yourself. (preferred)
> Use import foo from 'foo' instead of const foo = require('foo') to import the package.
> 2. If the package is used in an async context, you could use await import(…) instead of require(…).
> 3. Stay on the existing version of the package until you can move to ESM.

More information can be found [here](https://blog.sindresorhus.com/get-ready-for-esm-aa53530b3f77).

#365 incorrectly did option 1 as `"type": "module"` was not added to `package.json`. However, even with that change, it is pretty difficult for consumers of `@guardian/cdk` to use an ESM release as required tooling such as [ts-node](https://github.com/TypeStrong/ts-node/issues/1007) (see also [this](https://github.com/TypeStrong/ts-node/issues/935)), [ESLint](https://github.com/eslint/rfcs/pull/43) and [Jest](https://jestjs.io/docs/ecmascript-modules) have varying support.

This PR does option 3. That is, publishing `@guardian/cdk` in ESM format is tricky, it is easiest to stay on CommonJS. We'll have to stay on v7.0.1 of `read-pkg-up` to achieve this; v8.0.0 doesn't bring any new features or fixes, so this is safe.

Furthermore, after a couple of hours, I couldn't get option 2 to work, even though [top level await is supported](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#top-level-await). We can experiment later.

It's worth noting that https://github.com/guardian/cdk/pull/367 plans to introduce an integration test to this repository, which should make experimenting with various module formats with more confidence. Currently, this test will define an empty `GuStack` and attempt to synth it.

Lastly, although #365 suggests the change was tested locally, I suspect it was fluke that it worked.

No.

## How to test

n/a - we're getting back into the same state as we were before #356 and [`@guardian/cdk` v6.0.1](https://github.com/guardian/cdk/pull/356#issuecomment-809291683).

We publish a usable package again!

n/a

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6191,8 +6191,7 @@
     "hosted-git-info": {
       "version": "2.8.8",
       "resolved": "https://registry.npmjs.org/hosted-git-info/-/hosted-git-info-2.8.8.tgz",
-      "integrity": "sha512-f/wzC2QaWBs7t9IYqB4T3sR1xviIViXJRJTWBlx2Gf3g0Xi5vI7Yy4koXQ1c9OYDGHN9sBy1DQ2AB8fqZBWhUg==",
-      "dev": true
+      "integrity": "sha512-f/wzC2QaWBs7t9IYqB4T3sR1xviIViXJRJTWBlx2Gf3g0Xi5vI7Yy4koXQ1c9OYDGHN9sBy1DQ2AB8fqZBWhUg=="
     },
     "html-encoding-sniffer": {
       "version": "2.0.1",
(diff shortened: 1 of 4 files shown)


jacobwinch (approved):
Thanks for taking the time to explain this in detail 👍

akash1810:
> Thanks for taking the time to explain this in detail 👍

Module formats in Node are complicated! I hope my explanation is correct 😬 !

PR #371, opened 2021-04-01
feat: make Wazuh access security group a singleton

akash1810:
## What does this change?

We'd only ever want one of these resources defined in a stack as it can be re-used by anything that needs it. This helps keep the overall size of the template down and thus reduces chance of reaching the file size limits of cloudformation.

This change also refactors the security groups, splitting them across multiple files to make them easier to reason about.

Reviewing commit by commit might be easiest.

Yes. Usage of the Wazuh security group would need to change.

## How to test

See tests.

A simpler API and reduced chance of reaching the file size limits of cloudformation

n/a


jacobwinch (approved):
Nice work!

PR #374, opened 2021-04-01
fix: use high resolution timer for GuSSMIdentityParameter id

akash1810:
## What does this change?

We're seeing the `GuSSMIdentityParameter` tests sporadically fail (for example, [here](https://github.com/guardian/cdk/runs/2234965115?check_suite_focus=true#step:4:104)).

From the code:

> We need to create UIDs for the resources in this construct, as otherwise CFN will not trigger the lambda on updates for resources that appear to be the same

This PR switches to use `performance.now()` for a [high resolution timer](https://www.w3.org/TR/hr-time/) which provides more accuracy than `Date.now()` and thus increases the chance of uniqueness.

I suspect we could also use `Math.random`, but not sure.

This change also adds `ssm`'s exports to `core` and  moves the `id` function to a static function on `GuSSMIdentityParameter` to provide a more explicit namespace. This means usage becomes:

```typescript
import { GuSSMIdentityParameter } from "constructs/core";

GuSSMIdentityParameter.id(???)
```

vs.

```typescript
import { id } from "constructs/core";

id(???)
```

No, the changes are to the internal interfaces of the project.

## How to test

See tests.

We no longer see the sporadic failure of tests 🙏🏽 .

n/a

### DIFF
--- a/src/constructs/core/index.ts
+++ b/src/constructs/core/index.ts
@@ -1,3 +1,4 @@
 export * from "./mappings";
 export * from "./parameters";
 export * from "./stack";
+export * from "./ssm";
--- a/src/constructs/core/ssm.test.ts
+++ b/src/constructs/core/ssm.test.ts
@@ -48,11 +48,21 @@ describe("SSM:", () => {
 
   describe("the id function", function () {
     it("creates a unique ID given a string", function () {
-      expect(id("NameOfConstruct", "some-parameter")).toMatch(/NameOfConstruct-someparameter/i);
+      const actual1 = id("NameOfConstruct", "some-parameter");
+      const actual2 = id("NameOfConstruct", "some-parameter");
+
+      expect(actual1).toMatch(/NameOfConstruct-someparameter/i);
+      expect(actual2).toMatch(/NameOfConstruct-someparameter/i);
+      expect(actual1).not.toEqual(actual2);
     });
 
     it("will substitute a CDK token for an acceptable string", function () {
-      expect(id("NameOfConstruct", "${TOKEN}foobar")).toMatch(/NameOfConstruct-token/i);
+      const actual1 = id("NameOfConstruct", "${TOKEN}foobar");
+      const actual2 = id("NameOfConstruct", "${TOKEN}foobar");
+
+      expect(actual1).toMatch(/NameOfConstruct-token/i);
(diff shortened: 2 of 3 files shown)


jacobwinch (approved):
Great work - here's to future test runs which look like this!
✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅

akash1810:
ARGH! CD [failed](https://github.com/guardian/cdk/runs/2245552343?check_suite_focus=true#step:4:30) on this for the exact reason it's trying to fix. This library has an in-built April Fools 😂 .

```
FAIL src/constructs/core/ssm.test.ts
  ● SSM: › The GuSSMParameter construct › creates unique IDs using param and stack name for the parameters

    There is already a Construct with name 'Provider-pathtosomeparam-9993' in GuStack [Test]

      41 |     const { parameter } = props;
      42 | 
    > 43 |     const provider = new SingletonFunction(scope, GuSSMParameter.id("Provider", parameter), {
         |                      ^
      44 |       code: Code.fromInline(readFileSync(join(__dirname, "/custom-resources/runtime/lambda.js")).toString()),
      45 |       runtime: Runtime.NODEJS_12_X,
      46 |       handler: "index.handler",

      at Node.addChild (node_modules/constructs/src/construct.ts:534:13)
      at new Node (node_modules/constructs/src/construct.ts:75:22)
      at new ConstructNode (node_modules/@aws-cdk/core/lib/construct-compat.ts:277:24)
      at Object.createNode (node_modules/@aws-cdk/core/lib/construct-compat.ts:71:11)
      at new Construct (node_modules/constructs/src/construct.ts:575:26)
      at new Construct (node_modules/@aws-cdk/core/lib/construct-compat.ts:68:5)
      at new Resource (node_modules/@aws-cdk/core/lib/resource.ts:127:5)
      at new FunctionBase (node_modules/@aws-cdk/aws-lambda/lib/function-base.js:17:9)
      at new SingletonFunction (node_modules/@aws-cdk/aws-lambda/lib/singleton-lambda.ts:53:5)
      at new GuSSMParameter (src/constructs/core/ssm.ts:43:22)
      at Object.<anonymous> (src/constructs/core/ssm.test.ts:32:22)
```

PR #378, opened 2021-04-01
feat: allow GuSecurityGroup to be specified at the app level

akash1810:
## What does this change?

Refactor `GuSecurityGroup` into the vanilla abstract class `GuBaseSecurityGroup` and the concrete class `GuSecurityGroup` which has an `AppIdentity`.

This allows a stack to create a security group scoped to a particular app within a multi app stack.

`GuWazuhAccess` now extends `GuBaseSecurityGroup` as it is a singleton and applies to all apps within a stack.

Follows https://github.com/guardian/cdk/pull/326.

Yes, they will need to pass in an `AppIdentity` as the props for `GuSecurityGroup` have changed.

## How to test

No new tests added. Existing tests should pass.

It is possible to define a security group for a specific app within a stack and have it tagged with the `App` tag.

n/a

### DIFF
--- a/src/constructs/ec2/security-groups/base.ts
+++ b/src/constructs/ec2/security-groups/base.ts
@@ -1,6 +1,7 @@
 import type { CfnSecurityGroup, IPeer, SecurityGroupProps } from "@aws-cdk/aws-ec2";
 import { Peer, Port, SecurityGroup } from "@aws-cdk/aws-ec2";
 import type { GuStack } from "../../core";
+import { AppIdentity } from "../../core/identity";
 
 /**
  * A way to describe an ingress or egress rule for a security group.
@@ -27,11 +28,13 @@ export interface SecurityGroupAccessRule {
   description: string;
 }
 
-export interface GuSecurityGroupProps extends SecurityGroupProps {
+export interface GuBaseSecurityGroupProps extends SecurityGroupProps {
   overrideId?: boolean;
   ingresses?: SecurityGroupAccessRule[];
   egresses?: SecurityGroupAccessRule[];
 }
+
+export interface GuSecurityGroupProps extends GuBaseSecurityGroupProps, AppIdentity {}
 
 /**
  * Defining an AWS Security Group with ingress and egress rules.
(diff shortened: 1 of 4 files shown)


src/constructs/ec2/security-groups/base.ts
```diff
@@ -78,17 +89,19 @@ export class GuPublicInternetAccessSecurityGroup extends GuSecurityGroup {
...
+// TODO should this be a singleton?
 export class GuHttpsEgressSecurityGroup extends GuSecurityGroup {
-  constructor(scope: GuStack, id: string, props: SecurityGroupProps) {
+  constructor(scope: GuStack, id: string, props: GuSecurityGroupProps) {
     super(scope, id, {
-      vpc: props.vpc,
+      ...props,
       allowAllOutbound: false,
       description: "Allow all outbound HTTPS traffic",
+      ingresses: [],
```
akash1810, line 99: This has the effect of removing any ingresses provided on the `props`. Ideally the interface of `props` should omit this field; I'm planning that in a future bit of work!

src/constructs/ec2/security-groups/base.ts
```diff
@@ -68,8 +71,16 @@ export class GuSecurityGroup extends SecurityGroup {
...
 }
 
+export class GuSecurityGroup extends GuBaseSecurityGroup {
+  constructor(scope: GuStack, id: string, props: GuSecurityGroupProps) {
+    super(scope, AppIdentity.suffixText(props, id), props);
+    AppIdentity.taggedConstruct(props, this);
+  }
+}
+
+// TODO should this be a singleton?
```
jacobwinch, line 81: Yes, I think so

src/constructs/ec2/security-groups/base.ts
```diff
@@ -78,17 +89,19 @@ export class GuPublicInternetAccessSecurityGroup extends GuSecurityGroup {
   }
 }
 
+// TODO should this be a singleton?
```
jacobwinch, line 92: As above!

jacobwinch (approved):

src/constructs/ec2/security-groups/base.ts
akash1810: Will do this in a later PR.

PR #387, opened 2021-04-06
refactor!: Improve Mappings helper function(s)

jacobwinch:
## What does this change?

The `setStageDependentValue` and `getStageDependentValue` helper functions have been removed. In order to reduce boilerplate, the same functionality is now provided via a single function (`withStageDependentValue`).

Yes, this is a breaking change. Stacks which use `setStageDependentValue` and `getStageDependentValue` directly will need to use `withStageDependentValue` instead. Stacks which use this functionality via the ASG construct should be unaffected.

## How to test

[Existing ASG unit tests](https://github.com/guardian/cdk/blob/3bbd3124819876ef0dc1a4c8a3e361d7481e01c8/src/constructs/autoscaling/asg.test.ts#L181-L258) already validate this behaviour (and they still pass!).

* Less boilerplate is needed to use stage-dependent values.

As mentioned above, this is a breaking change. However, as this functionality is not widely adopted yet I think it's relatively low risk to make the change at this early stage of development.

### DIFF
--- a/src/constructs/autoscaling/asg.ts
+++ b/src/constructs/autoscaling/asg.ts
@@ -4,7 +4,7 @@ import type { ISecurityGroup, MachineImage, MachineImageConfig } from "@aws-cdk/
 import { InstanceType, OperatingSystemType, UserData } from "@aws-cdk/aws-ec2";
 import type { ApplicationTargetGroup } from "@aws-cdk/aws-elasticloadbalancingv2";
 import { Stage } from "../../constants";
-import type { GuStack, GuStageDependentValue } from "../core";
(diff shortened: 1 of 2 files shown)


jacobwinch:
> Stacks which use `setStageDependentValue` and `getStageDependentValue` directly will need to use `withStageDependentValue` instead.

cc @mbarton as I think you might be the only affected user so far! It would be good to know if you think that this is an improvement or not.

mbarton (approved):
👍 this makes it much less likely for people to try and do logic on it or assign the result to a variable

akash1810 (approved):

PR #390, opened 2021-04-07
feat: add ACM certificate construct

jacobwinch:
## What does this change?
This change adds a new construct for creating an ACM certificate. 

No.

## How to test
I've added unit tests to cover the new functionality.

We are able to create more resources using `@guardian/cdk`.

Adding a new construct is low risk.


src/constructs/acm/__snapshots__/certificate.test.ts.snap
```diff
@@ -0,0 +1,147 @@
...
+        "PROD",
+      ],
+      "Default": "CODE",
+      "Description": "Stage name",
+      "Type": "String",
+    },
+  },
+  "Resources": Object {
+    "TestCertificate6B4956B6": Object {
+      "DeletionPolicy": "Retain",
```
jacobwinch, line 28: I think that adding this deletion policy will improve safety and [allow us to import certificates created outside of CloudFormation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resource-import.html#resource-import-considerations) 🤞

src/constructs/acm/certificate.ts
```diff
@@ -0,0 +1,43 @@
+import { Certificate, CertificateValidation } from "@aws-cdk/aws-certificatemanager";
+import type { CertificateProps } from "@aws-cdk/aws-certificatemanager/lib/certificate";
+import { HostedZone } from "@aws-cdk/aws-route53";
+import { RemovalPolicy } from "@aws-cdk/core";
+import { Stage } from "../../constants";
+import type { GuStack } from "../core";
+import { GuMigratingResource } from "../core/migrating";
+import type { GuStatefulConstruct } from "../core/migrating";
+
+type GuCertificateProps = Record<Stage, GuDnsValidatedCertificateProps> & GuMigratingResource;
```
jacobwinch, line 10: This is my first attempt at using https://github.com/guardian/cdk/pull/364 - hopefully this is roughly what you were expecting @akash1810!

akash1810: Took a second to parse that, but yep looks correct 👍🏽 . You might want to add [`AppIdentity`](https://github.com/guardian/cdk/blob/e4a8dfa0faa765543e96b8fd28739be881f9132d/src/constructs/core/identity.ts#L9) too so we can add the `app` tag.

src/constructs/acm/certificate.ts
```diff
@@ -0,0 +1,87 @@
...
+  hostedZoneId?: string;
+}
+
+/**
+ * Construct which creates an ACM Certificate.
+ *
+ * If your DNS is managed via Route 53, then supplying `hostedZoneId` props will allow AWS to automatically
+ * validate your certificate.
+ *
+ * If your DNS is not managed via Route 53, or you omit the `hostedZoneId` props, then the CloudFormation
```
jacobwinch, line 24: This advice is based on https://github.com/aws/aws-cdk/blob/7966f8d48c4bff26beb22856d289f9d0c7e7081d/packages/%40aws-cdk/aws-certificatemanager/lib/certificate.ts#L101-L116. In the future, perhaps we can use a custom resource to create the appropriate records elsewhere (NS1?) automatically?

src/constructs/acm/certificate.ts
```diff
@@ -0,0 +1,87 @@
...
+
+export type GuCertificateProps = Record<Stage, GuDnsValidatedCertificateProps> & GuMigratingResource & AppIdentity;
+
+export interface GuDnsValidatedCertificateProps {
+  domainName: string;
+  hostedZoneId?: string;
+}
+
+/**
+ * Construct which creates an ACM Certificate.
```
akash1810, line 19: minor: Should we make it explicit that this is a DNS validated cert? Either by mentioning it in this comment or in the naming of the construct?

akash1810 (approved):
Fantastic!

src/constructs/acm/certificate.ts
jacobwinch: Good idea, I've [updated the docs](https://github.com/guardian/cdk/pull/390/commits/17a65a5c711727d39a613e3b8bab64e9d3024758).

PR #392, opened 2021-04-07
fix: Improve GuAutoScalingGroup's support of a multi-app stack

akash1810:
## What does this change?

For consistency with other constructs, suffix the `id` of `GuAutoScalingGroup` with the `app`. This ensures the logicalId is unique both on the auto-generated path and the `existingLogicalId` path.

Also adds a [custom Jest matcher](https://jestjs.io/docs/expect#custom-matchers-api) to simplify logicalId testing. It is fairly common to want to test the logicalId of a resource. This can be tricky as:
  - constructs that have an `AppIdentity` will suffix the `id` with the app value
  - not every construct supports static logicalIds and will always be auto-generated

Rather than passing in the `id` with a combination of `migratedFromCloudFormation` and `existingLogicalId` to get a static/deterministic logicalId, we can pass a regex to the new `toHaveResourceOfTypeAndLogicalId` matcher, asserting the stack has a resource of type x with logicalId y.

Yes. Their ASG's will have updated logicalIds (if auto-generated) and an additional tag.

## How to test

See new test.

More deterministic behaviour across the constructs in the library.

n/a

### DIFF
--- a/src/utils/test/jest.ts
+++ b/src/utils/test/jest.ts
@@ -1,0 +1,41 @@
+import { SynthUtils } from "@aws-cdk/assert";
+import type { GuStack } from "../../constructs/core";
+import type { SynthedStack } from "./synthed-stack";
+
+declare global {
+  // eslint-disable-next-line @typescript-eslint/no-namespace -- custom Jest matcher
+  namespace jest {
+    interface Matchers<R> {
+      toHaveResourceOfTypeAndLogicalId(resourceType: string, logicalId: string | RegExp): R;
+    }
+  }
+}
+
+expect.extend({
+  /**
+   * A helper function to assert a stack has a resource of a particular type and logicalId.
+   * Useful for when the logicalId is auto-generated.
+   * @param stack the stack to make an assertion on
+   * @param resourceType the AWS resource type string, for example "AWS::AutoScaling::AutoScalingGroup"
+   * @param logicalId a string or regex pattern to match against the resource's logicalId
+   */
+  toHaveResourceOfTypeAndLogicalId(stack: GuStack, resourceType: string, logicalId: string | RegExp) {
+    const json = SynthUtils.toCloudFormation(stack) as SynthedStack;
+
+    const matchResult = Object.entries(json.Resources).find(([key, { Type }]) => {
+      const logicalIdMatch = logicalId instanceof RegExp ? logicalId.test(key) : key === logicalId;
+      const typeMatch = Type === resourceType;
+      return logicalIdMatch && typeMatch;
+    });
+
+    return matchResult
+      ? {
+          pass: true,
+          message: () => "",
+        }
+      : {
+          pass: false,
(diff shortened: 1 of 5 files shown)


jacobwinch:
> Yes. Their ASG's will have updated logicalIds (if auto-generated) and an additional tag.

I'm not sure if we have any stacks which have an ASG with an auto-generated id (perhaps Typerighter?), but I think this change would require resource replacement? Do you think that's safe?

src/utils/test/jest.ts
```diff
@@ -0,0 +1,41 @@
...
+   * @param logicalId a string or regex pattern to match against the resource's logicalId
+   */
+  toHaveResourceOfTypeAndLogicalId(stack: GuStack, resourceType: string, logicalId: string | RegExp) {
+    const json = SynthUtils.toCloudFormation(stack) as SynthedStack;
+
+    const matchResult = Object.entries(json.Resources).find(([key, { Type }]) => {
+      const logicalIdMatch = logicalId instanceof RegExp ? logicalId.test(key) : key === logicalId;
+      const typeMatch = Type === resourceType;
+      return logicalIdMatch && typeMatch;
+    });
```
akash1810, line 29: I'm not overjoyed at the readability of this - opinions wanted!

jacobwinch:
The code looks good; I think the custom matcher will be very helpful! 

I just have one [question](https://github.com/guardian/cdk/pull/392#issuecomment-815500813) about side-effects of this change that we should probably consider before merging this in.

src/utils/test/jest.ts
jacobwinch: I found this simple enough to understand on first reading (plus it's a test helper function), so I wouldn't worry too much!

akash1810:
> > Yes. Their ASG's will have updated logicalIds (if auto-generated) and an additional tag.
> 
> I'm not sure if we have any stacks which have an ASG with an auto-generated id (perhaps Typerighter?), but I think this change would require resource replacement? Do you think that's safe?

Ah great point! I think we might just about get away with it as the Typerighter stack using CDK has a small audience, so disruption can be co-ordinated.

The [AWS docs on replacement](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-updating-stacks-update-behaviors.html#update-replacement) gives me hope that this event isn't too chaotic:

> AWS CloudFormation usually creates the replacement resource first, changes references from other dependent resources to point to the replacement resource, and then deletes the old resource.

jacobwinch (approved):

akash1810:
`toHaveResourceOfTypeAndLogicalId` makes tests so much more explicit 😍 .

From:

```typescript
expect(Object.keys(json.Resources)).toContain("AutoscalingGroup");
```

To:

```typescript
expect(stack).toHaveResourceOfTypeAndLogicalId("AWS::AutoScaling::AutoScalingGroup", "AutoscalingGroup");
```

PR #396, opened 2021-04-08
fix: ensure all generated logical ids conform to CloudFormation standard

jacobwinch:
## What does this change?
This PR ensures that the logical ids generated by this library always conform to [AWS's standard](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resources-section-structure.html):

> The logical ID must be alphanumeric (A-Za-z0-9).

No.

## How to test
Existing unit tests have been updated due to this change.

Previously, it was impossible to CloudForm the resources created by the `GuInstanceRole` construct if your `AppIdentity` included non-alphanumeric characters (e.g. when using `test-app`). After this change, users will be able to include hyphens (or other non-alphanumeric characters if they really must?!) in their app names.

I can't think of any risks which would result from making this change.

### DIFF
--- a/src/constructs/core/identity.ts
+++ b/src/constructs/core/identity.ts
@@ -15,7 +15,9 @@ export interface Identity extends StackStageIdentity, AppIdentity {}
 export const AppIdentity = {
   suffixText(appIdentity: AppIdentity, text: string): string {
     const titleCaseApp = appIdentity.app.charAt(0).toUpperCase() + appIdentity.app.slice(1);
-    return `${text}${titleCaseApp}`;
+    // CloudFormation Logical Ids must be alphanumeric, so remove any non-alphanumeric characters: https://stackoverflow.com/a/20864946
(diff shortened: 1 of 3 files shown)


src/constructs/core/identity.ts
```diff
@@ -15,7 +15,9 @@ export interface Identity extends StackStageIdentity, AppIdentity {}
 export const AppIdentity = {
   suffixText(appIdentity: AppIdentity, text: string): string {
     const titleCaseApp = appIdentity.app.charAt(0).toUpperCase() + appIdentity.app.slice(1);
-    return `${text}${titleCaseApp}`;
+    // CloudFormation Logical Ids must be alphanumeric, so remove any non-alphanumeric characters: https://stackoverflow.com/a/20864946
+    const alphanumericTitleCaseApp = titleCaseApp.replace(/[\W_]+/g, "");
```
akash1810, line 19: Lovely! This is a lot cleaner than my ascii code array idea!

akash1810 (approved):

PR #397, opened 2021-04-08
refactor: Stop producing migration messages in TEST

akash1810:
## What does this change?

As we migrate to the new logicalId logic introduced in #364, we'll see the migration messages more and more.
As these messages are expected, we can afford to stop producing them in TEST to avoid the build log becoming too noisy.

An example of noisy build log can be seen [here](https://github.com/guardian/cdk/runs/2294779782?check_suite_focus=true#step:4:45).

image

No.

## How to test

See the build log?

A quieter build log will highlight any genuine errors more easily.

n/a


jacobwinch (approved):
Great, thanks for this!

PR #399, opened 2021-04-08
feat: add first draft of EC2 App Pattern

jacobwinch:
## What does this change?
This PR adds a first attempt at our EC2 App pattern. It's still a work in progress, but we have confirmed that this minimal version of the pattern can be deployed successfully so it feels like a good time to get something merged.

No.

## How to test
We've added a basic unit test for this. More unit tests (and proper docs) will be added as the pattern develops.

We've also manually tested deploying a stack with the following code:

```typescript
import type { App } from "@aws-cdk/core";
import type { GuStackProps } from "@guardian/cdk/lib/constructs/core";
import { GuStack } from "@guardian/cdk/lib/constructs/core";
import { GuEc2App } from "@guardian/cdk/lib/patterns/ec2-app";

export class GuCDKEC2Pattern extends GuStack {
    super(scope, id, props);

    const port = 8000;

    new GuEc2App(this, {
      app: "TestApp",
      applicationPort: port,
      publicFacing: false,
      userData: `#!/usr/bin/env bash
        mkdir test
        cd test
        echo OK >> healthcheck
        python -m SimpleHTTPServer ${port}
      `,
    });
  }
}
```

This just allows us to serve a static file via `/healthcheck`, so that we can ensure that requests can be served correctly.

It should be much faster/easier to create a new EC2 app from scratch using this pattern.

Although this pattern is now functional, we don't want 'real' users to start experimenting with it at this stage. To mitigate against this:

1) We've added a [comment](https://github.com/guardian/cdk/pull/399/commits/ab1f0eb408a8653c945ab8ddfd9d20d7044776a7) to warn people that this is still under development
2) The pattern is not mentioned in https://github.com/guardian/cdk/blob/main/src/patterns/index.ts yet.

### DIFF
--- a/src/constants/library-info.ts
+++ b/src/constants/library-info.ts
@@ -10,8 +10,3 @@ export const TrackingTag = {
   Key: "gu:cdk:version",
   Value: LibraryInfo.VERSION,
 };
-
-export const TrackingTagWithPropagate = {
-  ...TrackingTag,
-  PropagateAtLaunch: true,
-};
--- a/src/constants/library-info.ts
+++ b/src/constants/library-info.ts
@@ -10,3 +10,8 @@ export const TrackingTag = {
   Key: "gu:cdk:version",
   Value: LibraryInfo.VERSION,
 };
+
+export const TrackingTagWithPropagate = {
+  ...TrackingTag,
+  PropagateAtLaunch: true,
+};
--- a/src/constructs/ec2/vpc.ts
+++ b/src/constructs/ec2/vpc.ts
@@ -7,7 +7,9 @@ interface VpcFromIdProps extends Omit<VpcAttributes, "availabilityZones"> {
   availabilityZones?: string[];
 }
 
-type VpcFromIdParameterProps = Omit<VpcFromIdProps, "vpcId">;
+interface VpcFromIdParameterProps extends Omit<VpcFromIdProps, "vpcId"> {
+  app?: string;
+}
 
 export enum SubnetType {
   PUBLIC = "Public",
@@ -16,19 +18,26 @@ export enum SubnetType {
 
 export interface GuSubnetProps {
   type?: SubnetType;
+  app?: string;
 }
+
+/*
+ * Some stacks maybe have multiple patterns (such as `GuEc2App`), therefore creating multiple VPC and subnet resources.
+ * This function ensures that, where desired, we can have unique IDs for each VPC and subnet declaration
+ */
+export const maybeApp = (props?: { app?: string }): string => props?.app ?? "";
 
 export class GuVpc {
   static subnets(scope: GuStack, subnets: string[]): ISubnet[] {
     /**
      *  As of ~1.77.0 of the aws/cdk, this line
-     * [ previosuly Subnet.fromSubnetId(scope, `subnet-${subnetId}`, subnetId)) ]
+     * [ previously Subnet.fromSubnetId(scope, `subnet-${subnetId}`, subnetId)) ]
      * resulted in the following error
      * "Found an encoded list token string in a scalar string context. Use 'Fn.select(0, list)' (not 'list[0]') to extract elements from token lists."
      * See https://github.com/aws/aws-cdk/issues/11945 for more information
      *
      * As a hacky workaround, we have moved to using the fromSubnetAttributes method and hardcoded an empty value
-     * for the routeTableId prop. This prevents the error and, when tested on existings stacks, results in no change to the CFN output
+     * for the routeTableId prop. This prevents the error and, when tested on existing stacks, results in no change to the CFN output
      *
      * TODO: Understand VPCs and Subnets better and develop a better solution to this problem
      */
@@ -40,7 +49,7 @@ export class GuVpc {
   static subnetsfromParameter(scope: GuStack, props?: GuSubnetProps): ISubnet[] {
     const type = props?.type ?? SubnetType.PRIVATE;
 
-    const subnets = new GuSubnetListParameter(scope, `${type}Subnets`, {
+    const subnets = new GuSubnetListParameter(scope, `${maybeApp(props)}${type}Subnets`, {
       description: `A list of ${type.toLowerCase()} subnets`,
       default: `/account/vpc/primary/subnets/${type.toLowerCase()}`,
       fromSSM: true,
@@ -57,7 +66,7 @@ export class GuVpc {
   }
 
   static fromIdParameter(scope: GuStack, id: string, props?: VpcFromIdParameterProps): IVpc {
-    const vpc = new GuVpcParameter(scope, "VpcId", {
+    const vpc = new GuVpcParameter(scope, `${maybeApp(props)}VpcId`, {
       description: "Virtual Private Cloud to run EC2 instances within",
       default: "/account/vpc/primary/id",
       fromSSM: true,
--- a/src/constructs/ec2/vpc.ts
+++ b/src/constructs/ec2/vpc.ts
@@ -7,6 +7,7 @@ interface VpcFromIdProps extends Omit<VpcAttributes, "availabilityZones"> {
   availabilityZones?: string[];
 }
 
+// TODO: Migrate this to use `AppIdentity`
 interface VpcFromIdParameterProps extends Omit<VpcFromIdProps, "vpcId"> {
   app?: string;
 }
@@ -16,6 +17,7 @@ export enum SubnetType {
   PRIVATE = "Private",
 }
 
+// TODO: Migrate this to use `AppIdentity`
 export interface GuSubnetProps {
   type?: SubnetType;
   app?: string;
--- a/src/patterns/ec2-app.test.ts
+++ b/src/patterns/ec2-app.test.ts
@@ -1,6 +1,6 @@
 import "@aws-cdk/assert/jest";
 import { SynthUtils } from "@aws-cdk/assert";
-import { TrackingTagWithPropagate } from "../constants/library-info";
+import { TrackingTag } from "../constants/library-info";
 import { alphabeticalTags, simpleGuStackForTesting } from "../utils/test";
 import { GuApplicationPorts, GuEc2App, GuNodeApp, GuPlayApp } from "./ec2-app";
 
@@ -38,7 +38,7 @@ describe("the GuEC2App pattern", function () {
         { Key: "Name", PropagateAtLaunch: true, Value: "Test/AutoScalingGroupPlayApp" },
         { Key: "Stack", PropagateAtLaunch: true, Value: "test-stack" },
         { Key: "Stage", PropagateAtLaunch: true, Value: { Ref: "Stage" } },
-        TrackingTagWithPropagate,
+        { ...TrackingTag, PropagateAtLaunch: true },
       ]),
     });
 
@@ -48,7 +48,7 @@ describe("the GuEC2App pattern", function () {
         { Key: "Name", PropagateAtLaunch: true, Value: "Test/AutoScalingGroupNodeApp" },
         { Key: "Stack", PropagateAtLaunch: true, Value: "test-stack" },
         { Key: "Stage", PropagateAtLaunch: true, Value: { Ref: "Stage" } },
-        TrackingTagWithPropagate,
+        { ...TrackingTag, PropagateAtLaunch: true },
       ]),
     });
   });
--- a/src/patterns/ec2-app.test.ts
+++ b/src/patterns/ec2-app.test.ts
@@ -1,0 +1,98 @@
+import "@aws-cdk/assert/jest";
+import { SynthUtils } from "@aws-cdk/assert";
+import { alphabeticalTags, simpleGuStackForTesting } from "../utils/test";
+import { GuApplicationPorts, GuEc2App } from "./ec2-app";
+
+describe("the GuEC2App pattern", function () {
+  it("should produce a functional EC2 app with minimal arguments", function () {
+    const stack = simpleGuStackForTesting();
+    new GuEc2App(stack, {
+      applicationPort: GuApplicationPorts.Node,
+      app: "test-gu-ec2-app",
+      publicFacing: false,
+      userData: "#!/bin/dev foobarbaz",
+    });
+    expect(SynthUtils.toCloudFormation(stack)).toMatchSnapshot();
+  });
+
+  it("can handle multiple EC2 apps in a single stack", function () {
+    const stack = simpleGuStackForTesting();
+    new GuEc2App(stack, {
+      applicationPort: GuApplicationPorts.Node,
+      app: "NodeApp",
+      publicFacing: false,
+      userData: "#!/bin/dev foobarbaz",
+    });
+
+    new GuEc2App(stack, {
+      applicationPort: GuApplicationPorts.Play,
+      app: "PlayApp",
+      publicFacing: false,
+      userData: "#!/bin/dev foobarbaz",
+    });
+
+    expect(stack).toHaveResource("AWS::AutoScaling::AutoScalingGroup", {
+      Tags: alphabeticalTags([
+        {
+          Key: "App",
+          PropagateAtLaunch: true,
+          Value: "PlayApp",
+        },
+        {
+          Key: "gu:cdk:version",
+          PropagateAtLaunch: true,
+          Value: "TEST",
+        },
+        {
+          Key: "Name",
+          PropagateAtLaunch: true,
+          Value: "Test/PlayAppAutoScalingGroupPlayApp",
+        },
+        {
+          Key: "Stack",
+          PropagateAtLaunch: true,
+          Value: "test-stack",
+        },
+        {
+          Key: "Stage",
+          PropagateAtLaunch: true,
+          Value: {
+            Ref: "Stage",
+          },
+        },
+      ]),
+    });
+
+    expect(stack).toHaveResource("AWS::AutoScaling::AutoScalingGroup", {
+      Tags: alphabeticalTags([
+        {
+          Key: "App",
+          PropagateAtLaunch: true,
+          Value: "NodeApp",
+        },
+        {
+          Key: "gu:cdk:version",
+          PropagateAtLaunch: true,
+          Value: "TEST",
+        },
+        {
+          Key: "Name",
+          PropagateAtLaunch: true,
+          Value: "Test/NodeAppAutoScalingGroupNodeApp",
+        },
+        {
+          Key: "Stack",
+          PropagateAtLaunch: true,
+          Value: "test-stack",
+        },
+        {
+          Key: "Stage",
+          PropagateAtLaunch: true,
+          Value: {
+            Ref: "Stage",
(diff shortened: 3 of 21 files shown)


src/patterns/ec2-app.ts
```diff
@@ -0,0 +1,77 @@
...
+    const certificateArn = new GuArnParameter(scope, "CertArn", {
+      description: "ARN of a TLS certificate to install on the load balancer",
+    });
+
+    const certificate = Certificate.fromCertificateArn(scope, "Certificate", certificateArn.valueAsString);
+
+    const asg = new GuAutoScalingGroup(scope, "AutoScalingGroup", {
+      vpc,
+      app: props.app,
+      stageDependentProps: {
```
jacobwinch, line 47: Perhaps this default should come from the `GuAutoScalingGroup` construct too?

src/patterns/ec2-app.ts
```diff
@@ -0,0 +1,77 @@
...
+import { GuVpc, SubnetType } from "../constructs/ec2";
+import { GuInstanceRole } from "../constructs/iam";
+import {
+  GuApplicationListener,
+  GuApplicationLoadBalancer,
+  GuApplicationTargetGroup,
+} from "../constructs/loadbalancing";
+
+interface GuEc2AppProps extends AppIdentity {
+  userData: GuUserData | string;
```
jacobwinch, line 18: If the user needs to download private config then we still need to wire up these permissions (for more details on that, see: https://github.com/guardian/cdk/pull/312)

src/patterns/ec2-app.ts
```diff
@@ -0,0 +1,77 @@
...
+    const asg = new GuAutoScalingGroup(scope, "AutoScalingGroup", {
+      vpc,
+      app: props.app,
+      stageDependentProps: {
+        CODE: { minimumInstances: 1 },
+        PROD: { minimumInstances: 3 },
+      },
+      role: new GuInstanceRole(scope, { app: props.app }),
+      healthCheck: HealthCheck.elb({ grace: Duration.minutes(2) }), // should this be defaulted at pattern or construct level?
+      userData: props.userData instanceof GuUserData ? props.userData.userData : props.userData,
```
akash1810, line 53: Might be nicer for `userData` in `GuAutoScalingGroup`'s props to be of type `GuUserData | string`.

src/patterns/ec2-app.ts
```diff
@@ -0,0 +1,77 @@
...
+    const certificate = Certificate.fromCertificateArn(scope, "Certificate", certificateArn.valueAsString);
+
+    const asg = new GuAutoScalingGroup(scope, "AutoScalingGroup", {
+      vpc,
+      app: props.app,
+      stageDependentProps: {
+        CODE: { minimumInstances: 1 },
+        PROD: { minimumInstances: 3 },
+      },
+      role: new GuInstanceRole(scope, { app: props.app }),
```
akash1810, line 51: Might be nicer for `GuAutoScalingGroup` to provide this role as default within the construct.

src/patterns/ec2-app.ts
```diff
@@ -0,0 +1,77 @@
...
+    scope.tags.setTag("App", props.app);
+
+    const vpc = GuVpc.fromIdParameter(scope, "VPC");
+    const privateSubnets = GuVpc.subnetsfromParameter(scope, { type: SubnetType.PRIVATE });
+
+    const certificateArn = new GuArnParameter(scope, "CertArn", {
+      description: "ARN of a TLS certificate to install on the load balancer",
+    });
+
+    const certificate = Certificate.fromCertificateArn(scope, "Certificate", certificateArn.valueAsString);
```
akash1810, line 42: Any reason to not create the certificate resource rather than adopting a resource?

src/patterns/ec2-app.ts
```diff
@@ -0,0 +1,77 @@
...
+  GuApplicationTargetGroup,
+} from "../constructs/loadbalancing";
+
+interface GuEc2AppProps extends AppIdentity {
+  userData: GuUserData | string;
+  publicFacing: boolean; // could also name it `internetFacing` to match GuApplicationLoadBalancer
+  applicationPort: GuApplicationPorts | number;
+}
+
+export enum GuApplicationPorts {
```
akash1810, line 23: minor: I think this can be simplified a little into:
- `GuEc2App` which takes in a `port`
- `GuPlayEc2App` which extends `GuEc2App` and sets `port` to 9000
- `GuNodeEc2App` which extends `GuEc2App` and sets `port` to 3000

This would mean the `applicationPort` in `GuEc2AppProps` becomes a `number` type, which is a little simpler.

src/patterns/ec2-app.ts
```diff
@@ -0,0 +1,77 @@
...
+    const loadBalancer = new GuApplicationLoadBalancer(scope, "LoadBalancer", {
+      vpc: vpc,
+      vpcSubnets: {
+        subnets: props.publicFacing ? GuVpc.subnetsfromParameter(scope, { type: SubnetType.PUBLIC }) : privateSubnets,
+      },
+    });
+
+    const targetGroup = new GuApplicationTargetGroup(scope, "TargetGroup", {
+      vpc,
+      protocol: ApplicationProtocol.HTTP,
```
akash1810, line 66: Why not HTTPS?

src/patterns/ec2-app.ts
jacobwinch: Yep, we started work on this before our certificate construct was ready! I think we should create the cert as part of this pattern but would suggest that we tackle that in a follow-up PR to avoid getting stuck on this feature branch for too long.

src/patterns/ec2-app.ts
jacobwinch: This communication is between the LB and the instances, so I think we just need HTTP here?

akash1810: Ah! Shall we rename `protocol` to make this more obvious? In a different PR.

akash1810 (commented):
This looks fantastic, great work!

Should we add a test for a single `GuStack` with multiple `GuEc2App`s too?

src/patterns/ec2-app.ts
jacobwinch: Yep, to be honest I wonder if it's another thing that we could just default at the construct level?

src/patterns/ec2-app.ts
stephengeller: #406 

PR out for this change.

src/patterns/ec2-app.ts
stephengeller: Also added to PR #406

src/patterns/ec2-app.ts
stephengeller: [I think it is already](https://github.com/guardian/cdk/blob/d48129ec99cbafd769eef2911a456a78c9ade53e/src/constructs/autoscaling/asg.ts#L33), what would look better here do you think?

src/patterns/ec2-app.ts
stephengeller: This has also been added to #406

src/patterns/ec2-app.ts
stephengeller: This has been implemented in a new commit, good suggestion!

stephengeller:
I made a change to how the IDs of VPC constructs are declared, as we ran into issues with declaring multiple `GuEC2App`s in the same stack, mainly due to the [IDs of VPCs and subnets constructs being hardcoed](https://github.com/guardian/cdk/blob/91ddc6dd052a6c8522f2c0546695b7dbbb6028ce/src/constructs/ec2/vpc.ts#L60). Would be keen to get opinions on the workaround implemented here.

src/constructs/ec2/vpc.ts
```diff
@@ -16,19 +18,26 @@ export enum SubnetType {
 
 export interface GuSubnetProps {
   type?: SubnetType;
+  app?: string;
```
akash1810, line 21: Could `GuSubnetProps` extend [`AppIdentity`](https://github.com/guardian/cdk/blob/d69bdbe181d90f0f2c502dee9b05095f1402301a/src/constructs/core/identity.ts#L9-L11) instead?

src/constructs/ec2/vpc.ts
```diff
@@ -40,7 +49,7 @@ export class GuVpc {
   static subnetsfromParameter(scope: GuStack, props?: GuSubnetProps): ISubnet[] {
     const type = props?.type ?? SubnetType.PRIVATE;
 
-    const subnets = new GuSubnetListParameter(scope, `${type}Subnets`, {
+    const subnets = new GuSubnetListParameter(scope, `${maybeApp(props)}${type}Subnets`, {
```
akash1810, line 52: [`AppIdentity.suffixText`](https://github.com/guardian/cdk/blob/d69bdbe181d90f0f2c502dee9b05095f1402301a/src/constructs/core/identity.ts#L16) in other places that are app aware, can this be reused here for consistency?

src/patterns/ec2-app.test.ts
```diff
@@ -0,0 +1,156 @@
...
+        {
+          Key: "App",
+          PropagateAtLaunch: true,
+          Value: "PlayApp",
+        },
+        {
+          Key: "gu:cdk:version",
+          PropagateAtLaunch: true,
+          Value: "TEST",
+        },
```
akash1810, line 45: Better to use the [mock](https://github.com/guardian/cdk/blob/d69bdbe181d90f0f2c502dee9b05095f1402301a/src/constants/__mocks__/library-info.ts) here.

This would become:

```typescript
import { TrackingTag } from "../../constants/library-info";

...

Tags: alphabeticalTags([
  TrackingTag,
  { Key: "App", ... },
  ...
]);
```

That is, this can be replaced:

```typescript
{
  Key: "gu:cdk:version",
  PropagateAtLaunch: true,
  Value: "TEST",
},
```

With:

```typescirpt
TrackingTag
```

Using the constant means we can update the naming of the tag in one place (should we want to).

src/patterns/ec2-app.test.ts
```diff
@@ -0,0 +1,156 @@
...
+        {
+          Key: "App",
+          PropagateAtLaunch: true,
+          Value: "NodeApp",
+        },
+        {
+          Key: "gu:cdk:version",
+          PropagateAtLaunch: true,
+          Value: "TEST",
+        },
```
akash1810, line 77: Same as above.

src/patterns/ec2-app.ts
```diff
@@ -0,0 +1,105 @@
...
+}
+
+/*
+This pattern is under development. Please don't attempt to use it yet.
+ */
+export class GuEc2App {
+  constructor(scope: GuStack, props: GuEc2AppProps) {
+    const id = (id: string) => `${props.app}${id}`;
+
+    scope.tags.setTag(id("App"), props.app);
```
akash1810, line 39: The [`AppIdentity.taggedConstruct`](https://github.com/guardian/cdk/blob/d69bdbe181d90f0f2c502dee9b05095f1402301a/src/constructs/core/identity.ts#L22) helper should be used here for consistency.

src/patterns/ec2-app.ts
```diff
@@ -0,0 +1,105 @@
...
+  Node = 3000,
+  Play = 9000,
+}
+
+/*
+This pattern is under development. Please don't attempt to use it yet.
+ */
+export class GuEc2App {
+  constructor(scope: GuStack, props: GuEc2AppProps) {
+    const id = (id: string) => `${props.app}${id}`;
```
akash1810, line 37: [`AppIdentity.suffixText`](https://github.com/guardian/cdk/blob/d69bdbe181d90f0f2c502dee9b05095f1402301a/src/constructs/core/identity.ts#L16) in other places that are app aware, can this be reused here for consistency?

src/patterns/ec2-app.ts
```diff
@@ -0,0 +1,105 @@
...
+
+    const certificateArn = new GuArnParameter(scope, id("CertArn"), {
+      description: "ARN of a TLS certificate to install on the load balancer",
+    });
+
+    const certificate = Certificate.fromCertificateArn(scope, id("Certificate"), certificateArn.valueAsString);
+
+    const asg = new GuAutoScalingGroup(scope, id("AutoScalingGroup"), {
+      vpc,
+      app: props.app,
```
akash1810, line 53: minor: we've destructured `app` above, so this can become:

```suggestion
      app,
```

src/patterns/ec2-app.ts
```diff
@@ -0,0 +1,105 @@
...
+    });
+  }
+}
+
+/*
+This pattern is under development. Please don't attempt to use it yet.
+ */
+export class GuPlayApp extends GuEc2App {
+  constructor(scope: GuStack, props: GuMaybePortProps) {
+    super(scope, { ...props, applicationPort: props.applicationPort ?? GuApplicationPorts.Play });
```
akash1810, line 94: I'd favour being strongly opinionated here and say Play always runs on `9000`. That is, don't accept `applicationPort` in the props.

src/patterns/ec2-app.ts
```diff
@@ -0,0 +1,105 @@
...
+  }
+}
+
+/*
+This pattern is under development. Please don't attempt to use it yet.
+ */
+export class GuNodeApp extends GuEc2App {
+  constructor(scope: GuStack, props: GuMaybePortProps) {
+    super(scope, { ...props, applicationPort: props.applicationPort ?? GuApplicationPorts.Node });
```
akash1810, line 103: Similar to above, I'd favour being strongly opinionated here and say Node always runs on 3000. That is, don't accept applicationPort in the props.

akash1810 (commented):
Added a couple of comments, mainly on code reuse.

src/constructs/ec2/vpc.ts
```diff
@@ -7,7 +7,9 @@ interface VpcFromIdProps extends Omit<VpcAttributes, "availabilityZones"> {
   availabilityZones?: string[];
 }
 
-type VpcFromIdParameterProps = Omit<VpcFromIdProps, "vpcId">;
+interface VpcFromIdParameterProps extends Omit<VpcFromIdProps, "vpcId"> {
```
akash1810, line 10: Could `VpcFromIdParameterProps` extend [`AppIdentity`](https://github.com/guardian/cdk/blob/d69bdbe181d90f0f2c502dee9b05095f1402301a/src/constructs/core/identity.ts#L9-L11) instead?

src/patterns/ec2-app.test.ts
```diff
@@ -0,0 +1,156 @@
...
+    new GuEc2App(stack, {
+      applicationPort: GuApplicationPorts.Node,
+      app: "test-gu-ec2-app",
+      publicFacing: false,
+      userData: "#!/bin/dev foobarbaz",
+    });
+    expect(SynthUtils.toCloudFormation(stack)).toMatchSnapshot();
+  });
+
+  it("can handle multiple EC2 apps in a single stack", function () {
```
akash1810, line 18: 🎉  lovely!

akash1810:
The commit history on this PR is 👌🏽 .

src/constructs/ec2/vpc.ts
stephengeller: As discussed IRL, this is not a trivial change and so we've elected to leave this for a future PR.

src/constructs/ec2/vpc.ts
stephengeller: As discussed IRL, this is not a trivial change and so we've elected to leave this for a future PR.

src/patterns/ec2-app.test.ts
stephengeller: Good shout. This doesn't currently work due to the missing `PropagateAtLaunch` keypair, but I've created a new constant that handles this called `TrackingTagWithPropagate`.

src/patterns/ec2-app.ts
stephengeller: Good shout, done.

src/constructs/ec2/vpc.ts
stephengeller: As discussed IRL, this is not a trivial change and so we've elected to leave this for a future PR.

src/patterns/ec2-app.ts
stephengeller: Yep, done!

src/patterns/ec2-app.ts
stephengeller: Good catch, done.

src/patterns/ec2-app.ts
stephengeller: Good shout, done.

src/patterns/ec2-app.ts
stephengeller: Good shout, done.

stephengeller:
This should now be ready for review again @akash1810 @jacobwinch !

src/constants/library-info.ts
```diff
@@ -10,3 +10,8 @@ export const TrackingTag = {
   Key: "gu:cdk:version",
   Value: LibraryInfo.VERSION,
 };
+
+export const TrackingTagWithPropagate = {
```
akash1810, line 14: I don't think this is correct, as I'm not sure we want to publicly export this variant as it's specifically for the tests and not for "daily use"? In the past we've done something like [this](https://github.com/guardian/cdk/blob/6159612ce108a7e8f4bc9d907896db9ffade42b3/src/constructs/autoscaling/asg.test.ts#L77).

I think there's an opportunity to add a [custom matcher](https://jestjs.io/docs/expect#custom-matchers-api) here (similar to [toHaveResourceOfTypeAndLogicalId](https://github.com/guardian/cdk/blob/73ef8ec96773c53a924e6795286a54565e9ad3ef/src/utils/test/jest.ts#L21)) as it looks like we want to assert a construct has the identity tags. That is, a "toHaveAnIdentityTaggedResource" matcher that'll take a stack, a resource type string and assert that it's tags contain all the [identity keys](https://github.com/guardian/cdk/blob/73ef8ec96773c53a924e6795286a54565e9ad3ef/src/constructs/core/identity.ts#L4-L11)?

jacobwinch: After discussing this via Chat, we decided to go with the [simpler option](https://github.com/guardian/cdk/pull/399/commits/1e9f5f0831023121684ed17d186e4ae979c9e8a6) for now, in order to get this merged. We'll look at implementing the custom matcher when everyone is around to work on it together.

akash1810 (approved):
Looks great!

👍🏽  to [this](https://github.com/guardian/cdk/pull/399/files#r612466018) too.

src/patterns/ec2-app.ts
jacobwinch: https://github.com/guardian/cdk/pull/445

src/patterns/ec2-app.ts
jacobwinch: https://github.com/guardian/cdk/pull/446

PR #400, opened 2021-04-08
fix: Adopt new logicalId override logic across the constructs

akash1810:
## What does this change?

In #364 we simplified and encapsulated the logic to override a construct's logicalId.

This change adopts the new logic into the constructs.

As a consequence to this a number of tests have been moved to using the auto-generated logicalId. This reduces the variables in play in our tests and therefore simplifies them.

Yes - the `overrideId` prop is being removed.

## How to test

See updated tests.

DRYer code and greater consistency in behaviour across constructs.

n/a

### DIFF
--- a/src/constructs/autoscaling/asg.test.ts
+++ b/src/constructs/autoscaling/asg.test.ts
@@ -6,8 +6,8 @@ import { ApplicationProtocol } from "@aws-cdk/aws-elasticloadbalancingv2";
 import { Stack } from "@aws-cdk/core";
 import { Stage } from "../../constants";
 import { TrackingTag } from "../../constants/library-info";
-import type { SynthedStack } from "../../utils/test";
-import { alphabeticalTags, simpleGuStackForTesting } from "../../utils/test";
+import type { Resource, SynthedStack } from "../../utils/test";
+import { alphabeticalTags, findResourceByTypeAndLogicalId, simpleGuStackForTesting } from "../../utils/test";
 import { GuSecurityGroup } from "../ec2";
 import { GuApplicationTargetGroup } from "../loadbalancing";
 import type { GuAutoScalingGroupProps } from "./asg";
@@ -194,10 +194,20 @@ describe("The GuAutoScalingGroup", () => {
 
   test("does not include the UpdatePolicy property", () => {
     const stack = simpleGuStackForTesting();
-    new GuAutoScalingGroup(stack, "AutoscalingGroup", { ...defaultProps, overrideId: true });
+    new GuAutoScalingGroup(stack, "AutoscalingGroup", { ...defaultProps });
 
-    const json = SynthUtils.toCloudFormation(stack) as SynthedStack;
-    expect(Object.keys(json.Resources.AutoscalingGroup)).not.toContain("UpdatePolicy");
+    expect(stack).toHaveResourceOfTypeAndLogicalId("AWS::AutoScaling::AutoScalingGroup", /AutoscalingGroup.+/);
+
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- the `toHaveResourceOfTypeAndLogicalId` line above confirms `asgResource` will not be `undefined`
+    const asgResource: Resource = findResourceByTypeAndLogicalId(
+      stack,
+      "AWS::AutoScaling::AutoScalingGroup",
+      /AutoscalingGroup.+/
+    )!;
+
+    // This is checking the properties of the ASG resource
+    // TODO improve the syntax
+    expect(Object.keys(Object.values(asgResource)[0])).not.toContain("UpdatePolicy");
   });
 
   test("overrides the id with the overrideId prop set to true", () => {
--- a/src/constructs/autoscaling/asg.test.ts
+++ b/src/constructs/autoscaling/asg.test.ts
@@ -139,12 +139,12 @@ describe("The GuAutoScalingGroup", () => {
   });
 
   test("adds any target groups passed through props", () => {
-    const stack = simpleGuStackForTesting();
+    const stack = simpleGuStackForTesting({ migratedFromCloudFormation: true });
 
     const targetGroup = new GuApplicationTargetGroup(stack, "TargetGroup", {
       vpc: vpc,
       protocol: ApplicationProtocol.HTTP,
-      overrideId: true,
+      existingLogicalId: "MyTargetGroup",
     });
 
     new GuAutoScalingGroup(stack, "AutoscalingGroup", {
@@ -155,7 +155,7 @@ describe("The GuAutoScalingGroup", () => {
     expect(stack).toHaveResource("AWS::AutoScaling::AutoScalingGroup", {
       TargetGroupARNs: [
         {
-          Ref: "TargetGroup",
+          Ref: "MyTargetGroup",
         },
       ],
     });
@@ -165,9 +165,9 @@ describe("The GuAutoScalingGroup", () => {
     const app = "Testing";
     const stack = simpleGuStackForTesting();
 
-    const securityGroup = new GuSecurityGroup(stack, "SecurityGroup", { vpc, overrideId: true, app });
-    const securityGroup1 = new GuSecurityGroup(stack, "SecurityGroup1", { vpc, overrideId: true, app });
-    const securityGroup2 = new GuSecurityGroup(stack, "SecurityGroup2", { vpc, overrideId: true, app });
+    const securityGroup = new GuSecurityGroup(stack, "SecurityGroup", { vpc, app });
+    const securityGroup1 = new GuSecurityGroup(stack, "SecurityGroup1", { vpc, app });
+    const securityGroup2 = new GuSecurityGroup(stack, "SecurityGroup2", { vpc, app });
 
     new GuAutoScalingGroup(stack, "AutoscalingGroup", {
       ...defaultProps,
@@ -180,13 +180,13 @@ describe("The GuAutoScalingGroup", () => {
           "Fn::GetAtt": [`GuHttpsEgressSecurityGroup${app}89CDDA4B`, "GroupId"],
         },
         {
-          "Fn::GetAtt": [`SecurityGroup${app}`, "GroupId"],
+          "Fn::GetAtt": [`SecurityGroup${app}A32D34F9`, "GroupId"],
         },
         {
-          "Fn::GetAtt": [`SecurityGroup1${app}`, "GroupId"],
+          "Fn::GetAtt": [`SecurityGroup1${app}CA3A17A4`, "GroupId"],
         },
         {
-          "Fn::GetAtt": [`SecurityGroup2${app}`, "GroupId"],
+          "Fn::GetAtt": [`SecurityGroup2${app}6436C75B`, "GroupId"],
         },
       ],
     });
@@ -210,13 +210,11 @@ describe("The GuAutoScalingGroup", () => {
     expect(Object.keys(Object.values(asgResource)[0])).not.toContain("UpdatePolicy");
   });
 
-  test("overrides the id with the overrideId prop set to true", () => {
-    const stack = simpleGuStackForTesting();
-    new GuAutoScalingGroup(stack, "AutoscalingGroup", { ...defaultProps, overrideId: true });
+  test("overrides the logicalId when existingLogicalId is set in a migrating stack", () => {
+    const stack = simpleGuStackForTesting({ migratedFromCloudFormation: true });
+    new GuAutoScalingGroup(stack, "AutoscalingGroup", { ...defaultProps, existingLogicalId: "MyASG" });
 
-    const json = SynthUtils.toCloudFormation(stack) as SynthedStack;
-
-    expect(Object.keys(json.Resources)).toContain("AutoscalingGroup");
+    expect(stack).toHaveResourceOfTypeAndLogicalId("AWS::AutoScaling::AutoScalingGroup", "MyASG");
   });
 
   test("does not override the id by default", () => {
--- a/src/constructs/autoscaling/asg.ts
+++ b/src/constructs/autoscaling/asg.ts
@@ -7,6 +7,8 @@ import { Stage } from "../../constants";
 import type { GuStack } from "../core";
 import { GuAmiParameter, GuInstanceTypeParameter } from "../core";
 import { AppIdentity } from "../core/identity";
+import { GuMigratingResource } from "../core/migrating";
+import type { GuStatefulConstruct } from "../core/migrating";
 import { GuHttpsEgressSecurityGroup } from "../ec2";
 
 // Since we want to override the types of what gets passed in for the below props,
@@ -25,7 +27,8 @@ export interface GuAutoScalingGroupProps
       | "desiredCapacity"
       | "securityGroup"
     >,
-    AppIdentity {
+    AppIdentity,
+    GuMigratingResource {
   stageDependentProps: GuStageDependentAsgProps;
   instanceType?: InstanceType;
   imageId?: GuAmiParameter;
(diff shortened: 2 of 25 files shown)


src/constructs/autoscaling/asg.test.ts
```diff
@@ -180,33 +180,41 @@ describe("The GuAutoScalingGroup", () => {
...
+    const asgResource: Resource = findResourceByTypeAndLogicalId(
+      stack,
+      "AWS::AutoScaling::AutoScalingGroup",
+      /AutoscalingGroup.+/
+    )!;
 
-    const json = SynthUtils.toCloudFormation(stack) as SynthedStack;
+    // This is checking the properties of the ASG resource
+    // TODO improve the syntax
+    expect(Object.keys(Object.values(asgResource)[0])).not.toContain("UpdatePolicy");
```
akash1810, line 210: 😷  sorry!

src/constructs/loadbalancing/alb.test.ts
```diff
@@ -1,4 +1,5 @@
 import "@aws-cdk/assert/jest";
```
akash1810, line 1: A lot of the tests in this file do the same thing, so have deleted the duplicates.

src/constructs/rds/instance.test.ts
```diff
@@ -1,4 +1,5 @@
 import "@aws-cdk/assert/jest";
```
akash1810, line 1: A lot of the tests in this file do the same thing, so have deleted the duplicates.

src/constructs/autoscaling/asg.ts
```diff
@@ -101,10 +105,11 @@ export class GuAutoScalingGroup extends AutoScalingGroup {
 
       // Do not use the default AWS security group which allows egress on any port.
       // Favour HTTPS only egress rules by default.
-      securityGroup: GuHttpsEgressSecurityGroup.forVpc(scope, props),
+      securityGroup: GuHttpsEgressSecurityGroup.forVpc(scope, { app: props.app, vpc: props.vpc }),
```
akash1810, line 108: Passing in `props` meant `existingLogicalId` was copied to the `GuHttpsEgressSecurityGroup` too. When it came time to synth the stack, it failed because two resources were being created with the same logicalId (the `GuHttpsEgressSecurityGroup` and the `GuAutoScalingGroup`).

This took the best part of two days to diagnose 😨 !

jacobwinch: 😱

jacobwinch: https://github.com/microsoft/TypeScript/issues/12936

src/constructs/ec2/security-groups/wazuh.ts
```diff
@@ -10,7 +10,6 @@ export class GuWazuhAccess extends GuBaseSecurityGroup {
     super(scope, "WazuhSecurityGroup", {
       vpc,
       description: "Wazuh agent registration and event logging",
-      overrideId: true,
```
jacobwinch, line 13: This seems like an odd default, given that the logical id was hardcoded... However, I guess anyone who had CloudFormed a resource called `WazuhSecurityGroup` would have inherited it by default before. Given how fiddly it is to replace security groups, do we need to provide a way to retain your existing Wazuh security group via the constructor?

src/constructs/iam/roles/instance-role.ts
```diff
@@ -19,7 +19,6 @@ interface GuInstanceRoleProps extends AppIdentity {
 export class GuInstanceRole extends GuRole {
   constructor(scope: GuStack, props: GuInstanceRoleProps) {
     super(scope, AppIdentity.suffixText(props, "InstanceRole"), {
-      overrideId: true,
```
jacobwinch, line 22: 👍  it should always be safe to replace this, right?

src/constructs/loadbalancing/alb.test.ts
```diff
@@ -171,13 +145,12 @@ describe("The GuApplicationListener class", () => {
...
       defaultAction: ListenerAction.forward([targetGroup]),
       certificates: [{ certificateArn: "" }],
     });
 
-    const json = SynthUtils.toCloudFormation(stack) as SynthedStack;
-    expect(Object.keys(json.Resources)).toContain("ApplicationListener");
+    expect(stack).toHaveResourceOfTypeAndLogicalId("AWS::ElasticLoadBalancingV2::Listener", "AppListener");
   });
 
   test("does not override the id if the prop is false", () => {
```
jacobwinch, line 156: Minor, should this now be something like:

`does not override the id if the existingLogicalId prop is not defined`

src/constructs/loadbalancing/alb.test.ts
```diff
@@ -83,36 +72,21 @@ describe("The GuApplicationTargetGroup class", () => {
...
-    new GuApplicationTargetGroup(stack, "ApplicationTargetGroup", { vpc });
+    new GuApplicationTargetGroup(stack, "ApplicationTargetGroup", { vpc, existingLogicalId: "ApplicationTargetGrp" });
 
-    const json = SynthUtils.toCloudFormation(stack) as SynthedStack;
-    expect(Object.keys(json.Resources)).toContain("ApplicationTargetGroup");
+    expect(stack).toHaveResourceOfTypeAndLogicalId("AWS::ElasticLoadBalancingV2::TargetGroup", "ApplicationTargetGrp");
   });
 
-  test("does not override the id if the stack migrated value is true but the override id value is false", () => {
+  test("does not override the id if the stack migrated value is true but existingLogicalId is not set", () => {
```
jacobwinch, line 82: I think that the behaviour under test is now the responsibility of [this code](https://github.com/guardian/cdk/blob/49e0cb06c9ee1f1ae86a0f50522e0ce7db757387/src/constructs/core/migrating.ts#L30-L68). I think it is covered sufficiently by [this test](https://github.com/guardian/cdk/blob/49e0cb06c9ee1f1ae86a0f50522e0ce7db757387/src/constructs/core/migrating.test.ts#L64-L78), so I would suggest removing tests like this from other constructs.

src/constructs/loadbalancing/alb.test.ts
jacobwinch: Actually, I wonder again if this test is really adding any value given that this is now controlled outside of this construct?

src/constructs/loadbalancing/alb.test.ts
```diff
@@ -195,27 +168,8 @@ describe("The GuApplicationListener class", () => {
...
-      certificates: [{ certificateArn: "" }],
-    });
-
-    const json = SynthUtils.toCloudFormation(stack) as SynthedStack;
-    expect(Object.keys(json.Resources)).toContain("ApplicationListener");
+    expect(stack).not.toHaveResourceOfTypeAndLogicalId("AWS::ElasticLoadBalancingV2::Listener", "AppListener");
+    expect(stack).toHaveResourceOfTypeAndLogicalId("AWS::ElasticLoadBalancingV2::Listener", /ApplicationListener.+/);
   });
 
   test("does not override the id if the stack migrated value is true but the override id value is false", () => {
```
jacobwinch, line 175: I think this is similar to the test(s) above: https://github.com/guardian/cdk/pull/400/files#r609853771

src/constructs/ec2/security-groups/wazuh.ts
akash1810: Maybe the Wazuh construct should hard-code the logicalId? It's a singleton, so it shouldn't matter too much.

src/constructs/loadbalancing/alb.test.ts
akash1810: I thought about that too 😄 . I opted to keep the tests here as a way of ensuring the magical `GuMigratingResource.setLogicalId(this, scope, props);` line is called in a construct's constructor as I forgot a few of these 😅 .

Maybe the test should be that `GuMigratingResource.setLogicalId` is called instead?

akash1810: Also just realised `GuMigratingResource.setLogicalId(this, scope, props);` doesn't take the lessons from my previous mistake onboard! Will adjust that.

src/constructs/loadbalancing/alb.test.ts
```diff
@@ -229,13 +183,12 @@ describe("The GuApplicationListener class", () => {
...
     new GuApplicationListener(stack, "ApplicationListener", {
       loadBalancer,
-      overrideId: false,
       defaultAction: ListenerAction.forward([targetGroup]),
       certificates: [{ certificateArn: "" }],
     });
 
-    const json = SynthUtils.toCloudFormation(stack) as SynthedStack;
-    expect(Object.keys(json.Resources)).not.toContain("ApplicationListener");
+    expect(stack).not.toHaveResourceOfTypeAndLogicalId("AWS::ElasticLoadBalancingV2::Listener", "AppListener");
```
jacobwinch, line 190: Should this be `ApplicationListener`?

jacobwinch: Ah sorry, I see now that we are passing `AppListener` in for other tests, but not here. I think we could remove this assertion and just have the regex check below?

src/constructs/loadbalancing/elb.test.ts
```diff
@@ -14,43 +15,31 @@ describe("The GuClassicLoadBalancer class", () => {
...
+    new GuClassicLoadBalancer(stack, "ClassicLoadBalancer", { vpc, existingLogicalId: "ClassicLB" });
 
-    const json = SynthUtils.toCloudFormation(stack) as SynthedStack;
-    expect(Object.keys(json.Resources)).toContain("ClassicLoadBalancer");
+    expect(stack).toHaveResourceOfTypeAndLogicalId("AWS::ElasticLoadBalancing::LoadBalancer", "ClassicLB");
   });
 
-  test("does not override the id if the stack migrated value is true but the override id value is false", () => {
+  test("has an auto-generated logicalId by default", () => {
     const stack = simpleGuStackForTesting({ migratedFromCloudFormation: true });
```
jacobwinch, line 33: Minor: we could remove the `{ migratedFromCloudFormation: true }` here just to keep things nice and simple

src/patterns/kinesis-lambda.ts
```diff
@@ -40,8 +41,7 @@ import { toAwsErrorHandlingProps } from "../utils/lambda";
  * existingKinesisStream: { externalKinesisStreamName: "KinesisStreamFromAnotherStack" }
  * ```
  */
-export interface ExistingKinesisStream {
-  logicalIdFromCloudFormation?: string;
+export interface ExistingKinesisStream extends GuMigratingResource {
```
jacobwinch, line 44: Some of the docs (in the comment above this interface) need to be updated due to this change

src/patterns/sns-lambda.ts
```diff
@@ -35,8 +36,7 @@ import { GuSnsTopic } from "../constructs/sns";
  *  existingSnsTopic: { externalTopicName: "MySnsTopicNameFromAnotherStack" }
  * ```
  */
-export interface ExistingSnsTopic {
-  logicalIdFromCloudFormation?: string;
+export interface ExistingSnsTopic extends GuMigratingResource {
```
jacobwinch, line 39: Sorry, but we need doc updates here too!

src/utils/test/jest.ts
```diff
@@ -20,13 +19,7 @@ expect.extend({
...
    */
   toHaveResourceOfTypeAndLogicalId(stack: GuStack, resourceType: string, logicalId: string | RegExp) {
-    const json = SynthUtils.toCloudFormation(stack) as SynthedStack;
-
-    const matchResult = Object.entries(json.Resources).find(([key, { Type }]) => {
-      const logicalIdMatch = logicalId instanceof RegExp ? logicalId.test(key) : key === logicalId;
-      const typeMatch = Type === resourceType;
-      return logicalIdMatch && typeMatch;
-    });
+    const matchResult = findResourceByTypeAndLogicalId(stack, resourceType, logicalId);
```
jacobwinch, line 22: I really grew to appreciate this helper function when reviewing this PR 😃

src/constructs/loadbalancing/alb.test.ts
jacobwinch: I see what you mean now, we could easily forget to call the function/remove the line accidentally during a refactor and that would have serious consequences. Perhaps we should leave these tests in for the purposes of this PR?

I don't think we need to tackle this in the current PR, but I wonder if there is a more general way to catch this type of problem across constructs... In theory any class which extends `GuMigratingResource` should call `GuMigratingResource.setLogicalId(this, scope, props)` as part of its constructor, right? I wonder if we could catch failures to do so with an assertion or lint rule.

src/constructs/kinesis/kinesis-stream.ts
```diff
@@ -1,15 +1,13 @@
-import type { CfnStream, StreamProps } from "@aws-cdk/aws-kinesis";
+import type { StreamProps } from "@aws-cdk/aws-kinesis";
 import { Stream } from "@aws-cdk/aws-kinesis";
 import type { GuStack } from "../core";
+import { GuMigratingResource } from "../core/migrating";
 
-export interface GuKinesisStreamProps extends StreamProps {
-  overrideId?: boolean;
-}
+export interface GuKinesisStreamProps extends StreamProps, GuMigratingResource {}
```
jacobwinch, line 6: I think we might want to implement `GuStatefulConstruct` here because:

* Kinesis streams can retain data for X days
* Stack A might use a Kinesis stream as way to accept input from stack B

Consequently I think it'd be helpful to receive [this warning](https://github.com/guardian/cdk/blob/54502809de5a111f26ad1574b78d488a77e37b85/src/constructs/core/migrating.ts#L49-L53) if the migration props have not been configured correctly.

src/constructs/loadbalancing/alb.test.ts
jacobwinch: Following our discussion above, it seems like it's worth keeping this test. If so, we might want to update the name here too:

`"does not override the id if the stack migrated value is true but the existingLogicalId prop is not defined"`

src/constructs/sns/sns-topic.ts
```diff
@@ -1,15 +1,13 @@
...
+import type { TopicProps } from "@aws-cdk/aws-sns";
 import type { GuStack } from "../core";
+import { GuMigratingResource } from "../core/migrating";
 
-interface GuSnsTopicProps extends TopicProps {
-  overrideId?: boolean;
-}
+interface GuSnsTopicProps extends TopicProps, GuMigratingResource {}
 
 export class GuSnsTopic extends Topic {
```
jacobwinch, line 8: We might want to implement `GuStatefulConstruct` here too? Stack A might use an SNS topic as way to accept input from stack B, so replacing may be unsafe?

src/constructs/ec2/security-groups/wazuh.ts
jacobwinch: I guess hardcoding will work fine as long as everyone who CloudForms this SG copy+pastes the [example logical id](https://github.com/guardian/security-hq/blob/main/hq/markdown/wazuh.md#outbound-traffic-on-ports-1514-and-1515) (seems reasonably likely!).

akash1810: The plot thickens!

An update to a security group's description is [destructive](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-security-group.html#cfn-ec2-securitygroup-groupdescription).

> Update requires: Replacement

The description we're using in this library differs from SecurityHQ's guide ("Wazuh agent registration and event logging" vs "Allow outbound traffic from wazuh agent to manager") 😱 .

I think it'll be easiest to change this library as adoption is still relatively low therefore we can more easily provide support to teams currently using it.

akash1810: https://github.com/guardian/cdk/pull/402

akash1810:
On reflection, the breakdown of this PR isn't great. I'll rework it to affect each construct in a commit and also action the feedback given. Moving to draft for now.

akash1810:
Closing as this change set is far too big and I'll break this into multiple PRs.

PR #404, opened 2021-04-09
fix: silence alarms in CODE and handle missing data correctly

jacobwinch:
## What does this change?

This PR makes two improvements to our alarms:
1. Disables alarm actions in `CODE` by default. This prevents us from spamming teams with alerts about their `CODE` environments (in my experience these just generate unactionable noise, as `CODE` environments are expected to break sometimes).
2. Handles missing data correctly for lambda-error-percentage alarms. No data in this scenario is fine; it just means that the lambda has not been invoked at all.

No. Any user who wants to keep (or temporarily test) alarm actions in their `CODE` environment can still do so if they wish by setting `actionsEnabledInCode` to `true`.

## How to test
I've tried upgrading https://github.com/guardian/tag-janitor/blob/main/cdk/lib/cdk-stack.ts (locally) in order to use this change and confirmed that the changes set looks sensible in AWS. I haven't tested actually deploying the change as there is no `CODE` environment for this stack. I'll finish the upgrade and check the deployment in `PROD` after releasing this change.

* We should receive fewer unactionable alarms
* Lambda-error-percentage alarms will be considered `OK` in scenarios where the lambda is not being invoked

There is a small risk that disabling these alarm actions in `CODE` (by default) will prevent us from noticing problems with alarm configuration. I think using the [`ActionsEnabled` property](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cw-alarm.html#cfn-cloudwatch-alarms-actionsenabled) allows us to minimise this risk as much as possible.

### DIFF
--- a/src/constructs/cloudwatch/__snapshots__/lambda-alarms.test.ts.snap
+++ b/src/constructs/cloudwatch/__snapshots__/lambda-alarms.test.ts.snap
@@ -2,6 +2,16 @@
 
 exports[`The GuLambdaErrorPercentageAlarm pattern should create the correct alarm resource with minimal config 1`] = `
 Object {
+  "Mappings": Object {
+    "stagemapping": Object {
+      "CODE": Object {
+        "alarmActionsEnabled": false,
+      },
+      "PROD": Object {
+        "alarmActionsEnabled": true,
+      },
+    },
+  },
   "Parameters": Object {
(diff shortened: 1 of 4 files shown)


akash1810 (approved):
Excellent improvement 👍🏽

PR #401, opened 2021-04-08
fix: Make it easier to use default props

akash1810:
WIP implementation of [this idea](https://www.typescriptlang.org/play?ssl=3&ssc=2&pln=1&pc=1#code/C4TwDgpgBA6glsAFgEQgMwIYFcA2wA8AKgHxQC8UA3gLABQUUyAogGICCAqgDKEBcUhOgF86dOADtgEAE6YAxtABC0gPYB3AM4yqdBuIwBbCPw3BpEgOYAaXVABuMjXBXj+4rAYBGM4aNpyXUyhldS1pfngkVExcAhDNGVIKGnpGVk4efhSGPUNjKAAiAEkmAptUhgdpJxd+ADZbEVomugDxIIsIYA4wtk7JfgAKT1UE6QB+fniwgEpyUlNzcQtyKGHRsKmN7QppmQA6ZnZuQjmyUmyoaS6saXEoAAMAEkoR0IP9IyEAehe3sf2VRq4iED18tFagWAUCwGAAjKtOt1ev1gIMZpD2tDYQAmRFdHoyPoQSSDShQT75ApyRCqIxleyOZyuKAIoQY2hAA).

PR #406, opened 2021-04-09
chore: add sensible defaults to GuAutoScalingGroup and GuApplicationLoadBalancer

stephengeller:
## What does this change?

This adds a few sensible defaults:

### 1. `GuInstanceRole`

We want/intend our autoscaling groups to have required permissions (eg AssumeRole for SSM) by default. This PR adds the role to the ASG construct so that you get it for free. When developing our EC2 pattern, it would be great to [have it created by default without explicitly specifying it in our pattern description](https://github.com/guardian/cdk/pull/399/files#r609673294).

### 2. `stageDependentProps` for `minimumInstances`

We also benefit from having sensible defaults for how we scale our instances based on stages. This adds a `minimumInstances` count of 1 for CODE (and therefore maximum of 2) and 3 for PROD (and therefore maximum of 6).

### 3. `ApplicationProtocol.HTTP` for `GuApplicationLoadBalancer`

By default, load balancers talk to autoscaling groups/instances via HTTP. This adds a sensible default to use HTTP that can be overridden.

I don't think so. I think you'd get a new role if you were to use the `GuAutoScalingGroup` construct _and_ didn't specify a role, but [every usage of this construct so far explicitly specifies a role](https://github.com/search?q=org%3Aguardian+%22new+GuAutoScalingGroup%22&type=code) so I think we're fine.

## How to test

`./script/test` from the top of the repository.

Less effort and work to get SSM & other goodies for free!

Can't think of any so far.

### DIFF
--- a/src/constructs/autoscaling/asg.test.ts
+++ b/src/constructs/autoscaling/asg.test.ts
@@ -10,6 +10,8 @@ import type { SynthedStack } from "../../utils/test";
 import { alphabeticalTags, simpleGuStackForTesting } from "../../utils/test";
 import type { AppIdentity } from "../core/identity";
 import { GuSecurityGroup } from "../ec2";
+import { GuAllowPolicy } from "../iam/policies";
+import { GuInstanceRole } from "../iam/roles";
 import { GuApplicationTargetGroup } from "../loadbalancing";
 import type { GuAutoScalingGroupProps } from "./asg";
 import { GuAutoScalingGroup } from "./";
@@ -224,26 +226,6 @@ describe("The GuAutoScalingGroup", () => {
     expect(Object.keys(json.Resources)).not.toContain("AutoscalingGroup");
   });
 
-  test("adds the correct mappings when provided with minimal capacity config", () => {
-    const stack = simpleGuStackForTesting();
-    new GuAutoScalingGroup(stack, "AutoscalingGroup", defaultProps);
-
-    const json = SynthUtils.toCloudFormation(stack) as SynthedStack;
-
-    expect(json.Mappings).toEqual({
-      stagemapping: {
-        CODE: {
-          minInstances: 1,
-          maxInstances: 2,
-        },
-        PROD: {
-          minInstances: 3,
-          maxInstances: 6,
-        },
-      },
-    });
-  });
-
   test("uses custom max capacities (if provided)", () => {
     const stack = simpleGuStackForTesting();
     new GuAutoScalingGroup(stack, "AutoscalingGroup", {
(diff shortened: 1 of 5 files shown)


stephengeller:
Closing this temporarily to incorporate another sensible default from PR #399

src/constructs/autoscaling/asg.ts
```diff
@@ -98,7 +99,7 @@ export class GuAutoScalingGroup extends AutoScalingGroup {
       machineImage: { getImage: getImage },
       instanceType: props.instanceType ?? new InstanceType(new GuInstanceTypeParameter(scope, props).valueAsString),
       userData,
-
+      role: new GuInstanceRole(scope, { app: props.app }),
```
jacobwinch, line 102: I think this change will prevent users from attaching a custom role to their ASG? If that's what we want, we should add "role" to the list of [omitted props](https://github.com/guardian/cdk/blob/d896ada6db37661863c0753b0986670492c08a06/src/constructs/autoscaling/asg.ts#L16-L27) to avoid confusing users.

However, I wonder if we should allow users to provide a custom role if they like, and default to `GuInstanceRole` if none is provided?

akash1810: > I wonder if we should allow users to provide a custom role if they like, and default to GuInstanceRole if none is provided?

Jinx! Had the same thought! Have been playing with the `WithDefault` idea too here https://github.com/guardian/cdk/pull/401/files.

stephengeller: I thought so too, but having written an additional test to check out this behaviour, it seems like passing in a role manually to `GuAutoScalingGroup` actually overrides the default. I would have assumed that, since `...props` is at the **top** of the object, it would be overridden by the below statements, but for some reason that didn't happen.

Regardless, this is a good point as we want usage to be clear to users. I've just pushed a change that should make it explicitly clear that `role` and `stageDependentProps` are optional and have sensible defaults if not specified.

src/constructs/autoscaling/asg.test.ts
```diff
@@ -295,4 +296,85 @@ describe("The GuAutoScalingGroup", () => {
...
+  });
+
+  test("passing in an instance role overrides the default", () => {
+    const stack = simpleGuStackForTesting();
+
+    new GuAutoScalingGroup(stack, "AutoscalingGroup", {
+      ...defaultProps,
+      app: "TestApp",
+      role: new GuInstanceRole(stack, {
+        app: "OverridenAppName",
```
akash1810, line 329: Oooh, why change the app in the role?

src/constructs/autoscaling/asg.test.ts
```diff
@@ -295,4 +296,85 @@ describe("The GuAutoScalingGroup", () => {
...
+      }),
+    });
+
+    expect(stack).toHaveResource("AWS::IAM::Role", {
+      Tags: [
+        { Key: "App", Value: "OverridenAppName" },
+        { Key: "gu:cdk:version", Value: "TEST" },
+        { Key: "Stack", Value: "test-stack" },
+        { Key: "Stage", Value: { Ref: "Stage" } },
+      ],
```
akash1810, line 339: There are a [couple of helpers](https://github.com/guardian/cdk/blob/b7b79365b5218b15d3bb798ac46e85b092af90a1/src/constructs/core/stack.test.ts#L41-L56) we can use for testing tags that we can use here.

src/constructs/autoscaling/asg.test.ts
```diff
@@ -295,4 +296,85 @@ describe("The GuAutoScalingGroup", () => {
...
+    });
+  });
+
+  test("has sensible default scaling capacities per stage based on minimum capacity", () => {
+    const stack = simpleGuStackForTesting();
+
+    new GuAutoScalingGroup(stack, "AutoscalingGroup", {
+      ...defaultProps,
```
jacobwinch, line 347: I think we need to remove some [default props from the test data](https://github.com/guardian/cdk/blob/c061c678305fec65660e0421216d441b962ad8be/src/constructs/autoscaling/asg.test.ts#L26-L32) to confirm that this test is working as expected.

src/constructs/autoscaling/asg.ts
```diff
@@ -93,12 +94,17 @@ export class GuAutoScalingGroup extends AutoScalingGroup {
     }
 
     const mergedProps = {
+      ...(props.stageDependentProps
+        ? wireStageDependentProps(scope, props.stageDependentProps)
+        : {
```
jacobwinch, line 99: I think these might also need to be converted to props that AWS will recognise using `wireStageDependentProps`... perhaps we can move that into a `const`?

src/constructs/autoscaling/asg.test.ts
stephengeller: You wouldn't, but thought it was a clear way to illustrate that the role being passed in as a prop would override the default instance role set.

src/constructs/autoscaling/asg.test.ts
stephengeller: Thanks, implemented in a new push.

src/constructs/autoscaling/asg.test.ts
stephengeller: Ah! Good catch, thanks. Tests and code have been updated to reflect this.

src/constructs/autoscaling/asg.ts
stephengeller: Yep, you're totally right. This has been implemented now.

src/constructs/autoscaling/asg.test.ts
```diff
@@ -301,4 +303,120 @@ describe("The GuAutoScalingGroup", () => {
...
+            Action: "some:Action",
+            Effect: "Allow",
+            Resource: "some:Resource",
+          },
+        ],
+      },
+    });
+  });
+
+  test("has sensible default scaling capacities per stage based on minimum capacity", () => {
```
jacobwinch, line 381: This test seems to overlap with an existing test (`adds the correct mappings when provided with minimal capacity config`) - perhaps we can combine them?

src/constructs/autoscaling/asg.test.ts
```diff
@@ -301,4 +303,120 @@ describe("The GuAutoScalingGroup", () => {
...
+      vpc,
+      stageDependentProps: {
+        [Stage.CODE]: { minimumInstances: 2 },
+        [Stage.PROD]: { minimumInstances: 5 },
+      },
+    });
+
+    const json = SynthUtils.toCloudFormation(stack) as SynthedStack;
+
+    expect(json.Mappings).toEqual({
```
jacobwinch, line 415: For completeness, both of these last two tests should probably also check that we are referencing these mappings correctly i.e. 

```typescript
expect(stack).toHaveResource("AWS::AutoScaling::AutoScalingGroup", {
  MinSize: {
    "Fn::FindInMap": [
      "stagemapping",
       {
          Ref: "Stage",
        },
        "minInstances",
      ],
    },
    MaxSize: {
      "Fn::FindInMap": [
        "stagemapping",
          {
          Ref: "Stage",
          },
          "maxInstances",
      ],
    },
});
```

If we add this here, then we can probably remove an existing test (`Uses Find In Map correctly to reference capacity mappings`) too.

src/constructs/autoscaling/asg.test.ts
stephengeller: True. I guess the difference with the existing test is that the first one [provides specific defaults](https://github.com/guardian/cdk/blob/053794f87cccc9f396c10554f3634507ae3653e5/src/constructs/autoscaling/asg.test.ts#L34), whereas this one doesn't? Would be happy to combine it into something sensible though.

src/constructs/autoscaling/asg.test.ts
stephengeller: Good shout, and done!

src/constructs/autoscaling/asg.test.ts
jacobwinch: I think we are currently testing 3 scenarios:

1. Using default capacity values gives you correct mappings & references (new test)
2. Using custom capacity values gives you correct mappings & references (new test)
3. Using custom capacity values that happen to be identical to the default values gives you correct mappings (existing test)

If 1 and 2 are working, I think it's highly unlikely that 3 will be broken, so I'd probably remove it for brevity.

src/constructs/loadbalancing/alb/application-target-group.ts
```diff
@@ -22,6 +22,7 @@ export class GuApplicationTargetGroup extends ApplicationTargetGroup {
     const { app } = props;
 
     const mergedProps = {
+      protocol: ApplicationProtocol.HTTP,
```
jacobwinch, line 25: Minor: should we add a test (and maybe docs?) for this?

src/constructs/loadbalancing/alb/application-listener.test.ts
```diff
@@ -246,4 +246,34 @@ describe("The GuHttpsApplicationListener class", () => {
       ],
     });
   });
+
+  test("uses HTTPS protocol by default", () => {
```
jacobwinch, line 250: I think this is already covered by [`sets default props`](https://github.com/guardian/cdk/blob/8125805fe5ece34996da2e031213e30a9a43711f/src/constructs/loadbalancing/alb/application-listener.test.ts#L121-L124)?

src/constructs/loadbalancing/alb/application-listener.test.ts
```diff
@@ -246,4 +246,34 @@ describe("The GuHttpsApplicationListener class", () => {
...
+      certificates: [{ certificateArn: "arn:aws:acm:eu-west-1:000000000000:certificate/123abc-0000-0000-0000-123abc" }],
+      ...app,
+    });
+
+    expect(stack).toHaveResource("AWS::ElasticLoadBalancingV2::Listener", {
+      Protocol: "HTTPS",
+    });
+  });
+
+  test("Can override default protocol with prop", () => {
```
jacobwinch, line 265: Minor: I would probably group this with [the other `GuApplicationListener` tests](https://github.com/guardian/cdk/blob/8125805fe5ece34996da2e031213e30a9a43711f/src/constructs/loadbalancing/alb/application-listener.test.ts#L36-L143).

jacobwinch (approved):
Looks good to me 👍 

I've left a few minor comments about tests, feel free to address/ignore as you see fit without seeking another review! 

We should probably improve the docs around these constructs, especially now we are defaulting more props. Perhaps we can have a separate card/PR for that so it doesn't get forgotten?

src/constructs/loadbalancing/alb/application-target-group.ts
stephengeller: Aha, yes this was a test that was missed when copying over to the new file structure. Re-added now.

src/constructs/autoscaling/asg.test.ts
stephengeller: Yeah, that makes sense, thanks for explaining it! I've removed the existing test as it is superfluous.

src/constructs/loadbalancing/alb/application-listener.test.ts
stephengeller: Correct, didn't catch this thanks.

src/constructs/loadbalancing/alb/application-listener.test.ts
stephengeller: Good shout, thanks.

stephengeller:
Good shout @jacobwinch , I've addressed the minor comments and will get a card written up for adding docs, as I think that's important to  do.

PR #409, opened 2021-04-09
fix: Make GuApplicationListeners support multiple apps

akash1810:
Requires https://github.com/guardian/cdk/pull/408.

## What does this change?

Use AppIdentity to identity the `GuApplicationListener` and `GuHttpsApplicationListener` constructs.

We're also making a call to `AppIdentity.taggedConstruct` here too even though it's a no-op as it's [not supported](https://docs.aws.amazon.com/ARG/latest/userguide/supported-resources.html#services-elasticloadbalancing).

This is for consistency across the project. If a construct's props extend `AppIdentity`, it should call `AppIdentity.taggedConstruct`.

Yes. The auto-generated logicalIds for `GuApplicationListener` and `GuHttpsApplicationListener` constructs have an updated naming scheme.

## How to test

See additional tests.

Consistency across the project.

n/a

### DIFF
--- a/src/constructs/loadbalancing/alb/application-listener.ts
+++ b/src/constructs/loadbalancing/alb/application-listener.ts
@@ -12,11 +12,11 @@ export interface GuApplicationListenerProps extends ApplicationListenerProps, Ap
 
 export class GuApplicationListener extends ApplicationListener {
   constructor(scope: GuStack, id: string, props: GuApplicationListenerProps) {
-    const { app } = props;
+    const { app, overrideId } = props;
 
     super(scope, AppIdentity.suffixText({ app }, id), { port: 443, protocol: ApplicationProtocol.HTTPS, ...props });
 
-    if (props.overrideId || (scope.migratedFromCloudFormation && props.overrideId !== false))
+    if (overrideId || (scope.migratedFromCloudFormation && overrideId !== false))
       (this.node.defaultChild as CfnListener).overrideLogicalId(id);
 
     /*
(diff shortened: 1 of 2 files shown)


src/constructs/loadbalancing/alb/application-listener.ts
```diff
@@ -3,19 +3,30 @@ import { ApplicationListener, ApplicationProtocol, ListenerAction } from "@aws-c
...
-    if (props.overrideId || (scope.migratedFromCloudFormation && props.overrideId !== false))
+    super(scope, AppIdentity.suffixText({ app }, id), { port: 443, protocol: ApplicationProtocol.HTTPS, ...props });
+
+    if (overrideId || (scope.migratedFromCloudFormation && overrideId !== false))
       (this.node.defaultChild as CfnListener).overrideLogicalId(id);
+
+    /*
+    AWS::ElasticLoadBalancingV2::Listener resources cannot be tagged.
+    Perform the call anyway for consistency across the project.
+    Who knows, maybe AWS will support it one day?!
```
jacobwinch, line 25: 🤞

src/constructs/loadbalancing/alb/application-listener.ts
```diff
@@ -3,19 +3,30 @@ import { ApplicationListener, ApplicationProtocol, ListenerAction } from "@aws-c
...
+    If so, tests will fail and we can celebrate!
```
jacobwinch, line 26: 🙃

jacobwinch (approved):

PR #423, opened 2021-04-12
fix: Update `GuApplicationListener` with revised logicalId logic

akash1810:
## What does this change?

In this change, we're updating the `GuApplicationListener` construct to adopt the new logic. As of #418 it's as simple as using the `GuStatefulMigratableConstruct` mixin!

The logicalId logic currently in `GuApplicationListener` is very complicated with multiple paths to overriding the logicalId.

This is massively simplified, as evidenced in the updated tests.

This is the first in a series of PRs to adopt the new logic. Hopefully this multi PR approach is better than the previous monster PR (#400)!

Possibly.

The overriding logic in `GuApplicationListener` has changed. If stacks are making use of the current (broken?) logic, they will need to be attended to.

## How to test

See tests?

A simpler, DRYer, code base?

### DIFF
--- a/src/constructs/loadbalancing/alb/application-listener.ts
+++ b/src/constructs/loadbalancing/alb/application-listener.ts
@@ -1,23 +1,20 @@
-import type { ApplicationListenerProps, CfnListener } from "@aws-cdk/aws-elasticloadbalancingv2";
+import type { ApplicationListenerProps } from "@aws-cdk/aws-elasticloadbalancingv2";
 import { ApplicationListener, ApplicationProtocol, ListenerAction } from "@aws-cdk/aws-elasticloadbalancingv2";
 import { RegexPattern } from "../../../constants";
+import { GuStatefulMigratableConstruct } from "../../../utils/mixin";
 import type { GuStack } from "../../core";
 import { GuCertificateArnParameter } from "../../core";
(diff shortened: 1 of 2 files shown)


src/constructs/loadbalancing/alb/application-listener.ts
```diff
@@ -1,24 +1,21 @@
...
+import type { GuMigratingResource } from "../../core/migrating";
 import type { GuApplicationTargetGroup } from "./application-target-group";
 
-export interface GuApplicationListenerProps extends ApplicationListenerProps, AppIdentity {
-  overrideId?: boolean;
-}
+export interface GuApplicationListenerProps extends ApplicationListenerProps, AppIdentity, GuMigratingResource {}
 
-export class GuApplicationListener extends ApplicationListener {
+export class GuApplicationListener extends GuStatefulMigratableConstruct(ApplicationListener) {
```
stephengeller, line 13: Did not know this is a thing, nice!

stephengeller (approved):
Took me a second to catch up on the new override ID logic works, but this works nicely.

src/constructs/loadbalancing/alb/application-listener.ts
akash1810: It's new 😄  #418.

PR #434, opened 2021-04-13
fix: Log messages as YAML comments

akash1810:
## What does this change?

When a stack is synthesized, AWS CDK will:
  - print the resulting CloudFormation template as YAML to the console
  - save the resulting CloudFormation template as JSON to disk

The AWS CDK library [does not support synthesizing a YAML template to disk](https://github.com/aws/aws-cdk/issues/2965). The only way to achieve this is is to redirect the result of `cdk synth` to disk.

For example:

```console
cdk synth > cloudformation.yaml
```

Since #364 we have started printing important messages to the console. This breaks the saving YAML to disk trick as the file on disk is not valid YAML.

For example:

```yaml
GuStack has 'migratedFromCloudFormation' set to false. MyDatabase is a stateful construct, it's logicalId will be auto-generated and AWS will create a new resource.
Parameters:
  Stage:
    Type: String
    Default: CODE
    AllowedValues:
      - CODE
      - PROD
    Description: Stage name
Resources:
  MyDatabaseA1B2C3D4:
    Type: AWS::RDS::DBInstance
```

For this reason, prefix console messages with "# ". This results in `cloudformation.yaml` still being valid YAML.

For example:

```yaml
# GuStack has 'migratedFromCloudFormation' set to false. MyDatabase is a stateful construct, it's logicalId will be auto-generated and AWS will create a new resource.
Parameters:
  Stage:
    Type: String
    Default: CODE
    AllowedValues:
      - CODE
      - PROD
    Description: Stage name
Resources:
  MyDatabaseA1B2C3D4:
    Type: AWS::RDS::DBInstance
```

See:
  - [AWS CDK](https://github.com/aws/aws-cdk/issues/2965)
  - [Typeerighter](https://github.com/guardian/typerighter/blob/fa90ef260cd71e0f4fa1b893d7bba9b87ff828ef/cdk/package.json#L12)

No.

## How to test

Not sure. Help wanted!

Was thinking the [integration-test](https://github.com/guardian/cdk/tree/main/integration-test) however not sure how to write the test.

The redirect trick to save a YAML template works.

n/a

### DIFF
--- a/src/utils/logger.ts
+++ b/src/utils/logger.ts
@@ -1,18 +1,58 @@
 /**
- * A small wrapper around `console`.
+ * A small wrapper around `console`, printing messages as YAML comments.
+ *
+ * When a stack is synthesized, AWS CDK will:
+ *   - print the resulting CloudFormation template as YAML to the console
+ *   - save the resulting CloudFormation template as JSON to disk
+ *
+ * In addition, any messages from `Logger` will be echoed on the console.
+ *
+ * The AWS CDK library does not support synthesizing a YAML template to disk.
+ * The only way to achieve this is is to redirect the result of `cdk synth` to disk.
+ * For example:
+ *
+ * ```console
+ * cdk synth > cloudformation.yaml
+ * ```
+ *
+ * For this reason, these messages will have a "# " prefix.
+ * This results in `cloudformation.yaml` still being valid YAML.
+ *
+ * For example:
+ *
+ * ```yaml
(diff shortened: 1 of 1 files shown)


jacobwinch (approved):
I have never used this trick before, but having a `#` in the log line doesn't cause any harm (and it might actually be useful to have these warnings at the top of a template), so 👍 for this simple workaround!

PR #435, opened 2021-04-13
fix: Update `GuApplicationTargetGroup` with revised logicalId logic

akash1810:
## What does this change?

In this change, we're updating the `GuApplicationTargetGroup` construct to adopt the new logic. As of #418 it's as simple as using the `GuStatefulMigratableConstruct` mixin!

This is another PR in this series. Favouring small PRs over the a massive one (#400).

Possibly.

The overriding logic in `GuApplicationTargetGroup` changed. If stacks are making use of the current (broken?) logic, they will need to be attended to.

## How to test

See tests?

A simpler, DRYer, more consistent code base?

### DIFF
--- a/src/constructs/loadbalancing/alb/application-target-group.test.ts
+++ b/src/constructs/loadbalancing/alb/application-target-group.test.ts
@@ -1,11 +1,9 @@
 import "@aws-cdk/assert/jest";
 import "../../../utils/test/jest";
-import { SynthUtils } from "@aws-cdk/assert";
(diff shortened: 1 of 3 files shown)


src/constructs/loadbalancing/alb/application-target-group.test.ts
```diff
@@ -1,11 +1,9 @@
 import "@aws-cdk/assert/jest";
```
akash1810, line 1: Loving the simplified logic of #364!

PR #439, opened 2021-04-13
feat: Add construct for EC2-based infrastructure alarm

jacobwinch:
## What does this change?
This PR adds a construct which allows users to easily create an alarm as per [this recommendation](https://github.com/guardian/recommendations/pull/35). In a subsequent PR we'll be able to create this alarm as part of the EC2 App pattern.

No.

## How to test

I've tested:

* CloudForming a stack with this alarm
* Triggering the alarm by forcing instances to serve 5XXs
* Triggering the alarm by forcing the load balancer to serve 5XXs

It should be easy to set up an alert as part of the EC2 App Pattern.

This should be a low risk change.

### DIFF
--- a/src/constructs/cloudwatch/ec2-alarms.test.ts
+++ b/src/constructs/cloudwatch/ec2-alarms.test.ts
@@ -24,7 +24,6 @@ describe("The Gu5xxPercentageAlarm construct", () => {
     const props = {
       tolerated5xxPercentage: 1,
       snsTopicName: "test-topic",
-      evaluationPeriods: 1, //FIXME
     };
     new Gu5xxPercentageAlarm(stack, "test", { ...app, loadBalancer: alb, ...props });
     expect(SynthUtils.toCloudFormation(stack)).toMatchSnapshot();
@@ -37,7 +36,6 @@ describe("The Gu5xxPercentageAlarm construct", () => {
       alarmDescription: "test-custom-alarm-description",
       tolerated5xxPercentage: 1,
       snsTopicName: "test-topic",
-      evaluationPeriods: 1, //FIXME
     };
     new Gu5xxPercentageAlarm(stack, "test", { ...app, loadBalancer: alb, ...props });
     expect(stack).toHaveResource("AWS::CloudWatch::Alarm", {
@@ -52,11 +50,24 @@ describe("The Gu5xxPercentageAlarm construct", () => {
       alarmName: "test-custom-alarm-name",
       tolerated5xxPercentage: 1,
       snsTopicName: "test-topic",
-      evaluationPeriods: 1, //FIXME
     };
     new Gu5xxPercentageAlarm(stack, "test", { ...app, loadBalancer: alb, ...props });
     expect(stack).toHaveResource("AWS::CloudWatch::Alarm", {
       AlarmName: "test-custom-alarm-name",
+    });
+  });
+
+  it("should adjust the number of evaluation periods if a custom value is provided", () => {
+    const stack = simpleGuStackForTesting();
+    const alb = new GuApplicationLoadBalancer(stack, "ApplicationLoadBalancer", { ...app, vpc });
+    const props = {
+      tolerated5xxPercentage: 1,
+      numberOfFiveMinutePeriodsToEvaluate: 3,
+      snsTopicName: "test-topic",
+    };
+    new Gu5xxPercentageAlarm(stack, "test", { ...app, loadBalancer: alb, ...props });
+    expect(stack).toHaveResource("AWS::CloudWatch::Alarm", {
+      EvaluationPeriods: 3,
     });
   });
 });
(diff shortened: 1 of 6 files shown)


src/constructs/cloudwatch/ec2-alarms.ts
```diff
@@ -0,0 +1,50 @@
...
+import type { AppIdentity } from "../core/identity";
+import type { GuApplicationLoadBalancer } from "../loadbalancing";
+import type { GuAlarmProps } from "./alarm";
+import { GuAlarm } from "./alarm";
+
+/**
+ * Creates an alarm which is triggered whenever the percentage of requests with a 5xx response code exceeds
+ * the specified threshold.
+ */
+export interface Gu5xxPercentageMonitoringProps
```
jacobwinch, line 13: This is the input we need from the user

src/constructs/cloudwatch/ec2-alarms.ts
```diff
@@ -0,0 +1,50 @@
...
+  extends Omit<GuAlarmProps, "evaluationPeriods" | "metric" | "period" | "threshold" | "treatMissingData">,
+    AppIdentity {
+  tolerated5xxPercentage: number;
+  numberOfFiveMinutePeriodsToEvaluate?: number;
+  noMonitoring?: false;
+}
+
+interface GuLoadBalancerAlarmProps extends Gu5xxPercentageMonitoringProps {
```
jacobwinch, line 21: The pattern should be able to pass through the load balancer in addition to the user defined props

src/constructs/cloudwatch/ec2-alarms.ts
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Creates an alarm which is triggered whenever the percentage of requests with a 5xx response code exceeds
+ * the specified threshold.
+ */
+export interface Gu5xxPercentageMonitoringProps
+  extends Omit<GuAlarmProps, "evaluationPeriods" | "metric" | "period" | "threshold" | "treatMissingData">,
+    AppIdentity {
+  tolerated5xxPercentage: number;
+  numberOfFiveMinutePeriodsToEvaluate?: number;
```
jacobwinch, line 17: AWS default to 5-minute evaluation periods - we could consider overriding this default (or allowing users to) if we felt that it was important?

src/constructs/cloudwatch/ec2-alarms.test.ts
```diff
@@ -0,0 +1,73 @@
...
+});
+
+const app: AppIdentity = {
+  app: "testing",
+};
+
+describe("The Gu5xxPercentageAlarm construct", () => {
+  it("should create the correct alarm resource with minimal config", () => {
+    const stack = simpleGuStackForTesting();
+    const alb = new GuApplicationLoadBalancer(stack, "ApplicationLoadBalancer", { ...app, vpc });
```
akash1810, line 23: tangent: I'm refactoring `GuApplicationLoadBalancer` to update it's logicalId override implementation and have been thinking the `id` can be defaulted and removed from the constructor as it's always `ApplicationLoadBalancer`.

src/constructs/cloudwatch/ec2-alarms.ts
akash1810: Not sure if you've seen how AWS CDK implement this? They have a [`private` directory](https://github.com/aws/aws-cdk/tree/master/packages/cdk-assets/lib/private) for private interfaces/types. That is, if something is defined outside there, it's a public interface. Not sure if we can cleanly copy that pattern (or even if we should?).

src/constructs/cloudwatch/ec2-alarms.ts
akash1810: I think 5 minutes is too slow for PROD. Would favour GuCDK providing a higher frequency, but allow it to be overridden.

IIUC the alarm will trip if `tolerated5xxPercentage` is exceeded within `numberOfFiveMinutePeriodsToEvaluate`? Might be overkill, but maybe we can print a warning when this is likely to be too long to be useful and other alarming strategies should be considered.

akash1810 (approved):
Couple of questions, nothing blocking. LGTM.

src/constructs/cloudwatch/ec2-alarms.ts
jacobwinch: > I think 5 minutes is too slow for PROD. Would favour GuCDK providing a higher frequency, but allow it to be overridden.

I'm inclined to agree - perhaps 1 minute would be better? 

We are using the [same default for lambda-based alerts](https://github.com/guardian/cdk/blob/1c648dcd8c0a0ed42096080c69e7405571031e21/src/constructs/cloudwatch/lambda-alarms.ts#L13). I will have a look at improving granularity for both in a subsequent PR, as it will require a bit of further experimentation/testing.

src/constructs/cloudwatch/ec2-alarms.ts
jacobwinch: Interesting, I hadn't looked into that before. 

> Not sure if we can cleanly copy that pattern (or even if we should?).

I'm not sure about this part either; will have a think about it!

PR #441, opened 2021-04-13
fix: Update `GuAutoScalingGroup` with revised logicalId logic

akash1810:
## What does this change?

In this change, we're updating the `GuAutoScalingGroup` construct to adopt the new logic. As of #418 it's as simple as using the `GuStatefulMigratableConstruct` mixin!

This is another PR in this series. Favouring small PRs over the a massive one (#400).

Possibly.

The overriding logic in `GuAutoScalingGroup` changed. If stacks are making use of the current (broken?) logic, they will need to be attended to.

## How to test

See tests?

A simpler, DRYer, more consistent code base?

### DIFF
--- a/src/constructs/autoscaling/asg.ts
+++ b/src/constructs/autoscaling/asg.ts
@@ -4,9 +4,11 @@ import type { ISecurityGroup, MachineImage, MachineImageConfig } from "@aws-cdk/
 import { InstanceType, OperatingSystemType, UserData } from "@aws-cdk/aws-ec2";
 import type { ApplicationTargetGroup } from "@aws-cdk/aws-elasticloadbalancingv2";
 import { Stage } from "../../constants";
+import { GuStatefulMigratableConstruct } from "../../utils/mixin";
 import type { GuStack } from "../core";
 import { GuAmiParameter, GuInstanceTypeParameter } from "../core";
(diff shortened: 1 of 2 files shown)


src/constructs/autoscaling/asg.ts
```diff
@@ -107,7 +109,7 @@ export class GuAutoScalingGroup extends AutoScalingGroup {
       userData,
       // Do not use the default AWS security group which allows egress on any port.
       // Favour HTTPS only egress rules by default.
-      securityGroup: GuHttpsEgressSecurityGroup.forVpc(scope, props),
+      securityGroup: GuHttpsEgressSecurityGroup.forVpc(scope, { app: props.app, vpc: props.vpc }),
```
akash1810, line 112: As [commented on](https://github.com/guardian/cdk/pull/400/files#r609667036) in #400:

> Passing in props meant existingLogicalId was copied to the GuHttpsEgressSecurityGroup too. When it came time to synth the stack, it failed because two resources were being created with the same logicalId (the GuHttpsEgressSecurityGroup and the GuAutoScalingGroup).
> 
> This took the best part of two days to diagnose 😨 !

jacobwinch (approved):

PR #445, opened 2021-04-14
feat: generate certificate as part of EC2 App pattern

jacobwinch:
## What does this change?
Previously, users had to create their ACM certificate outside of `cdk` and then pass it in via a parameter. Certificate generation is now handled by the EC2 App pattern*.

*If the DNS record required to validate the certificate can be created in a Route 53 hosted zone, users do not need to do anything. If the aforementioned DNS record needs to be created in NS1, this must be created manually before stack creation will complete. We should automate this step too but I think it's a separate, larger piece of work.

No, this pattern is still under development.

## How to test
Existing unit tests have been updated.

More resources are version controlled; less work needs to happen outside of pattern creation.

I think this is reasonably low risk. In order to reduce risk of keeping certificates under version control (and/or deploying them automatically) all certificates are created with `"DeletionPolicy": "Retain"`.


akash1810:
> If the DNS record required to validate the certificate can be created in a Route 53 hosted zone, users do not need to do anything. If the aforementioned DNS record needs to be created in NS1, this must be created manually before stack creation will complete. We should automate this step too but I think it's a separate, larger piece of work.

It would be good to print warnings about this, though this might be the role of RiffRaff or other deployment tooling.

NS1 work is automated for some of our domains with [dns-validation-lambda](https://github.com/guardian/dns-validation-lambda).

src/constructs/acm/certificate.ts
```diff
@@ -8,7 +8,8 @@ import type { GuStack } from "../core";
 import { AppIdentity } from "../core/identity";
 import type { GuMigratingResource } from "../core/migrating";
 
-export type GuCertificateProps = Record<Stage, GuDnsValidatedCertificateProps> & GuMigratingResource & AppIdentity;
+export type GuCertificateProps = Record<Stage, GuDnsValidatedCertificateProps> & GuMigratingResource;
+export type GuCertificatePropsWithApp = GuCertificateProps & AppIdentity;
```
akash1810, line 12: Feels a little horrible to make this split just to make the pattern props API simpler. The only alternative I can think of is to `GuEc2AppProps` to [`Omit`](https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys) the app field in `GuCertificateProps`, but not much better.

Might be one to address later once the AppIdentity mixin is written?

akash1810 (approved):

src/constructs/acm/certificate.ts
jacobwinch: > The only alternative I can think of is to `GuEc2AppProps` to [`Omit`](https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys) the app field in `GuCertificateProps`, but not much better.

Yep, I tried this alternative too and decided that the approach in this PR was 'least-bad' 🙈

jacobwinch: > Might be one to address later once the AppIdentity mixin is written?

I've already found myself needing to do this twice today ([see also](https://github.com/guardian/cdk/pull/446/files#diff-e4e7eb9118f3261284cbc3ed695b73b19661de3cbb7b438e102a8293a50719d1R19-R20)). So it's definitely going to be worth thinking about a general solution for this.

jacobwinch:
> NS1 work is automated for some of our domains with [dns-validation-lambda](https://github.com/guardian/dns-validation-lambda).

Oh that's great, I didn't know about this! Perhaps it's worth adding any missing domains to the lambda then? That way we can tell users 'it will just work' without needing to expose any of this complexity to them?

akash1810:
> > NS1 work is automated for some of our domains with [dns-validation-lambda](https://github.com/guardian/dns-validation-lambda).
> 
> Oh that's great, I didn't know about this! Perhaps it's worth adding any missing domains to the lambda then? That way we can tell users 'it will just work' without needing to expose any of this complexity to them?

I think it purposefully only operates on the gutools domains and not on *.theguardian.com. Is this right @sihil ?

jacobwinch:
Update: I've just tested this out with our new hosted zone and I can confirm that this creates the certificate and the appropriate CNAME automatically 🎉

PR #446, opened 2021-04-14
fix: improve support for GuUserData in EC2 App pattern

jacobwinch:
## What does this change?
Our [first draft of the EC2 App pattern](https://github.com/guardian/cdk/pull/399) works fine if `userData` is provided as a `string`. This PR fixes a [todo](https://github.com/guardian/cdk/pull/399#discussion_r609663035) so that our custom `userData` type (`GuUserDataProps`) can be used more easily. More specifically, if the user informs us that they need to download private config from s3, we now ensure that EC2 instances will (automatically) have the permissions to do this.

No.

## How to test
I've added a unit test to cover this scenario. We should probably start testing this pattern with a 'real' application soon too!

Users won't need to manually add permissions in a case like this, where we can figure out the permissions needed automatically.

I can't think of any...

### DIFF
--- a/src/patterns/ec2-app.test.ts
+++ b/src/patterns/ec2-app.test.ts
@@ -1,6 +1,7 @@
 import "@aws-cdk/assert/jest";
 import { SynthUtils } from "@aws-cdk/assert";
 import { TrackingTag } from "../constants/library-info";
+import { GuDistributionBucketParameter, GuPrivateConfigBucketParameter } from "../constructs/core";
 import { alphabeticalTags, simpleGuStackForTesting } from "../utils/test";
 import { GuApplicationPorts, GuEc2App, GuNodeApp, GuPlayApp } from "./ec2-app";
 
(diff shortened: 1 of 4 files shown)


src/patterns/ec2-app.test.ts
```diff
@@ -16,6 +17,64 @@ describe("the GuEC2App pattern", function () {
...
+      app: app,
+      publicFacing: false,
+      userData: {
+        distributable: {
+          bucket: GuDistributionBucketParameter.getInstance(stack),
+          fileName: "my-app.deb",
+          executionStatement: `dpkg -i /${app}/my-app.deb`,
+        },
+        configuration: {
+          bucket: new GuPrivateConfigBucketParameter(stack),
```
akash1810, line 34: for a different PR: should `GuPrivateConfigBucketParameter` be a singleton, similar to `GuDistributionBucketParameter`?

src/patterns/ec2-app.test.ts
```diff
@@ -16,6 +17,64 @@ describe("the GuEC2App pattern", function () {
     expect(SynthUtils.toCloudFormation(stack)).toMatchSnapshot();
   });
 
+  it("adds the correct permissions for apps which need to fetch private config from s3", function () {
+    const stack = simpleGuStackForTesting();
+    const app = "test-gu-ec2-app";
+    new GuEc2App(stack, {
```
akash1810, line 23: The interface to `GuEc2App` is looking really nice!

akash1810 (approved):
A couple of minor comments, LGTM 👍🏽 .

It looks like we're getting close to `GuUserData` being able to default all it's props? That is, in the basic instance, all that's needed to create a `GuUserData` is an `app` string?

src/patterns/ec2-app.test.ts
jacobwinch: Yes, I think so. Even if teams have private config split across multiple buckets it should be pretty trivial to copy & paste a couple of config files into a different bucket to avoid complexity in the template.

PR #448, opened 2021-04-15
feat: Include `__mocks__` in TS output (and published library)

akash1810:
## What does this change?

In #223 (and #233) we started adding a tag (`gu:cdk:version`) to track which version of `@guardian/cdk` is being used.

This tag helps us measure adoption:
  - the presence of the tag tells us the library is being used
  - the value of the tag tells us the version of the library being used

Whilst this is really useful, it does somewhat block users from using tools such as dependabot. This is because the snapshot tests will _always_ fail due to the value of the tag changing between version numbers.

In this repository, we have configured a global mock for this tag to ensure its value is static and thus simplifying the tests.

This change updates the TypeScript configuration to include (more specifically stop excluding) the `__mocks__` directories.

This results in all mocks making their way into the published library which can then be used within client stacks,
to also have a static value.

We currently have a single mock in the repository. If we add more, we may want to be more selective about which of them get published.

No.

## How to test

n/a

Keeping up to date with releases automatically becomes more viable. This reduces risk and [scary changes](https://github.com/guardian/prism/pull/155).

We, one day, add more mocks and start publishing them too?

### DIFF
--- a/docs/001-general-usage.md
+++ b/docs/001-general-usage.md
@@ -69,3 +69,53 @@ cdk synth --app="ts-node ./bin/my-app.ts" > cloudformation.yaml
 ```
 
 For app with multiple stacks, you can specify a stack to see the YAML output.
+
+## 🧪 Testing
+It is recommended to set up [Jest snapshot testing](https://jestjs.io/docs/snapshot-testing) for your stack and have the test run during CI.
+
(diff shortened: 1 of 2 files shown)


jacobwinch (approved):
Great idea!

PR #454, opened 2021-04-15
feat: use one minute granularity for 5XX alarms

jacobwinch:
## What does this change?

We are now using 1-minute periods (instead of 5-minute periods) for our high 5XX % alarms ([as per previous PR discussion](https://github.com/guardian/cdk/pull/439#discussion_r612501327)). By default, we will now move into an alarm state when errors have been above the user-specified threshold for 1 minute. Users can choose to make this alarm less sensitive if they wish (i.e. don't trigger the alarm unless the error threshold is exceeded for 5 consecutive minutes); they may choose to do this if they are defining infrastructure for a low priority, or low-traffic service, for example.

In theory this is a breaking change to the API, but I don't think anyone is actually using this construct yet as it's only existed for a couple of days.

## How to test
I've tested this by adding an alarm to our EC2 pattern. I confirmed that the alarm state is triggered more promptly than before (it still feels a little laggier than 1-minute, but I think that's down to AWS), and I've also tested that providing a custom value for `numberOfMinutesAboveThresholdBeforeAlarm` works as expected.

Previously it would have taken ~5 minutes to detect a production incident, now it should take ~1 minute.

Yes, there is a risk that we will generate noisy alarms by using more sensitive defaults. If this turns out to be the case we should consider relaxing the defaults.


src/constructs/cloudwatch/ec2-alarms.ts
```diff
@@ -43,7 +45,7 @@ export class Gu5xxPercentageAlarm extends GuAlarm {
       comparisonOperator: ComparisonOperator.GREATER_THAN_THRESHOLD,
       alarmName: props.alarmName ?? defaultAlarmName,
       alarmDescription: props.alarmDescription ?? defaultDescription,
-      evaluationPeriods: props.numberOfFiveMinutePeriodsToEvaluate ?? 1,
+      evaluationPeriods: props.numberOfMinutesAboveThresholdBeforeAlarm ?? 1,
```
jacobwinch, line 48: @sihil - I know you had some concerns about noisy alarms, so it'd be interesting to hear your take on this:

>  By default, we will now move into an alarm state when errors have been above the user-specified threshold for 1 minute.

akash1810 (approved):
The change looks good to me. I like the more explicit naming. Though it may lead to noisy alarms by default, the fact it's down to the minute means users have more fine grained control so can customise it as needed.

jacobwinch:
> Though it may lead to noisy alarms by default, the fact it's down to the minute means users have more fine grained control so can customise it as needed.

👍  We can always update the default in the future if necessary too.

PR #455, opened 2021-04-15
feat: Simplify props for `GuLambdaFunction`

akash1810:
## What does this change?

Simplifies the props needed to create a `GuLambdaFunction` by removing the bucket key in favour of using the `GuDistributionBucketParameter` singleton.

This brings `GuLambdaFunction` in line with our EC2 constructs and patterns.

`GuDistributionBucketParameter` adds an SSM parameter to a stack with a default path of `/account/services/artifact.bucket`.

BREAKING CHANGE: Shape change to the props for `GuLambdaFunction`

The `code.bucket` field has been removed.

Yes. They'll need to update, the diff shouldn't be too difficult though as it's just the removal of a property.

## How to test

CI should continue to pass.

Providing a simpler API that is consistent between EC2 and Lambda constructs.

n/a

### DIFF
--- a/src/constructs/cloudwatch/__snapshots__/lambda-alarms.test.ts.snap
+++ b/src/constructs/cloudwatch/__snapshots__/lambda-alarms.test.ts.snap
@@ -13,6 +13,11 @@ Object {
     },
   },
   "Parameters": Object {
+    "DistributionBucketName": Object {
+      "Default": "/account/services/artifact.bucket",
+      "Description": "SSM parameter containing the S3 bucket name holding distribution artifacts",
+      "Type": "AWS::SSM::Parameter::Value<String>",
+    },
     "Stage": Object {
       "AllowedValues": Array [
         "CODE",
@@ -31,7 +36,9 @@ Object {
       ],
       "Properties": Object {
         "Code": Object {
-          "S3Bucket": "bucket1",
+          "S3Bucket": Object {
+            "Ref": "DistributionBucketName",
+          },
           "S3Key": "folder/to/key",
         },
         "Handler": "handler.ts",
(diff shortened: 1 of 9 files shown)


jacobwinch (approved):
Good idea 👍 

It would be good to document this convention too; feel free to do that later / in a separate PR though if you are trying to make progress with other changes which are dependent on this.

akash1810:
> It would be good to document this convention too

Yes! Not sure where though. Any suggestions?

jacobwinch:
> Yes! Not sure where though. Any suggestions?

Maybe it should go with `GuFunctionProps`? The lambda-based pattern docs [typically point users in this direction](https://github.com/guardian/cdk/blob/2b3296f18a690b523be02ba40a076f92d40ecc83/src/patterns/kinesis-lambda.ts#L51), to avoid repeating the same docs in lots of places.

akash1810:
A thought: I think we'd want to enforce the path of the lambda distributable in the bucket, similar to we do for EC2. #259 tightened the `GuGetDistributablePolicy` to allow `s3:GetObject` on the `stack/stage/app/*` path.

akash1810:
Another thought, we might benefit from reusing [`GuUserDataS3DistributableProps`](https://github.com/guardian/cdk/blob/e79b15ce34717b1b0e3a17bead74ae34c24f5662/src/constructs/autoscaling/user-data.ts#L13) in `GuFunctionProps`? I think that would be +1 to having a top level `types` directory too.

akash1810:
> > Yes! Not sure where though. Any suggestions?
> 
> Maybe it should go with `GuFunctionProps`? The lambda-based pattern docs [typically point users in this direction](https://github.com/guardian/cdk/blob/2b3296f18a690b523be02ba40a076f92d40ecc83/src/patterns/kinesis-lambda.ts#L51), to avoid repeating the same docs in lots of places.

Added now.

PR #468, opened 2021-04-17
refactor!: Rename `library-info` to be more intention revealing

akash1810:
## What does this change?

Rename `library-info` to `tracking-tag` to be more descriptive and intention revealing.

This change makes the mocking advice added in #448 less confusing as the module name maps to the intent.

BREAKING CHANGE: `"@guardian/cdk/lib/constants/library-info"` renamed to `"@guardian/cdk/lib/constants/tracking-tag"`

Update the mock path in `jest.setup.js` - `"@guardian/cdk/lib/constants/library-info"` becomes `"@guardian/cdk/lib/constants/tracking-tag"`.

Yes. See breaking change description above.

## How to test

Tests continue to pass.

A more obvious and descriptive API.

n/a


stephengeller (approved):
Looks like a sensible change.

PR #478, opened 2021-04-19
feat: Simplify props for `GuUserData`

akash1810:
## What does this change?

Following from #455, simplify the props for `GuUserData` by removing the bucket key in favour of using the `GuDistributionBucketParameter` singleton.

BREAKING CHANGE: Shape change to the props for `GuUserData`

The `bucket` field has been removed.

Yes - the props to `GuUserData` has changed.

## How to test

Tests continue to pass.

A simpler, more consistent API.

With this change, we require an account to have the parameter store entry created. We do not have any tooling to assess an account's GuCDK readiness, so we're somewhat relying on our documentation for now.

### DIFF
--- a/src/constructs/autoscaling/user-data.ts
+++ b/src/constructs/autoscaling/user-data.ts
@@ -2,17 +2,21 @@ import type { S3DownloadOptions } from "@aws-cdk/aws-ec2";
 import { UserData } from "@aws-cdk/aws-ec2";
 import { Bucket } from "@aws-cdk/aws-s3";
 import type { GuPrivateS3ConfigurationProps } from "../../utils/ec2";
-import type { GuDistributionBucketParameter, GuStack } from "../core";
+import { GuDistributionBucketParameter } from "../core";
+import type { GuStack } from "../core";
 import type { AppIdentity } from "../core/identity";
 
-/**
- * Where to download a distributable from.
- * We'll look for `fileName` on the path "bucket/stack/stage/app/<fileName>".
- * `executionStatement` will be something like "dpkg -i application.deb` or `service foo start`.
- */
 export interface GuUserDataS3DistributableProps {
-  bucket: GuDistributionBucketParameter;
+  /**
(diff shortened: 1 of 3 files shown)


src/constructs/autoscaling/user-data.ts
```diff
@@ -11,7 +12,6 @@ import type { AppIdentity } from "../core/identity";
  * `executionStatement` will be something like "dpkg -i application.deb` or `service foo start`.
  */
 export interface GuUserDataS3DistributableProps {
-  bucket: GuDistributionBucketParameter;
   fileName: string;
```
jacobwinch, line 15: It's probably worth mentioning where the bucket name is coming from in this comment (i.e. something similar to [this one](https://github.com/guardian/cdk/blob/2692d040c70486fbe5acb11075e4fc0a027d86fa/src/constructs/lambda/lambda.ts#L23-L25)). Tangential, but this makes me wonder if we should fix the [`key`](https://github.com/guardian/cdk/blob/2692d040c70486fbe5acb11075e4fc0a027d86fa/src/constructs/lambda/lambda.ts#L23-L26)/`fileName` inconsistency too?

akash1810: > It's probably worth mentioning where the bucket name is coming from in this comment

👍🏽  thanks for the reminder!

> Tangential, but this makes me wonder if we should fix the [`key`](https://github.com/guardian/cdk/blob/2692d040c70486fbe5acb11075e4fc0a027d86fa/src/constructs/lambda/lambda.ts#L23-L26)/`fileName` inconsistency too?

Absolutely! Spoiler alert: this is going to be raised in a couple of PRs 😄 .

jacobwinch: I thought you'd probably notice it too! 😁

jacobwinch (approved):

akash1810:
Closing in favour of #479 which provides a change that can more easily be added to our lambdas.

PR #479, opened 2021-04-20
feat: Simplify props for `GuUserData`

akash1810:
An alternative to #478 .

## What does this change?

Following from #455, simplify the props for `GuUserData` by removing the bucket key in favour of using the `GuDistributionBucketParameter` singleton.

This is the start of a `GuLambdaFunction` refactoring which will, ultimately, answer [this TODO](https://github.com/guardian/cdk/blob/af4ab09e53f0695cd170b474e10ab1b4f95b5f1f/src/constructs/lambda/lambda.ts#L21).

This change introduces a new `src/types` directory to house our types. Having a dedicated directory makes it easier to share types between constructs.

BREAKING CHANGE: Shape change to the props for `GuUserData`

The `bucket` field has been removed and we'll automatically use `GuDistributionBucketParameter`.

Yes, see breaking change notes above.

## How to test

See updated tests.

A simpler API with a smaller surface area and more default values.

n/a

### DIFF
--- a/src/constructs/autoscaling/user-data.test.ts
+++ b/src/constructs/autoscaling/user-data.test.ts
@@ -3,7 +3,7 @@ import { Vpc } from "@aws-cdk/aws-ec2";
 import { Stack } from "@aws-cdk/core";
 import { Stage } from "../../constants";
 import { simpleGuStackForTesting } from "../../utils/test";
-import { GuDistributionBucketParameter, GuPrivateConfigBucketParameter } from "../core";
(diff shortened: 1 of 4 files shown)


jacobwinch (approved):

akash1810:
> This is the start of a `GuLambdaFunction` refactoring which will, ultimately, answer this TODO.

There are going to be a series of "breaking change" PRs for this bit of work. To simplify the release notes, I'd like to raise all the PRs and merge them into each other to, hopefully, create a single new release.

That is, this PR might sit open for a little while!

PR #481, opened 2021-04-20
feat: Simplify props for `GuLambdaFunction`

akash1810:
## What does this change?

Currently the props to `GuLambdaFunction` are a little confusing.

The props currently have an optional `rules` and `apis` property.

This means it is possible to create a lambda that:
  - Sits there with no event source
  - Runs on a cron
  - Runs behind API Gateway
  - Runs on a cron and behind API Gateway

We never want the first and last option to be possible.

This change moves the props for a cron lambda into the `GuScheduledLambda` pattern for a cleaner API.

BREAKING CHANGE: The shape of the `GuScheduledLambda` pattern has changed.

Yes. See breaking change note.

## How to test

See tests?

Ultimately, this is part of a stream of work to make defining an API Gateway backed lambda simpler. This will be the next PR, so success is if that PR makes sense?

n/a

### DIFF
--- a/src/patterns/scheduled-lambda.ts
+++ b/src/patterns/scheduled-lambda.ts
@@ -1,4 +1,6 @@
 import type { Schedule } from "@aws-cdk/aws-events";
+import { Rule } from "@aws-cdk/aws-events";
+import { LambdaFunction } from "@aws-cdk/aws-events-targets";
 import type { GuLambdaErrorPercentageMonitoringProps, NoMonitoring } from "../constructs/cloudwatch";
 import type { GuStack } from "../constructs/core";
 import { GuLambdaFunction } from "../constructs/lambda";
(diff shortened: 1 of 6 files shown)


src/patterns/scheduled-lambda.ts
```diff
@@ -48,8 +50,11 @@ import type { GuFunctionProps } from "../constructs/lambda";
  *  monitoringConfiguration: { noMonitoring: true }
  * ```
  */
-export interface GuScheduledLambdaProps extends Omit<GuFunctionProps, "rules" | "apis" | "errorPercentageMonitoring"> {
-  schedule: Schedule;
+export interface GuScheduledLambdaProps extends Omit<GuFunctionProps, "apis" | "errorPercentageMonitoring"> {
+  rules: Array<{
+    schedule: Schedule;
+    description: string;
```
jacobwinch, line 56: I would make this description optional. In many cases (e.g. the [example in the tests](https://github.com/guardian/cdk/pull/481/files#diff-8321919ed44468c832ea1017655c47c39ca6fe9943c8a84e56aab6e68e207988R19)), I don't find that providing a description adds any value.

jacobwinch (approved):

src/patterns/scheduled-lambda.ts
akash1810: Done.

PR #482, opened 2021-04-20
fix(guparameter): add a sensible default to GuParameters when using SSM

stephengeller:
## What does this change?

At the moment, there is no indication in the UI that the parameter you are filling in is an SSM parameter. Users may remember to add a sensible default or a hint in the description, but they also may not. 

This change ensures that, when not provided explicitly, users are hinted that the parameter is from SSM as well as what the ideal SSM path should be.

Nope.

## How to test

`./script/test`.

Users spend less time confused about how to handle CFN parameters in the UI.

I don't think there are any, as the default gets overridden if desired.

### DIFF
--- a/src/constructs/core/parameters/base.test.ts
+++ b/src/constructs/core/parameters/base.test.ts
@@ -25,6 +25,7 @@ describe("The GuParameter class", () => {
     const json = SynthUtils.toCloudFormation(stack) as SynthedStack;
 
     expect(json.Parameters.Parameter).toEqual({
+      Default: "/$STAGE/$STACK/$APP/parameter",
       Type: "AWS::SSM::Parameter::Value<Boolean>",
     });
   });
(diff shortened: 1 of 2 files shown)


akash1810:
This change has the risk of people seeing a populated field and blindly continuing.

What if we made the description for a `GuParameter` mandatory instead?

stephengeller:
> This change has the risk of people seeing a populated field and blindly continuing.
> 
> What if we made the description for a `GuParameter` mandatory instead?

This is true, although CFN UI does error out early enough if it can't find the SSM parameter: 
image

I think making the description mandatory could be more painful than necessary for users, but some logic that prepends the description with `"(From SSM)"` could be nice, eg:

### Before:

image

### After: 

image

It wouldn't be too complex code-wise, could even likely get away with something like:

```ts
{
   description: props.fromSSM ? `(From SSM) ${props.description || ""}` : props.description
}
```

stephengeller:
Bumping this

akash1810 (approved):
Should this be a fix (aka patch release) rather than chore? Otherwise LGTM 👍🏽 .

PR #483, opened 2021-04-20
feat: Add `GuApiLambda` pattern

akash1810:
## What does this change?

Refactor the props to `GuLambdaFunction` to create an explicit pattern for an API Gateway executed lambda.

BREAKING CHANGE: Define an API Gateway executed lambda with the explicit `GuLambdaFunction` pattern

Yes, see notes above.

## How to test

See tests.

The API to create a lambda is cleaner.

n/a

### DIFF
--- a/src/patterns/api-lambda.ts
+++ b/src/patterns/api-lambda.ts
@@ -10,10 +10,35 @@ interface ApiProps extends Omit<LambdaRestApiProps, "handler"> {
 }
 
 interface GuApiLambdaProps extends Omit<GuFunctionProps, "errorPercentageMonitoring"> {
+  /**
+   * A list of [[`LambdaRestApiProps`]] to configure for the lambda.
+   */
   apis: ApiProps[];
+
+  /**
+   * Configuration for an alarm based on the lambda's error percentage.
+   *
+   * For example:
+   *
+   * ```typescript
+   * monitoringConfiguration: {
+   *   toleratedErrorPercentage: <sensible_error_percentage_threshold>,
+   *   snsTopicName: "my-topic-for-cloudwatch-alerts",
+   * }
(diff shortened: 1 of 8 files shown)


jacobwinch (approved):
Great work!

I wonder if this pattern should create an ACM certificate for the API (and eventually a CNAME for the API Gateway instance) too? If we decide it should, I think we could add that in a follow-up PR.

src/patterns/api-lambda.ts
```diff
@@ -0,0 +1,31 @@
...
+interface ApiProps extends Omit<LambdaRestApiProps, "handler"> {
+  id: string;
+}
+
+interface GuApiLambdaProps extends Omit<GuFunctionProps, "errorPercentageMonitoring"> {
+  apis: ApiProps[];
+  monitoringConfiguration: NoMonitoring | GuLambdaErrorPercentageMonitoringProps;
+}
+
+export class GuApiLambda extends GuLambdaFunction {
```
jacobwinch, line 17: It's probably worth adding docs for this new pattern? Again, happy for that to follow in a separate PR if you'd prefer to get this merged.

akash1810:
> I wonder if this pattern should create an ACM certificate for the API (and eventually a CNAME for the API Gateway instance) too? If we decide it should, I think we could add that in a follow-up PR.

Great idea and I think it makes sense to be its own PR.

akash1810:
> I wonder if this pattern should create an ACM certificate for the API (and eventually a CNAME for the API Gateway instance) too? If we decide it should, I think we could add that in a follow-up PR.

Also thinking it would be good to add a CFN Output for API Gateway that results from this pattern to make DNS things a little easier. Will make that a small follow-up PR.

PR #485, opened 2021-04-23
feat: Add Wazuh security group to ASG by default

akash1810:
## What does this change?

Wazuh is required to be installed on an EC2 instance as part of our InfoSec program.

This security group allows the Wazuh agent to connect to the server.

See:
  - https://github.com/guardian/security-hq/blob/5e453f456b08763cabdfe5a0717346be8a24d556/hq/markdown/wazuh.md

It shouldn't do. Teams should have Wazuh running. This plus the fact that the Wazuh security group is a singleton should mean this change is a no-op for stacks.

## How to test

See updated tests.

More stuff comes for free 🎉 .

n/a

### DIFF
--- a/src/constructs/autoscaling/asg.test.ts
+++ b/src/constructs/autoscaling/asg.test.ts
@@ -187,6 +187,9 @@ describe("The GuAutoScalingGroup", () => {
           "Fn::GetAtt": [`GuHttpsEgressSecurityGroup${app}89CDDA4B`, "GroupId"],
         },
         {
+          "Fn::GetAtt": ["WazuhSecurityGroup", "GroupId"],
+        },
+        {
           "Fn::GetAtt": ["SecurityGroupTestingA32D34F9", "GroupId"], // auto-generated logicalId
         },
         {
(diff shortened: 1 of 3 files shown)


jorgeazevedo:
yay!! 🎉

jacobwinch (approved):
Looks good to me 👍 

IIUC, users should also get the required IAM permissions for Wazuh for free (via `GuInstanceRole`), unless they explicitly provide a different role _without_ these permissions?

akash1810:
> Looks good to me 👍
> 
> IIUC, users should also get the required IAM permissions for Wazuh for free (via `GuInstanceRole`), unless they explicitly provide a different role _without_ these permissions?

Yep 😄 . [These permissions](https://github.com/guardian/cdk/blob/112f34a4fb4fff0fd62669e1a825036331d4fec5/src/constructs/iam/policies/describe-ec2.ts#L15-L18) come for free:
- `autoscaling:DescribeAutoScalingInstances`
- `autoscaling:DescribeAutoScalingGroups`
- `ec2:DescribeTags`
- `ec2:DescribeInstances`

PR #486, opened 2021-04-23
feat: use one minute granularity for lambda error % alarms

jacobwinch:
## What does this change?
This PR updates our lambda-based alarms to use 1-minute periods (instead of 5-minute periods). It's the lambda version of https://github.com/guardian/cdk/pull/454!

Yes, some users will be required to update their alarm configuration:

```BREAKING CHANGE: GuLambdaErrorPercentageMonitoringProps no longer accepts `numberOfFiveMinutePeriodsToEvaluate`. `numberOfMinutesAboveThresholdBeforeAlarm` can be passed instead.```

## How to test
I've tested this out using a test stack in the Developer Playground.

Previously it would have taken ~5 minutes to detect a production incident, now it should take ~1 minute.

As with the EC2 equivalent, there is a risk that we will generate noisy alarms by using more sensitive defaults. If this turns out to be the case we should consider relaxing the defaults.


stephengeller (approved):
Looks sensible!

PR #494, opened 2021-04-26
docs: Add basic docs for GuAutoScalingGroup and GuInstanceRole

jacobwinch:
## What does this change?
Adds some simple docs which explain the permissions and rules that are granted by default when setting up an ASG. I'm sure there are more interesting things that we could mention in relation to these constructs, but I found this particularly tricky to figure out without reading the source code.

No.

## How to test
Run `./script/docs` and have a look at the changes?

It should be easier for users to understand which permissions/rules are granted by default (and why) without digging into the source code.

N/A

### DIFF
--- a/src/constructs/autoscaling/asg.ts
+++ b/src/constructs/autoscaling/asg.ts
@@ -80,6 +80,23 @@ function wireStageDependentProps(stack: GuStack, stageDependentProps: GuStageDep
   };
 }
 
+/**
+ * Construct which creates an Auto Scaling group.
+ *
+ * By default, all EC2 instances in this group will use [[`GuInstanceRole`]],
+ * which provides common permissions (e.g. the ability to download an artifact and write logs to our central ELK stack).
+ *
+ * If additional IAM permissions are required, a custom role can be provided via the `role` prop.
+ * You may wish to instantiate [[`GuInstanceRole`]] yourself as a basis for this custom role, as it allows custom permissions
+ * to be passed in.
+ *
+ * All EC2 instances in this group will be automatically associated with two security groups:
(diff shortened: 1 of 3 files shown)


src/constructs/autoscaling/asg.ts
```diff
@@ -80,6 +80,23 @@ function wireStageDependentProps(stack: GuStack, stageDependentProps: GuStageDep
   };
 }
 
+/**
+ * Construct which creates an Auto Scaling group.
+ *
+ * By default, all EC2 instances in this group will use [[`GuInstanceRole`]],
+ * which provides common permissions (e.g. the ability to download an artifact and write logs to our central ELK stack).
```
akash1810, line 87: minor: should it be "write logs to the account's logging kinesis stream"?

src/constructs/autoscaling/asg.ts
```diff
@@ -80,6 +80,23 @@ function wireStageDependentProps(stack: GuStack, stageDependentProps: GuStageDep
...
+ *
+ * If additional IAM permissions are required, a custom role can be provided via the `role` prop.
+ * You may wish to instantiate [[`GuInstanceRole`]] yourself as a basis for this custom role, as it allows custom permissions
```
akash1810, line 90: Not for this PR, but the ASG (or pattern) could take this prop and pass it down for ease.

akash1810 (approved):

src/constructs/autoscaling/asg.ts
jacobwinch: Yep, agreed; sometimes writing the docs is a good way to make yourself think "we could make this easier!" 😀

jacobwinch: cc @stephengeller, as I know you're currently looking at overriding defaults via the pattern

PR #495, opened 2021-04-26
docs: Add docs for ELB constructs

jacobwinch:
## What does this change?
This PR adds some docs for our ELB constructs.

No.

## How to test
Run `./script/docs` and have a look at the changes?

It should be harder for users to accidentally create these legacy resources (although I'm sure there's more we can do in this area). Users who must use these legacy resources should be able to understand the defaults (and how to override them) more easily than before.

N/A


stephengeller (approved):
Great work! Looks good and reads clearly to me.

PR #496, opened 2021-04-27
refactor: Improve `GuHttpsApplicationListener` and use it in `GuEc2App` pattern

jacobwinch:
## What does this change?

* Refactors `GuHttpsApplicationListener` so that it accepts a (mandatory) `GuCertificate`, instead of an optional ARN
* Uses the `GuHttpsApplicationListener` in our `GuEc2App` pattern
* Adds basic docs for all ALB-related constructs

Yes, this PR includes a [breaking change](https://github.com/guardian/cdk/commit/edb9b33672f312ec3b0aad96e41cf95321257148) to the `GuHttpsApplicationListener` construct. That said, [I don't think anyone is currently using this construct](https://github.com/search?q=org%3Aguardian+GuHttpsApplicationListener&type=code)!

## How to test

This should be covered by existing unit tests. Note that the CFN produced by the `GuEc2App` pattern does not change, the code is just simplified slightly.

* Pattern code should be easier to understand
* Users of the library are able to understand ALB constructs more easily
* The change to the `GuHttpsApplicationListener` construct improves type safety. It should also lead to fewer parameters being created and more resources being created by `@guardian/cdk`

I can't think of any significant risks associated with this change.

### DIFF
--- a/src/constructs/loadbalancing/alb/application-listener.test.ts
+++ b/src/constructs/loadbalancing/alb/application-listener.test.ts
@@ -1,12 +1,11 @@
 import "@aws-cdk/assert/jest";
 import "../../../utils/test/jest";
-import { SynthUtils } from "@aws-cdk/assert";
 import { Vpc } from "@aws-cdk/aws-ec2";
 import { ApplicationProtocol, ListenerAction } from "@aws-cdk/aws-elasticloadbalancingv2";
 import { Stack } from "@aws-cdk/core";
-import { RegexPattern } from "../../../constants";
-import type { SynthedStack } from "../../../utils/test";
+import { Stage } from "../../../constants";
 import { simpleGuStackForTesting } from "../../../utils/test";
+import { GuCertificate } from "../../acm";
 import type { GuStack } from "../../core";
 import type { AppIdentity } from "../../core/identity";
 import { GuApplicationListener, GuHttpsApplicationListener } from "./application-listener";
@@ -23,6 +22,18 @@ const app: AppIdentity = { app: "testing" };
 
 const getLoadBalancer = (stack: GuStack): GuApplicationLoadBalancer => {
   return new GuApplicationLoadBalancer(stack, "ApplicationLoadBalancer", { vpc, ...app });
+};
+
+const getCertificate = (stack: GuStack): GuCertificate => {
+  return new GuCertificate(stack, {
+    ...app,
+    [Stage.CODE]: {
+      domainName: "code-guardian.com",
+    },
+    [Stage.PROD]: {
+      domainName: "prod-guardian.com",
+    },
+  });
 };
 
 const getAppTargetGroup = (stack: GuStack): GuApplicationTargetGroup => {
@@ -132,6 +143,7 @@ describe("The GuHttpsApplicationListener class", () => {
 
     new GuHttpsApplicationListener(stack, "HttpsApplicationListener", {
       ...app,
+      certificate: getCertificate(stack),
       loadBalancer: getLoadBalancer(stack),
       targetGroup: getAppTargetGroup(stack),
     });
@@ -146,6 +158,7 @@ describe("The GuHttpsApplicationListener class", () => {
     const stack = simpleGuStackForTesting();
 
     new GuHttpsApplicationListener(stack, "ApplicationListener", {
+      certificate: getCertificate(stack),
       loadBalancer: getLoadBalancer(stack),
       targetGroup: getAppTargetGroup(stack),
       ...app,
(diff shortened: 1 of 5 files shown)


src/constructs/loadbalancing/alb/application-load-balancer.ts
```diff
@@ -7,6 +7,17 @@ import type { GuMigratingResource } from "../../core/migrating";
 
 interface GuApplicationLoadBalancerProps extends ApplicationLoadBalancerProps, AppIdentity, GuMigratingResource {}
 
+/**
+ * Construct which creates an Application Load Balancer.
+ *
+ * This construct should be used in conjunction with [[`GuApplicationListener`]] and [[`GuApplicationTargetGroup`]]
```
jacobwinch, line 13: I know the pattern provides a way to wire these 3 constructs together, but writing these docs made me wonder if we should provide a single construct which bundles them together for the common case (e.g. a EC2 App which only accepts traffic over HTTPS).

stephengeller (approved):
This looks good! Thanks for writing this up with all the documentation, it's all clear. I'm happy to approve this, but not 100% sure I understand why this change is needed. Was having the certificate be optionally passed in a problem for the construct?

src/constructs/loadbalancing/alb/application-load-balancer.ts
stephengeller: Yeah, that makes sense. It's tricky because I guess we have some layers of abstraction here:

1. Resource (CFN YAML)
2. CfnResource (CDK)
3. Construct (CDK)
4. GuConstruct (GuCDK)
5. GuPattern (GuCDK)

This feels like something in between 4 and 5, which is tricky but makes sense in this scenario. Our own GuConstructs can sometimes be a combination of multiple constructs, so I guess if we thought of the right naming/use-cases we could do it?

jacobwinch:
> why this change is needed

It's not strictly needed but I consider this to be an optimisation, here are a few more details:

* Currently you can omit the `certificate` prop. If you do this the construct will create a parameter for you, which means that you must pass the ARN in manually via the console. This tempts people to manage their certificates outside of version control (which is no longer necessary, due to `GuCertificate`) and makes the deployment process slower and more error prone. It also makes the code a little more complex (note that removing this option allows us to delete a test).
* If you provide the `certificate` prop, you can pass in any string and the code will compile. We had some code which performed checks and threw errors at synth time as a workaround for this. However, the feedback loop for this is slower than if we demand a proper type and the code is more complex (again, removing this and using a proper type allows us to delete a unit test).
* With the new approach users only have one option; they must create the certificate using`@guardian/cdk` (or inherit an existing one) and pass it in to the construct. If they do not do this the code will not compile. This allows us to simplify our code, enforce type safety and pushes users towards (what I consider to be!) best practice, in terms of creating and managing their certificates.

stephengeller:
Awesome, thanks for that explanation @jacobwinch ! Makes perfect sense, 🚢  it

src/constructs/loadbalancing/alb/application-listener.test.ts
```diff
@@ -165,69 +178,23 @@ describe("The GuHttpsApplicationListener class", () => {
...
-      ConstraintDescription: "Must be an ACM ARN resource",
-      Description: "The ARN of an ACM certificate for use on a load balancer",
-    });
-
     expect(stack).toHaveResource("AWS::ElasticLoadBalancingV2::Listener", {
       Certificates: [
         {
           CertificateArn: {
-            Ref: "TLSCertificateTesting",
+            Ref: "CertificateTesting28FCAC6D",
```
akash1810, line 194: Might be overkill, should we test the cert is added to the stack too?

src/patterns/ec2-app.ts
```diff
@@ -87,11 +87,11 @@ export class GuEc2App {
       port: props.applicationPort,
     });
 
-    new GuApplicationListener(scope, "Listener", {
+    new GuHttpsApplicationListener(scope, "Listener", {
```
akash1810, line 90: Should we have tests that cover this change?

akash1810 (approved):
Couple of minor points, but LGTM and thanks for the clear explanation of the change!

src/patterns/ec2-app.ts
jacobwinch: This is covered by the [snapshot test for this pattern](https://github.com/guardian/cdk/blob/314af33330c2445c4d8de779dbedd9ea6507bffe/src/patterns/ec2-app.test.ts#L9-L30). The resulting CFN is actually identical, the pattern code is just slightly simpler to read (IMO).

src/constructs/loadbalancing/alb/application-listener.test.ts
jacobwinch: I think this would probably be overkill, since adding the cert to the stack is not the responsibility of this construct. We [test cert creation](https://github.com/guardian/cdk/blob/314af33330c2445c4d8de779dbedd9ea6507bffe/src/constructs/acm/certificate.test.ts) as part of the tests for the `GuCertificate` construct, so I think it's probably cleaner to keep that separation of concerns here.

PR #497, opened 2021-04-27
chore: Add reminder about documentation to PR template

jacobwinch:
## What does this change?

Adds a reminder about documentation to our PR template.

No.

## How to test

Merge this PR?

It should be harder to forget about adding/editing the docs when making changes to this library.

Yes, there is a risk that my attempt at a friendly reminder will be off-putting to other contributors. Feel free to suggest different wording that could help here 😀


stephengeller:
Looks good! I wonder if at some point we could move towards using checkboxes if that makes more sense, eg:

- [ ] Documentation added
- [ ] Tested with `./script/test`
- [ ] Breaking changes

The downside of this approach is that for some bullet points (eg breaking changes) it would be better to have more detail than just a box tick, so this can wait till we have more checkbox-worthy things.

stephengeller (approved):

akash1810 (approved):

PR #500, opened 2021-04-27
feat(ec2-app-pattern): add support for different access types

stephengeller:
## What does this change?

Currently, we have no way of passing in whitelisted CIDR ranges to our EC2 pattern. This is something we'd like to be able to do as [some stacks such as AMIable require it](https://github.com/guardian/deploy-tools-platform/blob/ee1cfd98af1de71f0d2bb6700831827c7375942b/cdk/lib/amiable/amiable.ts#L81). 

Furthermore, the current logic of the pattern only accounts for 2 of the 3 possible application/load balancer access options:

1. Load balancer is open to the world (currently covered by `publicFacing: true`)
2. Load balancer is open to the world but IP-locked (not currently covered)
3. Load balancer is internal (covered by `publicFacing: false`)

This PR changes the `GuEc2App` props by replacing `publicFacing` with `access`. This new `access` field can be used in one of the three following ways:

### Usage

```ts
new GuEc2App(stack, {
      applicationPort: GuApplicationPorts.Node,
      app: app,
      // ONLY allows access to CIDR ranges provided
      access: { scope:  AccessScope.RESTRICTED, cidrRanges: [Peer.ipv4("192.168.1.1/32"), Peer.ipv4("8.8.8.8/32")] }, 
      certificateProps: getCertificateProps(),
      userData: "",
    });

new GuEc2App(stack, {
      applicationPort: GuApplicationPorts.Node,
      app: app,
      // Open to the world
      access: { scope:  AccessScope.PUBLIC}, 
      certificateProps: getCertificateProps(),
      userData: "",
    });

new GuEc2App(stack, {
      applicationPort: GuApplicationPorts.Node,
      app: app,
     // Errors, as this wouldn't make sense due to to the open-to-the-world CIDR range making any others pointless 
      access: { scope: AccessScope.RESTRICTED, cidrRanges: [Peer.ipv4("192.168.1.1/32"), Peer.ipv4("0.0.0.0/0")] }, 
      certificateProps: getCertificateProps(),
      userData: "",
    });
```

With this change, we can be more specific about the load balancer access types without adding too much noise to the API.

Nope.

Yes. Some comments have been added to document this usage, might need more though!

## How to test

`./script/test`. New tests have been added to ensure correct behaviour.

 - We can set correct access types for applications with ease.
 - We can restrict access on public load balancers 

No, but there may be some that further tests could dig out? Not sure.

### DIFF
--- a/src/patterns/ec2-app.test.ts
+++ b/src/patterns/ec2-app.test.ts
@@ -10,25 +10,27 @@ import { GuSecurityGroup } from "../constructs/ec2/security-groups";
 import { alphabeticalTags, simpleGuStackForTesting } from "../utils/test";
 import { GuApplicationPorts, GuEc2App, GuNodeApp, GuPlayApp } from "./ec2-app";
 
+const getCertificateProps = () => ({
+  [Stage.CODE]: {
+    domainName: "code-guardian.com",
+    hostedZoneId: "id123",
+  },
+  [Stage.PROD]: {
+    domainName: "prod-guardian.com",
+    hostedZoneId: "id124",
+  },
+});
+
 describe("the GuEC2App pattern", function () {
   it("should produce a functional EC2 app with minimal arguments", function () {
     const stack = simpleGuStackForTesting();
     new GuEc2App(stack, {
       applicationPort: GuApplicationPorts.Node,
       app: "test-gu-ec2-app",
-      access: { type: "RESTRICTED", cidrRanges: ["192.168.1.1/32", "8.8.8.8/32"] },
+      access: { type: "PUBLIC" },
       monitoringConfiguration: { noMonitoring: true },
       userData: "#!/bin/dev foobarbaz",
-      certificateProps: {
-        [Stage.CODE]: {
-          domainName: "code-guardian.com",
-          hostedZoneId: "id123",
-        },
-        [Stage.PROD]: {
-          domainName: "prod-guardian.com",
-          hostedZoneId: "id124",
-        },
-      },
+      certificateProps: getCertificateProps(),
     });
     expect(SynthUtils.toCloudFormation(stack)).toMatchSnapshot();
   });
@@ -40,16 +42,7 @@ describe("the GuEC2App pattern", function () {
       applicationPort: GuApplicationPorts.Node,
       app: app,
       access: { type: "PUBLIC" },
-      certificateProps: {
-        [Stage.CODE]: {
-          domainName: "code-guardian.com",
-          hostedZoneId: "id123",
-        },
-        [Stage.PROD]: {
-          domainName: "prod-guardian.com",
-          hostedZoneId: "id124",
-        },
-      },
+      certificateProps: getCertificateProps(),
       monitoringConfiguration: { noMonitoring: true },
       userData: {
         distributable: {
@@ -108,16 +101,7 @@ describe("the GuEC2App pattern", function () {
       applicationPort: GuApplicationPorts.Node,
       app: app,
       access: { type: "PUBLIC" },
-      certificateProps: {
-        [Stage.CODE]: {
-          domainName: "code-guardian.com",
-          hostedZoneId: "id123",
-        },
-        [Stage.PROD]: {
-          domainName: "prod-guardian.com",
-          hostedZoneId: "id124",
-        },
-      },
+      certificateProps: getCertificateProps(),
       monitoringConfiguration: {
         tolerated5xxPercentage: 5,
         snsTopicName: "test-topic",
@@ -127,29 +111,60 @@ describe("the GuEC2App pattern", function () {
     expect(stack).toHaveResource("AWS::CloudWatch::Alarm"); //The shape of the alarm is tested at construct level
   });
 
-  it("requires IP addresses to whitelist if set as a restricted access application", function () {
+  it("creates the appropriate ingress rules for a restricted access application", function () {
     const stack = simpleGuStackForTesting();
     const app = "test-gu-ec2-app";
     new GuEc2App(stack, {
       applicationPort: GuApplicationPorts.Node,
       app: app,
       access: { type: "RESTRICTED", cidrRanges: ["192.168.1.1/32", "8.8.8.8/32"] },
-      certificateProps: {
-        [Stage.CODE]: {
-          domainName: "code-guardian.com",
-          hostedZoneId: "id123",
-        },
-        [Stage.PROD]: {
-          domainName: "prod-guardian.com",
-          hostedZoneId: "id124",
-        },
-      },
+      certificateProps: getCertificateProps(),
       monitoringConfiguration: { noMonitoring: true },
       userData: "",
     });
 
-    expect(stack).toHaveResource("AWS::ElasticLoadBalancing::LoadBalancer", {
-      SecurityGroup: {},
+    expect(stack).toHaveResource("AWS::EC2::SecurityGroup", {
+      SecurityGroupIngress: [
+        {
+          CidrIp: "192.168.1.1/32",
+          Description: "0",
+          FromPort: 443,
+          IpProtocol: "tcp",
+          ToPort: 443,
+        },
+        {
+          CidrIp: "8.8.8.8/32",
+          Description: "1",
+          FromPort: 443,
+          IpProtocol: "tcp",
+          ToPort: 443,
+        },
+      ],
+    });
+  });
+
+  it("allows all connections if set to public", function () {
+    const stack = simpleGuStackForTesting();
+    const app = "test-gu-ec2-app";
+    new GuEc2App(stack, {
+      applicationPort: GuApplicationPorts.Node,
+      app: app,
+      access: { type: "PUBLIC" },
+      certificateProps: getCertificateProps(),
+      monitoringConfiguration: { noMonitoring: true },
+      userData: "",
+    });
+
+    expect(stack).toHaveResource("AWS::EC2::SecurityGroup", {
+      SecurityGroupIngress: [
+        {
+          CidrIp: "0.0.0.0/0",
+          Description: "Allow from anyone on port 443",
+          FromPort: 443,
+          IpProtocol: "tcp",
+          ToPort: 443,
+        },
+      ],
     });
   });
 
@@ -240,16 +255,7 @@ describe("the GuEC2App pattern", function () {
       access: { type: "PUBLIC" },
       monitoringConfiguration: { noMonitoring: true },
       userData: "#!/bin/dev foobarbaz",
-      certificateProps: {
-        [Stage.CODE]: {
-          domainName: "code-guardian.com",
-          hostedZoneId: "id123",
-        },
-        [Stage.PROD]: {
-          domainName: "prod-guardian.com",
-          hostedZoneId: "id124",
-        },
-      },
+      certificateProps: getCertificateProps(),
     });
 
     new GuEc2App(stack, {
@@ -258,16 +264,7 @@ describe("the GuEC2App pattern", function () {
       access: { type: "PUBLIC" },
       monitoringConfiguration: { noMonitoring: true },
       userData: "#!/bin/dev foobarbaz",
-      certificateProps: {
-        [Stage.CODE]: {
-          domainName: "code-guardian.com",
-          hostedZoneId: "id123",
-        },
-        [Stage.PROD]: {
-          domainName: "prod-guardian.com",
-          hostedZoneId: "id124",
-        },
-      },
+      certificateProps: getCertificateProps(),
     });
 
     expect(stack).toHaveResource("AWS::AutoScaling::AutoScalingGroup", {
--- a/src/patterns/ec2-app.test.ts
+++ b/src/patterns/ec2-app.test.ts
@@ -16,7 +16,7 @@ describe("the GuEC2App pattern", function () {
     new GuEc2App(stack, {
       applicationPort: GuApplicationPorts.Node,
       app: "test-gu-ec2-app",
-      publicFacing: false,
+      access: { type: "RESTRICTED", cidrRanges: ["192.168.1.1/32", "8.8.8.8/32"] },
       monitoringConfiguration: { noMonitoring: true },
       userData: "#!/bin/dev foobarbaz",
       certificateProps: {
@@ -39,7 +39,7 @@ describe("the GuEC2App pattern", function () {
     new GuEc2App(stack, {
       applicationPort: GuApplicationPorts.Node,
       app: app,
-      publicFacing: false,
+      access: { type: "PUBLIC" },
       certificateProps: {
         [Stage.CODE]: {
           domainName: "code-guardian.com",
@@ -107,7 +107,7 @@ describe("the GuEC2App pattern", function () {
     new GuEc2App(stack, {
       applicationPort: GuApplicationPorts.Node,
       app: app,
-      publicFacing: false,
+      access: { type: "PUBLIC" },
       certificateProps: {
         [Stage.CODE]: {
           domainName: "code-guardian.com",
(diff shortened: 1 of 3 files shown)


stephengeller:
Woohoo, PR #500! 🎉

src/patterns/ec2-app.ts
```diff
@@ -5,21 +5,48 @@ import type { GuCertificateProps } from "../constructs/acm";
...
+  type: typeof INTERNAL;
+}
+
+// Access whitelisted to the CIDR ranges passed in
+interface RestrictedAccess extends Access {
+  type: typeof RESTRICTED;
+  cidrRanges: string[];
+}
+
+type AppAccess = PublicAccess | InternalAccess | RestrictedAccess;
```
akash1810, line 45: Would an enum reduce the LOC here?

```typescript
enum AppAccess {
  PUBLIC = "PUBLIC",
  INTERNAL = "INTERNAL",
  RESTRICTED = "RESTRICTED",
}
```

akash1810: Oh wait, `RestrictedAccess` isn't just a string, it also has `cidrRanges`, ignore the above then!

src/patterns/ec2-app.ts
```diff
@@ -74,9 +102,17 @@ export class GuEc2App {
...
-        subnets: props.publicFacing ? GuVpc.subnetsfromParameter(scope, { type: SubnetType.PUBLIC }) : privateSubnets,
+        subnets: internetFacing ? GuVpc.subnetsfromParameter(scope, { type: SubnetType.PUBLIC, app }) : privateSubnets,
       },
+      ...(props.access.type === RESTRICTED && {
+        securityGroup: new GuSecurityGroup(scope, AppIdentity.suffixText({ app }, "SecurityGroup"), {
+          app,
+          vpc,
+          ingresses: transformToSecurityGroupAccessRule(Object.entries(props.access.cidrRanges), 443),
+        }),
+      }),
```
akash1810, line 115: Should this logic be pushed into the construct?

src/patterns/ec2-app.test.ts
```diff
@@ -123,42 +107,81 @@ describe("the GuEC2App pattern", function () {
     expect(stack).toHaveResource("AWS::CloudWatch::Alarm"); //The shape of the alarm is tested at construct level
   });
 
+  it("requires IP addresses to whitelist if set as a restricted access application", function () {
```
jacobwinch, line 110: Minor: I think `creates the appropriate ingress rules for a restricted access application` might be a more helpful test name.

src/patterns/ec2-app.ts
```diff
@@ -5,21 +5,48 @@ import type { GuCertificateProps } from "../constructs/acm";
...
+interface Access {
+  type: string;
+}
+
+// Everyone on the internet can access
+interface PublicAccess extends Access {
+  type: typeof PUBLIC;
+}
+
+// Access restricted to internal AWS networks
```
jacobwinch, line 34: As a developer setting up a new app, I'm unsure when I would want to use `INTERNAL` and when I'd want to use `RESTRICTED`, perhaps we can try to clarify that here?

src/patterns/ec2-app.ts
jacobwinch, line 45: Perhaps we should add a type which implements `RestrictedAccess` with CIDR ranges for all Guardian networks, as I guess this is the restricted access that most people would need?

src/patterns/ec2-app.ts
```diff
@@ -89,6 +125,7 @@ export class GuEc2App {
 
     new GuApplicationListener(scope, "Listener", {
       app,
+      open: props.access.type === PUBLIC, // this is true as default, which adds an ingress rule to allow all inbound traffic
```
jacobwinch, line 128: Nice catch!

src/patterns/ec2-app.test.ts
stephengeller: Nice, yeah that reads better.

src/patterns/ec2-app.ts
stephengeller: That sounds good, although that would need access to @guardian/private-infrastructure-config, which I believe would break CI as we don't have the appropriate SSH access set up in GitHub Actions, I _think_?

stephengeller:
After talks with @akash1810, @jacobwinch and @sihil , we concluded that this interface/API could be simplified into something like this:

```ts
interface Access {
  scope: "public" | "restricted",
  cidrRanges: [] // if left blank, default to ["0.0.0.0/0"] and assume PUBLIC access
}

// This is the default access for the patterns and constructs that accept this type
// Eg EC2 pattern, ALBs etc
const PUBLIC: Access = { 
 scope: "public",
 cidrRanges: ["0.0.0.0/0"]
}
```

A future PR could ensure that the `cidrRanges` param is of type `Peer.ipv4`, ensuring that valid CIDR ranges are passed in. We can also ensure that the ranges passed in make sense, e.g. if 0.0.0.0/0 is passed in, we should check and scream if other ranges are passed in (because it's already open to the world!)

stephengeller:
Comments and discussion points have been actioned, this is now ready for re-review @akash1810 @jacobwinch @sihil .

I've decided to maintain the initial approach of defining an interface for PUBLIC and RESTRICTED access, but implemented it so that not specifying an `access` type in the pattern props will result in an open-to-the-world application being created by default. 

Usage becomes the following (also shown in tests):

```ts
new GuEc2App(stack, {
      applicationPort: GuApplicationPorts.Node,
      app: app,
      // ONLY allows access to CIDR ranges provided
      access: { scope: "RESTRICTED", cidrRanges: [Peer.ipv4("192.168.1.1/32"), Peer.ipv4("8.8.8.8/32")] }, 
      certificateProps: getCertificateProps(),
      userData: "",
    });

new GuEc2App(stack, {
      applicationPort: GuApplicationPorts.Node,
      app: app,
      // Open to the world
      access: { scope: "PUBLIC"}, 
      certificateProps: getCertificateProps(),
      userData: "",
    });

new GuEc2App(stack, {
      applicationPort: GuApplicationPorts.Node,
      app: app,
     // Errors, as this wouldn't make sense due to to the open-to-the-world CIDR range making any others pointless 
      access: { scope: "RESTRICTED", cidrRanges: [Peer.ipv4("192.168.1.1/32"), Peer.ipv4("0.0.0.0/0")] }, 
      certificateProps: getCertificateProps(),
      userData: "",
    });
```

Would be good to get your thoughts on this, as I'm not sure this is what everyone had in mind!

src/patterns/ec2-app.ts
```diff
@@ -88,9 +136,16 @@ export class GuEc2App {
...
-      },
+      internetFacing: true,
+      vpcSubnets: { subnets: GuVpc.subnetsfromParameter(scope, { type: SubnetType.PUBLIC, app }) },
+      securityGroup: new GuSecurityGroup(scope, "SecurityGroup", {
+        app,
+        vpc,
+        ingresses:
+          props.access?.scope === RESTRICTED
+            ? restrictedCidrRanges(props.access.cidrRanges)
+            : [{ range: Peer.anyIpv4(), port: 443, description: "Allow from anyone on port 443" }],
```
akash1810, line 147: To keep things DRY, I think we can move the ternary up and reuse the [`GuPublicInternetAccessSecurityGroup` construct](https://github.com/guardian/cdk/blob/61b0c66af7c1b0992358738d8d0effaabd8f9812/src/constructs/ec2/security-groups/base.ts#L78-L87) for the public access scope here.

akash1810 (commented):
Looks good, let's add documentation to the pattern too to reduce surprise.

src/patterns/ec2-app.ts
```diff
@@ -9,17 +11,43 @@ import type { Gu5xxPercentageMonitoringProps, NoMonitoring } from "../constructs
...
+import { GuSecurityGroup, GuVpc, SubnetType } from "../constructs/ec2";
 import { GuGetPrivateConfigPolicy, GuInstanceRole } from "../constructs/iam";
 import {
   GuApplicationLoadBalancer,
   GuApplicationTargetGroup,
   GuHttpsApplicationListener,
 } from "../constructs/loadbalancing";
 
+const PUBLIC = "PUBLIC" as const;
+const RESTRICTED = "RESTRICTED" as const;
```
akash1810, line 23: Could we make this an enum or a const with multiple properties to enable the same method of interaction as [`Stage`](https://github.com/guardian/cdk/blob/7336acc94f541831c45d64468b78a53db52a1e8e/src/constants/stage.ts) or [`RegexPattern`](https://github.com/guardian/cdk/blob/7336acc94f541831c45d64468b78a53db52a1e8e/src/constants/regex-pattern.ts)? That is, prefer `AccessScope.PUBLIC` over `"PUBLIC"`.

akash1810:
Not necessarily for this PR, but I'm still trying to get the [`WithDefault`](https://www.typescriptlang.org/play?ssl=3&ssc=2&pln=1&pc=1#code/C4TwDgpgBA6glsAFgEQgMwIYFcA2wA8AKgHxQC8UA3gLABQUUyAogGICCAqgDKEBcUhOgF86dOADtgEAE6YAxtABC0gPYB3AM4yqdBuIwBbCPw3BpEgOYAaXVABuMjXBXj+4rAYBGM4aNpyXUyhldS1pfngkVExcAhDNGVIKGnpGVk4efhSGPUNjKAAiAEkmAptUhgdpJxd+ADZbEVomugDxIIsIYA4wtk7JfgAKT1UE6QB+fniwgEpyUlNzcQtyKGHRsKmN7QppmQA6ZnZuQjmyUmyoaS6saXEoAAMAEkoR0IP9IyEAehe3sf2VRq4iED18tFagWAUCwGAAjKtOt1ev1gIMZpD2tDYQAmRFdHoyPoQSSDShQT75ApyRCqIxleyOZyuKAIoQY2hAA) thing into the repository 😄 . [Here's](https://www.typescriptlang.org/play?#code/C4TwDgpgBA6glsAFgEQgMwIYFcA2wA8AKgHxQC8UA3gLABQUUyAogGICCAqgDKEBcUhOgF86dCADssAWyhsAxnIgBnJVwgA3CDip0GABQ4AhLgEkAwuSgAiA8fNWANLqgmAcoSYAlV2y6Wrbh7evlbCorRw4sAQAE6YirIKyko69FA4Glr88ooqapo4TmlycAAmMZ4Y4gDmEPxKwDGR1QDaALphtHRyAPbiDYm5SvzwSKiYuAQ5yaQUNGnM7Nx8qQwMGQXZSXmZOAB0tqZmRWtQJeWVNXVQLVYADHsPTwD0d1YdaUJQGCnTKp10SLROIYBKERAQPQxHpgFLzBhKHpSCAANQwOCw10kUgARrETt9tkoAPxbIYA2i9frAKBA2LidFQmHDAQQpmwyzwqCI5FojHXACMBNBQ34XPWuzJyXyWj2gS8Pi4BIY5wqVVq-FuAqej0ezwAzAAmd7OES0M3dPoDUrobB4dks8GQ6EcubOHmo9GY-hC5wi5JSlR7RacHidIA) a fun example of it.

src/patterns/ec2-app.ts
```diff
@@ -9,17 +11,43 @@ import type { Gu5xxPercentageMonitoringProps, NoMonitoring } from "../constructs
...
+
+export interface Access {
+  scope: string;
+}
+
+/*
+ * For when you want your application to be accessible to the world.
```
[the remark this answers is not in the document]
stephengeller: Good shout! Done.

src/patterns/ec2-app.ts
stephengeller: Nice, thanks. Didn't know about that one.

src/patterns/ec2-app.ts
stephengeller: Yep, think that's sensible.

stephengeller:
> Not necessarily for this PR, but I'm still trying to get the [`WithDefault`](https://www.typescriptlang.org/play?ssl=3&ssc=2&pln=1&pc=1#code/C4TwDgpgBA6glsAFgEQgMwIYFcA2wA8AKgHxQC8UA3gLABQUUyAogGICCAqgDKEBcUhOgF86dOADtgEAE6YAxtABC0gPYB3AM4yqdBuIwBbCPw3BpEgOYAaXVABuMjXBXj+4rAYBGM4aNpyXUyhldS1pfngkVExcAhDNGVIKGnpGVk4efhSGPUNjKAAiAEkmAptUhgdpJxd+ADZbEVomugDxIIsIYA4wtk7JfgAKT1UE6QB+fniwgEpyUlNzcQtyKGHRsKmN7QppmQA6ZnZuQjmyUmyoaS6saXEoAAMAEkoR0IP9IyEAehe3sf2VRq4iED18tFagWAUCwGAAjKtOt1ev1gIMZpD2tDYQAmRFdHoyPoQSSDShQT75ApyRCqIxleyOZyuKAIoQY2hAA) thing into the repository 😄 . [Here's](https://www.typescriptlang.org/play?#code/C4TwDgpgBA6glsAFgEQgMwIYFcA2wA8AKgHxQC8UA3gLABQUUyAogGICCAqgDKEBcUhOgF86dCADssAWyhsAxnIgBnJVwgA3CDip0GABQ4AhLgEkAwuSgAiA8fNWANLqgmAcoSYAlV2y6Wrbh7evlbCorRw4sAQAE6YirIKyko69FA4Glr88ooqapo4TmlycAAmMZ4Y4gDmEPxKwDGR1QDaALphtHRyAPbiDYm5SvzwSKiYuAQ5yaQUNGnM7Nx8qQwMGQXZSXmZOAB0tqZmRWtQJeWVNXVQLVYADHsPTwD0d1YdaUJQGCnTKp10SLROIYBKERAQPQxHpgFLzBhKHpSCAANQwOCw10kUgARrETt9tkoAPxbIYA2i9frAKBA2LidFQmHDAQQpmwyzwqCI5FojHXACMBNBQ34XPWuzJyXyWj2gS8Pi4BIY5wqVVq-FuAqej0ezwAzAAmd7OES0M3dPoDUrobB4dks8GQ6EcubOHmo9GY-hC5wi5JSlR7RacHidIA) a fun example of it.

This looks great! I think I'd prefer it to be merged as a separate PR if possible, as it would be good to see the logic in an isolated change-set. If/when this goes through it would make sense to me to do it straight after though.

src/patterns/ec2-app.ts
```diff
@@ -9,17 +11,45 @@ import type { Gu5xxPercentageMonitoringProps, NoMonitoring } from "../constructs
...
+  scope: AccessScope.RESTRICTED;
+  cidrRanges: IPeer[];
+}
+
+export type AppAccess = PublicAccess | RestrictedAccess;
+
 interface GuEc2AppProps extends AppIdentity {
   userData: GuUserDataProps | string;
-  publicFacing: boolean; // could also name it `internetFacing` to match GuApplicationLoadBalancer
+  access?: AppAccess; // If not specifying an access type, we can assume it to be open to the world
```
jacobwinch, line 52: Minor: personally I quite like the explicitness of a required prop for something like this.

src/patterns/ec2-app.ts
```diff
@@ -34,6 +64,24 @@ export enum GuApplicationPorts {
...
 }
 
+const OpenApplicationCidrError: Error = Error(`Your list of CIDR ranges includes 0.0.0.0/0, meaning all other ranges specified are unnecessary as
+your application is open to the world. Please either remove the open CIDR range or the other specified ones.`);
+
+function checkValidCidrRanges(access: RestrictedAccess) {
+  const openToWorld = access.cidrRanges.map((range) => range.uniqueId).includes("0.0.0.0/0");
+  const moreThanOneRange = access.cidrRanges.length > 1;
+
+  if (openToWorld && moreThanOneRange) throw OpenApplicationCidrError;
```
jacobwinch, line 74: I don't think we need to check if there is `moreThanOneRange` here. Specifying `RestrictedAccess` with `"0.0.0.0/0"` simply doesn't make sense so we should always throw an error in that scenario in my opinion.

src/patterns/ec2-app.ts
```diff
@@ -9,17 +11,45 @@ import type { Gu5xxPercentageMonitoringProps, NoMonitoring } from "../constructs
...
+  PUBLIC,
+  RESTRICTED,
+}
+
+export interface Access {
+  scope: AccessScope;
+}
+
+/*
+ * For when you want your application to be accessible to the world (0.0.0.0/0).
```
jacobwinch, line 32: These docs are very clear 👍

jacobwinch:
> Usage becomes the following (also shown in tests)

Thanks for explaining the new usage pattern. Would you mind updating the PR description and pattern docs in source code with this too (once we've settled on the final implementation!), just so that it's really accessible for anyone looking back at this change.

src/patterns/ec2-app.ts
stephengeller: Thinking about it, I think you're right. I've changed it to be mandatory for now, I think we can address sensible defaults (if any) in future PRs as it might require some more careful thought.

src/patterns/ec2-app.ts
```diff
@@ -34,6 +64,24 @@ export enum GuApplicationPorts {
   Play = 9000,
 }
 
+const OpenApplicationCidrError: Error = Error(`Your list of CIDR ranges includes 0.0.0.0/0, meaning all other ranges specified are unnecessary as
+your application is open to the world. Please either remove the open CIDR range or the other specified ones.`);
```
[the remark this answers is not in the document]
stephengeller: Good shout! That makes perfect sense.

src/patterns/ec2-app.ts
stephengeller: Correct, that's absolutely right! Good catch

stephengeller:
> > Usage becomes the following (also shown in tests)
> 
> Thanks for explaining the new usage pattern. Would you mind updating the PR description and pattern docs in source code with this too (once we've settled on the final implementation!), just so that it's really accessible for anyone looking back at this change.

Definitely! Good call. I've added the usage to the PR and above the class definition. The in-code docs will need to be buffed up a bit more later on, but this is a start at least while the pattern is still in development.

jacobwinch (approved):
👍

PR #501, opened 2021-04-28
chore: Dry run SemRel in CI to print next version info

akash1810:
It is sometimes hard to understand if a branch will trigger a release.

This change updates CI to use SemRel to analyse commits on the branch and print the next version, if any.


akash1810:
It's proving way too hard to get the branches config for semrel quite right. This looks related https://github.com/semantic-release/semantic-release/issues/1433.

Abandoning for now.

PR #506, opened 2021-04-28
feat: Output an ALB's DNS entry

akash1810:
## What does this change?

Creates a Cloudformation output for the DNS value of an ALB.

This is useful in two scenarios:
  1. When creating a new stack, as it saves you clicking around the AWS console
  2. An early warning that something stateful will change across versions of GuCDK as the snapshot will differ

No.

No.

## How to test

See tests?

Hopefully it makes spotting breaking changes to DNS easier.

n/a

### DIFF
--- a/src/constructs/cloudwatch/__snapshots__/ec2-alarms.test.ts.snap
+++ b/src/constructs/cloudwatch/__snapshots__/ec2-alarms.test.ts.snap
@@ -13,11 +13,8 @@ Object {
     },
   },
   "Outputs": Object {
-    "ApplicationLoadBalancerTestingApplicationLoadBalancerTestingdnsName79AE3FFF": Object {
+    "ApplicationLoadBalancerTestingDnsName": Object {
       "Description": "DNS entry for ApplicationLoadBalancerTesting",
-      "Export": Object {
-        "Name": "ApplicationLoadBalancerTestingDnsName",
-      },
       "Value": Object {
         "Fn::GetAtt": Array [
           "ApplicationLoadBalancerTesting172A253B",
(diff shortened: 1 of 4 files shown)


src/constructs/cloudwatch/__snapshots__/ec2-alarms.test.ts.snap
```diff
@@ -12,6 +12,20 @@ Object {
       },
     },
   },
+  "Outputs": Object {
+    "ApplicationLoadBalancerTestingApplicationLoadBalancerTestingdnsName79AE3FFF": Object {
+      "Description": "DNS entry for ApplicationLoadBalancerTesting",
+      "Export": Object {
```
jacobwinch, line 18: I'm not sure that we need to use [exports](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-stack-exports.html) to support either of the scenarios you've mentioned. Do you think this will be helpful for other use-cases?

jacobwinch (approved):
Great work 👍 

> Hopefully it makes spotting breaking changes to DNS easier.

Hopefully in the longer term, the CFN template will 'own' the DNS records too, and then it should become even easier to spot these problems. However this is a great (and simple!) interim step - good thinking!

src/constructs/cloudwatch/__snapshots__/ec2-alarms.test.ts.snap
akash1810: This was mainly because the logicalId is auto-generated, which made testing difficult.

I think we can safely override the logicalId for the output, not least because the auto-generated one is _really_ confusing now that I look at it! I've added a commit to do this.

jacobwinch (approved):

PR #507, opened 2021-04-29
fix: Support singletons across multiple stacks

akash1810:
## What does this change?

A singleton is something that will be instantiated once and once only.

Or rather, a singleton construct is one which will be instantiated once and once only in a `Stack`.

This change fixes a bug where a singleton could not be used within two `Stack`s that get instantiated using the same `App` instance.

Something like this was not possible as there would be a cyclic reference:

```typescript
class MyStack extends GuStack {
  constructor() {
    GuDistributionBucketParameter.getInstance(this);
  }
}

class MyOtherStack extends GuStack {
  constructor() {
    GuDistributionBucketParameter.getInstance(this);
  }
}

const app = new App();

const myStack = new MyStack(app, "MyStack", { stack: "deploy" });
const myOtherStack = new MyOtherStack(app, "MyOtherStack", { stack: "deploy" });

myStack.synth();
myOtherStack.synth();
```

We'd get an error like this:

> Error: 'MyStack' depends on 'MyOtherStack'. Adding this dependency (MyOtherStack -> MyStack/DistributionBucketName.Ref) would create a cyclic reference.

This change checks if a singleton exists on the `Stack`, rather than the `App` to fix this issue.

A test is added to the `GuDistributionBucketParameter` construct to demonstrate this.

Yes, only a version bump.

No.

## How to test

See added tests.

It is possible to share an `App` instance between `Stack`s. For example:

```typescript
const app = new App();

const prismStack = new PrismStack(app, "PrismPROD", { stack: "deploy" });
const amiableStack = new AmiableStack(app, "AmiablePROD", { stack: "deploy" });
```

Or more ambitiously (as more changes are needed to support this):

```typescript
const app = new App();

const prismCODE = new PrismStack(app, "PrismCODE", { stack: "deploy", stackName: "Prism-CODE" });
const prismPROD = new PrismStack(app, "PrismPROD", { stack: "deploy", stackName: "Prism-PROD" });
```

The change to the integration tests demonstrate this idea.

n/a

### DIFF
--- a/integration-test/bin/cdk.ts
+++ b/integration-test/bin/cdk.ts
@@ -3,4 +3,5 @@ import { App } from "@aws-cdk/core";
 import { IntegrationTestStack } from "../src/integration-test-stack";
 
 const app = new App();
-new IntegrationTestStack(app, "Cdk", { stack: "integration-test" });
+new IntegrationTestStack(app, "IntegrationTestStackCODE", { stack: "integration-test-CODE" });
+new IntegrationTestStack(app, "IntegrationTestStackPROD", { stack: "integration-test-PROD" });
--- a/src/utils/test/singleton.ts
+++ b/src/utils/test/singleton.ts
@@ -1,0 +1,41 @@
+import type { Stack } from "@aws-cdk/core/lib/stack";
+import type { GuStack } from "../../constructs/core";
+
+/**
+ * Check if an element exists in a [[`GuStack`]].
+ *
+ * Note: this explicitly tests presence in the Stack, not the App.
+ * This makes it is possible to create multiple Stacks in the same App;
+ * the singleton will new unique across the Stacks.
+ *
+ * ```typescript
+ * import { App } from "@aws-cdk/core";
+ *
+ * class SomeSingletonConstruct { }
+ *
+ * class MyStack extends GuStack {
+ *   constructor() {
+ *     // add SomeSingletonConstruct to stack
+ *   }
+ * }
+ *
+ * class MyOtherStack extends GuStack {
+ *   constructor() {
+ *     // add SomeSingletonConstruct to stack
+ *   }
+ * }
+ *
+ * const app = new App();
+ * const myStack = new MyStack(app, "MyStack", { stack: "deploy" });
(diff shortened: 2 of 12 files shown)


src/utils/test/singleton.ts
```diff
@@ -0,0 +1,41 @@
...
+ * ```
+ *
+ * @param stack an instance of [[``GuStack``]]
+ * @param maybeSingletonInstance the singleton instance
+ *
+ * @see https://github.com/aws/aws-cdk/blob/0ea4b19afd639541e5f1d7c1783032ee480c307e/packages/%40aws-cdk/core/lib/private/refs.ts#L47-L50
+ */
+export const isSingletonPresentInStack = (stack: GuStack, maybeSingletonInstance?: { stack: Stack }): boolean => {
+  // destructured `maybeSingletonInstance` to support `CfnElement`s (aka parameters) and `Resource`s, which do not share a type
+  return maybeSingletonInstance ? maybeSingletonInstance.stack.node === stack.node : false;
```
akash1810, line 40: This is the main change. From `instance.node.root === stack.node.root;` to `instance.stack.node === stack.node`. That is, comparing the `Stack` node, not the root `App` node.

integration-test/bin/cdk.ts
```diff
@@ -3,4 +3,5 @@ import { App } from "@aws-cdk/core";
 import { IntegrationTestStack } from "../src/integration-test-stack";
 
 const app = new App();
-new IntegrationTestStack(app, "Cdk", { stack: "integration-test" });
+new IntegrationTestStack(app, "IntegrationTestStackCODE", { stack: "integration-test-CODE" });
+new IntegrationTestStack(app, "IntegrationTestStackPROD", { stack: "integration-test-PROD" });
```
akash1810, line 7: We synthesise these stacks [in CI](https://github.com/guardian/cdk/blob/aeebaa8f60f5f3df08958686106be088903954a4/integration-test/package.json#L16), this step will exit non-zero on failure.

src/utils/test/singleton.ts
jacobwinch: Nice, thanks for moving this logic into it's own function!

jacobwinch (approved):

PR #521, opened 2021-05-04
feat: add runtime-specific versions of lambda construct (and example pattern)

jacobwinch:
## What does this change?
This PR adds runtime-specific versions of the `GuLambdaFunction` construct and the `GuScheduledLambda` pattern. I'm a little unsure about this approach, but it gives us [consistency with our EC2 App pattern](https://github.com/guardian/cdk/blob/a2a09bfd9b7fe5d05f49f3b917808551f0c84ee4/src/patterns/ec2-app.ts#L107-L123) and may help us to keep our infrastructure on newer AWS runtimes.

If others think this is a good idea, I'll make the same change for the other lambda-based patterns.

No, but existing projects can make use of the new constructs/patterns if they wish.

Yes, the docs have been updated accordingly.

## How to test
The new functionality being added here is minimal, I've added some simple snapshot tests to guard against silly mistakes which could be made whilst refactoring (e.g. using `latestJava` in the Node-based classes).

* Users have to specify one fewer prop when instantiating these constructs/patterns
* It should be easier for DevX to keep users on the latest runtimes

* If we introduce the concept of latest runtimes into this library, DevX need to take responsibility for noticing when AWS add support for a new runtime and making the corresponding change in our library. It's possible that we won't remember to do this regularly.
* It's possible that people will ignore my new docs, and will just use the runtime-agnostic construct/patterns directly. Perhaps we should make these `abstract`?


src/constructs/lambda/lambda.ts
```diff
@@ -71,3 +77,21 @@ export class GuLambdaFunction extends Function {
...
+export class GuJvmLambdaFunction extends GuLambdaFunction {
+  constructor(scope: GuStack, id: string, props: GuRuntimeSpecificFunctionProps) {
+    super(scope, id, { ...props, runtime: latestJava });
+  }
+}
+
+/**
+ * Creates a [[`GuLambdaFunction`]] with the latest Node runtime.
+ */
+export class GuNodeLambdaFunction extends GuLambdaFunction {
```
akash1810, line 93: AWS CDK has native [support for NodeJsFunctions](https://docs.aws.amazon.com/cdk/api/latest/docs/aws-lambda-nodejs-readme.html). A brief look at the docs suggests it has support for TS, which massively simplifies the build step!

Do we want to make use of that construct?

I think this is an example of "don't build your own constructs" that [this CDK Day talk](https://www.youtube.com/watch?v=1gxArqL_Eg4) describes.

stephengeller: I think I agree with Akash on the sentiment that we may want to consider creating fewer GuConstructs if we can get away with using AWS ones. I haven't watched the talk yet though, and also agree with the logic that the users should decide whether the constructs are fit for purpose or not, so not sure how strict we need to be with that.

So, I guess I wonder if the GuNodeLambdaFunction could just be a wrapper for the AWS CDK native construct, meaning we don't have to maintain it as much?

philmcmahon:
I'm not certain about `latestJava/Node` - they'd be useful on initial creation but for PROD stuff we'd probably then want to freeze the version so that people don't accidentally get upgraded on to an incompatible runtime

jacobwinch:
> we'd probably then want to freeze the version so that people don't accidentally get upgraded on to an incompatible runtime

I had planned to handle this by making updates to `latestXVersion` a breaking change, but perhaps this would still be confusing/problematic for users.

jacobwinch:
Closing this for now as it's not clear that this change has enough benefits for the user (i.e. a small amount of code saved) to make it worthwhile.

src/constructs/lambda/lambda.ts
jacobwinch: As the AWS construct seems to take a very different approach to building/deploying Node lambda code, I think it'd be worth understanding that in more detail (via a spike) before adding our own implementation of a Node lambda.

PR #530, opened 2021-05-05
feat: Log when any stateful resource is in stack

akash1810:
## What does this change?

A resource is a raw [CloudFormation item](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-template-resource-type-ref.html). A construct is CDK's L1 or L2 abstraction of a resource.

A stateful resource can be defined as something that holds state. This could be a database, a bucket, load balancer, message queue etc.

This change will, upon stack synthesis, walk the tree of resources and log a warning for all the stateful resources we have identified.

```console
The resource '<PATH IN TREE>' of type <TYPE> is considered stateful by @guardian/cdk. Care should be taken when updating this resource to avoid accidental replacement as this could lead to downtime.
```

This does mean we end up keeping a list of these resources, which is not ideal...

The [`GuStatefulMigratableConstruct` mixin](https://github.com/guardian/cdk/blob/74124e3bb0243b0e923d68f6ad24c363bd250b83/src/utils/mixin/migratable-construct-stateful.ts#L51) performs a similar role here, however that only operates against the constructs that exist in the library.

Ideally we'd be able to use [Stack Policies](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/protect-stack-resources.html) to protect these resources. However they are [not currently supported in CDK](https://github.com/aws/aws-cdk-rfcs/issues/72).

See:
  - https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_core.Construct.html#protected-prepare
  - https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/protect-stack-resources.html
  - https://github.com/aws/aws-cdk-rfcs/issues/72

No.

Yes and added.

## How to test

See added test.

We warn about updating _any_ stateful resource, even ones not covered by `GuStatefulMigratableConstruct`.

The risk is the `StatefulResourceTypes` is incomplete or becomes stale.

### DIFF
--- a/src/constants/stateful-resource-types.ts
+++ b/src/constants/stateful-resource-types.ts
@@ -1,0 +1,23 @@
+/**
+ * A list of resource types that should be considered stateful
+ * and care should be taken when updating them to ensure they
+ * are not accidentally replaced as this could lead to downtime.
+ *
+ * For example, if a load balancer is accidentally replaced,
+ * any CNAME DNS entry for it would now be invalid and downtime
+ * will be incurred for the TTL of the DNS entry.
+ *
+ * Currently, this list is used to generate warnings at synth time.
+ * Ideally we'd add a stack policy to stop the resource being deleted,
+ * however this isn't currently supported in CDK.
+ *
+ * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/protect-stack-resources.html
+ * @see https://github.com/aws/aws-cdk-rfcs/issues/72
+ */
+export const StatefulResourceTypes: string[] = [
+  "AWS::CertificateManager::Certificate",
+  "AWS::DynamoDB::Table",
+  "AWS::ElasticLoadBalancing::LoadBalancer",
+  "AWS::ElasticLoadBalancingV2::LoadBalancer",
+  "AWS::S3::Bucket",
+];
--- a/src/constructs/core/stack.ts
+++ b/src/constructs/core/stack.ts
@@ -1,7 +1,9 @@
-import type { App, StackProps } from "@aws-cdk/core";
-import { Stack, Tags } from "@aws-cdk/core";
+import type { App, IConstruct, StackProps } from "@aws-cdk/core";
+import { CfnResource, Stack, Tags } from "@aws-cdk/core";
 import { Stage } from "../../constants";
+import { StatefulResourceTypes } from "../../constants/stateful-resource-types";
 import { TrackingTag } from "../../constants/tracking-tag";
+import { Logger } from "../../utils/logger";
 import type { StackStageIdentity } from "./identity";
 import type { GuStageDependentValue } from "./mappings";
 import { GuStageMapping } from "./mappings";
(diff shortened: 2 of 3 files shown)


src/constants/stateful-resource-types.ts
```diff
@@ -0,0 +1,23 @@
...
+ * will be incurred for the TTL of the DNS entry.
+ *
+ * Currently, this list is used to generate warnings at synth time.
+ * Ideally we'd add a stack policy to stop the resource being deleted,
+ * however this isn't currently supported in CDK.
+ *
+ * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/protect-stack-resources.html
+ * @see https://github.com/aws/aws-cdk-rfcs/issues/72
+ */
+export const StatefulResourceTypes: string[] = [
```
akash1810, line 17: PR is in draft whilst we identify this list (and also discuss if this PR is a good idea, as I'm not entirely sold...).

src/constructs/core/stack.ts
```diff
@@ -120,4 +122,26 @@ export class GuStack extends Stack implements StackStageIdentity, GuMigratingSta
     this.addTag("Stack", this.stack);
     this.addTag("Stage", this.stage);
   }
+
+  protected prepare(): void {
```
akash1810, line 126: The [docs](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_core.Construct.html#protected-prepare) are a little scary, but we're only logging so should be ok?!

> Perform final modifications before synthesis.
>
> This method can be implemented by derived constructs in order to perform final changes before synthesis. prepare() will be called after child constructs have been prepared.
>
> This is an advanced framework feature. Only use this if you understand the implications.

src/constructs/core/stack.ts
```diff
@@ -120,4 +122,26 @@ export class GuStack extends Stack implements StackStageIdentity, GuMigratingSta
...
+    Ideally we'd add a stack policy to stop the resource being deleted,
+    however this isn't currently supported in CDK.
+
+    See:
+      - https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/protect-stack-resources.html
+      - https://github.com/aws/aws-cdk-rfcs/issues/72
+     */
+    this.node.findAll().forEach((construct: IConstruct) => {
+      if (CfnResource.isCfnResource(construct) && StatefulResourceTypes.includes(construct.cfnResourceType)) {
+        Logger.warn(
```
akash1810, line 141: We could call [`addDeletionOverride`](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_core.CfnResource.html#addwbrdeletionwbroverridepath) here, as we do for the [`GuAutoScalingGroup`](https://github.com/guardian/cdk/blob/31cea549bcf9a1a590bb217a3967c660c4084bf3/src/constructs/autoscaling/asg.ts#L143).

src/constants/stateful-resource-types.ts
jacobwinch: A few suggestions:

`AWS::SNS::Topic` (when used as an API for other services)
`AWS::SQS::Queue`
`AWS::Kinesis::Stream`
`AWS::RDS::DBInstance`

There must also be an API Gateway resource which should be included in this list (i.e. the resource which maintains the stable URL which is part of the CNAME DNS entry), but I'm struggling to figure out which resource  that is!

PR #532, opened 2021-05-05
chore: Use GitHub Action caching

akash1810:
## What does this change?

We're seeing builds fail sporadically with a stacktrace similar to this:

```console
npm ERR! code ENOENT
npm ERR! syscall open
npm ERR! path /home/runner/work/cdk/cdk/node_modules/npm/node_modules/spdx-expression-parse/package.json
npm ERR! errno -2
npm ERR! enoent ENOENT: no such file or directory, open '/home/runner/work/cdk/cdk/node_modules/npm/node_modules/spdx-expression-parse/package.json'
npm ERR! enoent This is related to npm not being able to find a file.
npm ERR! enoent

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/runner/.npm/_logs/2021-05-05T07_21_06_708Z-debug.log
Error: Process completed with exit code 254.
```

Although re-running the build is usually enough, it seems more frequent in recent days so we should fix it.

The NPM docs sort of hint towards a reason:

> npm is written to use resources efficiently on install, and part of this is that it tries to do as many things concurrently as is practical. Sometimes this results in race conditions and other synchronization issues.

This change adds the Cache Github Action to the build in an attempt to reduce the number of network requests as the cache will be used instead.

This change is more in hope than anything as it's not been possible to reliably reproduce the behaviour.

See:
  - https://docs.npmjs.com/common-errors#many-enoent--enotempty-errors-in-output
  - https://github.com/actions/cache/blob/main/examples.md#node---npm
  - https://github.com/npm/cli/issues/694

No.

No.

## How to test

CI should pass.

The builds are no less flaky than they currently are.

It may even result in a faster build!

n/a

### DIFF
--- a/.github/workflows/cd.yaml
+++ b/.github/workflows/cd.yaml
@@ -15,6 +15,12 @@ jobs:
         uses: actions/setup-node@v2.1.5
         with:
           node-version: ${{ matrix.node-version }}
+      - uses: actions/cache@v2
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-
       - name: CD
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


jacobwinch (approved):
Thanks for researching this - let's give it a try 👍

PR #533, opened 2021-05-06
docs: add warning about scheme property

jacobwinch:
## What does this change?
Adds some more docs about https://github.com/guardian/cdk/pull/510 to increase the visibility of this problem/risk.

 N/A

N/A

## How to test
N/A

Hopefully this will spare any more users from hitting this problem when migrating their stacks.

This PR probably decreases risk slightly!


akash1810 (approved):
👍🏽  great stuff!

PR #534, opened 2021-05-07
refactor: control incoming traffic via the listener and tighten egress rules

jacobwinch:
## What does this change?
In https://github.com/guardian/cdk/pull/500, we made some changes to the security group configuration of the EC2 App pattern. As part of this we have inadvertently added a [very permissive egress rule](https://github.com/guardian/cdk/blob/859ef3e6c3a0be0bab45ed4ee47ff1c684bcb278/src/patterns/__snapshots__/ec2-app.test.ts.snap#L604-L610) to our load balancer. We have also overridden some of the AWS CDK's default security group behaviour, which I now believe to be unnecessary.

Here's a [diff](https://github.com/guardian/cdk-playground/compare/jw-14.4.0-upgrade?expand=1) which shows the security group changes which occur when upgrading to `14.4.0` (from `14.2.0`).

Here's a much [smaller diff](https://github.com/guardian/cdk-playground/compare/jw-cdk-upgrade?expand=1) which shows the changes which occur when upgrading to use this branch (from `14.2.0`).

This shouldn't require changes to existing projects, as the pattern is still under development and (to my knowledge) no test projects have upgraded to the latest version yet.

No.

## How to test
I am hoping to spin up a new stack from scratch, which will allow us to test this change properly.

* Our security groups are less permissive
* AWS CDK takes care of more work for us
* Hopefully the pattern code is a little bit easier to reason about

Since this is predominantly a revert of some changes we recently introduced via https://github.com/guardian/cdk/pull/500, I think the risk is quite low.

### DIFF
--- a/src/patterns/ec2-app.test.ts
+++ b/src/patterns/ec2-app.test.ts
@@ -159,7 +159,7 @@ describe("the GuEC2App pattern", function () {
       SecurityGroupIngress: [
         {
           CidrIp: "0.0.0.0/0",
-          Description: "Allow all inbound traffic via HTTPS",
+          Description: "Allow from anyone on port 443",
           FromPort: 443,
           IpProtocol: "tcp",
           ToPort: 443,
@@ -221,18 +221,6 @@ describe("the GuEC2App pattern", function () {
       },
       monitoringConfiguration: { noMonitoring: true },
       userData: "UserData from pattern declaration",
-    });
-
-    expect(stack).toHaveResource("AWS::EC2::SecurityGroup", {
-      SecurityGroupIngress: [
-        {
-          CidrIp: "0.0.0.0/0",
-          Description: "Allow all inbound traffic via HTTPS",
-          FromPort: 443,
-          IpProtocol: "tcp",
-          ToPort: 443,
-        },
-      ],
     });
 
     const cfnLb = pattern.loadBalancer.node.defaultChild as CfnLoadBalancer;
--- a/src/patterns/ec2-app.test.ts
+++ b/src/patterns/ec2-app.test.ts
@@ -168,31 +168,6 @@ describe("the GuEC2App pattern", function () {
     });
   });
 
-  it("assume public application if no access specified", function () {
-    const stack = simpleGuStackForTesting();
-    const app = "test-gu-ec2-app";
-    new GuEc2App(stack, {
-      applicationPort: GuApplicationPorts.Node,
-      app,
-      access: { scope: AccessScope.PUBLIC },
-      certificateProps: getCertificateProps(),
-      monitoringConfiguration: { noMonitoring: true },
-      userData: "",
-    });
-
-    expect(stack).toHaveResource("AWS::EC2::SecurityGroup", {
-      SecurityGroupIngress: [
-        {
-          CidrIp: "0.0.0.0/0",
-          Description: "Allow all inbound traffic via HTTPS",
-          FromPort: 443,
-          IpProtocol: "tcp",
-          ToPort: 443,
-        },
-      ],
-    });
-  });
-
   it("errors if specifying open access as well as specific CIDR ranges", function () {
     const stack = simpleGuStackForTesting();
     const app = "test-gu-ec2-app";
--- a/src/patterns/ec2-app.ts
+++ b/src/patterns/ec2-app.ts
@@ -11,7 +11,7 @@ import type { Gu5xxPercentageMonitoringProps, NoMonitoring } from "../constructs
 import { Gu5xxPercentageAlarm } from "../constructs/cloudwatch";
 import type { GuStack } from "../constructs/core";
 import { AppIdentity } from "../constructs/core/identity";
-import { GuPublicInternetAccessSecurityGroup, GuSecurityGroup, GuVpc, SubnetType } from "../constructs/ec2";
+import { GuSecurityGroup, GuVpc, SubnetType } from "../constructs/ec2";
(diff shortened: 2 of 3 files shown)


src/patterns/ec2-app.ts
```diff
@@ -180,17 +180,6 @@ export class GuEc2App {
       vpc,
       internetFacing: true,
       vpcSubnets: { subnets: GuVpc.subnetsfromParameter(scope, { type: SubnetType.PUBLIC, app }) },
-      securityGroup:
```
jacobwinch, line 183: I think defining a custom security group here is unnecessary. By default, [AWS will configure a security group](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-elasticloadbalancingv2.ApplicationLoadBalancer.html#securitygroup) with rules which allow the load balancer to talk to instances (and vice versa) over the correct port only.

src/patterns/ec2-app.ts
```diff
@@ -180,17 +180,6 @@ export class GuEc2App {
...
-        props.access.scope === AccessScope.RESTRICTED
-          ? new GuSecurityGroup(scope, "SecurityGroup", {
```
jacobwinch, line 185: Both `GuSecurityGroup` and `GuPublicInternetAccessSecurityGroup` seem to add a very permissive egress rule by default:

```
"SecurityGroupEgress": Array [
  Object {
    "CidrIp": "0.0.0.0/0",
    "Description": "Allow all outbound traffic by default",
    "IpProtocol": "-1",
  },
],
```

In the case of the EC2 App Pattern, this allows the load balancer to communicate with targets which means that everything will still work on `14.4.0`. However, this rule also allows a lot of other traffic which is unnecessary. As far as I can tell, this happens because `allowAllOutbound` [defaults to true](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-ec2.SecurityGroup.html#allowalloutbound).

I think we should consider making the following changes in separate PRs:

1. Overriding this default as part of [`GuBaseSecurityGroup`](https://github.com/guardian/cdk/blob/859ef3e6c3a0be0bab45ed4ee47ff1c684bcb278/src/constructs/ec2/security-groups/base.ts#L40-L69).
2. Removing [`GuPublicInternetAccessSecurityGroup`](https://github.com/guardian/cdk/blob/859ef3e6c3a0be0bab45ed4ee47ff1c684bcb278/src/constructs/ec2/security-groups/base.ts#L78-L87) entirely, as the AWS CDK is capable of creating a practically identical security group by default (as shown in this PR). As far as I can see this construct is currently only used by `typerighter`.

src/patterns/ec2-app.ts
```diff
@@ -203,12 +192,24 @@ export class GuEc2App {
 
     const listener = new GuHttpsApplicationListener(scope, "Listener", {
       app,
-      open: props.access.scope === AccessScope.PUBLIC, // this is true as default, which adds an ingress rule to allow all inbound traffic
```
jacobwinch, line 206: This was confusing me because [`open`](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-elasticloadbalancingv2.ApplicationListener.html#open) is essentially doing the same thing as manually adding a `GuPublicInternetAccessSecurityGroup`.

src/patterns/ec2-app.ts
```diff
@@ -203,12 +192,24 @@ export class GuEc2App {
...
+      // When open=true, AWS will create a security group which allows all inbound traffic over HTTPS
+      open: props.access.scope === AccessScope.PUBLIC,
       loadBalancer: loadBalancer,
       certificate: certificate,
       targetGroup: targetGroup,
     });
 
+    // Since AWS won't create a security group automatically when open=false, we need to add our own
+    if (props.access.scope === AccessScope.RESTRICTED) {
+      listener.connections.addSecurityGroup(
```
jacobwinch, line 204: This is AWS's [recommended](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-elasticloadbalancingv2.ApplicationListener.html#open) way to set more restrictive rules:

> If you want to be more selective about who can access this load balancer, set this [open] to false and use the listener's connections object to selectively grant access to the listener.

jacobwinch: All of the associated unit tests still pass, so I guess it must work!

src/patterns/ec2-app.test.ts
```diff
@@ -248,18 +223,6 @@ describe("the GuEC2App pattern", function () {
       userData: "UserData from pattern declaration",
     });
 
-    expect(stack).toHaveResource("AWS::EC2::SecurityGroup", {
```
jacobwinch, line 251: This assertion feels unnecessary to me. This test is really checking that ingress rules can be overridden, which happens [here](https://github.com/guardian/cdk/pull/534/files#diff-39b32ce1aaa5782f5c37962a1acb30a4de4d7682adc29ea40b64bce8c8fe7d20L278-L288).

If we wanted to be really strict we could check that our custom ingress rule _doesn't_ yet exist at this point...

```
expect(stack).not.toHaveResource("AWS::EC2::SecurityGroup", {
  SecurityGroupIngress: [
    {
      CidrIp: "8.8.8.8/32",
      Description: "",
      FromPort: 1234,
      IpProtocol: "tcp",
      ToPort: 1234,
    },
  ],
});
```

...but I think this is overkill as the absurdity of this ingress rule means it's extremely unlikely that we'd ever add this by default!

PR #536, opened 2021-05-10
refactor: Create a Jest matcher for tag checking

akash1810:
## What does this change?

Create a Jest matcher to DRY out some of the tests.

By default, it will assert a resource has these tags:
  - Stack
  - Stage
  - gu:cdk:version

It also accepts a number of props:
  - appIdentity, to assert the App tag is present
  - additionalTags, for any other tags that should be present
  - resourceProperties, to check other properties in addition to tags

No.

No.

## How to test

- Tests should continue to pass
- Does the API for the matcher make sense? Does the comparison of other resource properties make it too overloaded?

DRYer codebase.

We're using a few things from the AWS library (principally `HaveResourceAssertion`). The API for these might change between versions. I'm not sure how likely this will be and if it does our CI step will fail, so we'll be able to resolve any issues before they land on `main`.


jacobwinch (approved):
Nice work, this looks really useful 👍

PR #537, opened 2021-05-10
feat: Add new tag for repository info

akash1810:
## What does this change?

Adds a new `gu:repo` tag to the GuStack and any resource added to it.

The purpose of this tag is to provide information on where to find  an
application.

The value of the tag comes from one of two places:
  1. CDK context
  2. Git config

If we cannot find the repository name from either, we log a message.

A Git URL can be in two main formats:
  1. https, for example `https://github.com/guardian/cdk.git`
  2. ssh, for example `git@github.com:guardian/cdk.git`

Use `git-url-parse` to handle the parsing of a Git URL and present it in the form `owner/name`. For example, `guardian/cdk` or `guardian/prism`.

This can be useful when using Prism to find lambdas running deprecated versions of Node, for example.

See:
  - https://docs.aws.amazon.com/cdk/latest/guide/context.html

Yes - they need to bump the version of @guardian/cdk in use and update their snapshot tests.

No.

## How to test

See build log - I've added a test for reading the value from CDK context and other tests have been updated.

It becomes easy to locate the repository of a running service.

n/a

### DIFF
--- a/src/constructs/core/stack.test.ts
+++ b/src/constructs/core/stack.test.ts
@@ -5,6 +5,8 @@ import "../../utils/test/jest";
 import { Role, ServicePrincipal } from "@aws-cdk/aws-iam";
 import { App } from "@aws-cdk/core";
 import { Stage, Stages } from "../../constants";
+import { ContextKeys } from "../../constants/context-keys";
+import { TagKeys } from "../../constants/tag-keys";
 import { simpleGuStackForTesting } from "../../utils/test";
 import type { SynthedStack } from "../../utils/test";
 import { GuParameter } from "./parameters";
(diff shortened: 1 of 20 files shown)


src/constructs/core/stack.ts
```diff
@@ -119,5 +124,30 @@ export class GuStack extends Stack implements StackStageIdentity, GuMigratingSta
...
+
+  /**
+   * Adds a tag to resources in the stack for the repository name.
+   * The value is retrieved in the following order:
+   *   1. From the context
+   *   2. From git config
+   *
+   * @private
+   */
+  private tryAddRepositoryTag(): void {
```
akash1810, line 139: This is the most interesting part of this PR.

src/constructs/core/stack.test.ts
```diff
@@ -41,6 +43,25 @@ describe("The GuStack construct", () => {
     expect(stack).toHaveGuTaggedResource("AWS::IAM::Role");
   });
 
+  it("should prefer context values for repository information", () => {
+    const stack = new GuStack(new App({ context: { [ContextKeys.REPOSITORY_URL]: "my-repository" } }), "Test", {
```
akash1810, line 47: This is atypical use of CDK context, as one would usually create a `cdk.json` file instead. We cannot do this here as we're building a library, rather than an application stack.

PR #540, opened 2021-05-10
fix(ec2-pattern): Ensure RESTRICTED applications have CIDR ingresses set to the load balancer security group

stephengeller:
## What does this change?

For a currently-unclear reason, using [`listener.connections.addSecurityGroup`](https://github.com/guardian/cdk/blob/d6be4df7e0fe95dc39eddf54b63ca7aef66bd819/src/patterns/ec2-app.ts#L204) when declaring `RESTRICTED` applications does not seem to attach the created security group to the listener or load balancer. The security group gets created, but doesn't seem to attach to anything. This results in `RESTRICTED` applications not being accessible by the CIDR ranges specified.

This PR changes the logic so that the load balancer security group is created with the intended ingresses _only if an application is `RESTRICTED`_, but creates its own security group if not, maintaining some of the benefits from #534. 

This may not be the best way to do this (and we may/should revert to using `connections.addSecurityGroup` once we get to the bottom of it), but this may be fine for now. Keen to hear people's thoughts though!

This also adds a second snapshot test for a `RESTRICTED` application (to help catch regressions) and updates a test to ensure we're expecting the load balancer's SG to have the right ingresses, not just any SG.

Nope.

Not that I can think of. I think the logic should be clear enough, but happy to add a comment above the change as the logic being removed had one!

## How to test

`./script/test`.

Stacks using this pattern can reliably declare CIDR ranges for restricted access.

None I can think of, but would be good to hear thoughts from those who reviewed #534 .

### DIFF
--- a/src/patterns/__snapshots__/ec2-app.test.ts.snap
+++ b/src/patterns/__snapshots__/ec2-app.test.ts.snap
@@ -1,5 +1,862 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`the GuEC2App pattern can produce a restricted EC2 app locked to specific CIDR ranges 1`] = `
+Object {
+  "Mappings": Object {
+    "stagemapping": Object {
+      "CODE": Object {
+        "domainName": "code-guardian.com",
+        "hostedZoneId": "id123",
+        "maxInstances": 2,
+        "minInstances": 1,
+      },
+      "PROD": Object {
+        "domainName": "prod-guardian.com",
+        "hostedZoneId": "id124",
+        "maxInstances": 6,
+        "minInstances": 3,
+      },
+    },
+  },
+  "Outputs": Object {
+    "LoadBalancerTestguec2appDnsName": Object {
+      "Description": "DNS entry for LoadBalancerTestguec2app",
+      "Value": Object {
+        "Fn::GetAtt": Array [
+          "LoadBalancerTestguec2appC77A055C",
+          "DNSName",
+        ],
+      },
+    },
+  },
+  "Parameters": Object {
+    "AMITestguec2app": Object {
+      "Description": "Amazon Machine Image ID for the app test-gu-ec2-app. Use this in conjunction with AMIgo to keep AMIs up to date.",
+      "Type": "AWS::EC2::Image::Id",
+    },
+    "DistributionBucketName": Object {
+      "Default": "/account/services/artifact.bucket",
+      "Description": "SSM parameter containing the S3 bucket name holding distribution artifacts",
+      "Type": "AWS::SSM::Parameter::Value<String>",
+    },
+    "InstanceTypeTestguec2app": Object {
+      "Default": "t3.small",
+      "Description": "EC2 Instance Type for the app test-gu-ec2-app",
+      "Type": "String",
+    },
+    "LoggingStreamName": Object {
+      "Default": "/account/services/logging.stream.name",
+      "Description": "SSM parameter containing the Name (not ARN) on the kinesis stream",
+      "Type": "AWS::SSM::Parameter::Value<String>",
+    },
+    "Stage": Object {
+      "AllowedValues": Array [
+        "CODE",
+        "PROD",
+      ],
+      "Default": "CODE",
+      "Description": "Stage name",
+      "Type": "String",
+    },
+    "testguec2appPrivateSubnets": Object {
+      "Default": "/account/vpc/primary/subnets/private",
+      "Description": "A list of private subnets",
+      "Type": "AWS::SSM::Parameter::Value<List<AWS::EC2::Subnet::Id>>",
+    },
+    "testguec2appPublicSubnets": Object {
+      "Default": "/account/vpc/primary/subnets/public",
+      "Description": "A list of public subnets",
+      "Type": "AWS::SSM::Parameter::Value<List<AWS::EC2::Subnet::Id>>",
+    },
+    "testguec2appVpcId": Object {
+      "Default": "/account/vpc/primary/id",
+      "Description": "Virtual Private Cloud to run EC2 instances within",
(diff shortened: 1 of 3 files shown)


src/patterns/ec2-app.test.ts
```diff
@@ -122,24 +136,36 @@ describe("the GuEC2App pattern", function () {
...
-      ],
-    });
+    const json = SynthUtils.toCloudFormation(stack) as SynthedStack;
+    const loadBalancerKey = Object.keys(json.Resources).filter(
+      (resource) => json.Resources[resource].Type === "AWS::ElasticLoadBalancingV2::LoadBalancer"
+    )[0];
+    const loadBalancer = json.Resources[loadBalancerKey];
+    const lbSecurityGroups = loadBalancer.Properties.SecurityGroups as [Record<string, never>];
+    const lbSecurityGroupID = lbSecurityGroups[0]["Fn::GetAtt"][0] as string;
+    const lbSecurityGroup = json.Resources[lbSecurityGroupID];
```
stephengeller, line 146: This is a bit of a hacky way to ensure the specific security group for the load balancer is the one we're asserting over, as the previous test failed to assert that the _specific_ security group was the one with the ingress rules.

src/patterns/ec2-app.ts
```diff
@@ -180,6 +180,13 @@ export class GuEc2App {
       vpc,
       internetFacing: true,
```
stephengeller, line 181: TODO: Add comment explaining why this is always true

src/patterns/ec2-app.ts
```diff
@@ -178,6 +178,8 @@ export class GuEc2App {
     const loadBalancer = new GuApplicationLoadBalancer(scope, "LoadBalancer", {
       app,
       vpc,
+      // This is always set to true, as it determines the load balancer scheme as `internet-facing` rather than internal.
+      // This does not result in public access to the load balancer in itself however, as that is handled by the listener's `open` prop
```
jacobwinch, line 182: 👍

src/patterns/ec2-app.ts
```diff
@@ -192,19 +194,20 @@ export class GuEc2App {
 
     const listener = new GuHttpsApplicationListener(scope, "Listener", {
       app,
+      loadBalancer,
+      certificate,
+      targetGroup,
       // When open=true, AWS will create a security group which allows all inbound traffic over HTTPS
```
jacobwinch, line 200: Correcting my own comment based on my research this morning!

```suggestion
      // When open=true, AWS will create an ingress rule which allows all inbound traffic over HTTPS
```

src/patterns/ec2-app.test.ts
```diff
@@ -122,24 +136,48 @@ describe("the GuEC2App pattern", function () {
...
+      // This will fail if we ever change the security group ID,
+      //if there's a cleaner way of identifying the security group we should do it
+      resource.toLowerCase().startsWith("restrictedingresssecuritygroup")
+    )[0];
+    const securityGroup = json.Resources[securityGroupKey];
+    const loadBalancerKey = Object.keys(json.Resources).filter(
+      (resource) => json.Resources[resource].Type === "AWS::ElasticLoadBalancingV2::LoadBalancer"
+    )[0];
+    const loadBalancer = json.Resources[loadBalancerKey];
+    const lbSecurityGroups = loadBalancer.Properties.SecurityGroups as Array<{
```
jacobwinch, line 150: I wonder if #536 could help here?

src/patterns/__snapshots__/ec2-app.test.ts.snap
```diff
@@ -1,5 +1,848 @@
...
+            "Ref": "InstanceRoleTestguec2appC325BE42",
+          },
+        ],
+      },
+      "Type": "AWS::IAM::Policy",
+    },
+    "RestrictedIngressSecurityGroupTestguec2appF4DE7574": Object {
+      "Properties": Object {
+        "GroupDescription": "Allow restricted ingress from CIDR ranges",
+        "SecurityGroupEgress": Array [
```
jacobwinch, line 670: I don't think we want [this permissive egress rule](https://github.com/guardian/cdk/pull/534#discussion_r628105168). I think we can avoid this by setting `allowAllOutbound` to `false` when creating the security group.

jacobwinch: (This is also my mistake from #534, sorry about that!)

src/patterns/ec2-app.ts
```diff
@@ -192,19 +194,21 @@ export class GuEc2App {
...
-      loadBalancer: loadBalancer,
-      certificate: certificate,
-      targetGroup: targetGroup,
     });
 
     // Since AWS won't create a security group automatically when open=false, we need to add our own
     if (props.access.scope === AccessScope.RESTRICTED) {
-      listener.connections.addSecurityGroup(
-        new GuSecurityGroup(scope, "SecurityGroup", {
+      loadBalancer.connections.addSecurityGroup(
```
jacobwinch, line 206: Minor: I think we can drop the `.connections` here, as `ApplicationLoadBalancer` has an `addSecurityGroup` [method](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-ec2.Connections.html#addwbrsecuritywbrgroupsecuritygroups).

stephengeller: Good catch - implemented this and tests & snapshots still pass so all good.

src/patterns/ec2-app.test.ts
stephengeller: We've agreed IRL that although this might help, it might not be super straightforward to implement and the current implementation successfully tests what we need it to, so we're going to leave it for now.

jacobwinch (approved):

PR #542, opened 2021-05-11
feat: allow EC2 App pattern users to customise role permissions

jacobwinch:
## What does this change?
This PR makes it simpler for users to customise the IAM role used by their EC2 App.

No, this PR adds a new, optional prop to the pattern.

Yes, docs have been added.

## How to test
I've added a unit test for this functionality. I've also tested this manually against [a `cdk-playground` branch](https://github.com/guardian/cdk-playground/tree/jw-testing) by:

1. Adding [a new route](https://github.com/guardian/cdk-playground/compare/jw-testing#diff-393bc34bcdc07a72485a3c629a4101a2baed157c9ce453ee1d606d0aa23cc2f1) which needs `sns:ListTopics` permissions to function correctly
2. Confirming that a 500 response is served when custom role configuration is not provided
3. [Adding the necessary permissions via the pattern](https://github.com/guardian/cdk-playground/commit/fb790747c73fa7425a56aacb461155e628aa93a8)
4. Confirming that a 200 response is now served

Users can pass custom IAM permissions to the role more easily.

I can't think of any risks associated with this PR.


src/constructs/iam/roles/instance-role.ts
```diff
@@ -11,11 +11,13 @@ import {
 } from "../policies";
 import { GuRole } from "./roles";
 
-interface GuInstanceRoleProps extends AppIdentity {
+export interface GuInstanceRoleProps {
   withoutLogShipping?: boolean;
   additionalPolicies?: GuPolicy[];
 }
 
+export type GuInstanceRolePropsWithApp = GuInstanceRoleProps & AppIdentity;
```
jacobwinch, line 19: Hopefully we can work on [this ticket](https://trello.com/c/QxhYoo4c/654-refactor-abstract-away-the-complexities-of-being-app-aware-construct) soon so that I can stop doing this 🙈

PR #547, opened 2021-05-11
feat(ec2-app): allow users to configure access logging

stephengeller:
## What does this change?

Many stacks [like to configure an access logging policy for their load balancers](https://github.com/search?p=2&q=org%3Aguardian+%22AccessLoggingPolicy%22&type=Code). This appears to be a fairly common trend among EC2 applications, and also has significant variability in terms of how it's configured. Although it would be great to enforce a sensible default (ie always logging to the same bucket and prefix), this is currently not the case across the Guardian and so it is worth providing an interface to users to configure it themselves.

This PR allows users to optionally configure access logging by specifying the bucket and prefix of where the logs are written.

Nope.

Yes, a comment has been added to explain usage, with an example provided.

## How to test

`./script/test`.

Users can specify access logging configuration if desired.

None that I can think of - the only risk is that we allow for more fragmentation across the department where different applications & teams have different preferences for using and defining access logging. However, it feels more important to have support for the feature for now, and potentially decide and enforce a best practice later down the line!

You are required to specify a region when using access logging - this has been reflected in our test, where we thread a new `env` prop through to our `simpleGuStackForTesting`. Documentation has been added to inform users of this.

### DIFF
--- a/src/patterns/ec2-app.ts
+++ b/src/patterns/ec2-app.ts
@@ -6,7 +6,6 @@ import { Duration } from "@aws-cdk/core";
 import { GuCertificate } from "../constructs/acm";
 import { GuAutoScalingGroup, GuUserData } from "../constructs/autoscaling";
 import { Gu5xxPercentageAlarm } from "../constructs/cloudwatch";
-import type { GuStack } from "../constructs/core";
 import { GuSSMParameter } from "../constructs/core";
 import { AppIdentity } from "../constructs/core/identity";
 import { GuSecurityGroup, GuVpc, SubnetType } from "../constructs/ec2";
@@ -66,8 +65,7 @@ export interface RestrictedAccess extends Access {
 export type AppAccess = PublicAccess | RestrictedAccess;
 
 export interface AccessLoggingProps {
-  // This provides a hint & incentive to users to use a recommended path
-  bucketSSMPath: string | "/account/services/access-logging/bucket";
+  enabled: boolean;
   prefix?: string;
 }
 
@@ -109,7 +107,8 @@ export interface AccessLoggingProps {
  * }
  * ```
  *
- * To enable access logging for your load balancer, you can specify the bucket and prefix to write the logs to.
+ * To enable access logging for your load balancer, you can specify the prefix to write the logs to.
+ * The S3 bucket used to hold these access logs must be specified in SSM at `/account/services/access-logging/bucket`
  * You must specify a region in your stack declaration if you are to use this prop, as specified here:
  * https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-elasticloadbalancingv2.ApplicationLoadBalancer.html#logwbraccesswbrlogsbucket-prefix
  * For example:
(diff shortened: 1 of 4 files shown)


akash1810:
> Although it would be great to enforce a sensible default (ie always logging to the same bucket and prefix), this is currently not the case across the Guardian and so it is worth providing an interface to users to configure it themselves.

Can we talk with teams to see how they use these logs? It the prefix is flexible we should be able to be opinionated here. Regarding bucket, could we get it from a standard location in parameter store similar to how we get the artifact bucket?

stephengeller:
Bumping this for review as it has now been tested against AMIable CODE and works well!

src/patterns/ec2-app.ts
```diff
@@ -48,6 +50,12 @@ interface RestrictedAccess extends Access {
 
 export type AppAccess = PublicAccess | RestrictedAccess;
 
+export interface AccessLoggingProps {
+  // This provides a hint & incentive to users to use a recommended path
+  bucketSSMPath: string | "/account/services/access-logging/bucket";
```
stephengeller, line 55: Since `GuSSMParameter` doesn't create a CFN parameter in the UI (and therefore means we can't set sensible defaults, what you write is what you get), this means we have a handy way of hinting that this is the preferred SSM path to use without enforcing it.

akash1810: Hmm, any reason to not enforce it?

stephengeller: I guess not, but if we enforce it, we risk users having less visibility on where the construct will look for the value in SSM. However, as long as we document it within the pattern, I suppose this could be okay? Users would either a) not have to worry about it anyways since the SSM parameter is created in the account, or b) click through to the prop definition, where this path is dictated.

jacobwinch: I would probably vote for enforcing this and adding it to the docs.

jacobwinch (approved):
I've left one opinion about this; feel free to merge as is if you disagree with it 👍

src/patterns/ec2-app.ts
akash1810: 👍🏽  to enforcing it. It'll be easier to make it flexible later once we have a use case, rather than restricting it later.

stephengeller: Yeah, good call both. Thinking about it, I agree that enforcing it makes sense. I've just pushed a change that enforces it as well as makes the prefix mandatory, as you'd otherwise have the following if you wanted to omit the prefix:

```typescript
{
// other props
  accessLogging: {}
}
```

Now, it looks like this:
```typescript
{
// other props
  accessLogging: { prefix: "some-prefix" }
}

{
// other props
  accessLogging: { prefix: false }
}
```

Would be keen to hear your thoughts on whether this is nice enough.

src/patterns/ec2-app.ts
```diff
@@ -48,6 +50,10 @@ interface RestrictedAccess extends Access {
 
 export type AppAccess = PublicAccess | RestrictedAccess;
 
+export interface AccessLoggingProps {
+  prefix: string | false;
```
jacobwinch, line 54: Now that we are using a single bucket per account I don't think `false` is going to be a sensible choice here (more generally, I am not super keen on odd union types like this). Perhaps we can provide a sensible default (`stack/stage/app`?) or just make the user give us a `string`?

stephengeller: Hmm, yeah, I think that's a sensible suggestion. To unblock ourselves, I'm going to add a `TODO` and merge this in so we can address this later.

stephengeller: On second thought, I think you're right in having the user give us a string. I've implemented this.

akash1810 (approved):

src/patterns/ec2-app.ts
akash1810: Agreed, using the presence of a string as a boolean proxy isn't ideal and it's better to be explicit, would something like this work:

```typescript
interface AccessLoggingProps {
  enabled: boolean;
  prefix?: string;
}
```

This has another benefit or making it easy to explicitly disable access logging, rather than relying on it being the default behaviour.

akash1810 (approved):

PR #548, opened 2021-05-11
chore!: remove `GuPublicInternetAccessSecurityGroup`

jacobwinch:
## What does this change?
`GuPublicInternetAccessSecurityGroup` has been removed, as AWS automatically creates a security group which
is almost identical to this by default when creating an Application Load Balancer.

Yes, this is a breaking change. As far as I can see it only affects [Typerighter](https://github.com/guardian/typerighter/blob/fa90ef260cd71e0f4fa1b893d7bba9b87ff828ef/cdk/lib/rule-manager/rule-manager.ts#L102-L107).

In the future, users should make use of the default security group that AWS automatically creates when creating an Application Load Balancer. However, moving from `GuPublicInternetAccessSecurityGroup`  to AWS's default requires a security group replacement, which is a disruptive change, so stacks which previously created an instance of `GuPublicInternetAccessSecurityGroup` should be able to retain the same custom security group, by passing the appropriate `id`, `ingresses` and `description` (see removed code in this PR) to `GuSecurityGroup` instead of using the `GuPublicInternetAccessSecurityGroup` helper.

No

## How to test
N/A

The codebase is simpler. Users won't be tempted to create custom security groups when AWS is capable of doing this for them.

Yes, there is a risk that this will disrupt the team who maintain Typerighter. I have tried to suggest a couple of ways forward in the PR description.

### DIFF
--- a/src/constructs/ec2/security-groups/base.test.ts
+++ b/src/constructs/ec2/security-groups/base.test.ts
@@ -3,7 +3,7 @@ import "../../../utils/test/jest";
 import { Peer, Port, Vpc } from "@aws-cdk/aws-ec2";
 import { Stack } from "@aws-cdk/core";
 import { simpleGuStackForTesting } from "../../../utils/test";
-import { GuHttpsEgressSecurityGroup, GuPublicInternetAccessSecurityGroup, GuSecurityGroup } from "./base";
+import { GuHttpsEgressSecurityGroup, GuSecurityGroup } from "./base";
 
 describe("The GuSecurityGroup class", () => {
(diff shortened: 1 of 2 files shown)


jacobwinch:
> Is it worth detailing the specific differences in the PR description (or a test?) to make any migration off `GuPublicInternetAccessSecurityGroup` easier?

Using the AWS default requires security group replacement and we only have one stack using this construct right now, so I'd probably advise them to use `GuSecurityGroup` instead for now _or_ wait and recreate/tear down the whole stack when moving to the pattern.

akash1810 (approved):
image

img

PR #558, opened 2021-05-13
fix(GuSSMParameter): Disable inline source mapping

stephengeller:
## What does this change?

Previously, our `GuSSMParameter` would break on usage due to the generated lambda.js file being long (over 4096 chars). This is due to the unnecessary source map comment at the end of the file. Having discussed IRL with @akash1810 , we feel that inline source mapping is unnecessary and generally against best practices, so this PR disables inline source mapping in favour of having dedicated map files.

Not that I can think of.

Not to this library, but we should update [recommendations](https://github.com/guardian/recommendations) to encourage the best practice of disabling inline source maps.

## How to test

`./script/test`.

We do not encounter issues with inline lambda code being too long due to inline source map comments! 

Not that I can think of, mainly due to the fact that I don't believe source maps are necessary for this repository.


akash1810:
> we should update recommendations

I think we can add a sentence here https://github.com/guardian/recommendations/blob/master/npm-packages.md#packaging

akash1810 (approved):

PR #560, opened 2021-05-18
docs: improve docs for EC2 app pattern

jacobwinch:
## What does this change?
Our EC2 app pattern seems to be working well now, so I've tidied up the docs and removed the warnings that we had in place during the early stages of development.

No.

## How to test
It'd be helpful if reviewers could run `./script/docs` and check for any broken links or obvious errors.

It should be easier for users to work with our pattern.

### DIFF
--- a/src/patterns/ec2-app.ts
+++ b/src/patterns/ec2-app.ts
@@ -29,19 +29,32 @@ export interface Access {
   scope: AccessScope;
 }
 
-/*
+/**
  * For when you want your application to be accessible to the world (0.0.0.0/0).
  * Your application load balancer will have a public IP address that can be reached by anyone,
  * so only use if you are aware and happy with the consequences!
- * */
-interface PublicAccess extends Access {
+ *
+ * Example usage:
+ * ```typescript
+ * { scope: AccessScope.PUBLIC }
+ * ```
+ */
+export interface PublicAccess extends Access {
   scope: AccessScope.PUBLIC;
 }
 
-/*
+/**
  * For when you want to restrict your application's access to a list of CIDR ranges.
- * */
-interface RestrictedAccess extends Access {
+ *
+ * Example usage:
+ * ```typescript
+ * {
+ *   scope: AccessScope.RESTRICTED,
+ *   cidrRanges: [Peer.ipv4("192.168.1.1/32"), Peer.ipv4("8.8.8.8/32")]
+ * }
+ * ```
+ */
+export interface RestrictedAccess extends Access {
   scope: AccessScope.RESTRICTED;
   cidrRanges: IPeer[];
 }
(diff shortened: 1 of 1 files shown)


src/patterns/ec2-app.ts
```diff
@@ -94,51 +125,108 @@ function restrictedCidrRanges(ranges: IPeer[]) {
...
  *
- * Usage:
+ * This pattern will grant your EC2 instances a number of commonly needed IAM permissions. For more information on
+ * this, see [[`GuInstanceRole`]]. To add additional permissions to your EC2 instances, see [[`GuEc2AppProps`]].
+ *
+ * This pattern will automatically create security groups/rules which allow for:
+ * 1. Incoming traffic over HTTPS (from the whole internet or from the specified CIDR ranges, depending on the
+ * [[`AppAccess`]] specified).
+ * 2. Communication between the load balancer and the EC2 instances over HTTP, via the specified application port.
+ * 3. Outbound traffic from your EC2 instances over HTTPs (to enable communication with third-party APIs).
```
akash1810, line 139: typo: `HTTPs` should be `HTTPS`

src/patterns/ec2-app.ts
```diff
@@ -94,51 +125,108 @@ function restrictedCidrRanges(ranges: IPeer[]) {
...
+ *     distributable: {
+ *       fileName: "app-name.deb",
+ *       executionStatement: `dpkg -i /app-name/app-name.deb`,
+ *     }
+ *   },
  * });
+ * ```
  *
- * // For public-facing applications
+ * Example usage for an app which is locked down to specific IP ranges:
```
akash1810, line 172: minor: should we use the same terminology as the interface? "locked down" -> "restricted", "IP ranges" -> "CIDR ranges"?

src/patterns/ec2-app.ts
```diff
@@ -94,51 +125,108 @@ function restrictedCidrRanges(ranges: IPeer[]) {
...
+ *     distributable: {
+ *       fileName: "app-name.deb",
+ *       executionStatement: `dpkg -i /app-name/app-name.deb`,
+ *     }
+ *   },
  * });
  * ```
- * */
+ *
+ * ### Customising the infrastructure generated by this pattern
```
akash1810, line 204: minor: I wonder if the escape hatch information should also live as an individual markdown file in the repo that we can link to in chat etc in future?

akash1810 (approved):
A couple of minor comments, but looks great overall!

src/patterns/ec2-app.ts
jacobwinch: I was wondering about this too. I guess we will need to add similar escape hatch behaviour for the lambda-based patterns in the future. Perhaps we can generalise the style and docs at that point?

PR #561, opened 2021-05-18
feat(ec2-app-pattern): allow users to optionally configure scaling policies per stage

stephengeller:
## What does this change?

At the moment, users of the GuEC2App have no way of configuring their scaling policies. This is something we can do [at the AutoScalingGroup construct level](https://github.com/guardian/cdk/blob/9b9a91bc03f507ba4d688024b4fc8059a617f9d0/src/constructs/autoscaling/asg.ts#L33), but have not wired up to the pattern.

This PR adds a `scaling` prop, allowing users to optionally specify minimum and maximum instance scaling policies for their applications. Like `GuAutoScalingGroup`, this prop only requires users to specify the minimum instance count and multiplies it by 2 unless otherwise specified.

None yet as it's an optional default, but this will further unlock AMIable's migration.

Docs have been added alongside the other props.

## How to test

`./script/test`.

Users can control their application scaling policies!

Not that I can think of.

### DIFF
--- a/src/patterns/ec2-app.test.ts
+++ b/src/patterns/ec2-app.test.ts
@@ -222,6 +222,38 @@ describe("the GuEC2App pattern", function () {
     ).toThrowError();
   });
 
+  it("can configure ASG scaling by stage if desired", function () {
+    const stack = simpleGuStackForTesting();
+    const app = "test-gu-ec2-app";
+    new GuEc2App(stack, {
+      applicationPort: GuApplicationPorts.Node,
+      access: { scope: AccessScope.PUBLIC },
+      app: app,
+      certificateProps: getCertificateProps(),
+      monitoringConfiguration: { noMonitoring: true },
+      userData: "",
+      scaling: {
+        CODE: { minimumInstances: 3 },
+        PROD: { minimumInstances: 5, maximumInstances: 12 },
+      },
+    });
+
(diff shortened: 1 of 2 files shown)


src/patterns/ec2-app.test.ts
```diff
@@ -222,6 +222,43 @@ describe("the GuEC2App pattern", function () {
...
+      certificateProps: getCertificateProps(),
+      monitoringConfiguration: { noMonitoring: true },
+      userData: "",
+      scaling: {
+        CODE: { minimumInstances: 3 },
+        PROD: { minimumInstances: 5, maximumInstances: 12 },
+      },
+    });
+
+    expect(stack).toHaveResource("AWS::AutoScaling::AutoScalingGroup", {
```
jacobwinch, line 241: Personally I wouldn't bother with this assertion here, as it feels like a responsibility of the construct, rather than the pattern (if I understand it correctly, this `FindInMap` stuff gets added regardless of whether the user overrides the defaults or not).

jacobwinch (approved):

src/patterns/ec2-app.test.ts
stephengeller: Yeah, good shout. I'll get rid of it as it was already copied from the `asg` tests.

PR #576, opened 2021-05-19
feat!: only remove ALB Type when necessary

jacobwinch:
## What does this change?

Whilst reviewing our constructs, I noticed that we were manipulating CFN to remove an ALB property (which `aws/cdk` adds automatically), but I wasn't sure why. After testing this, I can see that if the `Type` value is undefined in CFN, updating it to add the `application` value [triggers a resource replacement](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticloadbalancingv2-loadbalancer.html#cfn-elasticloadbalancingv2-loadbalancer-type), _even if_ the load balancer already looks like this:

image

This is very similar to the `Scheme` problem (https://github.com/guardian/cdk/pull/510) that we ran into for ELBs.

Clearly we need to support removing this property for migrations where the existing ALB must be retained, but for new stacks it's probably cleaner to use AWS's default behaviour, which adds `Type` explicitly. I don't know for sure if _removing_ the `Type` property from a stack which currently specifies `Type: application` also triggers replacement, but I suspect it might, so this PR also allows migrating stacks to retain this value if they want to.

BREAKING CHANGE: Any stack which uses a `GuApplicationLoadBalancer` should now set the `removeType` prop to `true` in order to avoid resource replacement.

We introduced the original behaviour in [PR 2](https://github.com/guardian/cdk/pull/2/files#diff-bdf93570f58af30de6a0b5b2652984897f888ffaf06474ec2389724572254e42R30), so I think it's OK to assume that all users of this construct are affected!

I've updated the docs in this PR.

## How to test

I've added a unit test to cover this scenario.

New infrastructure follows AWS defaults, but it's still possible to retain existing resources if necessary.

I have tried to mitigate this by adding docs. More generally, the more cases like this we encounter, the more I think that creating parallel stacks (and then replacing the old ones) is the safest option!


jacobwinch:
> BREAKING CHANGE: Any stack which uses a `GuApplicationLoadBalancer` should now set the `removeType` prop to `true` in order to avoid resource replacement.

Note: it's not currently possible to do this via the pattern (unless you use an escape hatch to achieve the same thing). I think we're currently still at a point where we can tear down / recreate these stacks, so I'm wondering if it's better to do that, rather than adding in more complexity to the API. Any thoughts are welcome!

akash1810 (approved):
The migration tooling is going to be complicated, isn't it?! 😭 

This change looks great though!

PR #580, opened 2021-05-20
fix(ec2-app): allow users to specify scaling for only one stage

stephengeller:
## What does this change?

When testing the pattern out, it struck me that part of the value of the `scaling` prop is to specify your `PROD` scaling exclusively, as you would often be happy with the default `CODE` scaling of 1 min & 2 max. At the moment, our pattern forces you to specify both CODE and PROD if you use the `scaling` prop, when you may often only want to control one environment. 

This PR makes both stages optional, so you can only specify one stage and benefit from the defaults for the other.

Nope.

Nope, as I think it's still fairly clear with the current docs.

## How to test

`./script/test`.

Users can optionally specify just one stage for scaling policies.

### DIFF
--- a/src/patterns/ec2-app.ts
+++ b/src/patterns/ec2-app.ts
@@ -141,8 +141,8 @@ export interface GuEc2AppProps extends AppIdentity {
   roleConfiguration?: GuInstanceRoleProps;
   monitoringConfiguration: NoMonitoring | Gu5xxPercentageMonitoringProps;
   scaling?: {
-    [Stage.CODE]: GuAsgCapacityProps;
-    [Stage.PROD]: GuAsgCapacityProps;
+    [Stage.CODE]?: GuAsgCapacityProps;
+    [Stage.PROD]?: GuAsgCapacityProps;
   };
   accessLogging?: AccessLoggingProps;
 }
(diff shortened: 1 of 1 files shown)


akash1810:
Is this a `fix` rather than `refactor`?

stephengeller:
> Is this a `fix` rather than `refactor`?

I mean, it's sort of both or neither. Nothing's _broken_ per se, but it's also changing functionality a tiny bit so it's not quite a refactor, so not sure. Happy to use either.

akash1810:
> > Is this a `fix` rather than `refactor`?
> 
> I mean, it's sort of both or neither. Nothing's _broken_ per se, but it's also changing functionality a tiny bit so it's not quite a refactor, so not sure. Happy to use either.

I'd suggest `fix` so it goes out as it's own patch release.

stephengeller:
> > > Is this a `fix` rather than `refactor`?
> > 
> > 
> > I mean, it's sort of both or neither. Nothing's _broken_ per se, but it's also changing functionality a tiny bit so it's not quite a refactor, so not sure. Happy to use either.
> 
> I'd suggest `fix` so it goes out as it's own patch release.

Sounds good, this has been updated to be a `fix`.

jacobwinch (approved):

PR #581, opened 2021-05-20
fix(ec2-app): use a traditional GuStringParameter for access logging

stephengeller:
## What does this change?

As discussed in chat, we are currently experiencing issues with `GuSSMParameter`'s implementation not playing nicely with snapshot tests, primarily due to `GuSSMParameter` forcing a new ID for its resources at every synth which then makes snapshot tests fail due to a mismatch in IDs.

As a temporary stop-gap while we fix this problem, this PR resorts to a "traditional" CFN string parameter with the default value being the same as the value fetched by `GuSSMParameter`. 

Nope.

Nope, don't think so.

## How to test

`./script/test`.

Snapshot tests still pass while users configure access logging.

### DIFF
--- a/src/patterns/ec2-app.test.ts
+++ b/src/patterns/ec2-app.test.ts
@@ -441,28 +441,14 @@ describe("the GuEC2App pattern", function () {
       accessLogging: { enabled: true, prefix: "access-logging-prefix" },
     });
 
-    const json = SynthUtils.toCloudFormation(stack) as SynthedStack;
-    const lbKey = Object.keys(json.Resources).find(
-      (resource) => json.Resources[resource].Type === "AWS::ElasticLoadBalancingV2::LoadBalancer"
-    );
-    expect(lbKey).toBeTruthy();
-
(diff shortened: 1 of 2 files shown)


src/patterns/ec2-app.test.ts
```diff
@@ -441,28 +441,14 @@ describe("the GuEC2App pattern", function () {
...
+      LoadBalancerAttributes: [
         { Key: "deletion_protection.enabled", Value: "true" },
         { Key: "access_logs.s3.enabled", Value: "true" },
-        {
-          Key: "access_logs.s3.bucket",
-          Value: {
-            "Fn::GetAtt": [expect.stringContaining("GuSSMParameter"), "Parameter.Value"],
-          },
-        },
+        { Key: "access_logs.s3.bucket", Value: { Ref: "AccessLoggingBucket" } },
```
stephengeller, line 448: Fortunately, this allows the test to be much cleaner!

jacobwinch (approved):

PR #585, opened 2021-05-24
chore: Use custom `tsconfig` for `ts-jest`

akash1810:
## What does this change?

By default, `ts-jest` will use [`tsconfig.json`](https://github.com/guardian/cdk/blob/6eb791548e727e9115991fbd53fa0931a4665c6f/tsconfig.json) for TypeScript configuration.

Our `tsconfig.json` sets [`noUnusedLocals`](https://www.typescriptlang.org/tsconfig#noUnusedLocals) to `true`. This means compilation will fail if we import something and don't use it.

If you're running tests locally (using [`script/start`](https://github.com/guardian/cdk/blob/6eb791548e727e9115991fbd53fa0931a4665c6f/script/start), for example) and comment out a block of code that results in an import not being used, a compile error will result and the tests will not run until resolved.

In reality you are going to use it, so you end up temporarily commenting out the import (once you've expanded them as the IDE collapses them 😅).

This whole process is slow.

In this change we instruct `ts-jest` to use a custom configuration, specifically allowing unused locals. This results in `jest` no longer complaining about unused imports and a faster feedback loop.

Unused locals will continue to get caught in the compile and lint steps, which are [run in CI](https://github.com/guardian/cdk/blob/6eb791548e727e9115991fbd53fa0931a4665c6f/script/ci#L24-L25). That is, this change simply improves the experience when running the tests.

See:
  - https://www.typescriptlang.org/tsconfig#noUnusedLocals
  - https://huafu.github.io/ts-jest/user/config/tsConfig

No.

No.

## How to test

- Checkout branch
- Run the tests in watch mode (`./script/start`)
- Add a random import to a file (`import fs from "fs";`)
- Witness the tests running
- Run the compile or lint step
- Witness this correctly failing due to the unused value

To confirm the other compiler options are still observed when running tests (i.e. this change is applied on top of `tsconfig.json`), we can add a random function to a file to test [`noUnusedParameters`](https://www.typescriptlang.org/tsconfig#noUnusedLocals):

```typescript
// error TS6133: 'modelID' is declared but its value is never read.
const createDefaultKeyboard = (modelID: number) => {
  const defaultModelID = 23;
  return { type: "keyboard", modelID: defaultModelID };
};
createDefaultKeyboard(1);
``` 

A faster feedback loop when running tests locally.

n/a

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -16,4 +16,9 @@ module.exports = {
     - #448
    */
   modulePathIgnorePatterns: ["<rootDir>/lib"],
+  globals: {
+    "ts-jest": {
+      tsconfig: "tsconfig.test.json",
+    },
+  },
 };
--- a/tsconfig.test.json
+++ b/tsconfig.test.json
@@ -1,0 +1,6 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "noUnusedLocals": false
+  }
+}


jacobwinch (approved):
💯

PR #588, opened 2021-05-26
refactor: Simplify the SSM custom resource

akash1810:
## What does this change?

We currently have quite a complicated setup in place to make a few tests work, namely we've checked in an empty `lambda.js` file and a symlink.

This change takes a different approach by using a mock in the tests.

No.

No.

## How to test

CI should continue to pass.

This change is part of a wider set of changes to introduce a CLI to GuCDK that checks if your AWS account is setup properly by checking [SSM parameters'](https://github.com/guardian/cdk/blob/4b007724c637bdf7db3a18574c8db18691fe4d5c/docs/005-default-parameter-store-locations.md) existence. Part of this process is to change the [tsconfig file](https://github.com/guardian/cdk/blob/4b007724c637bdf7db3a18574c8db18691fe4d5c/tsconfig.json#L19), outputting inline rather than in a separate `lib` directory. Once we do that `tsc` will overwrite `lambda.js` with actual JS code and we have to start dancing around with `.gitignore`/`.gitkeep` files in order to keep the empty `lambda.js` in VCS. It's simpler to remove it from VCS.

n/a

### DIFF
--- a/src/constructs/core/custom-resources/runtime/lambda.js
+++ b/src/constructs/core/custom-resources/runtime/lambda.js
@@ -1,4 +1,0 @@
-/*
-* The reason for this file is to appease the tests in `core/ssm.test.ts`.
-* We don't need this to contain anything useful because we're not testing the contents of it within that testing suite.
-* */
--- a/src/constructs/core/custom-resources/runtime/lambda.symlink.ts
+++ b/src/constructs/core/custom-resources/runtime/lambda.symlink.ts
@@ -1,1 +1,0 @@
-../runtime/lambda.ts
--- a/src/constructs/core/custom-resources/runtime/lambda.test.ts
+++ b/src/constructs/core/custom-resources/runtime/lambda.test.ts
@@ -1,10 +1,4 @@
-/*
- * You can't specify the file extension (eg .ts) of the file to import,
- * and lambda.js exists as a dummy file, so we'd always import that over lambda.ts (the actual code to be tested) if
- * we imported `lambda`
- * lambda.symlink is a symlink to lambda.ts, so we can test it without importing lambda.js
- * */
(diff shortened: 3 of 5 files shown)


jacobwinch (approved):
I'm not sure that I fully understand the original solution, but the mock approach looks fine to me 👍

akash1810:
> I'm not sure that I fully understand the original solution

Apologies, the PR description should have described this.

In it's constructor, `GuSSMParameter` [reads `lambda.js`](https://github.com/guardian/cdk/blob/f5c9104d30e0042ce3057c776a5dbd7320434750/src/constructs/core/ssm.ts#L39) from disk. This file doesn't exist when `ts-jest` executes the [tests](https://github.com/guardian/cdk/blob/f5c9104d30e0042ce3057c776a5dbd7320434750/src/constructs/core/ssm.test.ts#L22) and so instantiating a new `GuSSMParameter` will throw as the file it's reading cannot be found.

To fix this, we've checked in an empty `lambda.js` file.

This works, up until the point we want to [test the lambda itself](https://github.com/guardian/cdk/blob/f5c9104d30e0042ce3057c776a5dbd7320434750/src/constructs/core/custom-resources/runtime/lambda.test.ts).

Imports resolve alphabetically. Thus `import { flatten } from "./lambda"` imports the empty `js` file and not the desired `ts` file.

To fix this, we had created `lambda.symlink.ts` and imported that.

Pretty confusing stuff!

This change is simpler as the surface area is smaller (one test suite, vs two). The mock allows a new `GuSSMParameter` to be instantiated w/out file not found errors and the lambda tests now look like any other tests.

Lastly, the comment justifying this work around is colocated with its use case, which hopefully makes it easier to maintain.

jacobwinch:
Great, thanks for explaining all of the details!

PR #594, opened 2021-05-26
feat: add an alarm capable of catching cycling instances and instantiate within the EC2 App pattern

jacobwinch:
## What does this change?

* Creates a new alarm construct which helps to identify cycling instances (e.g. after a failed deploy or scaling event)
* Allows users to create this new alarm easily, via the EC2 App pattern
* Simplifies an existing alarm construct slightly (id is no longer required)

Perhaps this should have been split into separate PRs; I'm happy to do that if reviewers think it'd be helpful.

1. Yes, any user who currently instantiates a `Gu5xxPercentageAlarm` directly (i.e. not via the pattern) should no longer specify an id in the constructor
2. Any user who has already created a `Gu5xxPercentageAlarm` (either directly, or via the pattern) will also need to:
    1. Specify a custom name 
    2. Temporarily remove alarms
3. Any user who has configured an alarm via the pattern will need to refactor their `monitoringConfiguration` props slightly, as the API has been updated to offer better support for multiple alarms.
    
Note: 2 is required to avoid a situation where CFN attempts to create an alarm with a name which is identical to the first as part of resource replacement. Although this sounds like a real pain (it was during testing!), I don't think we actually have any users in this state!

Yes, the pattern docs/examples have been updated and (minimal) documentation has been added for the new construct.

## How to test

I've tested this change by deploying it to `cdk-playground` and breaking things in various ways!

Users should notice problems with cycling instances more quickly, which should help to keep apps in a healthier, more stable state.

As with other alarms, there is a risk that rolling this out across many stacks will generate too much noise.

### DIFF
--- a/src/constructs/cloudwatch/alarm.ts
+++ b/src/constructs/cloudwatch/alarm.ts
@@ -31,7 +31,7 @@ export class GuAlarm extends Alarm {
     };
     super(scope, id, { ...props, actionsEnabled: scope.withStageDependentValue(actionsEnabled) });
     const topicArn: string = `arn:aws:sns:${scope.region}:${scope.account}:${props.snsTopicName}`;
-    const snsTopic: ITopic = Topic.fromTopicArn(scope, "sns-topic-for-alarm-notifications", topicArn);
+    const snsTopic: ITopic = Topic.fromTopicArn(scope, `SnsTopicFor${id}`, topicArn);
     this.addAlarmAction(new SnsAction(snsTopic));
   }
 }
--- a/src/constructs/cloudwatch/ec2-alarms.test.ts
+++ b/src/constructs/cloudwatch/ec2-alarms.test.ts
@@ -5,7 +5,7 @@ import { ApplicationListener, ApplicationProtocol } from "@aws-cdk/aws-elasticlo
 import { Stack } from "@aws-cdk/core";
 import { simpleGuStackForTesting } from "../../utils/test";
 import { GuApplicationLoadBalancer, GuApplicationTargetGroup } from "../loadbalancing";
-import { Gu5xxPercentageAlarm, GuUnhealthyHostsAlarm } from "./ec2-alarms";
+import { Gu5xxPercentageAlarm, GuUnhealthyInstancesAlarm } from "./ec2-alarms";
 import type { AppIdentity } from "../core/identity";
 
 const vpc = Vpc.fromVpcAttributes(new Stack(), "VPC", {
@@ -73,7 +73,7 @@ describe("The Gu5xxPercentageAlarm construct", () => {
   });
 });
 
-describe("The GuUnhealthyHostsAlarm construct", () => {
+describe("The GuUnhealthyInstancesAlarm construct", () => {
   it("should create the correct alarm resource with minimal config", () => {
     const stack = simpleGuStackForTesting();
     const alb = new GuApplicationLoadBalancer(stack, "ApplicationLoadBalancer", { ...app, vpc });
@@ -84,7 +84,7 @@ describe("The GuUnhealthyHostsAlarm construct", () => {
       loadBalancer: alb,
       defaultTargetGroups: [targetGroup],
     });
-    new GuUnhealthyHostsAlarm(stack, { ...app, targetGroup: targetGroup, snsTopicName: "test-topic" });
+    new GuUnhealthyInstancesAlarm(stack, { ...app, targetGroup: targetGroup, snsTopicName: "test-topic" });
     expect(SynthUtils.toCloudFormation(stack)).toMatchSnapshot();
   });
 });
--- a/src/constructs/cloudwatch/ec2-alarms.test.ts
+++ b/src/constructs/cloudwatch/ec2-alarms.test.ts
@@ -1,10 +1,11 @@
-import "@aws-cdk/assert/jest";
 import { SynthUtils } from "@aws-cdk/assert";
+import "@aws-cdk/assert/jest";
 import { Vpc } from "@aws-cdk/aws-ec2";
+import { ApplicationListener, ApplicationProtocol } from "@aws-cdk/aws-elasticloadbalancingv2";
 import { Stack } from "@aws-cdk/core";
 import { simpleGuStackForTesting } from "../../utils/test";
-import { GuApplicationLoadBalancer } from "../loadbalancing";
-import { Gu5xxPercentageAlarm } from "./ec2-alarms";
+import { GuApplicationLoadBalancer, GuApplicationTargetGroup } from "../loadbalancing";
+import { Gu5xxPercentageAlarm, GuUnhealthyHostsAlarm } from "./ec2-alarms";
 import type { AppIdentity } from "../core/identity";
 
 const vpc = Vpc.fromVpcAttributes(new Stack(), "VPC", {
@@ -25,7 +26,7 @@ describe("The Gu5xxPercentageAlarm construct", () => {
       tolerated5xxPercentage: 1,
       snsTopicName: "test-topic",
     };
-    new Gu5xxPercentageAlarm(stack, "test", { ...app, loadBalancer: alb, ...props });
+    new Gu5xxPercentageAlarm(stack, { ...app, loadBalancer: alb, ...props });
     expect(SynthUtils.toCloudFormation(stack)).toMatchSnapshot();
   });
 
@@ -37,7 +38,7 @@ describe("The Gu5xxPercentageAlarm construct", () => {
       tolerated5xxPercentage: 1,
       snsTopicName: "test-topic",
     };
-    new Gu5xxPercentageAlarm(stack, "test", { ...app, loadBalancer: alb, ...props });
+    new Gu5xxPercentageAlarm(stack, { ...app, loadBalancer: alb, ...props });
     expect(stack).toHaveResource("AWS::CloudWatch::Alarm", {
       AlarmDescription: "test-custom-alarm-description",
     });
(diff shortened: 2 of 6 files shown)


src/constructs/cloudwatch/alarm.ts
```diff
@@ -31,7 +31,7 @@ export class GuAlarm extends Alarm {
     };
     super(scope, id, { ...props, actionsEnabled: scope.withStageDependentValue(actionsEnabled) });
     const topicArn: string = `arn:aws:sns:${scope.region}:${scope.account}:${props.snsTopicName}`;
-    const snsTopic: ITopic = Topic.fromTopicArn(scope, "sns-topic-for-alarm-notifications", topicArn);
+    const snsTopic: ITopic = Topic.fromTopicArn(scope, `SnsTopicFor${id}`, topicArn);
```
jacobwinch, line 34: Based on testing, it seems that this needs to be unique per alarm.

src/constructs/cloudwatch/ec2-alarms.test.ts
```diff
@@ -65,9 +66,25 @@ describe("The Gu5xxPercentageAlarm construct", () => {
...
     expect(stack).toHaveResource("AWS::CloudWatch::Alarm", {
       EvaluationPeriods: 3,
     });
   });
 });
+
+describe("The GuUnhealthyHostsAlarm construct", () => {
+  it("should create the correct alarm resource with minimal config", () => {
+    const stack = simpleGuStackForTesting();
+    const alb = new GuApplicationLoadBalancer(stack, "ApplicationLoadBalancer", { ...app, vpc });
```
jacobwinch, line 79: Unfortunately you need all 3 ALB components wired up in order to access the `metricUnhealthyHostCount()`metric!

src/constructs/cloudwatch/ec2-alarms.ts
```diff
@@ -46,6 +51,34 @@ export class Gu5xxPercentageAlarm extends GuAlarm {
...
+      ...props,
+      alarmName: alarmName,
+      alarmDescription: alarmDescription,
+      metric: props.targetGroup.metricUnhealthyHostCount(),
+      treatMissingData: TreatMissingData.NOT_BREACHING,
+      statistic: Statistic.MAXIMUM,
+      threshold: 1,
+      comparisonOperator: ComparisonOperator.GREATER_THAN_OR_EQUAL_TO_THRESHOLD,
+      period: Duration.minutes(5),
+      datapointsToAlarm: 6,
```
jacobwinch, line 79: Based on the [docs](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/AlarmThatSendsEmail.html) (and confirmed via testing):

> When you configure Evaluation Periods and Datapoints to Alarm as different values, you're setting an "M out of N" alarm. Datapoints to Alarm is ("M") and Evaluation Periods is ("N").

src/constructs/cloudwatch/ec2-alarms.ts
```diff
@@ -46,6 +51,34 @@ export class Gu5xxPercentageAlarm extends GuAlarm {
...
+    const alarmDescription = `${props.app}'s instances have failed healthchecks several times over the last hour.
+      This typically results in the AutoScaling Group cycling instances and can lead to problems with deployment,
+      scaling or handling traffic spikes.
+
+      Check ${props.app}'s application logs or ssh onto an unhealthy instance in order to debug these problems.`;
+    const alarmProps = {
+      ...props,
+      alarmName: alarmName,
+      alarmDescription: alarmDescription,
+      metric: props.targetGroup.metricUnhealthyHostCount(),
```
jacobwinch, line 73: ASG's have a [`GroupTerminatingInstances` metric](https://docs.aws.amazon.com/autoscaling/ec2/userguide/as-instance-monitoring.html#available-cloudwatch-metrics), which I originally planned on using. Unfortunately it doesn't seem possible to ascertain _why_ an instance is terminating using this metric. Consequently, it's not that helpful for this alarm because expected terminations (e.g. after a Riff-Raff deployment, or when an app scales down after a CPU-based scaling event) look exactly the same as instances being terminated due to failing health checks.

src/patterns/ec2-app.ts
```diff
@@ -393,11 +393,16 @@ export class GuEc2App {
     }
 
     if (!props.monitoringConfiguration.noMonitoring) {
-      new Gu5xxPercentageAlarm(scope, "Alarm", {
+      new Gu5xxPercentageAlarm(scope, {
         app,
         loadBalancer,
         ...props.monitoringConfiguration,
       });
+      new GuUnhealthyHostsAlarm(scope, {
```
jacobwinch, line 401: Since users don't need to supply any props to get this working (we only need an `snsTopicName` and the users has already provided one to get `Gu5xxPercentageAlarm` working), I have not changed the pattern's API at all. At the very least, I need to add docs for this, however there are a few other options:

1. Add a top-level boolean which allows users to opt-in/opt-out of this alert, regardless of whether they want a `Gu5xxPercentageAlarm` or not.
2. Change the type of `monitoringConfiguration: NoMonitoring | Gu5xxPercentageMonitoringProps;` to cover more alarms e.g. `monitoringConfiguration: NoMonitoring | GuSomeMoreGenericMonitoringInterface;`
3. Something else...

It'd be great to hear opinions on this.

akash1810: I think a set of booleans would be more explicit here:

```typescript
interface Alarms {
  5xxAlarm: boolean;
  unhealthyHostAlarm: boolean;
  anotherAlarm: boolean;
}
```

Downside is its another thing to maintain/keep up to date - maybe there's some funky generic TS we can write...

src/constructs/cloudwatch/ec2-alarms.ts
```diff
@@ -46,6 +51,34 @@ export class Gu5xxPercentageAlarm extends GuAlarm {
...
+      treatMissingData: TreatMissingData.NOT_BREACHING,
+      statistic: Statistic.MAXIMUM,
```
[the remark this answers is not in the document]
akash1810: Does this, or the alarm it self, need to be tied to the health check of the ASG?

src/constructs/cloudwatch/ec2-alarms.ts
```diff
@@ -46,6 +51,34 @@ export class Gu5xxPercentageAlarm extends GuAlarm {
...
+  }
+}
+
+/**
+ * Creates an alarm which is triggered whenever there have been several healthcheck failures within a single hour.
+ */
+export class GuUnhealthyHostsAlarm extends GuAlarm {
+  constructor(scope: GuStack, props: GuTargetGroupAlarmProps) {
+    const alarmName = `Unhealthy instances for ${props.app} in ${scope.stage}`;
+    const alarmDescription = `${props.app}'s instances have failed healthchecks several times over the last hour.
```
akash1810, line 64: Possibly overkill, but should we mention suspending the `ReplaceUnhealthy` process on the ASG to allow you to connect to an instance w/out fear of it being taken away from under your feet?

src/patterns/ec2-app.ts
```diff
@@ -393,11 +393,16 @@ export class GuEc2App {
...
+        app,
+        targetGroup,
+        snsTopicName: props.monitoringConfiguration.snsTopicName,
```
akash1810, line 404: I feel like this should become an account wide expectation that we lookup in Parameter Store, similar to the logging stream.

src/constructs/cloudwatch/ec2-alarms.ts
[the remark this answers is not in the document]
jacobwinch: I guess it wouldn't make much sense to use this alarm if your ASG only uses [`EC2` status checks](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-as-group.html#cfn-as-group-healthchecktype). Is this what you're referring to or did you have something else in mind here?

src/constructs/cloudwatch/ec2-alarms.ts
jacobwinch: Nice idea! I wondered how much detail we should get into here; my advice at the moment is slightly vague. It almost feels worth linking out to a runbook or something, as this is a common problem that I've helped to debug across many apps/teams.

src/patterns/ec2-app.ts
jacobwinch: I think the difficulty with doing this here is that a single account may host apps maintained by multiple teams. Sharing a logging stream is probably OK, but sharing an alerts topic could lead to a lot of noise. In recent times even Deploy Tools falls into this category (we are probably spamming Client-Side with a lot of alerts they don't care about right now). I think our long-term solution needs to look up appropriate targets based on tags (perhaps going via Anghammarad, which already has this capability).

src/patterns/ec2-app.test.ts
```diff
@@ -117,12 +117,54 @@ describe("the GuEC2App pattern", function () {
...
+  it("Skips alarm creation if the user explicitly opts-out", function () {
+    const stack = simpleGuStackForTesting();
+    const app = "test-gu-ec2-app";
+    new GuEc2App(stack, {
+      applicationPort: GuApplicationPorts.Node,
+      app,
+      access: { scope: AccessScope.PUBLIC },
+      certificateProps: getCertificateProps(),
+      monitoringConfiguration: {
+        snsTopicName: "test-topic",
```
jacobwinch, line 160: I would expect real users to use `{ noMonitoring: true }` in this scenario (as the `snsTopicName` is pointless), but this allows us to validate the behaviour in a single test.

akash1810 (approved):

PR #603, opened 2021-06-04
feat(ec2-app-pattern): add gu:cdk:pattern-name tag to AutoScalingGroup

stephengeller:
## What does this change?

At the moment, we have poor visibility on which stacks are using which patterns. As we roll out the pattern to more stacks, this will become poorer and poorer.

This PR is the first step in a piece of work to make surfacing pattern usage across the AWS estate easier. If & when this gets merged in, we can then build a lambda to surface this information in CloudWatch, which could then be queried by and visualised in Grafana dashboards.

No. We would need to bump the library version and re-deploy any stacks using the pattern (which there are not many of, as of yet).

'

Sort of - a comment has been left above the behaviour detailing why it is necessary.

## How to test

`./script/test` from the root of the repo.

The ASGs of stacks using the pattern begin to surface in AWS, and we are eventually able to display information on pattern adoption in Grafana.

Not that I can think of.

### DIFF
--- a/src/constants/tag-keys.ts
+++ b/src/constants/tag-keys.ts
@@ -1,4 +1,5 @@
 export const TagKeys = {
   TRACKING_TAG: "gu:cdk:version",
   REPOSITORY_NAME: "gu:repo",
+  PATTERN_NAME: "gu:cdk:pattern-name",
 };
--- a/src/patterns/__snapshots__/ec2-app.test.ts.snap
+++ b/src/patterns/__snapshots__/ec2-app.test.ts.snap
@@ -107,6 +107,11 @@ Object {
             "Value": "test-gu-ec2-app",
           },
           Object {
+            "Key": "gu:cdk:pattern-name",
+            "PropagateAtLaunch": true,
+            "Value": "GuEc2App",
+          },
+          Object {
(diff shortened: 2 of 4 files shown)


akash1810:
I wonder if we should make use of [Cloudformation metadata](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/metadata-section-structure.html) for this and the tracking tag as they are internal bits of information only for the GuCDK library.

[AWS CDK uses metadata to](https://docs.aws.amazon.com/cdk/latest/guide/cli.html#version_reporting):

> gain insight into how the CDK is used, the constructs used by CDK applications are collected and reported by using a resource identified as `AWS::CDK::Metadata`. This resource is added to AWS CloudFormation templates, and can easily be reviewed. This information can also be used by AWS to identify stacks using a construct with known security or reliability issues, and to contact their users with important information.
> ...
> The `Analytics` property is a gzipped, base64-encoded, prefix-encoded list of the constructs present in the stack.

Let's get this change in though and we can look at metadata separately.

akash1810 (approved):
👍🏽 

I think metadata would be a better fit for this, but we can get this in and refactor once we understand cfn metadata more.

PR #608, opened 2021-06-07
chore: check the npm version

akash1810:


akash1810:
This is just to check @JamieB-gu 's statement that the npm version is tied to the node version as described [here](https://nodejs.org/en/download/releases/). It is! That is we won't need to explicitly install npmv7 on the agent to use it.

Closing now that we know!

PR #609, opened 2021-06-07
feat(ec2-app-pattern): allow threading of ASG block device configuration

stephengeller:
## What does this change?

At the moment, some stacks like [Prism](https://github.com/guardian/prism/blob/3ca28b2329dd837cee07fb7152b74e7dd55b8185/cdk/lib/prism.ts#L93-L100) require specifying block device configurations. This isn't easily or obviously achievable by accessing the ASG sub-construct (eg `pattern.autoScalingGroup`) nor by escape hatches (eg `const cfnAsg = app.autoScalingGroup.node.defaultChild as unknown as CfnAutoScalingGroup`), and so the best method might be to thread it through the pattern props.

This PR adds a prop (nested under an `autoScalingGroup` key) that allows users to optionally configure block devices.

### DIFF
--- a/src/patterns/ec2-app.ts
+++ b/src/patterns/ec2-app.ts
@@ -156,9 +156,7 @@ export interface GuEc2AppProps extends AppIdentity {
     [Stage.PROD]?: GuAsgCapacityProps;
   };
   accessLogging?: AccessLoggingProps;
-  autoScalingGroup?: {
-    blockDevices?: BlockDevice[];
-  };
+  blockDevices?: BlockDevice[];
 }
 
 interface GuMaybePortProps extends Omit<GuEc2AppProps, "applicationPort"> {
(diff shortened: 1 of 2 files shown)


src/patterns/ec2-app.ts
```diff
@@ -145,6 +146,7 @@ export interface GuEc2AppProps extends AppIdentity {
     [Stage.PROD]: GuAsgCapacityProps;
   };
   accessLogging?: AccessLoggingProps;
+  autoScalingGroup?: { blockDevices?: BlockDevice[] };
```
jacobwinch, line 149: I would probably move `blockDevices` up a level (rather than nesting it within `autoScalingGroup`), as it seems comparable to other optional props at that level (e.g. `roleConfiguration` and `accessLogging`).

stephengeller: Think you're right on this, this has been changed.

stephengeller:
This is now ready for review.

akash1810:
FYI the linter violations should go away in https://github.com/guardian/cdk/pull/616

stephengeller:
Linting issues have been fixed and CI is passing, so this should be good for another review.

jacobwinch (approved):

PR #610, opened 2021-06-08
fix: Add a reason prop to GuMigratingResource

akash1810:
## What does this change?

Updates the shape of `GuMigratingResource` to include a reason field to explain why a resource's logical ID is being pinned.

This is on the back of https://github.com/guardian/prism/pull/212 which fixes [this mistake](https://github.com/guardian/prism/commit/dc69fbee724965a49d9db54ef356fb18867e368c#diff-14cbc0e0275d178f0a4cf31c3c0d4408758ee8daad333947c0ff38391bc7d600L19).

The `reason` field isn't used anywhere, it's more like using the compiler to force comments... now that I think about it, it's a possible misuse of the compiler 😬 !

Yes - there is an interface shape change. Projects cannot rely on Dependabot bumping the version of GuCDK as the CI step will definitely fail. The change isn't too onerous, however.

Yes - added.

## How to test

See tests?

Today's developers are kind to tomorrow's?

n/a

### DIFF
--- a/src/constructs/core/migrating.ts
+++ b/src/constructs/core/migrating.ts
@@ -32,7 +32,17 @@ export interface GuMigratingResource {
    * @see GuMigratingStack
    * @see GuStack
    */
-  existingLogicalId?: string;
+  existingLogicalId?: {
+    /**
+     * The logical ID to use in the synthesised template for this resource.
+     */
+    logicalId: string;
+
+    /**
+     * A short description to help developers understand why this resource's logical ID is being set.
+     */
+    reason: string;
+  };
 }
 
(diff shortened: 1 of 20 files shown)


src/constructs/core/migrating.ts
```diff
@@ -32,7 +32,17 @@ export interface GuMigratingResource {
    * @see GuMigratingStack
    * @see GuStack
    */
-  existingLogicalId?: string;
+  existingLogicalId?: {
```
akash1810, line 35: This is the main change, the other changes are updating of tests.

jacobwinch (approved):

PR #616, opened 2021-06-09
chore: Update `@guardian/eslint-config-typescript` to v0.6.0

akash1810:
## What does this change?

Since v0.6.0 `@guardian/eslint-config-typescript` directly installs its dependencies. This means we can remove some entries from `package.json`.

It is also pinning the versions of the deps so that we don't have to perform a dance each time `eslint-plugin-import` changes, for example.

As it is pinning `eslint-plugin-import` to v2.23.4, we have to do the `eslint --fix` dance one last time. This resolves [this PR](https://github.com/guardian/cdk/pull/598) and makes up the bulk of the changes in this PR.

See:
  - guardian/configs#66
  - https://github.com/guardian/cdk/pulls?q=is%3Apr+%22chore%28deps%29%3A+bump+eslint-plugin-import%22

No.

No.

## How to test

CI should pass.

Fewer dependencies for Dependabot to update.

Fewer chores, allowing us to focus on delivery rather than changing the order of import statements once a week.

n/a


package.json
```diff
@@ -22,20 +22,14 @@
...
     "@types/node": "15.6.1",
-    "@typescript-eslint/eslint-plugin": "^4.26.0",
-    "@typescript-eslint/parser": "^4.26.0",
     "cz-conventional-changelog": "^3.3.0",
     "eslint": "^7.27.0",
-    "eslint-config-prettier": "^8.3.0",
     "eslint-plugin-custom-rules": "file:eslint",
-    "eslint-plugin-eslint-comments": "^3.2.0",
-    "eslint-plugin-import": "^2.23.2",
-    "eslint-plugin-prettier": "^3.4.0",
```
akash1810, line 38: These are all direct dependencies of `@guardian/eslint-config-typescript` now and become transitive in this repository 🎉 .

package.json
```diff
@@ -22,20 +22,14 @@
     "update-aws-cdk": "ncu \"/^@?aws-cdk(/.*)?$/\" --upgrade --deep && npm install"
   },
   "devDependencies": {
-    "@guardian/eslint-config-typescript": "^0.5.0",
+    "@guardian/eslint-config-typescript": "^0.6.0",
```
akash1810, line 25: This is the only real change. It has a butterfly effect...

stephengeller (approved):

PR #626, opened 2021-06-22
feat: Add policy to allow writing to Anghammarad

akash1810:
## What does this change?

Create a policy that allows `sns:Publish` to the Anghammarad SNS topic.

The Anghammarad SNS topic is used across multiple stacks, so we can promote it to an account wide ("account/services") parameter store location.

We then wrap this account wide parameter into the `AnghammaradTopicParameter` singleton, which the policy will add when necessary.

Yes, they become a little DRYer as they no longer need to define this policy themselves.

Yes and documentation is included.

## How to test

Tests have been added that cover:
  - Automatic parameter addition (well, the implementation of the singleton)
  - The IAM policy

Creating a first class construct for Anghammarad permissions means downstream stacks become DRYer.

(I can also correctly spell Anghammarad!)

n/a

### DIFF
--- a/docs/005-default-parameter-store-locations.md
+++ b/docs/005-default-parameter-store-locations.md
@@ -3,13 +3,14 @@
 A number of constructs from the library define parameters to get configuration from Parameter Store. Each of these parameters configures a default path.
 The below table lists those paths, the parameter that sets them, the expected value type and which construct the parameter is defined within.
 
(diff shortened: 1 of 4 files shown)


jacobwinch (approved):
Great work 👍 

The documentation is very clear 💯

akash1810:
Just noticed a typo ("am" instead of "an"). Have force pushed that change. Have also enabled auto merge.

PR #628, opened 2021-06-22
fix: Consistently name the `AnghammaradSenderPolicy`

akash1810:
Follows #626.

## What does this change?

Update the name of `AnghammaradSenderPolicy` to `GuAnghammaradSenderPolicy` to be consistent with the library.

Also add exports to the index files to simplify how the construct gets imported.

No. Nothing is currently using this construct.

No.

## How to test

n/a

Consistent naming and importing to yield a simpler interface to the library.

n/a

### DIFF
--- a/src/constructs/iam/policies/anghammarad.ts
+++ b/src/constructs/iam/policies/anghammarad.ts
@@ -1,30 +1,30 @@
 import { isSingletonPresentInStack } from "../../../utils/test";
 import type { GuStack } from "../../core";
-import { AnghammaradTopicParameter } from "../../core/parameters/anghammarad";
+import { GuAnghammaradTopicParameter } from "../../core";
 import { GuAllowPolicy } from "./base-policy";
 
 /**
  * Creates an `AWS::IAM::Policy` to grant `sns:Publish` permission to the Anghammarad topic.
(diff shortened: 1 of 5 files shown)


src/constructs/iam/policies/anghammarad.ts
```diff
@@ -1,30 +1,30 @@
...
-export class AnghammaradSenderPolicy extends GuAllowPolicy {
-  private static instance: AnghammaradSenderPolicy | undefined;
+export class GuAnghammaradSenderPolicy extends GuAllowPolicy {
+  private static instance: GuAnghammaradSenderPolicy | undefined;
 
   private constructor(scope: GuStack) {
-    const anghammaradTopicParameter = AnghammaradTopicParameter.getInstance(scope);
+    const anghammaradTopicParameter = GuAnghammaradTopicParameter.getInstance(scope);
 
-    super(scope, "GuSESSenderPolicy", {
```
jacobwinch, line 19: I didn't notice this in the last PR review 😅 Thanks for fixing!

jacobwinch (approved):

src/constructs/iam/policies/anghammarad.ts
akash1810: You know know my secret 😊 .

PR #635, opened 2021-06-24
feat: Add a construct to ease S3 bucket creation

akash1810:
## What does this change?

Guardian best practice is to retain S3 buckets as buckets share a global namespace. `GuS3Bucket` will set the correct policy on the bucket to simplify this.

See:
  - https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-deletionpolicy.html

No.

Yes and it has been added.

## How to test

See the added tests?

It is easier to create Guardian flavoured buckets.

n/a

### DIFF
--- a/src/constructs/s3/__snapshots__/index.test.ts.snap
+++ b/src/constructs/s3/__snapshots__/index.test.ts.snap
@@ -1,0 +1,47 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`The GuS3Bucket construct should set the bucket's policy to 'retain' 1`] = `
+Object {
+  "Parameters": Object {
+    "Stage": Object {
+      "AllowedValues": Array [
+        "CODE",
+        "PROD",
+      ],
+      "Default": "CODE",
+      "Description": "Stage name",
+      "Type": "String",
+    },
+  },
+  "Resources": Object {
+    "MyBucketF68F3FF0": Object {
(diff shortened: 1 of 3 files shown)


src/constructs/s3/migrating-bucket.test.ts
```diff
@@ -0,0 +1,26 @@
+import "../../utils/test/jest";
+import { simpleGuStackForTesting } from "../../utils/test";
+import { GuMigratingS3Bucket } from "./migrating-bucket";
+
+describe("The GuMigratingS3Bucket construct", () => {
+  it("should auto-generate the logicalId by default", () => {
```
jacobwinch, line 6: I don't think we'd ever want this construct to create a new logical id (since it is explicitly for _migrating_ existing buckets)? Perhaps we could require the prop (or throw) instead of allowing this to happen?

akash1810: Good point! I actually don't think `GuMigratingS3Bucket` is in line with our other constructs, which are typically "Guardian flavoured versions of AWS constructs".

A Guardian flavoured bucket would be a tagged bucket with a retain policy. What if `GuMigratingS3Bucket` became `GuS3Bucket`, setting these props?

akash1810: Updated now. The diff is tiny as a result! 🎉

jacobwinch (approved):

PR #640, opened 2021-06-28
feat: Add EC2 pattern support for internal load balancers

jacobwinch:
@stephengeller and I paired on this change.

## What does this change?
This PR adds support for configuring internal load balancers via the EC2 App pattern, as some of our applications (e.g. Prism) need to make use of this scheme.

No, this change just adds new functionality. It should allow existing projects (e.g. Grafana and Prism) to make use of the EC2 app pattern soon.

Yes, this PR adds some new documentation.

## How to test
Unit tests have been added for new functionality. Further testing will be carried out as part of the Grafana and Prism migrations.

After this change is released, it will be possible to create applications with an internal load balancer.

Choosing when to use `RestrictedAccess` vs `InternalAccess` may be a little confusing for users. 

Due to variations in the way that our accounts are set up, the following block of code can produce significantly different results depending on the account which contains the infrastructure.

 ```typescript
 {
   scope: AccessScope.INTERNAL,
   cidrRanges: [Peer.ipv4("10.0.0.0/8")]
 }
 ```

In the Deploy Tools account, this will allow incoming traffic from:
1. Applications in the same VPC.
1. Applications in the `Frontend` VPC.
1. Users at Kings Place / on the VPN.

In the Mobile account, it will only allow incoming traffic from the same VPC.

This makes it hard to clearly document when each access type should/shouldn't be used.


src/patterns/ec2-app.ts
```diff
@@ -64,7 +66,26 @@ export interface RestrictedAccess extends Access {
...
+ *   cidrRanges: [Peer.ipv4("10.0.0.0/8")]
+ * }
+ * ```
+ */
+export interface InternalAccess extends Access {
+  scope: AccessScope.INTERNAL;
+  cidrRanges: IPeer[];
+}
+
+export type AppAccess = PublicAccess | RestrictedAccess | InternalAccess;
```
stephengeller, line 88: Are we still using this? If not we can maybe remove it, can't seem to spot usage in the changes.

```suggestion
```

jacobwinch: Looks like it is used in [`GuEc2AppProps`](https://github.com/guardian/cdk/pull/640/files#diff-161e4cde1954da9ec6ae32e3e8546e3f9befa14b6cc6d42824afe33fc7af6bc0R170)

stephengeller:
All looks good to me beyond the potential change left in the comment. I'll let @akash1810 get some eyes on it before I 👍🏻 , but would say it looks good enough to go. 

It's tough to articulate when to use `INTERNAL` access over `RESTRICTED` or vice versa, but @jacobwinch and I spoke about exploring other avenues of controlling access such [authentication at the ALB level](https://aws.amazon.com/blogs/aws/built-in-authentication-in-alb/), which might be the next sensible thing to investigate.

src/patterns/ec2-app.ts
stephengeller: Oh sorry 🤦🏻 missed that you weren't newly declaring this and it already existed, ignore this!

src/patterns/ec2-app.test.ts
```diff
@@ -68,6 +68,19 @@ describe("the GuEC2App pattern", function () {
...
+    const stack = simpleGuStackForTesting();
+    new GuEc2App(stack, {
+      applicationPort: GuApplicationPorts.Node,
+      app: "test-gu-ec2-app",
+      access: { scope: AccessScope.INTERNAL, cidrRanges: [Peer.ipv4("10.0.0.0/8")] },
+      monitoringConfiguration: { noMonitoring: true },
+      userData: "#!/bin/dev foobarbaz",
+      certificateProps: getCertificateProps(),
+    });
+    expect(SynthUtils.toCloudFormation(stack)).toMatchSnapshot();
```
akash1810, line 81: The snapshot for this test is pretty big (~900 lines), which makes it difficult to find what we're asserting. Could we use [`toHaveResourceOfTypeAndLogicalId`](https://github.com/guardian/cdk/blob/a5e1337b2b7f1d9ec1b5e1eb4e5b0adf0db1f393/src/utils/test/jest.ts#L42) to assert against the load balancer in isolation of the other resources that get created in the `GuEc2App`?

jacobwinch:
CI is currently failing due to `aws/cdk` version, which is fixed here: https://github.com/guardian/cdk/pull/641.

src/patterns/ec2-app.test.ts
```diff
@@ -68,6 +68,24 @@ describe("the GuEC2App pattern", function () {
...
+    new GuEc2App(stack, {
+      applicationPort: GuApplicationPorts.Node,
+      app: "test-gu-ec2-app",
+      access: { scope: AccessScope.INTERNAL, cidrRanges: [Peer.ipv4("10.0.0.0/8")] },
+      monitoringConfiguration: { noMonitoring: true },
+      userData: "#!/bin/dev foobarbaz",
+      certificateProps: getCertificateProps(),
+    });
+    expect(stack).toHaveResource("AWS::ElasticLoadBalancingV2::LoadBalancer", {
+      Scheme: "internal",
```
akash1810, line 82: 👍🏽

akash1810 (approved):

PR #570, opened 2021-05-19
chore: Add lockfile to `integration-test` project

akash1810:
## What does this change?

We previously had issues with lockfiles in CI for the integration-test project. This caused a few issues and led to the changes in #569.

This change reverts those commits and adds a lockfile (might be easiest to review commits in order). That is, I'm not sure what the previous issues were, but they don't seem to exist anymore and we can use lockfiles in CI and gain deterministic builds.

## Does this change require changes to existing projects or CDK CLI?

No.

## Does this change require changes to the library documentation?

No.

## How to test

Observe CI.

## How can we measure success?

- The addition of a lockfile and installation of dependencies using `npm ci` results in a deterministic build in CI
- The re-addition of linting keeps the integration-test project in line with the main project
 
## Have we considered potential risks?

n/a

PR #641, opened 2021-06-29
chore(deps)!: Update aws-cdk libraries to 1.110.1

jacobwinch:
BREAKING CHANGE: Update aws-cdk libraries to 1.110.1


akash1810 (commented):
There's a bug in the `update-aws-cdk` script as I think `integration-test/package-lock.json` should have changed too (since #570)? Probably [this line](https://github.com/guardian/cdk/blob/a5e1337b2b7f1d9ec1b5e1eb4e5b0adf0db1f393/script/update-aws-cdk#L54).

It also looks like this is a v2 lockfile, which comes from npmv7. Is this desired? See #340.

jacobwinch:
Thanks @akash1810 - I had trusted the script without checking. Will close this and have a proper look!

akash1810:
We might be able to move to NPMv7 as @JamieB-gu recently pointed out that the npm version is [tied](https://nodejs.org/en/download/releases/) to the node version ([demo](#608)).

jacobwinch:
> It also looks like this is a v2 lockfile, which comes from npmv7. Is this desired? See #340.

This part definitely wasn't desired, in fact I don't understand how it happened yet!

JamieB-gu:
> in fact I don't understand how it happened yet!

Which version of `npm` did you use to make this change? If you were on v7 could that have caused the migration?

jacobwinch:
> Which version of `npm` did you use to make this change? If you were on v7 could that have caused the migration?

See: https://github.com/guardian/cdk/pull/642#discussion_r660404644. `6.14.8` _and_ (accidentally) 7.x!

PR #642, opened 2021-06-29
chore: Attempt to fix update-aws-cdk script

jacobwinch:
## What does this change?

This PR attempts to fix some of the problems that we noticed in https://github.com/guardian/cdk/pull/641.

### DIFF
--- a/script/update-aws-cdk
+++ b/script/update-aws-cdk
@@ -35,7 +35,7 @@ updateAwsCdk() {
 }
 
 installAwsCdkUpdates() {
-  npm install && cd integration-test && npm install
+  npm install && cd integration-test && npm install && cd ..
 }
 
 checkForLocalChanges() {
--- a/script/update-aws-cdk
+++ b/script/update-aws-cdk
@@ -34,6 +34,10 @@ updateAwsCdk() {
   npm run update-aws-cdk
 }
 
+installAwsCdkUpdates() {
+  npm install && cd integration-test && npm install
+}
+
 checkForLocalChanges() {
   if git diff --quiet; then
     echo "No changes. No further action needed."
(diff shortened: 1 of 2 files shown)


script/update-aws-cdk
```diff
@@ -51,7 +55,7 @@ raisePR() {
 
   git checkout -b $BRANCH_NAME
 
-  for file in package.json package-lock.json integration-test/package.json; do
+  for file in package.json package-lock.json integration-test/package.json integration-test/package-lock.json; do
```
jacobwinch, line 58: https://github.com/guardian/cdk/pull/641#pullrequestreview-694701813

>  I think integration-test/package-lock.json should have changed too

Yep, this should fix it!

script/update-aws-cdk
```diff
@@ -34,6 +34,10 @@ updateAwsCdk() {
   npm run update-aws-cdk
 }
 
+installAwsCdkUpdates() {
+  npm install && cd integration-test && npm install
```
jacobwinch, line 38: > It also looks like this is a v2 lockfile, which comes from npmv7. Is this desired? 

When this part was being run from within `package.json`, npmv7 was being used. However:
```
npm --version
6.14.8
```
😕

akash1810: 🤔  that _is_ odd!

akash1810: Do we need to `cd ..` after this too to stay at the root? Alternatives:

```sh
(
  cd integration-test
  npm install
)
```

Or `pushd`/`popd`.

akash1810 (approved):

jacobwinch:
This seems to have worked successfully: https://github.com/guardian/cdk/pull/644.

PR #679, opened 2021-07-09
fix: Set the package description

akash1810:
## What does this change?

If not explicitly set, it looks like NPM will use the first line of the README.md as the description.

Using `npm view  @guardian/cdk --json | jq .description`, we can see the current description of `@guardian/cdk` in NPM is `"![npm][badge-npm] [![CD][badge-cd]][internal-cd-file]"`.

This change explicitly sets the description field.


jfsoul (approved):
Confident that this is an improvement on the current description 😆

PR #675, opened 2021-07-07
feat: Add a CLI to the repository

akash1810:
## What does this change?

Add a CLI to the repository 🎉 .

This is the start of a process to bring [`@guardian/cdk-cli`](https://github.com/guardian/cdk-cli) into this repository as it makes it easier to keep the CLI up to date (we should be able to use the compiler to tell us if we've changed a type but forgot to update the CLI). Yay mono-repos!

This change adds one command `aws-cdk-version` to print the version of `aws-cdk` libraries being used. We should be able to invoke it by doing `npx @guardian/cdk aws-cdk-version` 🤞🏽 .

There's also a helpful script to invoke the CLI in DEV - `./script/cli`.

This is heavily inspired by the [AWS CDK CLI](https://github.com/aws/aws-cdk/blob/71c0a4c413e77452f47c797d4e861aa542174ce9/packages/aws-cdk/bin/cdk.ts).

### Example output

```console
➜ npx @guardian/cdk --help
@guardian/cdk aws-cdk-version

Print the version of @aws-cdk libraries being used

Options:
      --verbose  Show verbose output                  [boolean] [default: false]
      --version  Show version number                                   [boolean]
  -h, --help     Show help                                             [boolean]
```

```console
➜ npx @guardian/cdk --version
22.0.1 (using @aws-cdk 1.110.1)
```

```console
➜ npx @guardian/cdk aws-cdk-version
1.110.1
```

```console
➜ npx @guardian/cdk aws-cdk-version --verbose | jq .
{
  "@aws-cdk/assert": "1.110.1",
  "@aws-cdk/aws-apigateway": "1.110.1",
  "@aws-cdk/aws-autoscaling": "1.110.1",
  "@aws-cdk/aws-cloudwatch-actions": "1.110.1",
  "@aws-cdk/aws-ec2": "1.110.1",
  "@aws-cdk/aws-elasticloadbalancing": "1.110.1",
  "@aws-cdk/aws-elasticloadbalancingv2": "1.110.1",
  "@aws-cdk/aws-events-targets": "1.110.1",
  "@aws-cdk/aws-iam": "1.110.1",
  "@aws-cdk/aws-kinesis": "1.110.1",
  "@aws-cdk/aws-lambda": "1.110.1",
  "@aws-cdk/aws-lambda-event-sources": "1.110.1",
  "@aws-cdk/aws-rds": "1.110.1",
  "@aws-cdk/aws-s3": "1.110.1",
  "@aws-cdk/core": "1.110.1"
}
[...]

PR #686, opened 2021-07-14
chore: Enable beta releases to NPM 

akash1810:
## What does this change?

Whilst it is possible to install this package from GitHub, it can be useful to install from NPM as the published package structure varies slightly, for example in #675 .

With this change, we will be able to publish to the `beta` channel by update the `beta` branch. This will enable us to to install our experimental changes from NPM with:

```
npm install @guardian/cdk@beta
```

I followed [these docs](https://semantic-release.gitbook.io/semantic-release/usage/workflow-configuration#prerelease).

No.

Yes and added.

## How to test

I don't think it's possible...

Once this PR is merged, we should be able to create a `beta` branch and observe it be published to `@guardian/cdk@beta`. We should also see updates to `main` continue to be released as normal.

1. Updates to the `main` branch continue to be automatically released to the default NPM channel.
2. Updates to the `beta` branch will get automatically released to the beta NPM channel.

CD becomes a little more complicated with the added `if` statements. An alternative to this change would be to copy the `./.github/workflows/cd.yaml` and `./script/cd` files into `./.github/workflows/cd-beta.yaml` and `./script/cd-beta` and remove the `if` guards. I think this would create more maintenance, however.


jacobwinch (approved):
Looks good to me 👍  Let's give it a try! 🚢

PR #687, opened 2021-07-14
chore: Set Dependabot's versioning-strategy

akash1810:
## What does this change?

We've started to see Dependabot raise PRs that only update the `package-lock.json` and not `package.json` (#683, #684). Setting `versioning-strategy` to `increase` will:

> Always increase the version requirement to match the new version.

That is, it should force both `package.json` and `package-lock.json` to be updated.

See:
  - https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#versioning-strategy

No.

No.

## How to test

n/a

More reliable Dependabot PRs.

n/a

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -16,6 +16,8 @@ updates:
       day: "wednesday"
     commit-message:
       prefix: "chore(deps): "
+    # force the lockfile and package.json to be updated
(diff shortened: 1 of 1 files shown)


jamie-lynch (approved):
👍

PR #712, opened 2021-07-28
feat: add SSM read to Lambda construct for app params

sihil:
## What does this change?
This adds a policy statement to the Lambda construct so that permission is always granted to read SSM parameters under  $stage/$stack/$app/* for a lambda.

No

Have added a note on the lambda docs to reflect new permission.

## How to test
Covered by snapshot tests.

Users don't need to concern themselves with granting permission to read configuration.

Risk is low due to test coverage.

### DIFF
--- a/src/constructs/iam/policies/parameter-store-read.ts
+++ b/src/constructs/iam/policies/parameter-store-read.ts
@@ -3,19 +3,21 @@ import type { GuStack } from "../../core";
 import { AppIdentity } from "../../core/identity";
 import { GuPolicy } from "./base-policy";
 
+export class GuParameterStoreReadPolicyStatement extends PolicyStatement {
+  constructor(scope: GuStack, props: AppIdentity) {
+    super({
+      effect: Effect.ALLOW,
+      actions: ["ssm:GetParametersByPath"],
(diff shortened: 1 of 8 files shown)


src/constructs/iam/policies/parameter-store-read.ts
```diff
@@ -3,19 +3,19 @@ import type { GuStack } from "../../core";
 import { AppIdentity } from "../../core/identity";
 import { GuPolicy } from "./base-policy";
 
+export function guParameterStoreReadPolicyStatement(scope: GuStack, app: string): PolicyStatement {
+  return new PolicyStatement({
+    effect: Effect.ALLOW,
+    actions: ["ssm:GetParametersByPath"],
+    resources: [`arn:aws:ssm:${scope.region}:${scope.account}:parameter/${scope.stage}/${scope.stack}/${app}`],
+  });
+}
```
akash1810, line 12: The library currently defines constructs as classes, can we convert this to a class for consistency?

sihil: Makes sense to be consistent. Can you have another look?

akash1810 (approved):
Looks great!

PR #738, opened 2021-08-13
fix: improve default healthcheck and deregistration settings

sihil:
## What does this change?
On ALBs the deregistration delay determines how quickly an instance can be removed from a target grouo. The default is five minutes which makes deploys take a lot longer than is needed.

This delay should be tuned to be longer than your longest ever expected request. As such 30s seems reasonable for most use cases but I could be convinced to push this up to 60s.

In addition we also flip the healthy/unhealthy default values which look wrong (an instance should go out of service no slower than come in) and increase the polling interval so that the load balancer is more responsive.

Existing projects will inherit these more sensible defaults when they upgrade.

These specific settings are not documented.

Deploys of apps like Prism are substantially faster.

It's possible that some apps rely on this. It's worth reviewing settings in use across the department to verify these make sense.

### DIFF
--- a/src/constructs/cloudwatch/__snapshots__/ec2-alarms.test.ts.snap
+++ b/src/constructs/cloudwatch/__snapshots__/ec2-alarms.test.ts.snap
@@ -403,11 +403,11 @@ Object {
     },
     "ApplicationTargetGroupTesting080C2FF2": Object {
       "Properties": Object {
-        "HealthCheckIntervalSeconds": 30,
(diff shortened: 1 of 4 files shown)


jacobwinch:
This change looks sensible to me - I think we'll just need to update some of the tests in `src/constructs/loadbalancing/alb/application-target-group.test.ts` before merging this in

jacobwinch (approved):

PR #749, opened 2021-08-20
chore: Report test coverage

akash1810:
## What does this change?

Collect test coverage to see how thoroughly the codebase is tested.


akash1810:
Probably not much use unless we have a plan on how to remediate any actions...

PR #761, opened 2021-08-26
feat: Add Fastly Logs IAM and related constructs

frankie297:
## What does this change?
Adds an s3 put object construct, and creates a Fastly Logs IAM pattern. This is specifically to allow Fastly to log directly to s3 rather than using keys and having to keep them rotated. See documentation: https://docs.fastly.com/en/guides/creating-an-aws-iam-role-for-fastly-logging

No

Yes

## How to test
Snapshot test added

Manual key rotation is no longer needed, so less security issues.

Paired with @jfsoul


jfsoul:
@akash1810 we missed the requirement for angular commit messages - have we considered adding a commit hook if that's a pretty strict expectation for contributors?

akash1810:
> @akash1810 we missed the requirement for angular commit messages - have we considered adding a commit hook if that's a pretty strict expectation for contributors?

IIRC we discussed commit hooks in the early days and opted against after experience on other repositories (they're more restrictive than helpful); keen to hear from @jacobwinch and @sihil to see if this still holds.

I'm not too sure how semantic-release will parse this PR, either:
- It will parse the `feat` part of the merge commit and release to npm
- Parse each commit and not deem the change as not needing a release

I'm really interested to find out! Happy to merge as is and we can update any docs to reflect our new learning.

I think a sure fire way to make semantic-release happy is to [squash and merge](https://docs.github.com/en/github/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges#squash-and-merge-your-pull-request-commits).

akash1810 (dismissed):
🎉

jfsoul:
> I'm really interested to find out! Happy to merge as is and we can update any docs to reflect our new learning.

It's a tempting option to experiment, but I think our commits are not ideal for the history in this case so I was always assuming it'd be a squash and merge for this PR!

akash1810:

src/constants/ssm-parameter-paths.ts
```diff
@@ -1,6 +1,7 @@
 export interface SsmParameterPath {
   path: string;
   description: string;
+  optional?: boolean;
```
jfsoul, line 4: The `account-readiness` CLI command currently expects all these parameters to be required, but actually we're introducing an optional, account-wide parameter here (account-wide because you might have multiple instances of the fastly IAM logging construct in a single account)

docs/005-default-parameter-store-locations.md
```diff
@@ -3,14 +3,15 @@
...
+|-----------------------------------------|------------------------------| ---------------------------- |----------------------------|
+| /account/vpc/primary/id                 | VpcId                        | AWS::EC2::VPC::Id            | GuVpc.fromIdParameter      |
+| /account/vpc/primary/subnets/public     | PublicSubnets                | List\<AWS::EC2::Subnet::Id\> | GuVpc.subnetsFromParameter |
+| /account/vpc/primary/subnets/private    | PrivateSubnets               | List\<AWS::EC2::Subnet::Id\> | GuVpc.subnetsFromParameter |
+| /account/services/artifact.bucket       | DistributionBucketName       | String                       | GuGetDistributablePolicy   |
+| /account/services/logging.stream.name   | LoggingStreamName            | String                       | GuLogShippingPolicy        |
+| /account/services/anghammarad.topic.arn | AnghammaradSnsArn            | String                       | AnghammaradSenderPolicy    |
+| /account/external/fastly/customer.id    | GuFastlyCustomerIdParameter  | String                       | GuFastlyLogsIam            |
```
akash1810, line 14: We should really auto-generate this list!

src/constants/ssm-parameter-paths.ts
```diff
@@ -38,4 +39,9 @@ export const SSM_PARAMETER_PATHS: Record<string, SsmParameterPath> = {
     path: `${VPC_SSM_PARAMETER_PREFIX}/primary/subnets/public`,
     description: "A list of public subnets",
   },
+  FastlyCustomerId: {
+    path: "/account/external/fastly/customer.id",
+    description: "SSM parameter containing the Fastly Customer ID",
```
akash1810, line 44: Is this unique per Fastly service? And/or is it worth adding how one can obtain this value?

src/constructs/iam/fastly-logs-iam.test.ts
```diff
@@ -0,0 +1,15 @@
...
+import { SynthUtils } from "@aws-cdk/assert";
+import { simpleGuStackForTesting } from "../../utils/test";
+import { GuFastlyLogsIam } from "./fastly-logs-iam";
+
+describe("The GuFastlyLogsIam pattern", () => {
+  it("correctly wires up the policy", () => {
+    const stack = simpleGuStackForTesting();
+    new GuFastlyLogsIam(stack, "FastlyS3LoggingIam", {
+      bucketName: "test",
+      path: "",
```
akash1810, line 11: Can we avoid the empty string at all?

src/constructs/iam/fastly-logs-iam.ts
```diff
@@ -0,0 +1,68 @@
...
+ * ```typescript
+ * new GuFastlyLogsIam(stack, "FastlyS3LoggingMyApp", {
+ *   bucketName: "gu-mobile-logs"
+ *   path: "fastly/my-app/PROD/*"
+ * })
+ * ```
+ *
+ * See https://docs.fastly.com/en/guides/creating-an-aws-iam-role-for-fastly-logging
+ */
+export class GuFastlyLogsIam {
```
akash1810, line 52: I think this should be extend `GuRole` rather than being a plain class.

src/constants/ssm-parameter-paths.ts
jfsoul: It's a global value, I'll add a link to finding it here

src/constructs/iam/fastly-logs-iam.ts
jfsoul: I've made that change, although it doesn't make a whole lot of sense now because I have a policy being created inside a role constructor. I think semantically it was correct before - a construct formed of 2 resources (policy + role) which both used `Gu` constructs themselves (but the containing construct was its own entity).

Tempted to revert to previous implementation now!

akash1810: Extending `GuRole` follows the current convention of the repo, for example the [`GuInstanceRole`](https://github.com/guardian/cdk/blob/bf89943f4ea655994088870d9caa4e08bd5e2089/src/constructs/iam/roles/instance-role.ts#L36). This also makes me realise that this should probably be called `GuFastlyLogRole` too!

To nitpick a little:

> I think semantically it was correct before - a construct formed of 2 resources

This isn't correct as before we had a _vanilla class_ formed of two resources. In `CDK` there is a [`Construct` class](https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_core.Construct.html) which almost everything extends from 🤯 .

I think generally we'd only create vanilla classes for patterns, rather than constructs - but even then, I think patterns could extend `Construct` (see [this example](https://github.com/aws-samples/aws-cdk-examples/blob/master/typescript/static-site/static-site.ts#L24) of a pattern from AWS).

One of the benefits from using the `Construct` type is you can hook into `validate` more easily. Additionally any exceptions that you might throw in the constructor of a `Construct` get pretty printed with a breadcrumb so it's a little easier to debug.

src/constructs/iam/fastly-logs-iam.ts
```diff
@@ -0,0 +1,57 @@
...
+export class GuFastlyLogsIamRole extends GuRole {
+  constructor(scope: GuStack, props: GuFastlyLogsIamRoleProps) {
+    const fastlyCustomerId = GuFastlyCustomerIdParameter.getInstance(scope).valueAsString;
+    super(scope, "GuFastlyLogsIamRole", {
+      assumedBy: new AccountPrincipal(FASTLY_AWS_ACCOUNT_ID),
+      externalIds: [fastlyCustomerId],
+    });
+    const policy = new GuPutS3ObjectsPolicy(scope, "GuFastlyLogsIamRolePolicy", {
+      bucketName: props.bucketName,
+      paths: props.path ? [props.path] : undefined,
```
akash1810, line 52: minor: I've recently started to set defaults via destructuring and find it more explicit and easier to follow, e.g.

```ts
const { path = "*" } = props;
```

Not massive, but this method would make the description of `path` above more true, specifically `@default - '*'` as at the moment it defaults to undefined, and its `GuPutS3ObjectsPolicy` that is setting it to `*`.

akash1810 (approved):
🎉

PR #762, opened 2021-08-26
fix: Ensure the healthcheck for a Target Group is legal

akash1810:
## What does this change?

A Target Group's healthcheck must have a timeout lower than the interval.

Annoyingly, AWS CDK doesn't catch this and will happily synth a stack that violates this rule. This change checks that `timeout` is less than `interval` and throws otherwise to prevent synth from completing.

See: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-elb-health-check.html#cfn-elb-healthcheck-timeout.

This follows the changes made in #738.

No.

I don't think so.

## How to test

See added test.

Illegal CloudFormation is caught early (at build time rather than runtime).

n/a


akash1810:
This has been fixed upstream now too in v1.121.0 - https://github.com/aws/aws-cdk/pull/16107.

PR #781, opened 2021-09-08
fix: make it possible to define multiple lambdas per stack

amyhughes:
## What does this change?
Since the `ErrorPercentageAlarmForLambda` is hard coded to a value which does not contain a unique identifier per lambda in a stack, it is not currently possible to define more than one per stack. This change
prefixes the `id` so that multiple lambdas may be created per stack.

I don't think so, I think fresh cloudformation including this change will spin up a load of new resources with new ids anyway, but I'm not sure.

No

## How to test

We reference this code version for generating a stack with two lambdas [here](https://github.com/guardian/uk-coronavirus-data-alerts/blob/email-type-verified/package.json#L5) on a branch of the uk-coronavirus-data-alerts project where we define a stack with two lambdas:
image

The cloudformation generated can be successfully deployed:

image

### DIFF
--- a/src/constructs/lambda/lambda.ts
+++ b/src/constructs/lambda/lambda.ts
@@ -75,7 +75,7 @@ export class GuLambdaFunction extends Function {
     });
 
     if (props.errorPercentageMonitoring) {
-      new GuLambdaErrorPercentageAlarm(scope, "ErrorPercentageAlarmForLambda", {
+      new GuLambdaErrorPercentageAlarm(scope, `${id}-ErrorPercentageAlarmForLambda`, {
         ...props.errorPercentageMonitoring,
         lambda: this,
       });


jacobwinch (approved):
Thanks for fixing this 👍 

> I don't think so, I think fresh cloudformation including this change will spin up a load of new resources with new ids anyway, but I'm not sure.

I think the new logical id will cause any existing `GuLambdaErrorPercentageAlarm`s to be replaced (i.e. the old alarm will be deleted and a new one will be created). In the case of alarms this should be totally safe; we'd only need to worry about changing the logical id if it belonged to a stateful resource like a load balancer!

akash1810 (approved):
Amazing! Thanks for the contribution!

I'm a little surprised that no tests need updating with this change, I assume we've just never tested this line before 😱 .

PR #784, opened 2021-09-09
feat: allow GuEc2App users to access VPC

jacobwinch:
## What does this change?

This PR allows users of the `GuEc2App` (and its subclasses) to access the underlying VPC. This means that users can, amongst other things, more easily wire up additional security groups. For example:

```typescript

const playApp = new GuPlayApp(params);

const anotherSg = new GuSecurityGroup(this, "PostgresSecurityGroup", {
  app: "my-app",
  vpc: playApp.vpc,
  allowAllOutbound: false,
  egresses: [{
    range: Peer.anyIpv4(),
    port: 5432,
    description: "My Postgres SG",
  }],
});

playApp.autoScalingGroup.addSecurityGroup(anotherSg)
```

cc @jonathonherbert 

No.

No.

## How to test
I've tested this via [`cdk-playground`](https://github.com/guardian/cdk-playground/compare/jw-test?expand=1).

It should be easier to teams to configure additional resources which require access to the VPC (e.g. security groups).

N/A

### DIFF
--- a/src/patterns/ec2-app.ts
+++ b/src/patterns/ec2-app.ts
@@ -1,6 +1,6 @@
 import type { BlockDevice } from "@aws-cdk/aws-autoscaling";
 import { HealthCheck } from "@aws-cdk/aws-autoscaling";
-import type { IPeer } from "@aws-cdk/aws-ec2";
+import type { IPeer, IVpc } from "@aws-cdk/aws-ec2";
 import { Port } from "@aws-cdk/aws-ec2";
 import { ApplicationProtocol } from "@aws-cdk/aws-elasticloadbalancingv2";
 import { Bucket } from "@aws-cdk/aws-s3";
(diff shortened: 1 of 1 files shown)


akash1810 (approved):
IIRC we've made it so a stack only has one VPC? (this might be a bad decision!). One could also use [`GuVpc.fromIdParameter`](https://github.com/guardian/cdk/blob/37c9eca50ec0fae400b8ee420f354a6f096db18f/src/constructs/ec2/vpc.ts#L70) to access the VPC. Quite a lot of work though, this is far nicer! 🎉

jacobwinch:
I also considered adding a new prop to the pattern to follow an existing convention. Something like:

```typescript
interface GuAdditionalSecurityGroupProps {
    ingresses?: SecurityGroupAccessRule[];
    egresses?: SecurityGroupAccessRule[];
}

export interface GuEc2AppProps extends AppIdentity {
  securityGroupConfiguration?: GuAdditionalSecurityGroupProps; // this is new
  roleConfiguration?: GuInstanceRoleProps; // this already exists
  monitoringConfiguration: Alarms | NoMonitoring; // this already exists
  // other props
}
```

This would allow us to instantiate the `GuSecurityGroup`(s) under the hood, and would avoid users directly interacting with the VPC. However, users would still have to pass the following:

```typescript
securityGroupConfiguration: {
  egresses: [{
    range: Peer.anyIpv4(),
    port: 5432,
    description: "My Postgres SG",
  }]
}
```

...so I decided that it probably wasn't enough of an improvement to justify the change to the pattern.

PR #789, opened 2021-09-09
chore: fix typo

akash1810:
TBF we are _unsure_ about the types of `args` and we need to _ensure_ they are correct before continuing...


jacobwinch (approved):
😁

PR #790, opened 2021-09-10
feat!: Make instanceType a mandatory prop for EC2-based infrastructure

jacobwinch:
## What does this change?

BREAKING CHANGE: The following constructs and patterns now require an `instanceType` prop:

* `GuAutoScalingGroup`
* `GuEc2App`
* `GuPlayApp`
* `GuNodeApp`

Previously these constructs and patterns would automatically create a CloudFormation parameter for the instance type. This parameter has now been removed and the instance type must be provided via the constructor's props. For example:

```typescript
new GuEc2App(stack, {
  applicationPort: GuApplicationPorts.Node,
  app: "test-gu-ec2-app",
  access: { scope: AccessScope.PUBLIC },
  instanceType: InstanceType.of(InstanceClass.T4G, InstanceSize.MEDIUM), //This is now required
  userData: "#!/bin/dev foobarbaz",
  certificateProps: getCertificateProps(),
});
```

Yes, all projects which use these constructs/patterns will need to add the `instanceType` prop, as demonstrated above.

Yes, this PR also updates the docs.

## How to test

Existing unit tests have been amended and should be sufficient.

Whilst configuring the `GuEc2App` pattern with @TomPretty and @jonathonherbert, the fact that the instance type was being configured via a CloudFormation parameter caused some confusion. We felt that providing a default instance type was particularly confusing, as this made it easy to overlook the fact that this parameter was being added automatically by the pattern when reviewing the changes in the console. This PR should also help to make things more explicit.

Losing the CFN parameter also means that another piece of runtime configuration is kept in version control. 

Finally, it removes a manual step (populating this parameter with an appropriate value in the CFN UI) from the migration process.

Although this is breaking change it should be straightforward for all users to adopt it, so I think it's fairly low risk.


src/constructs/autoscaling/asg.ts
```diff
@@ -21,7 +21,6 @@ export interface GuAutoScalingGroupProps
       | "imageId"
       | "osType"
       | "machineImage"
-      | "instanceType"
```
jacobwinch, line 24: This is mandatory on AWS's `AutoScalingGroupProps`, so we no longer need to omit it.

src/patterns/ec2-app.ts
```diff
@@ -369,6 +372,7 @@ export class GuEc2App {
     const autoScalingGroup = new GuAutoScalingGroup(scope, "AutoScalingGroup", {
       app,
       vpc,
+      instanceType: props.instanceType,
```
akash1810, line 375: Should we make this stage aware so that CODE can use a smaller instance than PROD?

sihil: Presumably users can do this themselves without the need for direct support? Perhaps a general "stage switch" function that can be plugged in as a user desires?

akash1810: Sort of. As stage is a parameter, it's a token at compile time, so we have to use CFN Mappings. GuCDK already supports mappings (https://github.com/guardian/cdk/pull/252) 🎉 .

jacobwinch: In my experience some teams/apps vary instance type based on stage and others do not. As we typically run fewer instances (often a single instance) in `CODE` I don't think the cost-savings here are very significant. Consequently, I am unconvinced that we should actively encourage teams to vary instance type based on stage*, as every difference between `PROD` and `CODE` introduces more possibilities for stage-specific bugs. 

As it stands users would need to do [something like this](https://github.com/guardian/cdk-playground/blob/700edcdb5298e966ef084af226c1f5ff291bd37c/cdk/lib/cdk-playground.ts#L23-L34), so I don't think the workaround is too problematic for teams who feel strongly about this.

akash1810: Your reasoning makes sense @jacobwinch and the Prism data (e.g. `prism deploy -f 'specification.instanceType' | sort -k 3`) seems to back it up.

I wonder if it's worth documenting that work around (also happy to leave it until it's needed).

akash1810 (approved):

src/patterns/ec2-app.ts
jacobwinch: I think it would be better to document that in its own right, rather than specifically linking it to this change - perhaps [here](https://github.com/guardian/cdk/tree/main/docs) or in a more general FAQs section? I will do that in a new PR as I've also noticed some other docs that need to be updated whilst thinking about this!

PR #791, opened 2021-09-13
feat: Implement GuScheduledEcsTask pattern

philmcmahon:
## What does this change?
This adds a new pattern to the cdk library - a fargate task that is run on a schedule using a step function. 

The purpose of the step function is to make alarming/alerting easier if the task times out or fails. 

I've written quite a bit of documentation within the code files explaining the different props of this task but shout if something doesn't make sense. 

See [here](https://github.com/guardian/lurch/compare/ingest-scheduler#diff-b02ac84a0d662a31497d1edaa71f9887f7879d2c6088abdaccd8521964a27efa) (private repo so shout if you need access) for some example code using this pattern. That branch is destined for mergification fairly soon.

Hopefully it makes sense for this to live here, otherwise I can dump it back in the lurch project

No

I have added documentation in comments to the task code - maybe I need to add it somewhere else?

## How to test
This is being used already in the lurch CODE and soon to be PROD environments. I've added a basic test file copied from the GuScheduledLambda tests.

A team other than investigations uses this

Is this a generic enough need? In theory yes - anyone wanting to do a scheduled thing that takes longer than the 15 minute lambda time out.


src/patterns/scheduled-ecs-task.test.ts
```diff
@@ -0,0 +1,67 @@
...
+    new GuScheduledEcsTask(stack, {
+      schedule: Schedule.rate(Duration.minutes(1)),
+      containerConfiguration: { id: "node:10" },
+      vpc: vpc,
+      stack: "test",
+      stage: "TEST",
+      app: "ecs-test",
+    });
+    expect(SynthUtils.toCloudFormation(stack)).toMatchSnapshot();
+    expect(stack).toHaveResource("AWS::IAM::Policy");
```
akash1810, line 29: This assertion feels a bit odd - the snapshot assertion above it should provide enough confidence that an IAM Policy has been created? Unless we want to assert on specific properties of the Policy?

src/patterns/scheduled-ecs-task.ts
```diff
@@ -0,0 +1,239 @@
...
+ *     repository: Repository.fromRepositoryArn("<repository arn>"),
+ *     version: '1'
+ * }
+ * ```
+ */
+interface ContainerConfiguration {
+  id?: string;
+  repository?: IRepository;
+  version?: string;
+}
```
akash1810, line 51: Should all these properties be optional? Would it be clearer to make the usage of `ContainerConfiguration` be of type `ContainerConfiguration | undefined`?

src/patterns/scheduled-ecs-task.ts
```diff
@@ -0,0 +1,239 @@
...
+ * Note that if your task reliably completes in less than 15 minutes then you should probably use a [[`GuScheduledLambda`]] instead. This
+ * pattern was mainly created to work around the 15 minute lambda timeout.
+ */
+export class GuScheduledEcsTask {
+  constructor(scope: GuStack, props: GuScheduledEcsTaskProps) {
+    const distBucket = GuDistributionBucketParameter.getInstance(scope).valueAsString;
+
+    const timeout = props.taskTimeoutInMinutes ?? 15;
+
+    const cpu = props.cpu ?? 2048;
```
akash1810, line 145: `2048` is a massive CPU (does that even exist?!). Is this correct?

src/patterns/scheduled-ecs-task.ts
```diff
@@ -0,0 +1,239 @@
...
+    const memory = props.memory ?? 4096;
```
akash1810, line 146: minor: pulling this (and the above default values) into constants might be more maintainable. For example:

```typescript
const defaultMemory = 4 * 1024; // 4GB
```

src/patterns/scheduled-ecs-task.ts
```diff
@@ -0,0 +1,239 @@
...
+
+    const cluster = new Cluster(scope, `${props.app}-cluster`, {
```
akash1810, line 148: Could we use the [`AppIdentity.suffixText`](https://github.com/guardian/cdk/blob/d1933e39a5d12395c17f236406a2905f35d04582/src/constructs/core/identity.ts#L16) helper here (and other places where `app` is used for the construct ID) for consistency with the rest of the repository? (Again, we've got some work in the backlog to somewhat automate this away as it's annoying to remember!).

src/patterns/scheduled-ecs-task.ts
```diff
@@ -0,0 +1,239 @@
...
+      definition: task,
+      stateMachineName: `${props.app}-${props.stage}`,
+    });
+
+    new Rule(scope, `${props.app}-schedule-rule`, {
+      schedule: props.schedule,
+      targets: [new SfnStateMachine(stateMachine)],
+    });
+
+    if (props.alarmSnsTopicArn) {
```
akash1810, line 200: Other [constructs](https://github.com/guardian/cdk/blob/55c689e584b08e600772eea74f624112f34ec7f7/src/patterns/api-lambda.ts#L35) and patterns follow a slightly different approach here; alarms are always on unless `noMonitoring: true`. Could we take a similar approach here?

akash1810 (commented):
A few questions added.

Similar to #761, I'd be interested to see how these commits get parsed (https://github.com/guardian/cdk/pull/761#issuecomment-913443651) and improve the docs afterwards.

src/patterns/scheduled-ecs-task.ts
```diff
@@ -0,0 +1,239 @@
...
+    const taskDefinition = new TaskDefinition(scope, `${props.app}-task-definition`, {
+      compatibility: Compatibility.FARGATE,
+      cpu: cpu.toString(),
+      memoryMiB: memory.toString(),
+    });
+
+    taskDefinition.addContainer("scheduled-task-container", {
+      image: getContainer(props.containerConfiguration),
+      entryPoint: props.taskCommand ? ["/bin/sh"] : undefined,
+      command: props.taskCommand ? ["-c", `${props.taskCommand}`] : undefined,
```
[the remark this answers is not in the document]
philmcmahon: `command` is optional as the container might have a 'CMD' already specified in the docker file. Really I don't want to set it at all if `taskCommand` is undefined. Do you think there's a better way of doing this?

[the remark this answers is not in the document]
akash1810: > Do you think there's a better way of doing this?

The only thing I can think of is to read and parse the docker file, but that doesn't sound especially sensible or worth the effort.

To make it more obvious for future travellers, maybe we can add an inline comment here to say the `CMD` directive in the Dockerfile will be used if `taskCommand` hasn't been set.

src/patterns/scheduled-ecs-task.ts
itsibitzi: fwiw - fargate is provisioning in "cpu value", 2048 is equivalent to 2 vCPUs.

src/patterns/scheduled-ecs-task.ts
philmcmahon: Good point. I think I've made things clearer in https://github.com/guardian/cdk/pull/791/commits/cb6a6fb20db19655d6410080a996c10b06517856

src/patterns/scheduled-ecs-task.ts
```diff
@@ -0,0 +1,239 @@
...
+      memoryLimitMiB: props.memory,
+      logging: LogDrivers.awsLogs({
+        streamPrefix: props.app,
+        logRetention: 14,
+      }),
+    });
+
+    const fetchArtifact = new PolicyStatement();
+    fetchArtifact.addActions("s3:GetObject");
+    fetchArtifact.addResources(`arn:aws:s3:::${distBucket}/*`);
```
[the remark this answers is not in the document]
philmcmahon: I wasn't certain what to do there - LMK if this makes sense https://github.com/guardian/cdk/pull/791/commits/068a9dcaaee2092e312204ae929f8a2e52925630

src/patterns/scheduled-ecs-task.ts
philmcmahon: I've pulled out and added a bit more documentation

src/patterns/scheduled-ecs-task.ts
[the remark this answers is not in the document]
philmcmahon: Sounds good https://github.com/guardian/cdk/pull/791/commits/d0bdd6ae48424de98e8f116dffff27fb7b4b7b54

philmcmahon:
@adamnfish we had a stab at some better tests - dropped the simple snapshot one and instead tested for a couple of things that we care about. It could be more thorough but I think this is a good start

src/patterns/scheduled-ecs-task.ts
philmcmahon: 👍🏻 addressed in https://github.com/guardian/cdk/pull/791/commits/01099d273e7a80db9c294f2ccb1b28c0fbb569fd

src/patterns/scheduled-ecs-task.ts
akash1810: > 2048 is equivalent to 2 vCPUs

I don't understand, but happy to blindly accept.

src/patterns/scheduled-ecs-task.ts
```diff
@@ -0,0 +1,255 @@
...
+  securityGroups?: ISecurityGroup[];
+  customTaskPolicies?: PolicyStatement[];
+}
+
+/**
+ * Containers can be specified either via a repository and version or by container id (in which case it will
+ * be fetched from docker hub).
+ */
+const getContainer = (config: ContainerConfiguration) => {
+  if (config.type === "repository") {
```
akash1810, line 131: I think a [user defined type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards) (essentially duck typing) would be more idomatic here.

src/patterns/scheduled-ecs-task.ts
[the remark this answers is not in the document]
akash1810: Looks good. We can probably make `GuGetS3ObjectsPolicy` a little DRYer now too by re-using `GuGetDistributablePolicyStatement`. Happy for that to be done in a separate PR.

src/patterns/scheduled-ecs-task.ts
akash1810: Much nicer! 👍🏽

akash1810 (commented):
The changes look great. Added a small comment around type guards. Once the build passes, happy to approve - the linter is currently failing 😢  we'd recommend configuring your IDE to run eslint and prettier on save.

src/patterns/scheduled-ecs-task.ts
philmcmahon: Thanks Akash, I attempted this here https://github.com/guardian/cdk/pull/791/commits/dfe51f534e9e61e4de69888e33056948f8dd16d3 but vscode is complaining about `(config as RepositoryContainer).version !== undefined` being an unncessary check. Would you have implemented it differently?

akash1810: Hmm, I think this is because `ContainerConfiguration` is a union type? If you make `config` an `any` type it should be happier (something similar [here](https://github.com/guardian/cdk/blob/1f1735d3eea11686e24cccb6ee6bdc2aef661362/src/types/migrating.ts#L14-L16)).

Having said that, [these docs](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html) follow the same pattern as before (switching on  the `type` field). I'm not sure what's best anymore! I'd go for whatever is easiest, so probably the switching. Sorry!

philmcmahon: Thanks! I'm going to be lazy as at least it's quite readable - those docs were the ones I followed originally

philmcmahon:
Looks like the build is happy now. I've attempted to shorten the commit history a bit too and added the relevant `chore/fix` tags

akash1810 (approved):
Amazing, thanks for the contribution!

src/patterns/scheduled-ecs-task.ts
philmcmahon: Sorry I did reply to this but I think my internet broke when I hit comment. Thanks for the feedback! I decided to leave it as is - isSomething is a more readable notation but in this instance where we're only checking it once doing the type == thing seems ok to me - or at least that's how I justified my laziness

PR #804, opened 2021-09-21
fix: Don't tag the entire stack with pattern app tag in GuScheduledEcsTask

philmcmahon:
## What does this change?
This removes the AppIdentity.taggedConstruct call from `GuScheduledEcsTask`. On Lurch this caused some problems as we have multiple apps in one stack, and `AppIdentity.taggedConstruct` replaced all of those tags with the app tag of our `GuScheduledEcsTask`. We're doing this https://github.com/guardian/lurch/blob/main/packages/infra/LurchCDK.ts#L359 to work around it at the moment but that's not very good!

Just Lurch

Gu Patterns should not interfere with other resources within the same stack

Tagging will be less thorough :(

### DIFF
--- a/src/patterns/__snapshots__/scheduled-ecs-task.test.ts.snap
+++ b/src/patterns/__snapshots__/scheduled-ecs-task.test.ts.snap
@@ -31,6 +31,10 @@ Object {
         "ClusterName": "ecs-test-cluster-TEST",
         "Tags": Array [
           Object {
+            "Key": "App",
+            "Value": "ecs-test",
+          },
+          Object {
             "Key": "gu:cdk:version",
             "Value": "TEST",
           },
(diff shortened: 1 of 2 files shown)


jacobwinch (approved):
Good spot! I think we might be making the same mistake in the `GuEc2App` pattern: https://github.com/guardian/cdk/blob/a01cd04a015fbcce343ebcef68ea92bd0ca5b3b4/src/patterns/ec2-app.ts#L348

If you want to retain the tagging I think it would be safe to apply this at the construct level within the pattern e.g.

```typescript
const cluster = new Cluster(scope, AppIdentity.suffixText(props, "Cluster"), {
  clusterName: `${props.app}-cluster-${props.stage}`,
  enableFargateCapacityProviders: true,
  vpc: props.vpc,
});

AppIdentity.taggedConstruct({ app: props.app }, cluster);
```

philmcmahon:
Thanks Jacob! Sorry to miss this yesterday. I'll push up some changed with the tagging at construct level

jacobwinch (approved):
Great work 👍

PR #805, opened 2021-09-21
fix: Don't tag the entire stack with pattern app tag in GuEc2App

jacobwinch:
This fixes the problem reported in https://github.com/guardian/cdk/pull/804 for the `GuEc2App` pattern (and subclasses).

### DIFF
--- a/src/patterns/__snapshots__/ec2-app.test.ts.snap
+++ b/src/patterns/__snapshots__/ec2-app.test.ts.snap
@@ -830,10 +830,6 @@ Object {
         ],
         "Tags": Array [
           Object {
-            "Key": "App",
-            "Value": "test-gu-ec2-app",
-          },
-          Object {
             "Key": "gu:cdk:version",
             "Value": "TEST",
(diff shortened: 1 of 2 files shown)


src/patterns/__snapshots__/ec2-app.test.ts.snap
```diff
@@ -829,10 +829,6 @@ Object {
           },
         ],
         "Tags": Array [
-          Object {
```
jacobwinch, line 832: The app tag is only being removed from `WazuhSecurityGroup`s. This construct is [implemented as a singleton anyway](https://github.com/guardian/cdk/blob/a12758512cc6942cfe8234a8ef413edb04596e0c/src/constructs/ec2/security-groups/wazuh.ts#L9-L11) (i.e. it might be shared by multiple apps in the same stack) so it doesn't make sense to add the app tag here in all cases anyway!

philmcmahon (approved):
LGTM!

PR #807, opened 2021-09-22
feat: Perform VPC analysis in account-readiness

akash1810:
## What does this change?

GuCDK uses values in SSM parameter store to know into which VPC resources should be launched.

If the parameters do not exist, GuCDK will produce a template that cannot be executed.

This change adds some VPC analysis to the `account-readiness` CLI command. It:
  - Finds all VPCs in the AWS region
  - Determines if the VPC is in use (logic copied from [Prism](https://github.com/guardian/prism/blob/a9f3ecf47a1f2fd4361f61c39832c936a170b6f7/app/collectors/vpc.scala))
  - Expects three SSM parameters to exist for each in use VPC:
    1. `^/account/vpc/([A-z0-9.-_])+/id$`
    2. `^/account/vpc/([A-z0-9.-_])+/subnets/private$`
    3. `^/account/vpc/([A-z0-9.-_])+/subnets/public$`

Typically, an account has one in use VPC.

image

This _is_ the CLI 😈 .

No.

## How to test

- Checkout branch
- Run `./script/cli account-readiness --profile deployTools`
- Observe output

Migrating to GuCDK is that much easier.

### DIFF
--- a/src/bin/commands/account-readiness/vpc.ts
+++ b/src/bin/commands/account-readiness/vpc.ts
@@ -1,0 +1,58 @@
+import AWS from "aws-sdk";
+import chalk from "chalk";
+import type { AwsAccountReadiness } from "../../../types/cli";
+import { doesDefaultVpcSsmParameterExist, getSsmParametersForVpc, getVpcsInDetail } from "../../../utils/cli/vpc";
+
+const UNKNOWN = "unknown";
+
+export const vpcReadiness = async ({ credentialProvider, region }: AwsAccountReadiness): Promise<number> => {
+  const ec2Client = new AWS.EC2({
+    credentialProvider,
+    region,
+  });
+  const ssmClient = new AWS.SSM({
+    credentialProvider,
+    region,
+  });
+
+  const vpcs = await getVpcsInDetail(ec2Client);
+  const inUseVpcs = vpcs.filter((vpc) => vpc.isUsed);
+  const ssmParamsFromAccount = await getSsmParametersForVpc(ssmClient);
+  const usingDeprecatedSsmParameter = doesDefaultVpcSsmParameterExist(ssmParamsFromAccount);
+
+  console.group(chalk.bold("\nVPC Summary"));
+
+  if (usingDeprecatedSsmParameter) {
+    console.log(
+      `❌ SSM parameters with path '/account/vpc/default/*' found. Please rename to '/account/vpc/primary/*'`
+    );
+  }
+
+  const totalSsmParametersExpected = inUseVpcs.length * 3;
+  const hasExpectedSsmParameters = ssmParamsFromAccount.length === totalSsmParametersExpected;
+
+  if (!hasExpectedSsmParameters) {
+    console.log(
+      `❌ Expected to find ${chalk.bold(
+        totalSsmParametersExpected
+      )} SSM Parameters (3 per in use VPC) but found ${chalk.bold(ssmParamsFromAccount.length)}`
+    );
+  }
+
+  const vpcsForLogging = vpcs.map((vpc) => ({
+    VpcId: vpc.VpcId ?? UNKNOWN,
+    Region: region,
+    IsAwsDefaultVpc: vpc.IsDefault,
+    InUse: vpc.isUsed,
+    CidrBlock: vpc.CidrBlock,
+  }));
+  console.table(vpcsForLogging);
+
+  console.groupEnd();
+
+  if (!usingDeprecatedSsmParameter && hasExpectedSsmParameters) {
+    return Promise.resolve(0);
+  } else {
+    return Promise.resolve(1);
+  }
+};
--- a/src/bin/commands/account-readiness/vpc.ts
+++ b/src/bin/commands/account-readiness/vpc.ts
@@ -1,5 +1,6 @@
 import AWS from "aws-sdk";
 import chalk from "chalk";
+import { LibraryInfo } from "../../../constants/library-info";
 import type { AwsAccountReadiness } from "../../../types/cli";
 import { doesDefaultVpcSsmParameterExist, getSsmParametersForVpc, getVpcsInDetail } from "../../../utils/cli/vpc";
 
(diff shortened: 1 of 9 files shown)


src/bin/commands/account-readiness/vpc.ts
```diff
@@ -0,0 +1,58 @@
...
+      `❌ Expected to find ${chalk.bold(
+        totalSsmParametersExpected
+      )} SSM Parameters (3 per in use VPC) but found ${chalk.bold(ssmParamsFromAccount.length)}`
+    );
+  }
+
+  const vpcsForLogging = vpcs.map((vpc) => ({
+    VpcId: vpc.VpcId ?? UNKNOWN,
+    Region: region,
+    IsDefault: vpc.IsDefault,
```
jacobwinch, line 45: Minor: since default has become a bit overloaded in this context, would `IsAwsDefaultVpc` (or similar) be slightly clearer here?

src/bin/commands/account-readiness/vpc.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  console.group(chalk.bold("\nVPC Summary"));
+
+  if (usingDeprecatedSsmParameter) {
+    console.log(
+      `❌ SSM parameters with path '/account/vpc/default/*' found. Please rename to '/account/vpc/primary/*'`
+    );
+  }
+
+  const totalSsmParametersExpected = inUseVpcs.length * 3;
```
jacobwinch, line 31: I'm wondering if this assumption is correct. Currently the CLI complains about the Deploy Tools account because we don't have _enough_ SSM parameters. I believe this is because the default VPC is technically in use at the moment, but it is not referenced in SSM? 

IIUC, we don't want teams to use the default VPC at all? However, the CLI output implies that to 'fix' things, I could quickly add SSM parameters which point to the default VPC. 

Although there is another potential action here (migrate everything away from the default VPC!), this is non-trivial and is also not strictly required in order to start using `@guardian/cdk` (in fact I might be able to use the library to help me to spin up new infrastructure in the correct VPC as part of a migration).

So perhaps it would be more appropriate to check:

* The number of SSM parameters == `inUseNonDefaultVpcs.length * 3`

And (for bonus points!):

* There are no `/account/vpc/*` SSM parameters that point to the default VPC

jacobwinch:
> Typically, an account has one in use VPC.

It would be good to get a count of accounts in this state (where the one in use VPC is not the default). If this is the majority of accounts then it would allow us to manually check the shape of the VPCs and (if correct), set up the SSM parameters without needing to work with teams on this individually.

src/bin/commands/account-readiness/vpc.ts
akash1810: > So perhaps it would be more appropriate to check:
>
> - The number of SSM parameters == inUseNonDefaultVpcs.length * 3

Yes I think this makes sense.

You've also reminded me of a further check that needs implementing:
  - If there are multiple VPCs in use, a param of `/account/vpc/primary/*` should not exist

I'm also not entirely happy by hard-coding `3` here and would prefer for it to be a `.length` of something. Will have a think.

akash1810:
> > Typically, an account has one in use VPC.
> 
> Not strictly related to this PR, but it would be good to get a count of accounts in this state (where the one in use VPC is not the default). If this is the majority of accounts then it would allow us to manually check the shape of the VPCs and (if correct), quickly set up the SSM parameters without needing to discuss this with teams.

@sihil produced [this](https://docs.google.com/spreadsheets/d/14CRwzIKS-boZAA74NUOWlAdT7F4nHDdH0q_DU4rS7Pc/edit#gid=1560011512) wonderful Sheet previously, which I think helps answer this?

src/bin/commands/account-readiness/vpc.ts
jacobwinch: > You've also reminded me of a further check that needs implementing:
> 
> * If there are multiple VPCs in use, a param of `/account/vpc/primary/*` should not exist

I agree with this in principle, but in practice it's also possible that there are 2 VPCs currently in use but the team are trying to consolidate into one, so I'd be inclined to chat with teams about this on a case-by-case basis rather than encoding it as a strict rule.

akash1810: 👍🏽  (not only because it's less work)

akash1810: > So perhaps it would be more appropriate to check:
> 
> - The number of SSM parameters == inUseNonDefaultVpcs.length * 3
>
> And (for bonus points!):
> 
> - There are no /account/vpc/* SSM parameters that point to the default VPC

Thinking about this a little more, this would mean accounts would need to migrate away from the default VPC first in order to adopt GuCDK. That's quite a lot of work! According to Prism, we have 11 accounts using the default VPC - 1/3 of all our accounts!

What if we log/warn if:
  - Default VPCs are used
  - SSM Parameters point to default VPCs

jacobwinch: Yes I think that's a sensible compromise 👍

akash1810: Updated now.

src/bin/commands/account-readiness/vpc.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  if (isDefaultVpcInUse) {
+    console.warn(
+      `⚠️ The default VPC is in use. Whilst you're still able to use ${LibraryInfo.NAME}, it's recommended to use a custom VPC. See the docs for more information.`
+    );
+  }
+
+  if (isDefaultVpcInSsmParameters) {
+    console.warn(
+      `⚠️ The default VPC is referenced within SSM Parameters. Whilst you're still able to use ${LibraryInfo.NAME}, it's recommended to use a custom VPC. See the docs for more information.`
```
akash1810, line 77: TODO: create these docs!

jacobwinch (approved):

PR #815, opened 2021-09-22
feat: add CDK constructs for Guardian::DNS::RecordSet

jacobwinch:
## What does this change?

This PR allows users to create DNS records in NS1 via `@guardian/cdk`.

No, this is new functionality. Some projects will now be able to 'inherit' their existing DNS records, but this is not mandatory.

Yes, I've added some minimal documentation.

## How to test

I've added unit tests for this. I've also confirmed that the CFN that gets produced can create a `Guardian::DNS::RecordSet` in the Developer Playground.

Apps which use `@guardian/cdk` to define their infrastructure can now create and manage CNAMEs in their source code, with minimal boilerplate.

I think the risks associated with this change are managed via https://github.com/guardian/cfn-private-resource-types. This PR just makes it easier to use the private resource type!

### DIFF
--- a/src/constructs/dns/__snapshots__/dns-records.test.ts.snap
+++ b/src/constructs/dns/__snapshots__/dns-records.test.ts.snap
@@ -1,0 +1,65 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`The GuCname construct should create the correct resources with minimal config 1`] = `
+Object {
+  "Parameters": Object {
+    "Stage": Object {
+      "AllowedValues": Array [
+        "CODE",
+        "PROD",
+      ],
+      "Default": "CODE",
+      "Description": "Stage name",
+      "Type": "String",
+    },
+  },
+  "Resources": Object {
+    "TestRecord": Object {
+      "Properties": Object {
+        "Name": "banana.example.com",
+        "RecordType": "CNAME",
+        "ResourceRecords": Array [
+          "apple.example.com",
+        ],
+        "Stage": Object {
+          "Ref": "Stage",
+        },
+        "TTL": 3600,
+      },
+      "Type": "Guardian::DNS::RecordSet",
+    },
+  },
+}
+`;
+
+exports[`The GuDnsRecordSet construct should create the correct resources with minimal config 1`] = `
+Object {
+  "Parameters": Object {
+    "Stage": Object {
+      "AllowedValues": Array [
+        "CODE",
+        "PROD",
+      ],
+      "Default": "CODE",
+      "Description": "Stage name",
+      "Type": "String",
+    },
+  },
+  "Resources": Object {
+    "TestRecord": Object {
+      "Properties": Object {
+        "Name": "banana.example.com",
+        "RecordType": "CNAME",
+        "ResourceRecords": Array [
+          "apple.example.com",
+        ],
+        "Stage": Object {
+          "Ref": "Stage",
+        },
+        "TTL": 3600,
+      },
+      "Type": "Guardian::DNS::RecordSet",
+    },
+  },
+}
+`;
--- a/src/constructs/dns/dns-records.test.ts
+++ b/src/constructs/dns/dns-records.test.ts
@@ -1,5 +1,6 @@
 import "@aws-cdk/assert/jest";
 import { SynthUtils } from "@aws-cdk/assert";
+import { Duration } from "@aws-cdk/core";
 import { simpleGuStackForTesting } from "../../utils/test";
 import { GuCname, GuDnsRecordSet, RecordType } from "./dns-records";
 
@@ -10,21 +11,9 @@ describe("The GuDnsRecordSet construct", () => {
       name: "banana.example.com",
       recordType: RecordType.CNAME,
       resourceRecords: ["apple.example.com"],
+      ttl: Duration.hours(1),
     });
     expect(SynthUtils.toCloudFormation(stack)).toMatchSnapshot();
-  });
-
-  it("should allow users to override the default TTL", () => {
-    const stack = simpleGuStackForTesting();
-    new GuDnsRecordSet(stack, "TestRecord", {
-      name: "banana.example.com",
-      recordType: RecordType.CNAME,
-      resourceRecords: ["apple.example.com"],
-      ttl: 123,
-    });
-    expect(stack).toHaveResourceLike("Guardian::DNS::RecordSet", {
-      TTL: 123,
-    });
   });
 });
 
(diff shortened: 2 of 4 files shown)


src/constructs/dns/dns-records.ts
```diff
@@ -0,0 +1,57 @@
...
+}
+
+/**
+ * Do not use this construct directly. Use [[`GuCname`]] instead in order to reduce boilerplate.
+ */
+export class GuDnsRecordSet {
+  constructor(scope: GuStack, id: string, props: GuDnsRecordSetProps) {
+    // The spec for this private resource type can be found here:
+    // https://github.com/guardian/cfn-private-resource-types/tree/main/dns/guardian-dns-record-set-type/docs#syntax
+    new CfnResource(scope, id, {
```
jacobwinch, line 22: Based on https://docs.aws.amazon.com/cdk/latest/guide/use_cfn_public_registry.html#use_cfn_public_registry_add

src/constructs/dns/dns-records.ts
```diff
@@ -0,0 +1,57 @@
...
+      },
+    });
+  }
+}
+
+export interface GuCnameProps {
+  /** The name of the record, for example: xyz.example.com */
+  name: string;
+  /** The record your CNAME should point to, for example your Load Balancer DNS name */
+  resourceRecord: string;
```
jacobwinch, line 39: Since you can only have one `CNAME` for a given name, we use different props to prevent users from making mistakes. An alternative would be to throw an error inside `GuDnsRecordSet`, but preventing these mistakes via the type system felt more appropriate to me.

src/constructs/dns/dns-records.test.ts
```diff
@@ -0,0 +1,40 @@
...
+    expect(stack).toHaveResourceLike("Guardian::DNS::RecordSet", {
+      TTL: 123,
+    });
+  });
+});
+
+describe("The GuCname construct", () => {
+  it("should create the correct resources with minimal config", () => {
+    const stack = simpleGuStackForTesting();
+    new GuCname(stack, "TestRecord", {
```
jacobwinch, line 34: Perhaps it's overkill to have a separate `GuCname` class, but it allows us to eliminate a bit of boilerplate, as shown here

src/constructs/dns/dns-records.ts
```diff
@@ -0,0 +1,57 @@
...
+  constructor(scope: GuStack, id: string, props: GuDnsRecordSetProps) {
+    // The spec for this private resource type can be found here:
+    // https://github.com/guardian/cfn-private-resource-types/tree/main/dns/guardian-dns-record-set-type/docs#syntax
+    new CfnResource(scope, id, {
+      type: "Guardian::DNS::RecordSet",
+      properties: {
+        Name: props.name,
+        ResourceRecords: props.resourceRecords,
+        RecordType: props.recordType,
+        TTL: props.ttl ?? 3600,
```
jacobwinch, line 28: This seemed like a sensible default to me but other opinions are welcome.

src/constructs/dns/dns-records.ts
```diff
@@ -0,0 +1,57 @@
+import { CfnResource } from "@aws-cdk/core";
+import type { GuStack } from "../core";
+
+export enum RecordType {
+  CNAME = "CNAME",
```
jacobwinch, line 5: Using an `enum` in the hope that we support more record types soon 😅

akash1810:
This is amazing!!

I think this creates a small issue where we're creating a resource for a private repository now. Should we see if we can open the private resource type? Or wire this construct in such a way that it can be omitted? Or address it later?

src/constructs/dns/dns-records.ts
sihil: I am not convinced that we should have a default. Perhaps one option might be to provide some presets with purposes?

jacobwinch: Some kind of presets/guidance would certainly be helpful. I would guess that most users (and I include myself in this group!) won't know what a sensible choice is here.

Do you think it makes sense to have defaults for different record types (i.e. provide a default for CNAMEs) or is that still unhelpful?

sihil: NS records should certainly be longer. But I'm not entirely convinced that there will be a use case for that with a single plane of control. It's generally about purpose. We could have a default of 3600 on the GuCname type. Think I'd be somewhat happier with that, but again we'd likely want to reduce it during migrations.

src/constructs/dns/dns-records.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+}
+
+export interface GuCnameProps {
+  /** The name of the record, for example: xyz.example.com */
+  name: string;
+  /** The record your CNAME should point to, for example your Load Balancer DNS name */
+  resourceRecord: string;
+  /** The time to live in seconds for the DNS record - this will be set to 3600 seconds (1 hour) by default */
+  ttl?: number;
```
akash1810, line 41: Would it be appropriate to use AWS's [`Duration`](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_core.Duration.html#static-hoursamount) type here? It becomes a bit easier to define an hour - `Duration.hours(1)`.

src/constructs/dns/dns-records.ts
akash1810: I wonder if we provide an enum or similar:

```typescript
const TTLs = {
  InMigrationMode: 60,
  StableApplication: 3600
}
```

The intention being to stop you having to think what a sensible value should be. (Obvs. better names needed!)

akash1810 (commented):
LGTM

src/constructs/dns/dns-records.ts
jacobwinch: Nice idea 👍

src/constructs/dns/dns-records.ts
jacobwinch: > We could have a default of 3600 on the GuCname type. Think I'd be somewhat happier with that, but again we'd likely want to reduce it during migrations.

I've gone for applying a default at `GuCname` level for now. I agree that we will need to drop this for migrations, but in order to adopt the record in the first place users will need to match the TTL that's currently configured in NS1 (and hopefully this will frequently align with the 1 hour default!)

jacobwinch:
> I think this creates a small issue where we're creating a resource for a private repository now. Should we see if we can open the private resource type? Or wire this construct in such a way that it can be omitted? Or address it later?

I think we'll need to address this when we link this construct up to our patterns - users will probably need to be able to opt out of DNS record creation (e.g. if they don't use NS1 for DNS, or if they are non-Guardian users).

akash1810 (approved):
Amazing work!

img

src/constructs/dns/__snapshots__/dns-records.test.ts.snap
```diff
@@ -0,0 +1,65 @@
...
+        "CODE",
+        "PROD",
+      ],
+      "Default": "CODE",
+      "Description": "Stage name",
+      "Type": "String",
+    },
+  },
+  "Resources": Object {
+    "TestRecord": Object {
```
akash1810, line 49: This logicalId doesn't have a GUID appended to it like the native CDK constructs. Is this likely to cause any issues?

PR #819, opened 2021-09-24
fix: include app in scheduled ecs pattern alarm names 

philmcmahon:
## What does this change?
This adds the app-stage prefix to alarms generated by the scheduled ecs task pattern. Without this it is not possible to have more than one GuScheduledEcsTask in the same stack

Currently this pattern is only in use by Lurch so all OK I think!

No

## How to test
I've tested this using the lurch stack

Being able to have >1 guscheduledecstask in a stack


philmcmahon:
Thanks @akash1810 ! Sorry for slow response - have pushed a test up now.

akash1810 (approved):
Lovely 👍🏽 .

PR #828, opened 2021-09-29
chore: Swap `actions/setup-node` for `guardian/actions-setup-node`

akash1810:
## What does this change?

[`guardian/actions-setup-node`](https://github.com/guardian/actions-setup-node) is a fork of `actions/setup-node` and adds node version auto-discovery. Using it results in simpler GitHub workflows.

It also makes changing the node version as simple as updating `.nvmrc`, keeping `node-version` within the workflow files in sync with `.nvmrc`.

Additionally `actions/setup-node` (and thus ``guardian/actions-setup-node`) [uses `actions/cache` under the hood](https://github.com/actions/setup-node#caching-packages-dependencies). This means we can further simplify the workflows.

No.

No.

## How to test

CI should continue to function.

Simpler configuration and a simpler story for node upgrades in the future.

We introduced caching to solve a specific issue (#532). Does this change impact that? Hopefully not as, ultimately, we still use `actions/cache`.

We'll need to update our branch protection rule as the check `CI (14.15.1)` will no longer exist and is replaced with `CI / CI (pull_request)`.

image

### DIFF
--- a/.github/workflows/cd.yaml
+++ b/.github/workflows/cd.yaml
@@ -7,21 +7,12 @@ on:
 jobs:
   CD:
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        node-version: [14.15.1]
     steps:
       - uses: actions/checkout@v2
-      - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v2.4.0
+      - name: Setup node
+        uses: guardian/actions-setup-node@v2.4.1
         with:
-          node-version: ${{ matrix.node-version }}
-      - uses: actions/cache@v2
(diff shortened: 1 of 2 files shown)


jacobwinch (approved):
Great work! 

> We'll need to update our branch protection rule as the check CI (14.15.1) will no longer exist and is replaced with CI / CI (pull_request).

Feel free to bypass the checks / change them over now in order to get this merged 👍

akash1810:
This has unearthed a curious [issue](https://github.com/guardian/cdk/runs/3754735385?check_suite_focus=true) not seen before:

image

#829 should fix this.

PR #829, opened 2021-09-30
chore: Update node to 14.18.0 (latest LTS)

akash1810:
## What does this change?

Updates node to the [latest LTS](https://nodejs.org/en/about/releases/).

No.

No.

## How to test

CI should continue to pass.

Staying up to date.

n/a


jacobwinch:
It looks like we are also setting the node version in our workflow files e.g.

https://github.com/guardian/cdk/blob/060d4f73ea2c2fe9d58aeccf32cc2be8fe208fc9/.github/workflows/cd.yaml#L10-L18

Perhaps we should move to https://github.com/guardian/actions-setup-node as part of this change?

akash1810:
> It looks like we are also setting the node version in our workflow files e.g.
> 
> https://github.com/guardian/cdk/blob/060d4f73ea2c2fe9d58aeccf32cc2be8fe208fc9/.github/workflows/cd.yaml#L10-L18
> 
> Perhaps we should move to https://github.com/guardian/actions-setup-node as part of this change?

I should have mentioned this PR builds on top of #828, which makes that move.

jacobwinch:
> I should have mentioned this PR builds on top of #828, which makes that move.

Ah yes, sorry, I see that now!

jacobwinch (approved):
👍

PR #839, opened 2021-10-08
Experiment with a better way to map a Record<Stage, V>

sihil:
Is something like this useful?


akash1810:

akash1810:
This has gotten a little too stale. Closing.

PR #840, opened 2021-10-08
feat!: Make mandatory dns constructor in ec2-app pattern

michaelwmcnamara:
## What does this change?
BREAKING CHANGE: changed the properties in GuEc2AppProps

This change adds the ability to automatically setup the DNS with in the EC2App pattern. 

Yes: To migrate, change certificateProps to domainNameProps in the EC2App pattern.

The migration guide needs to be updated apropriately:

## How to test
run all scripted tests

EC2 apps are starting up succesfully and do not need to have their domain names defined in a .yaml file.

Accounts that do not have private resource type will not be able to use this version


jacobwinch:
I'm going to close this as we opted to change the approach in the end (see https://github.com/guardian/cdk/discussions/913 and https://github.com/guardian/cdk/pull/912). Feel free to re-open or add comments to the discussion if you disagree with this @michaelwmcnamara!

PR #871, opened 2021-10-29
chore: Add a composite GitHub Action for CI of GuCDK infrastructure

akash1810:
## What does this change?

[Composite Actions](https://github.blog/changelog/2020-08-07-github-actions-composite-run-steps/) allow us to simplify workflows, specifically by encapsulating logic in a single place for use elsewhere.

This change introduces a composite action for CDK, it will:
- Install dependencies
- Run tests
- Synth the stack to disk

This action requires there to be a `build`, `lint`, `test` and `synth` script in a `package.json` file. At first glance this is a bit fragile, however with the upcoming changes to GuCDK project creation, we can guarantee the presence of them.

This vastly simplifies downstream repos, for example:
- https://github.com/guardian/tag-janitor/pull/230 (a yarn project)
- https://github.com/guardian/cdk-playground/pull/110 (a npm project)

That is, one can simply add this to a GHA workflow:

```yaml
- uses: guardian/cdk@main
```

We can later combine this with [reusable workflows](https://docs.github.com/en/actions/learn-github-actions/reusing-workflows) to simplify CI of repositories even further.

Sort of. It commits us to adding a number of scripts to a `package.json` when we start the project generation work.

No.

## How to test

See above linked example PRs, their CI should pass.

DRYer GHA for CI definitions in repositories.

The advice given [here](https://github.com/actions/toolkit/blob/master/docs/action-versioning.md) is to:

> Warning: do not reference `main` since that is the latest code and can be carrying breaking changes of the next major version.

We publish releases for this repository fairly frequently. If a downstream repository has configured Dependabot to run against GHA workflows, it is likely to get multiple PRs a week.

For this reason, the example usage is `guardian/cdk@main`, committing us to not releasing breaking changes to it. Given its relative simplicity, I think that's possible to fulfil.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -1,0 +1,54 @@
+# A composite GitHub Action, performing CI of GuCDK infrastructure.
+#
+# Supports npm and yarn.
+# Usage when using npm:
+# - uses: guardian/cdk@main
+#
+# Usage when using yarn:
+# - uses: guardian/cdk@main
+#   with:
+#     packageManager: yarn
+#     lockfile: cdk/yarn.lock
+#
+# See https://docs.github.com/en/actions/creating-actions/creating-a-composite-action
+
+name: CI for GuCDK
+description: CI for infrastructure defined using GuCDK
+inputs:
+  packageManager:
+    description: 'Which node package manager to use'
+    default: 'npm'
+    required: false
+  lockfile:
+    description: 'Location of the lockfile for the GuCDK project'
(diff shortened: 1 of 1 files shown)


nicl (commented):
If we want to really respect the scripts pattern, ought we to call the scripts rather than directly targeting the underlying yarn/npm commands?

akash1810:

PR #872, opened 2021-11-01
refactor: Use `GuAppAwareConstruct` mixin to DRY codebase

akash1810:
## What does this change?

Refactors some constructs to make use of the `GuAppAwareConstruct` mixin introduced in #853. This mixin ensures the logicalId of a construct includes the app identifier and also ensures the construct gets an `App` tag.

This is a no-op, as evidenced in the diff - there are only _additional_ tests and assertions in previous tests continue to hold true.

Note: This doesn't change all app aware constructs in the codebase, just the ones where refactoring is a no-op. There are more changes incoming to update the other constructs.

No.

No.

## How to test

Observe CI, it should go green.

A DRYer codebase and simpler construct definition as we only need to add a mixin rather than mutating the `id` and adding tags.

n/a

### DIFF
--- a/src/constructs/ec2/security-groups/base.test.ts
+++ b/src/constructs/ec2/security-groups/base.test.ts
@@ -12,7 +12,7 @@ describe("The GuSecurityGroup class", () => {
     publicSubnetIds: [""],
   });
 
-  it("applies the Stage, Stage and App tags", () => {
+  it("applies the Stack, Stage and App tags", () => {
     const stack = simpleGuStackForTesting({ migratedFromCloudFormation: true });
     new GuSecurityGroup(stack, "TestSecurityGroup", {
       vpc,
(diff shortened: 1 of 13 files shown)


src/constructs/ec2/security-groups/base.test.ts
```diff
@@ -12,6 +12,17 @@ describe("The GuSecurityGroup class", () => {
     publicSubnetIds: [""],
   });
 
+  it("applies the Stage, Stage and App tags", () => {
```
jacobwinch, line 15: Minor correction: 

```suggestion
  it("applies the Stack, Stage and App tags", () => {
```

jacobwinch (approved):
Great work 💯

src/constructs/ec2/security-groups/base.test.ts
akash1810: Argh! I thought I crafted this changeset so well!

PR #874, opened 2021-11-02
feat!: Make `GuStageMapping` app aware

### LINKED ISSUE #850: GuEc2App cannot differentiate between apps with different domains
At the moment, it looks there's only one namespace per stack for mappings produced by the GuEc2App pattern. When `synth`ing [this CDK](https://github.com/guardian/typerighter/blob/fb6cb1ad7e084ff8020b30515e590028ae8a51e3/cdk/lib/index.ts), with two apps with different certificates, the current output is:

```json
"stagemapping": {
  "CODE": {
    "domainName": "checker.typerighter.code.dev-gutools.co.uk",
    "minInstances": 1,
    "maxInstances": 2
  },
  "PROD": {
    "domainName": "checker.typerighter.gutools.co.uk",
    "minInstances": 3,
    "maxInstances": 6
  }
}
```

We'd expect space for two apps here (the other is `manager.typerighter.gutools.co.uk`), but it looks like we're missing a dimension to express that, and as a result the last app declaration wins – if I swap the order of the declared apps, I get `manager` in the map above, with no mention of a `checker`.

akash1810:
Fixes #850.

## What does this change?

As reported in #850 it isn't possible to define stage dependent values for multiple apps in a single stack.

Let's say we have two apps: `api` and `login`. The mappings currently produced look like:

```json
"stagemapping": {
  "CODE": {
    "domainName": "login.code.example.com",
    "minInstances": 1,
    "maxInstances": 2
  },
  "PROD": {
    "domainName": "login.example.com",
    "minInstances": 3,
    "maxInstances": 6
  }
}
```

That is the `Mappings` for `api` have been lost as they're overwritten by those from `login` as it's last write wins.

This change makes `Mappings` app aware, resulting in this:

```json
"api": {
  "CODE": {
    "domainName": "api.code.example.com",
    "minInstances": 1,
    "maxInstances": 2
  },
  "PROD": {
    "domainName": "api.example.com",
    "minInstances": 5,
    "maxInstances": 10
  }
},
"login": {
  "CODE": {
    "domainName": "login.code.example.com",
    "minInstances": 1,
    "maxInstances": 2
  },
  "PROD": {
    "domainName": "login.example.com",
    "minInstances": 3,
    "maxInstances": 6
  }
}
```

BREAKING CHANGE: `AppIdentity` is required in more places.

Arguments for `withStageDependentValue` on `GuStack` now require an `AppIdentity`. Add an `app` property to the function arguments.

From:

```ts
scope.withStageDependentValue({
  variableName: "something",
  stageValues: {
    [Stage.CODE]: "value-for-code-stage",
    [Stage.PROD]: "value-for-prod-stage",
  },
})
```

To:

```ts
scope.withStageDependentValue({
  app: "my-app", // <- new
  variableName: "something",
  stageValues: {
    [Stage.CODE]: "value-for-code-stage",
    [Stage.PROD]: "value-for-prod-stage",
  },
})
```

`GuAlarm` now requires an `AppIdentity` when instantiated.

No.

No.

## How to test

See CI and the added tests.

It is possible to define different stage dependent values for different apps within a single stack.

As [commented](https://github.com/guardian/cdk/issues/850#issuecomment-957167109) it would be preferable to follow the Stack Stage App specificity in the mapping. This isn't possible, however as we'd require to know the value of stage at compile time, but it is a [token](https://docs.aws.amazon.com/cdk/latest/guide/tokens.html) as the result of it being a parameter of the CloudFormation stack. We don't lose too much with this solution.

### DIFF
--- a/src/constructs/core/mappings.ts
+++ b/src/constructs/core/mappings.ts
@@ -1,14 +1,46 @@
 import { CfnMapping } from "@aws-cdk/core";
 import type { Stage } from "../../constants";
+import type { AppIdentity } from "./identity";
 import type { GuStack } from "./stack";
 
-export interface GuStageDependentValue<T extends string | number | boolean> {
+export type GuMappingValue = string | number | boolean;
+
+export interface GuStageMappingValue<T extends GuMappingValue> extends AppIdentity {
   variableName: string;
   stageValues: Record<Stage, T>;
 }
 
-export class GuStageMapping extends CfnMapping {
+export class GuStageMapping {
+  private readonly scope: GuStack;
+
+  // the key in this Map is an app string
+  private mappings: Map<string, CfnMapping>;
+
   constructor(scope: GuStack) {
-    super(scope, "stage-mapping");
+    this.scope = scope;
+    this.mappings = new Map<string, CfnMapping>();
+  }
+
+  private getCfnMapping(app: string): CfnMapping {
+    if (!this.mappings.has(app)) {
+      const emptyCfnMapping = new CfnMapping(this.scope, app);
+      this.mappings.set(app, emptyCfnMapping);
+      return emptyCfnMapping;
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- `this.mappings.has` above is `true`, so we can safely assert non-null
+    return this.mappings.get(app)!;
+  }
+
+  withValue<T extends GuMappingValue>(mappingValue: GuStageMappingValue<T>): T {
+    const { app, variableName, stageValues } = mappingValue;
+
+    const cfnMapping = this.getCfnMapping(app);
+
+    for (const [stage, value] of Object.entries(stageValues)) {
+      cfnMapping.setValue(stage, variableName, value);
+    }
+
+    return cfnMapping.findInMap(this.scope.stage, variableName) as T;
   }
 }
--- a/src/patterns/ec2-app.test.ts
+++ b/src/patterns/ec2-app.test.ts
@@ -349,15 +349,19 @@ describe("the GuEC2App pattern", function () {
     const json = SynthUtils.toCloudFormation(stack) as SynthedStack;
 
     expect(json.Mappings).toEqual({
-      stagemapping: {
-        CODE: expect.objectContaining({
+      testguec2app: {
+        CODE: {
+          hostedZoneId: "id123",
+          domainName: "code-guardian.com",
           minInstances: 3,
           maxInstances: 6,
-        }) as Record<string, number>,
-        PROD: expect.objectContaining({
+        },
+        PROD: {
+          hostedZoneId: "id124",
+          domainName: "prod-guardian.com",
           minInstances: 5,
           maxInstances: 12,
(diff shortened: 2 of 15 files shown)


src/patterns/ec2-app.test.ts
```diff
@@ -349,15 +349,19 @@ describe("the GuEC2App pattern", function () {
     const json = SynthUtils.toCloudFormation(stack) as SynthedStack;
 
     expect(json.Mappings).toEqual({
-      stagemapping: {
-        CODE: expect.objectContaining({
+      testguec2app: {
```
akash1810, line 352: The `app` in this test is `"test-gu-ec2-app"` (defined just above). CDK removes `-`.

src/constructs/core/mappings.ts
```diff
@@ -1,14 +1,46 @@
 import { CfnMapping } from "@aws-cdk/core";
 import type { Stage } from "../../constants";
+import type { AppIdentity } from "./identity";
 import type { GuStack } from "./stack";
 
-export interface GuStageDependentValue<T extends string | number | boolean> {
+export type GuMappingValue = string | number | boolean;
+
+export interface GuStageMappingValue<T extends GuMappingValue> extends AppIdentity {
```
akash1810, line 8: `extends AppIdentity` is proving pretty common/useful!

src/constructs/core/mappings.ts
```diff
@@ -1,14 +1,46 @@
...
+  private getCfnMapping(app: string): CfnMapping {
+    if (!this.mappings.has(app)) {
+      const emptyCfnMapping = new CfnMapping(this.scope, app);
+      this.mappings.set(app, emptyCfnMapping);
+      return emptyCfnMapping;
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- `this.mappings.has` above is `true`, so we can safely assert non-null
+    return this.mappings.get(app)!;
+  }
```
akash1810, line 33: This is the main change in this PR - we now maintain a [`Map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) of `app -> CfnMapping`.

The rest of this PR is "noise" identified via the compiler and tests.

src/constructs/core/mappings.ts
```diff
@@ -1,14 +1,46 @@
...
-export interface GuStageDependentValue<T extends string | number | boolean> {
+export type GuMappingValue = string | number | boolean;
+
+export interface GuStageMappingValue<T extends GuMappingValue> extends AppIdentity {
   variableName: string;
   stageValues: Record<Stage, T>;
 }
 
-export class GuStageMapping extends CfnMapping {
+export class GuStageMapping {
```
akash1810, line 13: Although this class has significant changes, it isn't explicitly tested. Instead we're implicitly testing it - see tests added/updated within this PR.

jonathonherbert (approved):
Code looks fine, and test output matches your description 🎉  v. happy to test on a beta release if that's desirable, but otherwise LGTM!

PR #875, opened 2021-11-03
fix: Ensure lambdas are app aware

akash1810:
Requires #872.

## What does this change?

Add the `GuAppAwareConstruct` mixin, introduced in #853, to the `GuLambdaFunction` construct. This results in construct, and sub-classes, get tagged with the app identifier and also having a logicalId with the app identifier.

No.

No.

## How to test

See updated tests.

  - receive the `App` tag

The logical id for lambdas change with this PR, this will cause a replacement once deployed. This should be fine as lambdas do not hold state.

### DIFF
--- a/src/constructs/cloudwatch/__snapshots__/lambda-alarms.test.ts.snap
+++ b/src/constructs/cloudwatch/__snapshots__/lambda-alarms.test.ts.snap
@@ -29,10 +29,10 @@ Object {
     },
   },
   "Resources": Object {
-    "lambda8B5974B5": Object {
+    "lambdaTestingDF42012E": Object {
       "DependsOn": Array [
-        "lambdaServiceRoleDefaultPolicyBF6FA5E7",
-        "lambdaServiceRole494E4CA6",
+        "lambdaTestingServiceRoleDefaultPolicy6824E72F",
+        "lambdaTestingServiceRoleA8ACCBD2",
       ],
       "Properties": Object {
         "Code": Object {
@@ -65,7 +65,7 @@ Object {
         "MemorySize": 512,
         "Role": Object {
           "Fn::GetAtt": Array [
-            "lambdaServiceRole494E4CA6",
+            "lambdaTestingServiceRoleA8ACCBD2",
             "Arn",
           ],
         },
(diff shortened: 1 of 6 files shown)


jacobwinch:
> The logical id for lambdas change with this PR, this will cause a replacement once deployed. This should be fine as lambdas do not hold state.

I think we need to be a bit careful with this assumption. Although lambdas do not hold state there are a couple of considerations to keep in mind here:

1. Presumably replacement means: add a new one, _then_ delete the old one. Consequently, there will be a (very) brief period where 2 identical lambdas exist. In most cases this should be fine, but we have recently learnt that this can cause problems, so we should ensure that teams consider this before upgrading!
2. Although most lambda invocations happen via another resource (for example: SNS, API Gateway or SQS), it is also possible to invoke a lambda directly (via name/ARN). If we upgrade the [DNS Change Executor](https://github.com/guardian/cfn-private-resource-types/blob/main/dns/dns-change-executor/cdk/lib/deploy.ts) (and replace the lambda) then we will need to update the private resource type config (in all accounts) so that it points to the new lambda.

akash1810:
> > The logical id for lambdas change with this PR, this will cause a replacement once deployed. This should be fine as lambdas do not hold state.
> 
> I think we need to be a bit careful with this assumption. Although lambdas do not hold state there are a couple of considerations to keep in mind here:
> 
> 1. Presumably replacement means: add a new one, _then_ delete the old one. Consequently, there will be a (very) brief period where 2 identical lambdas exist. In most cases this should be fine, but we have recently learnt that this can cause problems, so we should ensure that teams consider this before upgrading!
> 2. Although most lambda invocations happen via another resource (for example: SNS, API Gateway or SQS), it is also possible to invoke a lambda directly (via name/ARN). If we upgrade the [DNS Change Executor](https://github.com/guardian/cfn-private-resource-types/blob/main/dns/dns-change-executor/cdk/lib/deploy.ts) (and replace the lambda) then we will need to update the private resource type config (in all accounts) so that it points to the new lambda.

I'm so glad we have peer review as I hadn't connected the recent log lambda incident to this! Will have a think.

PR #876, opened 2021-11-03
fix: Ensure kinesis streams are app aware

akash1810:
Requires #875.

## What does this change?

Apply the `GuAppAwareConstruct` mixin, introduced in #853, to the `GuKinesisStream` construct.

No.

No.

## How to test

See updated tests.

  - receive the `App` tag

This changes the logical id, which is a replacement operation. In reality, I don't think this effects any stack today as I don't think we've used GuCDK to create a new Kinesis stream, but rather we have migrated existing streams into a GuCDK defines stack.

### DIFF
--- a/src/constructs/kinesis/kinesis-stream.test.ts
+++ b/src/constructs/kinesis/kinesis-stream.test.ts
@@ -6,14 +6,19 @@ import { GuKinesisStream } from "./kinesis-stream";
 describe("The GuKinesisStream construct", () => {
   test("auto-generates the logicalId by default", () => {
     const stack = simpleGuStackForTesting();
-    new GuKinesisStream(stack, "LoggingStream");
+    new GuKinesisStream(stack, "LoggingStream", { app: "testing" });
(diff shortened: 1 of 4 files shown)


jacobwinch:
I'm a little unsure about this change. Kinesis streams are typically used as a bridge/connection between more than one app. If I understand it correctly, a single stream can also have multiple consumers? Consequently I don't know if it makes sense for this resource to be so tightly coupled to a single app - what do you think?

akash1810:
> I'm a little unsure about this change. Kinesis streams are typically used as a bridge/connection between more than one app. If I understand it correctly, a single stream can also have multiple consumers? Consequently I don't know if it makes sense for this resource to be so tightly coupled to a single app - what do you think?

I think our typical setup is for a single Kinesis stream, there is:
  - 1 producer
  - 1+ consumer

In this setup, the consumer only does work as the direct result of producer's work. That is, I think it is sensible for the producer to be the owner of the stream.

Admittedly, our logging stream breaks this convention, which is argument to not make this change. Maybe this is one for the [ADRs](https://github.com/guardian/cdk/tree/main/docs/architecture-decision-records)?

PR #877, opened 2021-11-03
fix: Ensure sns topics are app aware

akash1810:
Requires #876.

## What does this change?

Add the `GuAppAwareConstruct` mixin, introduced in #853, to the `GuSnsTopic` construct.

No.

No.

## How to test

See updated tests.

  - receive the `App` tag

This changes the logical id, which is a replacement operation. In reality, I don't think this effects any stack today as I don't think we've used GuCDK to create a new SNS topic, but rather we have migrated existing topics into a GuCDK defines stack.


akash1810:
Closing. See https://github.com/guardian/cdk/pull/876#issuecomment-959272311.

PR #882, opened 2021-11-04
chore: Update to Node v16

akash1810:
## What does this change?

Node 16 is now active LTS. It also ships with npm v8.

npm v8 uses a different lockfile format, denoted by the value of `lockfileVersion` going from 1 (in npm <v6) and 2 (in nom >=v7).

This means we can remove the `check-npm-version` script which was added (#340) to protect the lockfile as bad things happened when a v2 lockfile was used with npm v6 which required a v1 lockfile.

No.

No.

## How to test

CI should pass. It should also be possible to [install this repository from git](https://github.com/guardian/cdk-playground/blob/main/script/switch-cdk) which it is https://github.com/guardian/cdk-playground/pull/114.

Beating dependabot at keeping up to date?

n/a


nicl (approved):
🔢 🎈

PR #884, opened 2021-11-08
Nicl/topic

nicl:
## What does this change?

Adds SSM parameter paths for alarm and notification topics. Motivated by Security HQ work where it felt odd that there are default dist and config etc but nothing for alarm notifications.

TODO: we should integrate these with our existing patterns.

No. (Though the account-readiness check will automatically change.)

No.

## How to test

N/A.

Eventual usage.


src/constants/ssm-parameter-paths.ts
```diff
@@ -26,6 +26,16 @@ export const SSM_PARAMETER_PATHS: Record<string, SsmParameterPath> = {
...
     description: "SSM parameter containing the S3 bucket name holding the app's private configuration",
   },
+  AlarmTopic: {
+    path: "/account/services/alarm.topic.name",
+    description:
+      "SSM parameter containing the Name (not ARN) of an SNS topic to use for alarms. Alarms are for things that require an immediate response.",
+  },
+  NotificationTopic: {
+    path: "/account/services/notification.topic.name",
+    description:
```
akash1810, line 36: Anghammarad has a [node client](https://github.com/guardian/anghammarad/blob/master/anghammarad-client-node/README.md) too...

src/constructs/core/parameters/sns.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    return this.instance;
+  }
+}
+
+/**
+ * Creates a CloudFormation parameter which references an SNS topic to send
+ * notifications to. By default, the bucket name is stored in an SSM Parameter
+ * called `/account/services/alarm.topic.name`.
+ */
```
akash1810, line 35: > By default, the bucket name is stored

I think this doc string is out of date? (same for `GuAlarmTopicParameter` above).

Should we want to make reference to [`GuAnghammaradTopicParameter`](https://github.com/guardian/cdk/blob/147ad77533ca384ef504cc0719aa87d8c84742be/src/constructs/core/parameters/anghammarad.ts) here?

nicl: @akash1810 ah bucket is a copy+paste+edit error! Thanks. Re Anghammarad, yes good shout - wil mirror the original parameter description.

src/constants/ssm-parameter-paths.ts
```diff
@@ -26,6 +26,16 @@ export const SSM_PARAMETER_PATHS: Record<string, SsmParameterPath> = {
     path: "/account/services/private.config.bucket",
     description: "SSM parameter containing the S3 bucket name holding the app's private configuration",
   },
+  AlarmTopic: {
+    path: "/account/services/alarm.topic.arn",
```
jacobwinch, line 30: I'm not entirely convinced that it makes sense to have a default topic for alarms. In my experience multiple teams often share the same AWS account, but not the same topic for alerts (or the same mailing list).

src/constructs/cloudwatch/alarm.ts
```diff
@@ -2,13 +2,27 @@ import { Alarm } from "@aws-cdk/aws-cloudwatch";
...
+   */
+  snsTopicArn?: string;
+
+  /**
+   * Topic Name.
+   *
+   * @deprecated prefer {@link GuAlarmProps.snsTopicArn} instead.
+   * @default account alarm topic - see {@link GuAlarmTopicParameter}.
+   */
+  snsTopicName?: string;
```
jacobwinch, line 24: I think the reason that I originally preferred name to `ARN` was because it allows teams to encode this directly in the template (i.e. it avoids using CFN parameters, which often need to be populated manually) without leaking private information (like account number).

nicl: Thanks, that makes sense. Perhaps we shouldn't deprecate this field but support both.

src/constants/ssm-parameter-paths.ts
akash1810: Worth noting this will be the default value of the CFN parameter - teams that share accounts would be able to customise it if needed.

akash1810:
What's the status of this PR? I found myself doing similar work to this for the cost anomaly work; shall we get this reviewed/merged?

src/constants/ssm-parameter-paths.ts
jacobwinch: Yes, that's a good point; I don't think the concern I previously raised should block this PR

akash1810:

nicl:
Closing as out of date. Let's review alerting/dashboards as a bigger piece and perhaps pick this up again then.

PR #886, opened 2021-11-09
feat: add (optional) throttling alarm to lambda construct

nicl:
Note, it makes sense to update the lambda-related patterns too here, but keen to keep that to a separate PR as still have some thinking about the specific approach there, and also to help keep this PR small.

## What does this change?

Adds a throttling alarm to our lambda construct.

## How to test

TODO (test on Playground.)

Initial motivation was to require a concurrency limit to our lambda construct, but it is difficult to know how to set this and not always desirable. An alternative is, as here, to alarm on throttling. This should let teams know when they hit lambda contention/account limits. They can then apply limits or increase quotas (or fix issues) as appropriate.

N/A.

## Checklist

Note, we should update the lambda patterns too. (Current thinking is that a follow-on PR for that will make reviewing things easier.)

### DIFF
--- a/src/constructs/cloudwatch/lambda-alarms.ts
+++ b/src/constructs/cloudwatch/lambda-alarms.ts
@@ -46,3 +46,50 @@ export class GuLambdaErrorPercentageAlarm extends GuAlarm {
     super(scope, id, alarmProps);
   }
 }
+
+export interface GuLambdaThrottlingMonitoringProps
+  extends Omit<
+    GuAlarmProps,
+    "metric" | "threshold" | "comparisonOperator" | "evaluationPeriods" | "treatMissingData" | "app"
+  > {
+  /**
+   * Sum of thottled invocations above which to alarm.
+   *
+   * @defaultValue 0
+   */
+  toleratedThrottlingCount?: number;
+
+  /**
+   * Evaluation period in minutes for alarm.
+   *
+   * @defaultValue 1
+   */
+  numberOfMinutesAboveThresholdBeforeAlarm?: number;
+
+  noMonitoring?: false;
+}
+
+interface GuLambdaThrottlingAlarmProps extends GuLambdaThrottlingMonitoringProps {
+  lambda: GuLambdaFunction;
+}
+
+export class GuLambdaThrottlingAlarm extends GuAlarm {
+  constructor(scope: GuStack, id: string, props: GuLambdaThrottlingAlarmProps) {
+    super(scope, id, {
+      ...props,
+      app: props.lambda.app,
+      alarmName:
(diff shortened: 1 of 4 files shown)


src/constructs/cloudwatch/lambda-alarms.ts
```diff
@@ -46,3 +46,49 @@ export class GuLambdaErrorPercentageAlarm extends GuAlarm {
...
+interface GuLambdaThrottlingAlarmProps extends GuLambdaThrottlingMonitoringProps {
+  lambda: GuLambdaFunction;
+}
+
+export class GuLambdaThrottlingAlarm extends GuAlarm {
+  constructor(scope: GuStack, id: string, props: GuLambdaThrottlingAlarmProps) {
+    super(scope, id, {
+      ...props,
+      app: props.lambda.app,
+      alarmName: props.alarmName ?? "Lambda throttling alarm",
```
jacobwinch, line 81: I think we might want to put the stage into the default alarm name (I'm 99% sure alarm names must be unique within the account/region i.e. I don't think you can deploy an alarm with the same name via `CODE` and `PROD` stacks).

src/constructs/cloudwatch/lambda-alarms.ts
```diff
@@ -46,3 +46,49 @@ export class GuLambdaErrorPercentageAlarm extends GuAlarm {
...
+      alarmDescription:
+        props.alarmDescription ?? "Alarm when lambda is throttled (which causes requests to fail with a 429).",
+      threshold: props.toleratedThrottlingCount ?? 1,
+      evaluationPeriods: props.numberOfMinutesAboveThresholdBeforeAlarm ?? 1,
+      metric: props.lambda.metricThrottles({
+        period: Duration.seconds(60),
+        statistic: "sum",
+      }),
+      treatMissingData: TreatMissingData.NOT_BREACHING,
+      comparisonOperator: ComparisonOperator.GREATER_THAN_OR_EQUAL_TO_THRESHOLD,
```
jacobwinch, line 91: Minor: for consistency with other alarms (and because of the name `toleratedThrottlingCount`) I think we want this to be `GREATER_THAN` https://github.com/guardian/cdk/blob/b1913a5a23cf6d9c3ff64584b63feaa434fc62d3/src/constructs/cloudwatch/lambda-alarms.ts#L41

jacobwinch (approved):
Great work 💯 

> Note, it makes sense to update the lambda-related patterns too here, but keen to keep that to a separate PR as still have some thinking about the specific approach there, and also to help keep this PR small.

👍

src/constructs/cloudwatch/lambda-alarms.ts
nicl: good shout. 👍

PR #887, opened 2021-11-09
fix: check update-aws-cdk script dependencies before running

nicl:
Note, PRs currently failing because we are not on latest aws-cdk so ran the script and hit this.

## What does this change?

Checks some of the less common deps for this script. Note, I haven't checked `git` or `npm` as those are expected for a TS project already.

## How to test

Attempt to run the script without the dependencies and get a descriptive error.

### DIFF
--- a/script/update-aws-cdk
+++ b/script/update-aws-cdk
@@ -5,6 +5,18 @@ set -e
 DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 ROOT_DIR=$DIR/..
 
+checkDependencies() {
+  if ! [ -x "$(command -v gh)" ]; then
+    echo 'Error: gh is not installed but required. To install run `brew install gh` or see https://cli.github.com/.' >&2
+    exit 1
+  fi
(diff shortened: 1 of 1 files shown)


akash1810:
I wonder if we should create a [`Brewfile`](https://github.com/Homebrew/homebrew-bundle) for use with `brew bundle`, and then add `brew bundle` to `script/setup`?

akash1810:

nicl:
@akash1810 I've rebased but a bit relunctant to put more time into this for now, although I agree a Brew Bundle could be nice. Are you happy if I merge this as an improvement over the status quo at least?

akash1810:
> @akash1810 I've rebased but a bit relunctant to put more time into this for now, although I agree a Brew Bundle could be nice. Are you happy if I merge this as an improvement over the status quo at least?

Sounds good!

akash1810 (approved):

akash1810:
I'm also happy to pair to add a Brewfile, it should be a fairly small change.

PR #888, opened 2021-11-09
chore(deps)!: Update aws-cdk libraries to 1.132.0

nicl:
BREAKING CHANGE: Update aws-cdk libraries to 1.132.0


src/constructs/cloudwatch/__snapshots__/ec2-alarms.test.ts.snap
```diff
@@ -439,6 +439,10 @@ Object {
             "Key": "deregistration_delay.timeout_seconds",
             "Value": "30",
           },
+          Object {
```
jacobwinch, line 442: It seems like AWS are [explicitly setting a default here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-elasticloadbalancingv2-targetgroup-targetgroupattribute.html) (that is, I don't think this will cause any functional change to our stacks):

> stickiness.enabled - Indicates whether sticky sessions are enabled. The value is true or false. The default is false.

In the past we have seen cases where changing an implicit default to the same value explicitly has caused resources to be replaced (😱), but I think we're [safe in this case](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticloadbalancingv2-targetgroup.html#cfn-elasticloadbalancingv2-targetgroup-targetgroupattributes).

jacobwinch (approved):
Thanks for updating this 👍

akash1810:
Can we orchestrate this release with https://github.com/guardian/cdk/pull/889#issuecomment-964912343?

PR #912, opened 2021-11-15
feat!: Improve the GuCname construct

jacobwinch:
## What does this change?

This PR updates the `GuCname` construct so that it is optimised for our most typical use case. That is, creating CNAMEs which point to a load balancer or API gateway instance for the `CODE` and `PROD` stages. 

It is still possible to create CNAMEs which don't align with this use-case (e.g. using a non-standard `Stage`), but as that is moving away from the paved road, it must be achieved via the lower-level `GuDnsRecordSet` (which requires a little more thought and boilerplate).

In order to achieve this, we have made **breaking changes**:

* Setting the TTL is now mandatory when you create a `GuCname`
* `GuCname` now requires an app name
* Domain names for `CODE` and `PROD` stages are also required

## How to test

I have tested this by releasing to beta and [making the change to AMIgo](https://github.com/guardian/amigo/pull/655/files). The snapshot tests confirm that there are no changes to the generated CloudFormation template.

* Users should be able to start managing their DNS records more easily. More specifically, as the construct creates an appropriate mapping for users under the hood, it should be harder to [make mistakes when wiring up the stage-dependent](https://github.com/guardian/lurch/pull/15#discussion_r596952548) domain name.
* Users are also forced to explicitly think about/acknowledge the TTL, which is important for reducing risk as part of the migration from CloudFormation to `@guardian/cdk`.
* Although users will need to work with this construct directly (that is, [we don't plan to create this construct via the pattern](https://github.com/guardian/cdk/discussions/913)), this change ensures that we're not adding [much code/complexity](https://github.com/guardian/amigo/pull/655/files#diff-78d894e83ca4cb48af587f94e9f8986cc0592d412273ef99246b30c644ae5098R275-R279) to the infrastructure definition.

I can't think of any particular risks associated with this PR. A GitHub search suggests that AMIgo is the only project using this construct in `main` at the moment.

## Checklist

### DIFF
--- a/src/constructs/dns/dns-records.test.ts
+++ b/src/constructs/dns/dns-records.test.ts
@@ -1,6 +1,8 @@
 import "@aws-cdk/assert/jest";
+import "../../utils/test/jest";
 import { SynthUtils } from "@aws-cdk/assert";
 import { Duration } from "@aws-cdk/core";
+import { Stage } from "../../constants";
 import { simpleGuStackForTesting } from "../../utils/test";
 import { GuCname, GuDnsRecordSet, RecordType } from "./dns-records";
 
@@ -15,27 +17,30 @@ describe("The GuDnsRecordSet construct", () => {
     });
     expect(SynthUtils.toCloudFormation(stack)).toMatchSnapshot();
   });
+  it("should use the exact logical id that is passed in", () => {
+    const stack = simpleGuStackForTesting();
+    new GuDnsRecordSet(stack, "ThisExactLogicalId", {
+      name: "banana.example.com",
+      recordType: RecordType.CNAME,
+      resourceRecords: ["apple.example.com"],
+      ttl: Duration.hours(1),
+    });
+    expect(stack).toHaveResourceOfTypeAndLogicalId("Guardian::DNS::RecordSet", "ThisExactLogicalId");
+  });
 });
 
 describe("The GuCname construct", () => {
   it("should create the correct resources with minimal config", () => {
     const stack = simpleGuStackForTesting();
     new GuCname(stack, "TestRecord", {
-      name: "banana.example.com",
+      domainNameProps: {
+        [Stage.CODE]: { domainName: "xyz.code-guardian.com" },
+        [Stage.PROD]: { domainName: "xyz.prod-guardian.com" },
+      },
+      app: "my-test-app",
       resourceRecord: "apple.example.com",
+      ttl: Duration.hours(1),
     });
     expect(SynthUtils.toCloudFormation(stack)).toMatchSnapshot();
-  });
-  it("should allow users to override the default TTL", () => {
-    const stack = simpleGuStackForTesting();
-    new GuDnsRecordSet(stack, "TestRecord", {
-      name: "banana.example.com",
-      recordType: RecordType.CNAME,
-      resourceRecords: ["apple.example.com"],
-      ttl: Duration.minutes(1),
-    });
-    expect(stack).toHaveResourceLike("Guardian::DNS::RecordSet", {
-      TTL: 60,
-    });
   });
 });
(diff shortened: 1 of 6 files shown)


src/constructs/dns/dns-records.test.ts
```diff
@@ -15,27 +17,30 @@ describe("The GuDnsRecordSet construct", () => {
     });
     expect(SynthUtils.toCloudFormation(stack)).toMatchSnapshot();
   });
+  it("should use the exact logical id that is passed in", () => {
```
jacobwinch, line 20: We were already getting this behaviour thanks to `CfnResource`, but I have added this test to make it more explicit and ensure that we don't accidentally drop this requirement in the future.

src/types/domain-names.ts
```diff
@@ -0,0 +1,8 @@
+import type { Stage } from "../constants";
+
+export type GuDomainNameProps = Record<Stage, GuDomainName>;
+
+export interface GuDomainName {
+  domainName: string;
+  hostedZoneId?: string;
```
jacobwinch, line 7: The `hostedZoneId?` is here because:

1. `GuCertificate` and `GuCname` share this interface 
2. We need the hosted zone id for the former if the DNS record required for certificate validation will reside in `Route53`.

It's slightly tangential to this PR, but it's a bit odd that we offer proper support for creating the certificate in a Route 53 zone, but we _don't_ offer the same support for managing your CNAME... Perhaps we should consider making the `GuCname` construct DNS-provider agnostic in the future?

src/constructs/dns/dns-records.test.ts
akash1810: Brilliant!

src/constructs/dns/dns-records.ts
```diff
@@ -13,7 +17,10 @@ export interface GuDnsRecordSetProps {
 }
 
 /**
- * Do not use this construct directly. Use [[`GuCname`]] instead in order to reduce boilerplate.
+ * This construct can be used to create DNS records in NS1.
+ *
+ * If you are creating a CNAME for your CODE and PROD stages, for example when creating a CNAME
+ * for your load balancer, [[`GuCname`]] may be used instead to reduce boilerplate.
```
akash1810, line 23: I think we should be more assertive here:

```
Prefer to use [[`GuCname`]] when creating a CNAME for a CODE or PROD load balancer as this requires less boilerplate.
```

src/types/domain-names.ts
akash1810: > It's slightly tangential to this PR, but it's a bit odd that we offer proper support for creating the certificate in a Route 53 zone, but we _don't_ offer the same support for managing your CNAME... Perhaps we should consider making the `GuCname` construct DNS-provider agnostic in the future?

Should we drop Route53 support entirely as Route53 isn't on our paved road? AWS's constructs for Route53 are pretty good too, so being agnostic might yield more complicated code?

I've not thought about this that much though - not blocking this PR but would love to discuss.

akash1810 (commented):
One blocking comment, otherwise LGTM!

src/constructs/dns/dns-records.ts
jacobwinch: Good idea, thanks for suggesting this new phrasing 👍

akash1810 (approved):

src/types/domain-names.ts
jacobwinch: > AWS's constructs for Route53 are pretty good too, so being agnostic might yield more complicated code?

Yes, you're probably right. I know that we tried to do this before, but I guess we need to understand more generally whether we should use/support Route53 in the department. It would certainly be simpler to have a single DNS provider, but I worry that we have so much stuff in Route53 now that migrating away might be unfeasible.

jacobwinch: > not blocking this PR but would love to discuss.

Definitely agree that we should discuss this more outside of the PR though!

PR #915, opened 2021-11-15
feat: add `scheduleEnabled` to ECS task props

MarSavar:
## What does this change?
This PR introduces a `scheduleEnabled` prop to the `GuScheduledEcsTaskProps` interface.
Explicitly setting the prop to `false` disables the rule, otherwise it is enabled by default.

The use case for this change is letting the user determine whether to run a schedule or not, for example depending on the stage. 
We have an example of this in Investigations & Reporting where we only need to run a certain task on PROD but not on CODE. We could use an if statement around `new GuScheduledEcsTask ...` to check the stage, but passing this prop feels like a more sensible way of doing it.

The prop is optional and doesn't break existing scheduled tasks.

## How can you measure success?
Less code required to determine whether to run the task.

### DIFF
--- a/src/patterns/scheduled-ecs-task.test.ts
+++ b/src/patterns/scheduled-ecs-task.test.ts
@@ -1,5 +1,5 @@
-import "@aws-cdk/assert/jest";
 import { SynthUtils } from "@aws-cdk/assert";
+import "@aws-cdk/assert/jest";
 import type { IVpc } from "@aws-cdk/aws-ec2";
 import { SecurityGroup, Vpc } from "@aws-cdk/aws-ec2";
 import { Schedule } from "@aws-cdk/aws-events";
@@ -7,8 +7,8 @@ import { Effect, PolicyStatement } from "@aws-cdk/aws-iam";
 import { Duration } from "@aws-cdk/core";
 import type { GuStack } from "../constructs/core";
 import { simpleGuStackForTesting } from "../utils/test";
-import { GuScheduledEcsTask } from "./scheduled-ecs-task";
 import "../utils/test/jest";
+import { GuScheduledEcsTask } from "./scheduled-ecs-task";
 
 const makeVpc = (stack: GuStack) =>
   Vpc.fromVpcAttributes(stack, "VPC", {
@@ -57,6 +57,24 @@ describe("The GuScheduledEcsTask pattern", () => {
     });
 
     expect(stack).toHaveResourceLike("AWS::Events::Rule", { ScheduleExpression: "rate(1 minute)" });
+    expect(stack).toHaveResourceLike("AWS::Events::Rule", { State: "ENABLED" });
+  });
+
+  it("should be disabled when scheduleEnabled is set to false", () => {
+    const stack = simpleGuStackForTesting();
+
+    new GuScheduledEcsTask(stack, {
+      schedule: Schedule.rate(Duration.minutes(1)),
+      containerConfiguration: { id: "node:10", type: "registry" },
+      monitoringConfiguration: { noMonitoring: true },
+      vpc: makeVpc(stack),
+      stack: "test",
+      stage: "TEST",
+      app: "ecs-test",
+      scheduleEnabled: false,
+    });
+
+    expect(stack).toHaveResourceLike("AWS::Events::Rule", { State: "DISABLED" });
   });
 
   const generateComplexStack = (stack: GuStack, app: string, vpc: IVpc) => {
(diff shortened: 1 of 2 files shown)


akash1810:
Thanks for the contribution!

> We could use an if statement around `new GuScheduledEcsTask ...` to check the stage

This won't work as `Stage` is a [`Token`](https://docs.aws.amazon.com/cdk/latest/guide/tokens.html) and only known at runtime.

Do you have an example of how this'll get used across different stages? Can we add to a test to demonstrate usage? I suspect this will be a [Mapping](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/mappings-section-structure.html) via [`withStageDependentValue`](https://github.com/guardian/cdk/blob/cfe6517bbc4015f38a0418db45489d6762219030/src/constructs/core/stack.ts#L77-L79).

MarSavar:
Thanks @akash1810 for reviewing this so quickly! I've added a couple of tests.
The stage was just an example - the idea was just to introduce this property as sometimes it is useful to have an ECS task to run on demand as opposed to having it enabled by default.

akash1810:
> the idea was just to introduce this property as sometimes it is useful to have an ECS task to run on demand as opposed to having it enabled by default.

In this scenario, newing up a `GuScheduledEcsTask` only to have it be manually invoked is quite a confusing API.

I'd suggest we separate this up a little, creating two constructs: `GuEcsTask` and `GuScheduledEcsTask` where `GuScheduledEcsTask` would extend `GuEcsTask` adding the schedule.

Then, the pattern would instantiate one of them based on the value of the new `scheduleEnabled` prop.

See [`GuScheduledLambda`](https://github.com/guardian/cdk/blob/cfe6517bbc4015f38a0418db45489d6762219030/src/patterns/scheduled-lambda.ts#L66-L84) for a similar example.

philmcmahon:
@akash1810 there are a few scenarios we're trying to deal with 

1. As Mario said, when we probably rarely want to run a particular task on CODE, but want to keep our CODE infra the same as PROD - which would need some kind of mapping
2. When one of our ingest jobs is broken because companies house has stopped temporarily publishing the data - in this scenario it's helpful to disable the task till they fix the problem. Or, because we've messed up our database so that the ingest jobs start failing

2) we've been dealing with up till now by just making manual changes in the eventbridge console, but these sometimes get overridden by cloudformation deploys (we're not sure why - possibly because a branch has been deployed that removes all the infrastructure, then when it is added back the event gets re-enabled. This feature would put that data in version control rather than manually in the console so we can rely on it remaining consistent.

akash1810:
Thinking about it some more, my previous suggestion to create `GuEcsTask` and `GuScheduledEcsTask` and get the pattern to conditionally instantiate one wouldn't work as it relies on knowing the value of `Stage` at compile time, which we don't.

> want to keep our CODE infra the same as PROD

💯 

> This feature would put that data in version control rather than manually in the console so we can rely on it remaining consistent.

I don't think it will as currently it is not stage aware. If it's enabled/disabled for one stage, it's enabled/disabled for all stages. Changing the type of the prop to `Record<Stage, boolean>` would enable being stage aware.

The pattern employed by [`GuAlarm`](https://github.com/guardian/cdk/blob/cfe6517bbc4015f38a0418db45489d6762219030/src/constructs/cloudwatch/alarm.ts) might also be worth looking at.

philmcmahon:
> I don't think it will as currently it is not stage aware. If it's enabled/disabled for one stage, it's enabled/disabled for all stages. Changing the type of the prop to Record<Stage, boolean> would enable being stage aware.

that sounds good, but we don't want it to be only based off the stage - sometimes it will be hard coded to false for a few days whilst we e.g. fix the DB. the 'actionEnabledInCode' param that GuAlarm has wouldn't allow us to switch off an action in PROD

akash1810:
> we've been dealing with up till now by just making manual changes in the eventbridge console, but these sometimes get overridden by cloudformation deploys (we're not sure why - possibly because a branch has been deployed that removes all the infrastructure, then when it is added back the event gets re-enabled. This feature would put that data in version control rather than manually in the console so we can rely on it remaining consistent.

> sometimes it will be hard coded to false for a few days whilst we e.g. fix the DB

I've interpreted this as we'd like to keep the `main` branch as the source of truth for whether a schedule is enabled or not? If so, the current changes in this PR will only work if we're disabling CODE and PROD in tandem, as demonstrated by the added test. Is this desired?

If we're not seeking to track a schedule's enablement via `main`, can I suggest we do.

philmcmahon:
>  If so, the current changes in this PR will only work if we're disabling CODE and PROD in tandem
Yes this is the desired behaviour I think - basically it's a 'turn this thing off everywhere whilst we wait till we fix the DB' or similar

akash1810:

PR #919, opened 2021-11-17
refactor!: rename stackName to cloudFormationStackName in GuStackProps

jacobwinch:
## What does this change?

BREAKING CHANGE: Removes the `stackName` prop and adds an optional `cloudFormationStackName` prop.
This helps to remove ambiguity between the Guardian concept of a stack and a CloudFormation stack.

## How to test

This is a trivial change so we haven't added a test.

Before there were props called `stack` and `stackName` that referred to two completely separate things - very confusing! Hopefully renaming the latter should make this distinction clearer.

This should be low risk as only a couple of projects currently use this prop (and they are both owned by Developer Experience!). We'll update them once this change is released.

## Checklist

### DIFF
--- a/src/constructs/core/stack.ts
+++ b/src/constructs/core/stack.ts
@@ -13,8 +13,16 @@ import type { GuMappingValue, GuStageMappingValue } from "./mappings";
 import { GuStageParameter } from "./parameters";
 import type { GuParameter } from "./parameters";
 
-export interface GuStackProps extends StackProps, Partial<GuMigratingStack> {
+export interface GuStackProps extends Omit<StackProps, "stackName">, Partial<GuMigratingStack> {
+  /**
+   * The Guardian stack being used (as defined in your riff-raff.yaml).
+   * This will be applied as a tag to all of your resources.
+   */
   stack: string;
+  /**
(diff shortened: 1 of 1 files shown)


src/constructs/core/stack.ts
```diff
@@ -13,8 +13,16 @@ import type { GuMappingValue, GuStageMappingValue } from "./mappings";
 import { GuStageParameter } from "./parameters";
 import type { GuParameter } from "./parameters";
 
-export interface GuStackProps extends StackProps, Partial<GuMigratingStack> {
+export interface GuStackProps extends Omit<StackProps, "stackName">, Partial<GuMigratingStack> {
+  /**
+   * The Guardian stack being used (as defined in your riff-raff.yaml).
+   * This will be applied as a tag to all of your resources.
+   */
   stack: string;
```
jacobwinch, line 21: It was tempting to rename this too, but:

a) This would mean updating every single stack using `@guardian/cdk`, which is a lot of effort
b) The tag that gets applied is still called `Stack` and in the riff-raff.yaml the property is called `stacks`, so we can't fully escape from this ambiguity anyway

akash1810 (approved):

PR #921, opened 2021-11-22
feat: Introduce GuEcsTask (pulled from GuScheduledEcsTask)

philmcmahon:
## What does this change?
This pulls out the guts of GuScheduledEcsTask pattern and moves them into a new GuEcsTask construct. This is to support the creation of fargate tasks which do not run on a schedule. 

The investigations team needs this to support our our migrations task - which we run every time we need to make a change to the database. This is currently a lambda but can take longer than 15 minutes. We could run it on our bastion/locally, but this would mean adding psql and node to the bastion which we'd rather avoid for the usual security reasons. The downside of running them locally (as is done for discussion and flexible-content) is that someone could switch their laptop off/lose network connection part way through.

Note the change set isn't as big as it seems - this is mostly a copy paste job of scheduled-ecs-task to ecs-task, with some stuff then pulled back into scheduled-ecs-task

## How to test
I've tested this against https://github.com/guardian/lurch which is the only project using this pattern  as far as I know

## Checklist

- [ x I have listed any breaking changes, along with a migration path [^1]

### DIFF
--- a/src/patterns/scheduled-ecs-task.ts
+++ b/src/patterns/scheduled-ecs-task.ts
@@ -5,6 +5,7 @@ import type { GuStack } from "../constructs/core";
 import { AppIdentity } from "../constructs/core/identity";
 import type { GuEcsTaskProps } from "../constructs/ecs/ecs-task";
 import { GuEcsTask } from "../constructs/ecs/ecs-task";
+import { GuAppAwareConstruct } from "../utils/mixin/app-aware-construct";
 
 /**
  * Configuration options for the [[`GuScheduledEcsTask`]] pattern.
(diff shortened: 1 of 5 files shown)


philmcmahon:
Note, thanks to @akash1810 for the inspiration for this here https://github.com/guardian/cdk/pull/915#issuecomment-969025869. I think there might still be a use case for a 'disabled' scheduled task, but pulling out the ecstask makes a lot of sense

src/patterns/scheduled-ecs-task.ts
```diff
@@ -145,116 +47,15 @@ const getContainer = (config: ContainerConfiguration) => {
...
-    [cluster, task, taskDefinition, stateMachine, rule].forEach((c) =>
-      AppIdentity.taggedConstruct({ app: props.app }, c)
-    );
-
-    new CfnOutput(scope, AppIdentity.suffixText(props, "StateMachineArnOutput"), {
-      value: stateMachine.stateMachineArn,
+      targets: [new SfnStateMachine(this.stateMachine)],
+      enabled: true,
     });
+    AppIdentity.taggedConstruct({ app: props.app }, rule);
```
akash1810, line 59: Since

akash1810 (approved):

PR #922, opened 2021-11-22
docs: Explain the use of npx to check the expected aws-cdk version

shtukas:
## What does this change?

This is a readme update to explain the use of `npx` (to read the expected version of aws-cdk) while preparing a migration from aws-cdk to @guardian/cdk.


jacobwinch (approved):
Thanks @shtukas 👍

PR #934, opened 2021-11-25
feat: introduce constructs/core/parameters/base/GuNumberParameter

shtukas:
## What does this change?
Introduce constructs/core/parameters/base/GuNumberParameter. We already have GuStringParameter but I am migrating a code base that has

```
const dbPort = new CoreCfnParameter(this, "DBPort", {
  type: "Number",
  description: "The port that the RDS DB will be available on.",
  default: 5432,
});
```
With this change I will be able to write 

```
const dbPort = new GuNumberParameter(this, "DBPort", {
  description: "The port that the RDS DB will be available on.",
  default: 5432,
});
```

Instead of using GuParameter.


akash1810:
Hmm. I'd suggest, for the database port example, we just use a variable in TypeScript (`const dbPort = 5432`). Are there any other use cases for a `GuNumberParameter`?

shtukas:
https://github.com/guardian/baton/pull/221

PR #948, opened 2021-11-30
feat: Add `GuStackForInfrastructure` where `Stage` is always `INFRA`

akash1810:
Possibly easiest to review commit by commit.

## What does this change?

Creates `GuStackForInfrastructure`, a sub-class of `GuStack` designed for infrastructure based stacks.

An infrastructure based stack is a singleton - there should only be one in an AWS account (well, per region, per account).

Examples of infrastructure stacks include:
  - [VPC stacks](https://github.com/guardian/cdk/blob/57d427b09a3c3d5fcf9b406f1ad2ecaf927e6152/src/constructs/vpc/vpc.ts)
  - [cloudwatch-logs-management](https://github.com/guardian/cloudwatch-logs-management)
  - [elasticsearch-node-rotation](https://github.com/guardian/elasticsearch-node-rotation)
  - (coming soon) cost anomaly detection

A `GuStack` can have a `Stage` value of either `CODE` or `PROD`. It is not designed for infrastructure/singleton stacks. This means, for example, we currently have to remember there is (and should only be) a `PROD` VPC stack and creating a `CODE` one could cause disruption.

We could use restrictions in Riff-Raff to help us remember this, but we can do better via types! Introducing `GuStackForInfrastructure` ✨ .

A `GuStackForInfrastructure` only has one stage - `INFRA`. It is a sub-class of `GuStack`, thus inherits all the benefits of `GuStack`, such as rich tagging.

For infrastructure/singleton stacks we can use the new `GuStackForInfrastructure` instead of `GuStack` to provide more meaning, clearer tagging and less implicit knowledge.

In this change `GuVpc` has been updated to use `GuStackForInfrastructure`.

Our custom linting rule has been updated too (with tests added).

## How to test

See added tests, in particular the update to the `GuVpc` snapshot.

Singleton stacks are better represented.

Is the API clear enough? As noted in the comments and tests, if one uses the wrong type when calling `withStageDependentValue` on an instance of `GuStackForInfrastructure` an exception will be raised. Is this clear enough?

## Checklist

### DIFF
--- a/src/constructs/core/mappings.ts
+++ b/src/constructs/core/mappings.ts
@@ -1,5 +1,5 @@
 import { CfnMapping } from "@aws-cdk/core";
-import type { Stage, StageForInfrastructure } from "../../constants";
+import type { StageAwareValue } from "../../types/stage";
 import type { AppIdentity } from "./identity";
 import type { GuStack } from "./stack";
 
@@ -7,7 +7,7 @@ export type GuMappingValue = string | number | boolean;
 
 export interface GuStageMappingValue<T extends GuMappingValue> extends AppIdentity {
   variableName: string;
-  stageValues: Record<Stage, T> | Record<StageForInfrastructure, T>;
+  stageValues: StageAwareValue<T>;
 }
 
 export class GuStageMapping {
--- a/src/constructs/core/mappings.ts
+++ b/src/constructs/core/mappings.ts
@@ -1,5 +1,5 @@
 import { CfnMapping } from "@aws-cdk/core";
-import type { Stage } from "../../constants";
+import type { Stage, StageForInfrastructure } from "../../constants";
 import type { AppIdentity } from "./identity";
 import type { GuStack } from "./stack";
 
@@ -7,7 +7,7 @@ export type GuMappingValue = string | number | boolean;
 
 export interface GuStageMappingValue<T extends GuMappingValue> extends AppIdentity {
   variableName: string;
-  stageValues: Record<Stage, T>;
+  stageValues: Record<Stage, T> | Record<StageForInfrastructure, T>;
 }
 
 export class GuStageMapping {
--- a/src/constructs/core/stack.test.ts
+++ b/src/constructs/core/stack.test.ts
@@ -3,16 +3,22 @@ import "@aws-cdk/assert/jest";
 import { SynthUtils } from "@aws-cdk/assert";
 import "../../utils/test/jest";
 import { Role, ServicePrincipal } from "@aws-cdk/aws-iam";
-import { App } from "@aws-cdk/core";
-import { Stage, Stages } from "../../constants";
+import { Annotations, App } from "@aws-cdk/core";
+import { Stage, StageForInfrastructure, Stages } from "../../constants";
 import { ContextKeys } from "../../constants/context-keys";
 import { TagKeys } from "../../constants/tag-keys";
 import { simpleGuStackForTesting } from "../../utils/test";
 import type { SynthedStack } from "../../utils/test";
 import { GuParameter } from "./parameters";
-import { GuStack } from "./stack";
+import { GuStack, GuStackForInfrastructure } from "./stack";
 
 describe("The GuStack construct", () => {
+  const warn = jest.spyOn(Annotations.prototype, "addWarning");
+
+  afterEach(() => {
+    warn.mockReset();
+  });
+
   it("requires passing the stack value as props", function () {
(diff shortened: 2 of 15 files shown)


akash1810:
CI is failing for a separate reason and fixed in #949. This branch would need rebasing once #949 is merged.

src/constructs/core/stack.test.ts
```diff
@@ -78,4 +84,41 @@ describe("The GuStack construct", () => {
...
+    expect(() => {
+      stack.withStageDependentValue({
+        app: "test",
+        variableName: "test",
+        stageValues: {
+          [Stage.CODE]: 1,
+          [Stage.PROD]: 2,
+        },
+      });
+    }).toThrowError("Mapping doesn't contain top-level key 'INFRA'");
```
akash1810, line 122: This is an error thrown by AWS CDK. We could provide a custom error message, but it feels a little unnecessary as we can rely on the strongly typed `stageValues` instead.

src/constructs/core/mappings.ts
```diff
@@ -1,13 +1,13 @@
...
+import type { Stage, StageForInfrastructure } from "../../constants";
 import type { AppIdentity } from "./identity";
 import type { GuStack } from "./stack";
 
 export type GuMappingValue = string | number | boolean;
 
 export interface GuStageMappingValue<T extends GuMappingValue> extends AppIdentity {
   variableName: string;
-  stageValues: Record<Stage, T>;
+  stageValues: Record<Stage, T> | Record<StageForInfrastructure, T>;
```
akash1810, line 10: `stageValues` can be:

```typescript
{
  "CODE": 1,
  "PROD": 2,
}
```

or

```typescript
{
  "INFRA": 1
}
```

That is, via the type system, stage is either `CODE or PROD` or `INFRA`.

src/constructs/core/stack.ts
```diff
@@ -83,6 +83,13 @@ export class GuStack extends Stack implements StackStageIdentity, GuMigratingSta
    * Mappings to work around this limitation.
    */
   withStageDependentValue<T extends GuMappingValue>(mappingValue: GuStageMappingValue<T>): T {
+    const isInfraStageProvided = Object.keys(mappingValue.stageValues).includes(StageForInfrastructure);
+    if (isInfraStageProvided) {
+      Annotations.of(this).addWarning(
```
jacobwinch, line 88: It's worth noting that some of our constructs currently use (`withStageDependentValue`) under the hood - [ASGs are one example which come to mind](https://github.com/guardian/cdk/blob/d001f357814afc905c59c397dc60ae9a278cf200/src/constructs/autoscaling/asg.ts#L63-L86). That is, I think this warning will sometimes appear when users are not able to avoid it?

That said, I don't think this problem applies to any of the example `INFRA` services that you've listed, so perhaps we can revisit this in the future. (In theory it would apply to a service like Security HQ, which is currently `PROD`-only but might be better suited to the `INFRA` stage).

jacobwinch (approved):
Great work 💯

src/constructs/core/stack.ts
akash1810: Oops, you mentioned this on Chat too and I just didn't clock it, sorry!

Yes, this is a legit concern. As you said, it's currently unlikely that we'd be impacted on any of our current infrastructure stacks. I think we can make two improvements to this PR to help this:
- Add a TODO on the function you've linked to
- Add an annotation to the overridden `withStageDependentValue` function as a way to point users to consider using a vanilla variable rather than a CFN mapping

akash1810: Looking at it [more closely](https://github.com/guardian/cdk/search?q=withStageDependentValue) usage of `withStageDependentValue` is a bit wider than I had realised. I think we can make these usages support `INFRA` fairly easily, will give it a go.

akash1810: > Looking at it [more closely](https://github.com/guardian/cdk/search?q=withStageDependentValue) usage of `withStageDependentValue` is a bit wider than I had realised. I think we can make these usages support `INFRA` fairly easily, will give it a go.

Done on https://github.com/guardian/cdk/pull/948/commits/e684aa6c6c85f412880d401217c0a1448c908351. Existing tests continue to pass, which is enough, right?! 😅

jacobwinch: Thanks for revisiting this, the latest changes look good to me 👍

jacobwinch: > Existing tests continue to pass, which is enough, right?! 😅

Generally our test coverage is pretty good, so let's hope so!

PR #949, opened 2021-12-01
ci: Explicitly add DOM types

akash1810:
## What does this change?

CI for the `integration-test` project is currently failing with:

```console
Error: node_modules/@types/jsdom/base.d.ts(24,40): error TS2304: Cannot find name 'DocumentFragment'.
Error: node_modules/@types/jsdom/base.d.ts(41,28): error TS2304: Cannot find name 'Node'.
Error: node_modules/@types/jsdom/base.d.ts(172,19): error TS2304: Cannot find name 'HTMLScriptElement'.
Error: node_modules/@types/jsdom/base.d.ts(172,39): error TS2304: Cannot find name 'HTMLLinkElement'.
Error: node_modules/@types/jsdom/base.d.ts(172,57): error TS2304: Cannot find name 'HTMLIFrameElement'.
Error: node_modules/@types/jsdom/base.d.ts(172,77): error TS2304: Cannot find name 'HTMLImageElement'.
Error: node_modules/@types/jsdom/base.d.ts(181,38): error TS2304: Cannot find name 'Window'.
Error: node_modules/@types/jsdom/base.d.ts(185,24): error TS2304: Cannot find name 'Window'.
```

This suggests TypeScript cannot find types definitions for JsDOM. JsDom is a transitive dependency via Jest.

In this change we explicitly add the DOM types to tsconfig to solve this.

I don't fully understand why this has suddenly started happening. I suspect our lack of a lockfile on the `integration-test` project might have something to do with it.

I also don't fully understand why we didn't see CI issues in #931 which bumped the version of TypeScript...

See: https://www.typescriptlang.org/tsconfig#high-level-libraries

## How to test

See CI.

Or more specifically, first witness the failing case:
- Checkout `main`
- Run `./script/ci`
- Witness failure similar to above

Then witness the passing case:
- Checkout this branch
- Run `./script/ci`
- Witness it succeeding.

CI should pass.

Other than a lack of understanding the root cause...

## Checklist

### DIFF
--- a/tools/integration-test/tsconfig.json
+++ b/tools/integration-test/tsconfig.json
@@ -8,7 +8,7 @@
     "target": "ES2020",
     "module": "ES2020",
     "moduleResolution": "node",
-    "lib": ["ES2020"],
+    "lib": ["ES2020", "DOM"],
     "declaration": true,
     "strict": true,
     "noImplicitAny": true,


jacobwinch (approved):
Thanks for figuring this out 🕵️👍

akash1810:
Reading https://github.com/facebook/jest/issues/12098, it looks like others are experiencing similar issues and have made similar fixes. Alternative suggestions on that issue is to downgrade to Jest v26.

We don't need JsDOM and it looks like [Jest v28](https://github.com/facebook/jest/issues/12098#issuecomment-982693710) would be better suited for us.

Will merge this to unblock `main`.

PR #950, opened 2021-12-01
feat: Add cost anomaly monitor construct

akash1810:
## Checklist


src/constants/ssm-parameter-paths.ts
```diff
@@ -38,4 +38,8 @@ export const SSM_PARAMETER_PATHS: Record<string, SsmParameterPath> = {
     path: `${VPC_SSM_PARAMETER_PREFIX}/primary/subnets/public`,
     description: "A list of public subnets",
   },
+  BillingEmail: {
+    path: "/account/services/billing/notification.email",
```
akash1810, line 42: TODO update docs.

PR #962, opened 2021-12-06
refactor: Create skeleton for 'new' command in preparation for copying this command from cdk-cli

michaelwmcnamara:
## What does this change?
This adds the ability to accept and parse a 'new' command for cdk.
It is an implementation for yargs of the 'new' command from the cdk-cli:
https://github.com/guardian/cdk-cli/blob/dd3460d17937a9de3c24bfe11f8bad0bba0f66e5/src/commands/new.ts#L41-L70

The command accepts and processes all the inputs the real 'new' command will use, but just displays some testing text to show the arguments have been received.
 
This is the first part of the "merge cdk-cli into cdk" trello card: https://trello.com/c/xSsYX4zR/992-merge-cdk-cli-into-cdk
This change also refactors the profile and region yargs from the global scope to within the 'account-readiness' code as this is the only command that makes use of these.

## How to test

-
 
- To test the 'new' command:
Run `./script/cli new --multiApp true --app "myApp" --stack "myStack" --yaml-template-location "thing"`

- To test the  'account-readiness' command:
Run `./script/cli account-readiness --profile deployTools`

It will be easier to port across the full functionality for the 'new' command.
All other commands are unaffected by this change

This change has negligible risk.

## Checklist

- [] I have updated the documentation as required for the described changes [^2] - this will be done as part of the implementation of the 'new' functionality in the next PR

### DIFF
--- a/src/bin/index.ts
+++ b/src/bin/index.ts
@@ -39,7 +39,7 @@ const parseCommandLineArguments = () => {
       )
       .command(Commands.New, "Creates a new CDK stack", (yargs) =>
         yargs
-          .option("multiApp", {
+          .option("multi-app", {
             type: "boolean",
             description:
               "Create the stack files within sub directories as the project defines multiple apps (defaults to false)",
@@ -61,7 +61,7 @@ const parseCommandLineArguments = () => {
               "The Guardian stack being used (as defined in your riff-raff.yaml). This will be applied as a tag to all of your resources.",
             demandOption: true,
           })
-          .option("yamlTemplateLocation", {
+          .option("yaml-template-location", {
             type: "string",
             description: "Path to the YAML CloudFormation template",
           })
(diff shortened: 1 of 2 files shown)


src/bin/index.ts
```diff
@@ -36,6 +39,36 @@ const parseCommandLineArguments = () => {
...
+          .option("stack", {
+            type: "string",
+            description:
+              "The Guardian stack being used (as defined in your riff-raff.yaml). This will be applied as a tag to all of your resources.",
+            demandOption: true,
+          })
+          .option("yaml-template-location", {
+            type: "string",
+            description: "Path to the YAML CloudFormation template",
+            default: "",
```
akash1810, line 69: Do we need the empty string as a default?

src/bin/index.ts
```diff
@@ -36,6 +39,36 @@ const parseCommandLineArguments = () => {
...
+          .option("multi-app", {
+            type: "boolean",
+            description:
+              "Create the stack files within sub directories as the project defines multiple apps (defaults to false)",
+            default: false,
+          })
+          .option("init", {
+            type: "boolean",
+            description: "Create the cdk directory before building the app and stack files (defaults to true)",
+            default: true,
```
akash1810, line 53: For a clearer API, I'd suggest this be `no-init` and default it to false, as we're then effectively using the _presence_ of the flag as the value.

This means we go from:

```
npx @guardian/cdk@latest new --init false ...
```

To:

```
npx @guardian/cdk@latest new --no-init
```

akash1810 (approved):
Added a suggestion around CLI flag naming, but happy for this to go in as it's a skeletal change.

src/bin/index.ts
akash1810: TIL yargs treats `no-` as the inverse of a boolean flag! So please ignore this suggestion.

PR #963, opened 2021-12-06
feat: Add SNS topic policy to allow external publishing 

akash1810:
## What does this change?

Allow an SNS topic to be written to by external accounts or organisations.

These topics can be defined as a centralised topic.

This is a pattern employed by stacks such as [Anghammarad](https://github.com/guardian/anghammarad/blob/4b441ccd7e92b7bd8dc0161c0a2256918249e1a7/cloudformation/anghammarad.template.yaml#L75-L87) and the intention is to employ this for the cost anomaly work.

I'm a little unsure if a CFN output should also be added to the stack if external publishing to an SNS topic gets configured.

## How to test

See added unit tests.

It is possible to easily define centralised SNS topics.

n/a

## Checklist

### DIFF
--- a/src/constructs/sns/sns-topic.ts
+++ b/src/constructs/sns/sns-topic.ts
@@ -1,10 +1,26 @@
-import { Topic } from "@aws-cdk/aws-sns";
+import { AccountPrincipal, OrganizationPrincipal } from "@aws-cdk/aws-iam";
 import type { TopicProps } from "@aws-cdk/aws-sns";
+import { Topic } from "@aws-cdk/aws-sns";
+import { RegexPattern } from "../../constants";
 import { GuStatefulMigratableConstruct } from "../../utils/mixin";
 import type { GuStack } from "../core";
 import type { GuMigratingResource } from "../core/migrating";
 
-interface GuSnsTopicProps extends TopicProps, GuMigratingResource {}
+interface GuSnsTopicProps extends TopicProps, GuMigratingResource {
+  /**
+   * List of AWS account ids to grant `SNS:Publish` permission to.
+   *
+   * Note: if you need to add conditions to the Topic Policy, consider calling `this.grantPublish` directly.
+   */
+  accountsAllowedToPublish?: string[];
+
+  /**
+   * List of AWS organisations to grant `SNS:Publish` permission to.
+   *
+   * Note: if you need to add conditions to the Topic Policy, consider calling `this.grantPublish` directly.
+   */
+  organisationsAllowedToPublish?: string[];
+}
 
 /**
  * Construct which creates an SNS Topic.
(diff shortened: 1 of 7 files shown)


jacobwinch (dismissed):
Fantastic work 💯 

I think this PR closes [this card](https://trello.com/c/B1zpA6Fr/527-add-construct-for-cross-account-sns-permissions) and [unblocks this one](https://trello.com/c/HhI3DgZw/528-add-optional-cross-account-access-prop-to-gusnslambda-pattern).

jacobwinch:
> I'm a little unsure if a CFN output should also be added to the stack if external publishing to an SNS topic gets configured.

Perhaps we can take care of this at the pattern-level?

akash1810:
> Fantastic work 💯
> 
> I think this PR closes [this card](https://trello.com/c/B1zpA6Fr/527-add-construct-for-cross-account-sns-permissions) and [unblocks this one](https://trello.com/c/HhI3DgZw/528-add-optional-cross-account-access-prop-to-gusnslambda-pattern).

Oh wonderful, hadn't thought about that! Annoyingly, this PR doesn't quite do what I need for my use case right now, hence draft status. I think this is more due me not being quite sure what I need atm. Will keep this open whilst I think about it some more.

akash1810:
> Annoyingly, this PR doesn't quite do what I need for my use case right now, hence draft status.

My use case required setting conditions too:

```typescript
    new TopicPolicy(this, `${id}TopicOrgPublishAccess`, {
      topics: [snsTopic],
      policyDocument: new PolicyDocument({
        statements: Object.values(Constants.AWS_ORG_IDS).map((orgId) => {
          return new PolicyStatement({
            resources: [snsTopic.topicArn],
            actions: ["SNS:Publish"],
            effect: Effect.ALLOW,
            principals: [
              new PrincipalWithConditions(
                new ServicePrincipal("costalerts.amazonaws.com"),
                { StringEquals: { "aws:PrincipalOrgID": orgId } }
              ),
            ],
          });
        }),
      }),
    });
```

> I think this PR closes [this card](https://trello.com/c/B1zpA6Fr/527-add-construct-for-cross-account-sns-permissions) and [unblocks this one](https://trello.com/c/HhI3DgZw/528-add-optional-cross-account-access-prop-to-gusnslambda-pattern).

I've not thought enough about these cards to say if this PR is still useful or not...

akash1810:
There's actually already a [`grantPublish` function](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-sns.Topic.html#grantwbrpublishgrantee) on the L2 construct `Topic` that does exactly this PR!

Example usage:

```typescript
const topic = new GuSnsTopic(this, "cross-account-topic");

// allow sns:Publish to an entire OU
topic.grantPublish(new OrganizationPrincipal("o-abcde12345"))

// allow sns:Publish to a specific account
topic.grantPublish(new AccountPrincipal("012345678900"));
```

The only thing this PR adds is input validation via regex patterns, however this shouldn't be needed if we publish a list of AWS OUs and account ids:

```typescript
import { AwsOrgUnits, AwsAccountIds } from "some-package";

const topic = new GuSnsTopic(this, "cross-account-topic");

AwsOrgUnits.forEach(orgId => topic.grantPublish(new OrganizationPrincipal(orgId)));

AwsAccountIds.forEach(accountId => topic.grantPublish(new AccountPrincipal(accountId)));
```

akash1810:
> There's actually already a [`grantPublish` function](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-sns.Topic.html#grantwbrpublishgrantee) on the L2 construct `Topic` that does exactly this PR!

Updated this PR to use `grantPublish`. I'm not fully convinced this change adds much to the repository and happy to be convinced either way.

src/constructs/sns/sns-topic.ts
```diff
@@ -1,10 +1,26 @@
...
 import type { GuMigratingResource } from "../core/migrating";
 
-interface GuSnsTopicProps extends TopicProps, GuMigratingResource {}
+interface GuSnsTopicProps extends TopicProps, GuMigratingResource {
+  /**
+   * List of AWS account ids to grant `SNS:Publish` permission to.
+   *
+   * Note: if you need to add conditions to the Topic Policy, consider calling `this.grantPublish` directly.
+   */
+  accountsAllowedToPublish?: string[];
```
jacobwinch, line 15: I wonder if it'd be better to surface this at the [pattern level](https://github.com/guardian/cdk/blob/794ff425b8e540807833b8f514f460e8d15d1cdf/src/patterns/sns-lambda.ts#L69-L72) instead/as well?

akash1810:

src/constructs/sns/sns-topic.ts
akash1810: Agreed. I think ideally a pattern's props should extend all the props of it's child construct's props to keep things a bit DRY. This would be quite the refactor so I'd be tempted to do this in a separate PR. WDYT?

jacobwinch: > I think ideally a pattern's props should extend all the props of it's child construct's props to keep things a bit DRY.

I'm not sure about surfacing _all_ props in all cases - that could produce some pretty overwhelming interfaces/documentation (I'm thinking about a complex pattern like `GuEc2App` which has many children here). Happy to think about how to apply this to the pattern in a separate PR though!

jacobwinch (approved):

PR #971, opened 2021-12-08
feat: Add "new" command for cdk creation (ported from cdk-cli repo)

michaelwmcnamara:
## What does this change?

This PR follows on from https://github.com/guardian/cdk/pull/962
It adds in the actual functionality to generate a cdk project. This is almost identical to what was in the cdk-cli repository except for the fact we are using yargs instead of oclif for parsing the input arguments.

A future pr will refactor this to optimise the code and remove any unused parts.

## How to test

From the cdk directory run ./script/cli new --app my-app --stack my-stack and confirm cdk directory is created successfully

The "new" script successfully creates a cdk folder and proper elements when run.

 minimal risk. This is porting over existing functionality to a new repo. The change is for stand-alone functionality. No impact on other things.

## Checklist

- [x] I have listed any breaking changes, along with a migration path.[^1] _There are no breaking changes in this PR._


akash1810:
Super excited by this!

I wonder if we should label this PR as `chore` or similar to _not_ create a new release? This is because there are elements of the CLI that no longer apply, such as the `multi-app` feature, or the use of `yarn` and possibly a move away from the `script/task` pattern as we've discovered it doesn't always fit GHA for CI.

Alternatively, we could work against the `beta` branch to create an [experimental npm release](https://github.com/guardian/cdk/blob/794ff425b8e540807833b8f514f460e8d15d1cdf/docs/006-testing.md#would-your-change-update-the-layout-of-the-npm-package), merging to `main` once we've applied the improvements to the CLI.

jacobwinch:
> I wonder if we should label this PR as `chore` or similar to _not_ create a new release?

Interesting thought; I was assuming that we'd release this as it is. Although the tooling is imperfect as it stands, releasing it as is would allow us to formally deprecate [cdk-cli](https://github.com/guardian/cdk-cli) which feels like a big step in the right direction to me.

More generally, I feel less worried about applying a strict release process (or proper semantic versioning) to the `cli` parts of this repo. Since we will always advise people to run `npx @guardian/cdk@latest new --init --stack <STACK> --app <APP> --yaml-template-location <PATH TO YAML>` as a one-off command, I don't think we need to worry too much about making breaking changes (e.g. removing the multi-app flag) between versions.

.eslintrc.js
```diff
@@ -18,4 +18,12 @@ module.exports = {
   },
   root: true,
   ignorePatterns: ["**/*.js", "node_modules"],
+  overrides: [
+    {
+      files: ["src/bin/**"],
+      rules: {
+        "custom-rules/valid-constructors": 0,
```
jacobwinch, line 25: This custom linting rule is designed to enforce [conventions](https://github.com/guardian/cdk/blob/main/docs/architecture-decision-records/002-component-constuctors.md) for the constructs/patterns which are published as part of this library (e.g. `GuPlayApp`), in order to improve consistency for users. 

It is not relevant to the `cli` part of this repo (since users never work with those classes directly), so this change turns it off for irrelevant files/directories.

akash1810:
> Although the tooling is imperfect as it stands, releasing it as is would allow us to formally deprecate [cdk-cli](https://github.com/guardian/cdk-cli) which feels like a big step in the right direction to me.

I hadn't considered this. This reasoning makes a lot of sense.

> More generally, I feel less worried about applying a strict release process (or proper semantic versioning) to the cli parts of this repo. ... I don't think we need to worry too much about making breaking changes (e.g. removing the `multi-app` flag) between versions.

This also makes a lot of sense. My main concerns are that we don't manage to complete the CLI migration before next migration fortnight. We have ~8 weeks for that, which should be enough time. Let's get this in then.

akash1810 (approved):
Looks great!

Following https://github.com/guardian/cdk/pull/998#issuecomment-1000128013, I've enabled "check-aws-cdk" as a required check, meaning this branch would need to be rebased against `main` now. We can add this check after this PR is in, if that's simpler too!

PR #972, opened 2021-12-09
feat: expose patterns in typedoc

nicl:
(For discussion - thought showing the change would be easier to reason about than merely describing it.)

## What does this change?

At the moment we expose patterns directly under /src, but this is non-obvious and inconsistent with how we expose our other constructs etc. The intention is (I'm guessing here) to promote the use of patterns, but I'm not sure this is helping here - as it makes things difficult to find in the autogenerated docs.

Note, projects that use patterns already, will need to update their import paths after this. E.g.

    import { GuEc2App } from '@guardian/cdk'; // old
    import { GuEc2App } from '@guardian/cdk/patterns // new

Some missing `index.ts` files (for ecs and cli) were added too, to ensure we generate docs for these. (It could be argued that we don't want to expose docs for `cli` but I think it is worth it as it is part of our publicly exposed code atm. If we don't want people to use it we need to ensure this more robustly.)

Before:

Screenshot 2021-12-09 at 14 13 33

After:
Screenshot 2021-12-09 at 14 10 10

## How to test

    ./script/start-docs

Easier to find things in the auto-generated docs.

This is a breaking change (as breaks import paths for patterns) so adds a little friction to updating @guardian/cdk version.

## Checklist

### DIFF
--- a/docs/001-general-usage.md
+++ b/docs/001-general-usage.md
@@ -1,6 +1,7 @@
 # General Usage for @guardian/cdk
 
 ## 🖥 Installation
+
 This library can be installed from [NPM](https://www.npmjs.com/package/@guardian/cdk).
 
 ```
@@ -16,13 +17,17 @@ yarn add @guardian/cdk
 We recommend using TypeScript with `@guardian/cdk` as type safety is wonderful!
 
 ## 🏟 Patterns
+
 Patterns are high level classes which compose a number of constructs to produce standard architectures.
 
 For example, you should be able to get all the resources you need to deploy a new lambda function from one `GuLambdaStack` class.
 
-Patterns should be your default entry point to this library and can be found [here](../src/patterns).
+Patterns should be your default entry point to this library and can be found
+[here](/modules/patterns) (note, link only works when browsing auto-generated
+documentation).
 
 ## 🏗 Constructs
+
 Constructs are lower level classes which will create one or more resources to produce one element of a stack.
 
 For example, the `GuDatabaseInstance` will create an RDS instance as well as a parameter group, if required.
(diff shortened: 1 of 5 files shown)


docs/001-general-usage.md
```diff
@@ -1,6 +1,7 @@
 # General Usage for @guardian/cdk
 
 ## 🖥 Installation
+
```
nicl, line 4: Hmm, these must have been automatically added by VS Code.

akash1810:
Would we also need to update the [`main` entrypoint](https://github.com/guardian/cdk/blob/main/package.json#L5)?

My 2c - users should be encouraged to use patterns as their default (see also [this ADR](https://github.com/guardian/cdk/blob/main/docs/architecture-decision-records/001-constructs-and-patterns.md) and [this one](https://github.com/guardian/cdk/blob/main/docs/architecture-decision-records/005-package-structure.md)). Importing patterns from the highest namespace available (i.e. `import { GuEc2App } from '@guardian/cdk';`) encourages that. Additionally, this also make it easy to tell if a stack is doing something custom by performing a quick scan of the imports. Having said that, this is just my opinion which is open to change!

I agree that discovering patterns within our TypeDoc docs isn't particularly obvious - I wonder if there is some TypeDoc config we can do to alias `index` to `patterns` instead? Or if there's a TypeDoc alternative we can use?

akash1810:
Actually, I suspect we can keep the `import { Pattern } from '@guardian/cdk';` imports _and_ provide a better typedoc experience by changing `main` in `package.json`.

akash1810:

akash1810:
https://github.com/guardian/cdk/pull/1132 updated the typedoc site and retained the `import { GuEc2App } from '@guardian/cdk'` import path. Is this PR still needed?

akash1810:
Closing in favour of #1132.

PR #977, opened 2021-12-13
feat: Add Stack Set support

akash1810:
## What does this change?

Create constructs to support the creation of `AWS::CloudFormation::StackSet` resources. Unfortunately, AWS CDK does not (yet) have a L2 construct for Stack Sets 😢 .

Usage:
```typescript
// the infrastructure to create in target accounts
class AccountAlertTopic extends GuStackForStackSetInstance {
  constructor(id: string, props: GuStackProps) {
    super(id, props);

    new GuSnsTopic(this, "topic-for-alerts");
  }
}

// the infrastructure to create in the root account
class ParentStack extends GuStackForInfrastructure {
    super(scope, id, props);

    new GuStackSet(this, `${id}StackSet`, {
      stackSetInstance: new AccountAlertTopic("Alerts", { stack: props.stack }),
      name: "centralised-alarms",
      description: "Provisioning of standard account alerting resources",
      organisationUnitTargets: [ "o-abcde12345" ]
    });
  }
}

// contents of `bin/cdk.ts`
new ParentStack(new App(), "AccountAlarmResources", { stack: "alarms" })
```

This will produce a CloudFormation template like this:
```yaml
Resources:
  AccountAlarmResourcesStackSet:
    Type: AWS::CloudFormation::StackSet
    Properties:
      PermissionModel: SERVICE_MANAGED
      StackSetName: centralised-alarms
      AutoDeployment:
        Enabled: true
        RetainStacksOnAccountRemoval: false
      Description: Provisioning of standard account alerting resources
      Parameters: []
      StackInstancesGroup:
        - DeploymentTargets:
            OrganizationalUnitIds:
              - o-abcde12345
          Regions:
            - Ref: AWS::Region
      Tags:
        - Key: gu:cdk:version
          Value: TEST
        - Key: gu:repo
          Value: guardian/cdk
        - Key: Stack
          Value: alarms
        - Key: Stage
          Value: INFRA
      TemplateBody: |-
        {
          "Resources": {
            "topicforalerts57330FBE": {
              "Type": "AWS::SNS::Topic",
              "Properties": {
                "Tags": [
                  {
                    "Key": "gu:cdk:version",
                    "Value": "TEST"
                  },
                  {
                    "Key": "gu:repo",
                    "Value": "guardian/cdk"
                  },
                  {
                    "Key": "Stack",
                    "Value": "alarms"
                  },
                  {
                    "Key": "Stage",
                    "Value": "INFRA"
                  }
                ]
              }
            }
          }
        }
```

See https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cloudformation-stackset.html.

Note: this does not support provisioning of stack sets to specific accounts as I'm not sure why we'd ever do this.

## How to test

See added tests.

We get a step closer to a better stack set deployment story as, by provisioning stack sets as just another resource in a CloudFormation template, we're closer to being able to deploy them via RIff-Raff like any other infrastructure.

Do the construct names make sense and are they intention revealing enough?

## Checklist

### DIFF
--- a/src/constructs/core/stack.ts
+++ b/src/constructs/core/stack.ts
@@ -1,5 +1,6 @@
-import { Annotations, Stack, Tags } from "@aws-cdk/core";
-import type { App, StackProps } from "@aws-cdk/core";
+import { SynthUtils } from "@aws-cdk/assert";
+import { Annotations, App, Stack, Tags } from "@aws-cdk/core";
+import type { StackProps } from "@aws-cdk/core";
 import execa from "execa";
 import gitUrlParse from "git-url-parse";
 import { StageForInfrastructure } from "../../constants";
@@ -126,6 +127,10 @@ export class GuStack extends Stack implements StackStageIdentity, GuMigratingSta
     return this.params.get(key) as T;
   }
 
+  get parameterKeys(): string[] {
+    return Array.from(this.params.keys());
+  }
+
   // eslint-disable-next-line custom-rules/valid-constructors -- GuStack is the exception as it must take an App
   constructor(app: App, id: string, props: GuStackProps) {
     const mergedProps = {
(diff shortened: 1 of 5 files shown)


src/constructs/core/stack.ts
```diff
@@ -196,3 +201,19 @@ export class GuStackForInfrastructure extends GuStack {
     return super.withStageDependentValue(mappingValue);
   }
 }
+
+/**
+ * A GuStack but designed for Stack Set instances.
+ *
+ * In a stack set application, `GuStackForStackSetInstance` is used to represent the infrastructure to provision in target AWS accounts.
```
jacobwinch, line 208: I find this naming very clear 👍

jacobwinch (approved):
Great work 💯

PR #997, opened 2021-12-22
chore(deps)!: Update aws-cdk libraries to 1.137.0

akash1810:
## What does this change?

Moving from 1.132.0 to 1.137 brings a few new features and bug fixes, most notably tag support added to `AWS::Logs::LogGroup` resources!

There are a few snapshot updates to reflect these new features and bug fixes. See the individual commits for more information.

We should update our Lambda constructs to create an explicit `LogGroup` to take advantage of tag support. See also https://github.com/guardian/cloudwatch-logs-management/issues/19.

BREAKING CHANGE: Update aws-cdk libraries to 1.137.0

## How to test

See CI pass.

Note: CI is now witnessing a lot of `console.log`s, this gets fixed in https://github.com/aws/aws-cdk/pull/18136.

## How can we measure success?

Keeping dependencies up to date (and unblocking https://github.com/guardian/cdk/pull/995).

## Have we considered potential risks?

n/a

## Checklist

- [ ] I have listed any breaking changes, along with a migration path [^1]
- [ ] I have updated the documentation as required for the described changes [^2]

[^1]: Consider whether this is something that will mean changes to projects that have already been migrated, or to the CDK CLI tool. If changes are required, consider adding a checklist here and/or linking to related PRs.
[^2]: If you are adding a new construct or pattern, has new documentation been added? If you are amending defaults or changing behaviour, are the existing docs still valid?

PR #998, opened 2021-12-23
ci: Move AWS CDK check into it's own GHA job

akash1810:
## What does this change?

Currently CI will fail when the version of AWS CDK libraries being used are over 5 releases out of date. This forces us to keep somewhat up to date with releases, however it does introduce an odd situation:
  1. Raise PR
  2. CI passes on PR
  3. PR gets approved
  4. New AWS CDK released
  5. PR merged
  6. Semantic Release creates a new release
  7. Robot that keeps `version` in `package.json` up to date raises PR
  8. Robot PR fails CI and cannot be automatically merged

The latest example of this can be seen on:
  - #977 
  - #995 
  - #997.

Where #977 passed CI (step 2) and was merged (step 5) and #995 (step 7) is now failing it's CI checks (step 8).

When this happens we usually update the branch of the robot's PR to remove the AWS CDK check, allowing CI to pass. This isn't great as we put the check back in the next PR. This is noise.

This change moves the AWS CDK check into a discrete job in GHA. This should mean we can resolve the above solution by updating the branch protection rules, allowing the AWS CDK check to fail but still require the tests to pass.

This yields less noise in the repository.

## How to test

CI should pass and there should be another build report.

https://github.com/guardian/cdk/pull/995 can be merged before https://github.com/guardian/cdk/pull/997 and thus the revision history stays truthful.

More importantly though, we able to ignore the AWS CDK check a little more easily.

???

## Checklist

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -17,6 +17,11 @@ jobs:
           validateSingleCommit: true
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  check-aws-cdk:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - run: ./script/check-aws-cdk
   CI:
     runs-on: ubuntu-latest
     steps:
--- a/script/ci
+++ b/script/ci
@@ -7,7 +7,6 @@ ROOT_DIR=$DIR/..
 
 preamble() {
   "${DIR}"/check-yarn-lock
-  "${DIR}"/check-aws-cdk
 }
 
 runIntegrationTest() {


akash1810:
There's a little dance to do with this PR:
  - [ ] Merge this in (the "check-aws-cdk" check will be failing)
  - [ ] Rebase https://github.com/guardian/cdk/pull/995 with `main` and merge it (the "check-aws-cdk" check will be failing)
  - [ ] Update branch protection rules to require "check-aws-cdk" to pass before a PR can be merged
  - [ ] Rebase https://github.com/guardian/cdk/pull/997 with `main` and merge it
  - [ ] Rebase any open branches with `main`
  - [ ] Done

jacobwinch:
> This change moves the AWS CDK check into a discrete job in GHA. This should mean we can resolve the above solution by updating the branch protection rules, allowing the AWS CDK check to fail but still require the tests to pass.

We could also bypass this check for machine users, for example: https://github.com/guardian/cdk/pull/1000/files#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddR22.

jacobwinch (approved):

akash1810:
> We could also bypass this check automatically for machine users, for example: https://github.com/guardian/cdk/pull/1000/files#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddR22.

YES! Even more reason to move away from `script/ci`!

PR #1006, opened 2021-12-23
chore: Forgo keeping `version` in `package.json` up to date

akash1810:
## What does this change?

Currently when a new release is made (via Semantic Release), a robot will raise and merge a PR to update `version` in `package.json`.

This is handy as we can browse the file to see what the latest version is.
I don't think it's necessary for the release process - Semantic Release gets the latest version number from npm.

It comes at a slight cost, however:
  - A bit of PR noise. We're not expected to interacted with the robot's PRs, but the `CODEOWNERS` rules mean we get added as reviewers.
  - An odd scenario where our CI checks don't pass and the robot's PR cannot be automatically merged. See #998.

This change moves to a simpler model of keeping `version` in `package.json` completly static between releases.
If one would like to find the number of the latest version, npm or GitHub releases can be used.

## How to test

Might be one to try on the `beta` branch?

Releasing a new version of the library produces less noise.

This should not have any impact on the release process as [Semantic Release gets the latest number from npm](https://github.com/semantic-release/npm/blob/708c29bf735d731f6855b2425ba9f12b618873dd/lib/prepare.js#L10-L17), rather than reading form the `package.json`.

## Checklist

### DIFF
--- a/.github/workflows/cd.yaml
+++ b/.github/workflows/cd.yaml
@@ -18,17 +18,3 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
         run: ./script/cd
-
-      # Only performed if `main` is updated
-      # See https://docs.github.com/en/actions/reference/environment-variables#determining-when-to-use-default-environment-variables-or-contexts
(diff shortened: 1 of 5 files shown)


jacobwinch (approved):

akash1810:
Bah! The version number is being used on the documentation site!

image

PR #1007, opened 2021-12-23
feat: Add error annotation when a deprecated lambda runtime is used

akash1810:
## What does this change?

We should never use a deprecated or non-LTS runtime for lambdas. In this change, we add an error annotation to this effect.

Note: the annotation does not prevent the stack from being synthesised, to do this, we'd want to `throw`. I'm not sure if that's desirable behaviour atm though?

## How to test

See added tests.

We are discouraged from using deprecated or non-LTS lambda runtimes.

We have to maintain the `DEPRECATED_RUNTIMES` list.

## Checklist

### DIFF
--- a/src/constructs/lambda/lambda.ts
+++ b/src/constructs/lambda/lambda.ts
@@ -1,7 +1,7 @@
-import { Code, Function, RuntimeFamily } from "@aws-cdk/aws-lambda";
-import type { FunctionProps, Runtime } from "@aws-cdk/aws-lambda";
+import { Code, Function, Runtime, RuntimeFamily } from "@aws-cdk/aws-lambda";
+import type { FunctionProps } from "@aws-cdk/aws-lambda";
 import { Bucket } from "@aws-cdk/aws-s3";
-import { Duration } from "@aws-cdk/core";
+import { Annotations, Duration } from "@aws-cdk/core";
 import { GuDistributable } from "../../types/distributable";
 import { GuLambdaErrorPercentageAlarm, GuLambdaThrottlingAlarm } from "../cloudwatch";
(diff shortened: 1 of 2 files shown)


akash1810:
Raised this as a way to test https://github.com/guardian/cdk/pull/1006 hasn't negatively impacted the release process.

src/constructs/lambda/lambda.ts
```diff
@@ -10,6 +10,14 @@ import type { GuStack } from "../core";
 import { AppIdentity } from "../core/identity";
 import { GuParameterStoreReadPolicyStatement } from "../iam";
 
+const DEPRECATED_RUNTIMES: Runtime[] = [
+  Runtime.NODEJS_4_3,
+  Runtime.NODEJS_6_10,
+  Runtime.NODEJS_8_10,
+  Runtime.NODEJS_10_X,
+  Runtime.NODEJS_12_X,
```
jacobwinch, line 18: IIUC, this [version](https://nodejs.org/en/about/releases/) becomes EOL on 2022-04-30 - do we want to deprecate it before then?

jacobwinch:
> Note: the annotation does not prevent the stack from being synthesised, to do this, we'd want to throw. I'm not sure if that's desirable behaviour atm though?

I guess this is something that we (or the Security team) might want to consider in the future but I think it's better to start with this less disruptive approach.

jacobwinch (approved):

src/constructs/lambda/lambda.ts
akash1810: Ah, I thought it was already deprecated! I think it's fine to do so before then as:
  - We're not blocking the synth step
  - It acts as a reminder to update _before_ the deadline
  - There aren't too many lambdas defined using GuCDK today (and 0 using this version of the library)

PR #1008, opened 2021-12-23
fix: include template for project generation tooling when publishing

jacobwinch:
## What does this change?

I noticed a problem when testing https://github.com/guardian/cdk/pull/971 after it was released; the template files were omitted from the published `npm` package, which causes the `new` command to fail. This PR ensures that the necessary template files are included.

I've taken a similar approach to https://github.com/guardian/cdk-cli/pull/48, but I think using `cp` will be fine for our purposes.

## How to test

Create a temporary directory and run `npx @guardian/cdk@beta new --stack my-stack --app my-app`.

The `new` command works correctly!

N/A

## Checklist

- [x] I have listed any breaking changes, along with a migration path [^1] N/A
- [x] I have updated the documentation as required for the described changes [^2] N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
     "test:custom-lint-rule": "eslint tools/eslint/rules/valid-constructors.test.ts",
     "test:dev": "jest --detectOpenHandles --runInBand --watch",
     "prepare": "tsc",
+    "prepack": "cp -r src/bin/commands/template lib/bin/commands/template",
     "release": "semantic-release",
     "release:docs": "rm -rf target && typedoc && gh-pages -d target -u \"github-actions-bot <support+actions@github.com>\"",
     "serve:docs": "rm -rf target && typedoc && serve target",


package.json
```diff
@@ -20,6 +20,7 @@
     "test:custom-lint-rule": "eslint tools/eslint/rules/valid-constructors.test.ts",
     "test:dev": "jest --detectOpenHandles --runInBand --watch",
     "prepare": "tsc",
+    "prepack": "cp -r src/bin/commands/template lib/bin/commands/template",
```
akash1810, line 23: I wonder if adding `src/bin/commands/template` to [`files`](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#files) is a more intention revealing version of this? Especially as JSON files cannot be commented 😞 .

jacobwinch: Unfortunately that includes the files under `src/bin/commands/template`, but we need them under `lib/bin/commands/template`

akash1810 (approved):

PR #1014, opened 2021-12-29
fix: Only create a CFN Mapping when necessary

akash1810:
Might be easier to review commit by commit.

## What does this change?

The `withStageDependentValue` function on a `GuStack` can be used to add a [CloudFormation Mapping](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/mappings-section-structure.html) to the template where a value is dependant on the stage the template is being run in.

This makes sense for a `GuStack` where `stage` is either CODE or PROD. However creating a mapping for a `GuStackForInfrastructure` where `stage` is always INFRA is redundant.

In this change we only create the CloudFormation Mapping where necessary; if stage is INFRA we return a value directly, rather than via a Mapping.

## How to test

See the additional tests where we assert that Mappings for `GuStackForInfrastructure` is `undefined`.

A shorter and simpler synthesised CloudFormation template.

n/a

## Checklist

### DIFF
--- a/src/constructs/acm/certificate.test.ts
+++ b/src/constructs/acm/certificate.test.ts
@@ -54,22 +54,6 @@ describe("The GuCertificate class", () => {
   });
 
   it("should not create a CloudFormation Mapping when used in a GuStackForInfrastructure", () => {
-    const stack = simpleGuStackForTesting();
-    new GuCertificate(stack, {
-      app: "testing",
-      [Stage.CODE]: {
-        domainName: "code-guardian.com",
-        hostedZoneId: "id123",
-      },
-      [Stage.PROD]: {
-        domainName: "prod-guardian.com",
-        hostedZoneId: "id124",
-      },
-    });
-
-    const json = SynthUtils.toCloudFormation(stack) as SynthedStack;
-    expect(json.Mappings).toBeDefined();
-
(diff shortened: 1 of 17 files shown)


src/constructs/acm/certificate.test.ts
```diff
@@ -52,4 +52,33 @@ describe("The GuCertificate class", () => {
...
+        hostedZoneId: "id123",
+      },
+      [Stage.PROD]: {
+        domainName: "prod-guardian.com",
+        hostedZoneId: "id124",
+      },
+    });
+
+    const json = SynthUtils.toCloudFormation(stack) as SynthedStack;
+    expect(json.Mappings).toBeDefined();
```
jacobwinch, line 71: I would probably drop this assertion (and related code) as it should already be covered by earlier snapshot tests

akash1810: Good shout! Done.

akash1810:
For future reference - the commits on this PR are not exclusive to this bit of work. This is because the base branch of https://github.com/guardian/cdk/pull/1018 was this branch and not `main`. I can't remember why I did that! It should not cause any issues issuing a new release, it's just confusing! Sorry!

jacobwinch (approved):

PR #1017, opened 2021-12-29
feat: Add support for a static site hosted by S3 and fronted by CloudFront w/custom domain

akash1810:
## What does this change?

Adds a new construct `GuStaticSite` to produce a static site hosted by CloudFront with an S3 origin, creating the "Using a REST API endpoint as the origin, with access restricted by an OAI" layout described [here](https://aws.amazon.com/premiumsupport/knowledge-center/cloudfront-serve-static-website/).

There are few nuances with this setup, which this construct codifies:
  - An ACM certificate for a CloudFront distribution must be in the `us-east-1` region (until https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/523 ships)
  - Account numbers are considered private and should not be added to VCS
  - Riff-Raff's `aws-s3` type can only upload to a single bucket and will shard the bucket by `stage`. That is, CloudFront's origin path should not be the root of the bucket. 

The construct supports use of a pre-existing bucket and a pre-existing ACM certificate.

Example usage:

```typescript
new GuStaticSite(scope, "StaticSite", {
  app: "my-static-site",
  domainNameProps: {
    CODE: {
      domainName: "my-static-site.code.domain.co.uk",
    },
    PROD: {
      domainName: "my-static-site.prod.domain.co.uk",
    },
  },
});
```

Example usage, using a pre-existing ACM certificate:

```typescript
new GuStaticSite(scope, "StaticSite", {
  app: "my-static-site",
  domainNameProps: {
    CODE: {
      domainName: "my-static-site.code.domain.co.uk",
    },
    PROD: {
      domainName: "my-static-site.prod.domain.co.uk",
    },
  },
  certificateArn: {
    CODE: `arn:aws:acm:us-east-1:${scope.account}:certificate/abcde-01234-fghij-56789`,
    PROD: `arn:aws:acm:us-east-1:${scope.account}:certificate/01234-abcde-56789-fghij`,
  }
});
```

Example usage, using a pre-existing S3 bucket:

```typescript
new GuStaticSite(scope, "StaticSite", {
  app: "my-static-site",
  domainNameProps: {
    CODE: {
      domainName: "my-static-site.code.domain.co.uk",
    },
    PROD: {
      domainName: "my-static-site.prod.domain.co.uk",
    },
  },
  originBucketName: "my-static-site-origin",
});
```

In this instance, the bucket's policy needs to be updated separately to grant `s3:GetObject` permission to the CloudFront Origin Access Identity - the canonical user id is added as a CloudFormation Output for use like so:

```json
{
  "Effect": "Allow",
  "Action": "s3:GetObject"
  "Principal": { 
    "CanonicalUser": "64-digit-alphanumeric-value"
  },
  "Resource": "arn:aws:s3:::my-static-site-origin/*"
}
```

See also https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-restricting-access-to-s3.html.

## How to test

See added tests and https://github.com/guardian/cdk-static-site/blob/main/cdk/lib/static-site.ts for a real demo.

Creation of static sites is simpler.

This construct assumes DNS is managed outside of Route53.

## Checklist

### DIFF
--- a/src/constructs/static-site/static-site.ts
+++ b/src/constructs/static-site/static-site.ts
@@ -67,6 +67,11 @@ export interface GuStaticSiteProps
    * This is useful if you want to control the TTL, which defaults to 1 hour.
    */
   withoutDns?: boolean;
+
+  /**
+   * Name of the bucket to create, otherwise the bucket name will be auto-generated.
+   */
+  bucketName?: string;
 }
 
 /**
@@ -208,12 +213,12 @@ export class GuStaticSite extends GuAppAwareConstruct(CloudFrontWebDistribution)
 
   private static getBucket(
     scope: GuStack,
-    { app, preExistingOriginBucketName }: GuStaticSiteProps,
+    { app, preExistingOriginBucketName, bucketName }: GuStaticSiteProps,
     { cloudFrontOriginAccessIdentityS3CanonicalUserId }: OriginAccessIdentity
   ) {
     const bucket = preExistingOriginBucketName
       ? Bucket.fromBucketName(scope, AppIdentity.suffixText({ app }, "OriginBucket"), preExistingOriginBucketName)
-      : new GuS3Bucket(scope, "OriginBucket", { app });
+      : new GuS3Bucket(scope, "OriginBucket", { app, bucketName });
 
     if (preExistingOriginBucketName) {
       /*
--- a/src/constructs/static-site/static-site.ts
+++ b/src/constructs/static-site/static-site.ts
@@ -61,6 +61,12 @@ export interface GuStaticSiteProps
    * Defaults to compress responses and redirect http to https.
    */
   behaviours?: Behavior[];
+
+  /**
+   * Flag to opt-out of automatically creating a [[ `GuCname` ]]. You'll have to create it yourself.
+   * This is useful if you want to control the TTL, which defaults to 1 hour.
+   */
+  withoutDns?: boolean;
 }
 
 /**
@@ -252,6 +258,7 @@ export class GuStaticSite extends GuAppAwareConstruct(CloudFrontWebDistribution)
           viewerProtocolPolicy: ViewerProtocolPolicy.REDIRECT_TO_HTTPS,
         },
       ],
+      withoutDns = false,
     } = props;
     const { stage } = scope;
 
@@ -298,11 +305,13 @@ export class GuStaticSite extends GuAppAwareConstruct(CloudFrontWebDistribution)
       ],
     });
 
-    new GuCname(scope, AppIdentity.suffixText({ app }, "DnsRecord"), {
-      app,
-      domainNameProps,
-      resourceRecord: this.distributionDomainName,
-      ttl: Duration.hours(1),
-    });
+    if (!withoutDns) {
+      new GuCname(scope, AppIdentity.suffixText({ app }, "DnsRecord"), {
+        app,
+        domainNameProps,
+        resourceRecord: this.distributionDomainName,
+        ttl: Duration.hours(1),
+      });
+    }
   }
 }
--- a/src/constructs/static-site/static-site.ts
+++ b/src/constructs/static-site/static-site.ts
@@ -1,0 +1,308 @@
+import { Certificate } from "@aws-cdk/aws-certificatemanager";
+import {
+  CloudFrontWebDistribution,
+  OriginAccessIdentity,
+  ViewerCertificate,
+  ViewerProtocolPolicy,
+} from "@aws-cdk/aws-cloudfront";
+import type { Behavior, CloudFrontWebDistributionProps } from "@aws-cdk/aws-cloudfront/lib/web-distribution";
+import { CanonicalUserPrincipal, PolicyStatement } from "@aws-cdk/aws-iam";
+import { Bucket } from "@aws-cdk/aws-s3";
+import { Annotations, Arn, ArnFormat, CfnOutput, Duration, Token } from "@aws-cdk/core";
+import { Stage } from "../../constants";
+import type { GuDomainNameProps } from "../../types/domain-names";
+import { StageAwareValue } from "../../types/stage";
+import { GuAppAwareConstruct } from "../../utils/mixin/app-aware-construct";
+import { GuCertificate } from "../acm";
+import type { GuStack } from "../core";
+import { AppIdentity } from "../core/identity";
+import { GuCname } from "../dns";
+import { GuS3Bucket } from "../s3";
+
+const GLOBAL_AWS_REGION = "us-east-1";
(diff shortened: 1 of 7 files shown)


akash1810:
cc @tjmw

src/constructs/static-site/static-site.ts
```diff
@@ -0,0 +1,322 @@
...
+            originAccessIdentity,
+            originPath: `/${stage}/${app}`,
+          },
+          behaviors: behaviours,
+        },
+      ],
+    });
+
+    if (!withoutDns) {
+      new GuCname(scope, AppIdentity.suffixText({ app }, "DnsRecord"), {
```
jacobwinch, line 314: We [decided *not* to create the `GuCname`](https://github.com/guardian/cdk/discussions/913) as part of our other patterns. I'm not sure if the same drawbacks apply here, or whether we want to follow the same approach for consistency?

src/constructs/static-site/static-site.ts
```diff
@@ -0,0 +1,322 @@
...
+  /**
+   * Behaviours of the S3 bucket origin.
+   *
+   * Defaults to compress responses and redirect http to https.
+   */
+  behaviours?: Behavior[];
+
+  /**
+   * Flag to opt-out of automatically creating a [[ `GuCname` ]]. You'll have to create it yourself.
+   * This is useful if you want to control the TTL, which defaults to 1 hour.
```
jacobwinch, line 67: > This is useful if you want to control the TTL, which defaults to 1 hour.

If this is the case it might be better to allow users to pass in a custom TTL instead of opting out completely?

I guess there is another key use case for opting out too: if you are using Route53 to manage DNS?

src/constructs/static-site/static-site.ts
```diff
@@ -0,0 +1,322 @@
...
+   Validate an ACM ARN and return the resource name for later use in a CFN Mapping.
+   This is because a CFN Mapping cannot include the `AccountId` pseudo function.
+
+   @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/pseudo-parameter-reference.html#cfn-pseudo-param-accountid
+   @see https://docs.aws.amazon.com/cdk/v2/guide/tokens.html
+    */
+  private static getResourceNameFromArn(certificateArn: string): string {
+    const { account, region, resource, resourceName } = Arn.split(certificateArn, ArnFormat.SLASH_RESOURCE_NAME);
+
+    if (!Token.isUnresolved(account)) {
```
jacobwinch, line 149: This is very neat 👏

src/constructs/static-site/static-site.ts
```diff
@@ -0,0 +1,322 @@
...
+  /**
+   * ARN of a pre-existing ACM certificate to attach to CloudFront. It must:
+   *   - exist in the us-east-1 region
+   *   - not have a hard coded account number
+   *
+   * If not provided, a new certificate is created. Note, this means the stack must be deployed in the us-east-1 region.
+   *
+   * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cloudfront-distribution-viewercertificate.html#cfn-cloudfront-distribution-viewercertificate-acmcertificatearn
+   */
+  preExistingCertificateArn?: StageAwareValue<string>;
```
jacobwinch, line 45: Is there any particular reason that we want to support the usage of pre-existing certificates? 

IIUC it's possible to have more than one certificate with the same domain name and creating this as part of the pattern ensures that it is under version control with the rest of the infrastructure.

I guess by removing this option we'd be forcing people to deploy all static sites into `us-east-1`, but I'm not convinced that's much of a problem.

src/constructs/static-site/static-site.ts
```diff
@@ -0,0 +1,322 @@
...
+  }
+
+  private static getBucket(
+    scope: GuStack,
+    { app, preExistingOriginBucketName, bucketName }: GuStaticSiteProps,
+    { cloudFrontOriginAccessIdentityS3CanonicalUserId }: OriginAccessIdentity
+  ) {
+    const bucket = preExistingOriginBucketName
+      ? Bucket.fromBucketName(scope, AppIdentity.suffixText({ app }, "OriginBucket"), preExistingOriginBucketName)
+      : new GuS3Bucket(scope, "OriginBucket", { app, bucketName });
```
jacobwinch, line 221: I think this will create a bucket per stage? But in the PR description you mentioned that:

> Riff-Raff's aws-s3 type can only upload to a single bucket and will shard the bucket by stage. 

I'm wondering if we should force users to pass in a pre-existing bucket unless/until we change Riff-Raff?

jacobwinch: Alternatively we could use a [condition](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/conditions-section-structure.html) to create the bucket in one stage only, but that feels pretty gross!

src/constructs/static-site/static-site.ts
akash1810: This is a bit of a rabbit hole... In order to fully support the deployment of a static site into `us-east-1` directly, we'd need to [enable `Guardian::DNS::RecordSet` private type into that region](https://github.com/guardian/cfn-private-resource-types/issues/25). https://github.com/guardian/cfn-private-resource-types/pull/24 attempted to do that, but [Riff-Raff wasn't happy](https://riffraff.gutools.co.uk/deployment/view/0da119d3-0c18-4945-b9bf-fbb2370a150b), so we'd need to fix that first.

The `preExistingCertificateArn` option is the ~laziest~ simplest and also means a static site can be defined in the same stack as a Play app (which we run in `eu-west-1`).

jacobwinch: > ...means a static site can be defined in the same stack as a Play app (which we run in eu-west-1).

I hadn't thought of this, thanks for explaining 👍

src/constructs/static-site/static-site.ts
akash1810: The only reason I opted to create it by default is we're already creating the CloudFront distribution and passing in the `domainNameProps`. Consistency would be better, however so will revisit this.

src/constructs/static-site/static-site.ts
akash1810: Ah, hadn't noticed this, great spot! Conditions are pretty horrible imo and I'd prefer to keep this kind of logic in the domain of TypeScript (i.e. compile time).

Maybe an alternative approach could be to require an [`IBucket`](https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_aws-s3.IBucket.html) and leave it to the user to either create a new bucket or use `Bucket.fromBucketName`?

akash1810: For further discussion - https://github.com/guardian/recommendations/discussions/74.

akash1810:
Closing as stale. Please re-open if still needed though!

PR #1018, opened 2021-12-30
fix!: Make S3 buckets follow best practices

akash1810:
## What does this change?

Updates `GuS3Bucket` to be private by default and also become app aware, which results in the bucket receiving the `App` tag.

BREAKING CHANGE: The `app` prop for a `GuS3Bucket` is now required.

## How to test

See added tests.

Buckets follow best practice by default.

By becoming app aware, the auto generated logical id changes. This results in resource replacement 😱 ! [Usage](https://github.com/search?q=org%3Aguardian+GuS3Bucket&type=code) of this construct isn't very big and those stacks that are using it are also preserving a logical id from a previously YAML defined stack.

## Checklist


nicl (approved):
Sometimes breaking changes are worth it! 👍

PR #1019, opened 2021-12-30
fix: `GuEcsTask` uses `stack` and `stage` from parent `GuStack` for consistency

akash1810:
## What does this change?

We do not need to specify `stack` and `stage` at the component/pattern level as these properties are available on `GuStack`. Using the `GuStack` values also increases the type-safety as `stage` is _not_ a `string`.

This change also refactors to use destructing of the `props` object, which yields simpler code for default values.

## How to test

CI should continue to pass.

Increased type-safety and consistency.

n/a

## Checklist

### DIFF
--- a/src/constructs/ecs/ecs-task.ts
+++ b/src/constructs/ecs/ecs-task.ts
@@ -18,7 +18,6 @@ import { EcsFargateLaunchTarget, EcsRunTask } from "@aws-cdk/aws-stepfunctions-t
 import { CfnOutput, Duration } from "@aws-cdk/core";
 import type { NoMonitoring } from "../cloudwatch";
 import type { GuStack } from "../core";
-import type { Identity } from "../core/identity";
 import { AppIdentity } from "../core/identity";
 import { GuGetDistributablePolicyStatement } from "../iam";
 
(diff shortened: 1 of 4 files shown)


src/constructs/ecs/ecs-task.ts
```diff
@@ -139,112 +138,120 @@ export class GuEcsTask {
...
   constructor(scope: GuStack, id: string, props: GuEcsTaskProps) {
-    const timeout = props.taskTimeoutInMinutes ?? 15;
+    const {
+      app,
 
-    // see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-taskdefinition.html#cfn-ecs-taskdefinition-cpu for details
-    const defaultCpu = 2048; // 2 cores and from 4-16GB memory
-    const defaultMemory = 4096; // 4GB
+      // see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-taskdefinition.html#cfn-ecs-taskdefinition-cpu for details
+      cpu = 2048, // 2 cores and from 4-16GB memory
```
nicl, line 145: Somehow I never realised defaults with destructuring was possible! But it's really nice.

akash1810: Yeah it's really nice! I was inspired by an [old HTTP203 episode](https://youtu.be/te3Zm4bHBVs?t=240).

nicl (approved):
Really like the use of destructuring throughout to improve readability. And the core change to use stage/stack from the scope makes sense.

PR #1027, opened 2022-01-05
revert: "chore: Forgo keeping `version` in `package.json` up to date"

jacobwinch:
This PR reverts https://github.com/guardian/cdk/pull/1006. Although we don't need the version number in `package.json` for releasing, it is [used by `LibraryInfo`](https://github.com/guardian/cdk/blob/a33a14bd05528a202c47208b1a6d930e95a36f0f/src/constants/library-info.ts#L8-L31), so I think it's worth reinstating this.

### DIFF
--- a/.github/workflows/cd.yaml
+++ b/.github/workflows/cd.yaml
@@ -18,3 +18,17 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
         run: ./script/cd
+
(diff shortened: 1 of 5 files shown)


akash1810 (approved):
`package.json` will look correct in the published module on npm. We can confirm this via this convoluted script 😄:

```sh
cd /tmp \
&& npm pack --silent @guardian/cdk \
&& tar -xzf guardian-cdk-33.0.0.tgz \
&& cat package/package.json | jq .
```

It's probably more confusing to remember this than it's worth, so let's get this in.

PR #1028, opened 2022-01-05
feat: Use latest versions of @guardian/cdk and aws-cdk when generating projects

jacobwinch:
## What does this change?

This PR ensures that the project generation tooling always pulls in the latest `@guardian/cdk` version and the appropriate `aws-cdk` versions.

## How to test

Run `./script/cli new --app "myApp" --stack "myStack"` locally. Note that this relies on https://github.com/guardian/cdk/pull/1027 (or at the very least, temporarily hardcoding the version in the root-level `package.json`) to work properly locally.

## How can we measure success?

New projects use the latest `cdk` dependencies by default.

## Have we considered potential risks?

Yes, there is a risk that by automatically updating these library versions we'll break the project generation tooling. We're planning to add some integration tests (in a future PR) to mitigate this risk.

## Checklist

- [x] I have listed any breaking changes, along with a migration path [^1] N/A
- [x] I have updated the documentation as required for the described changes [^2] N/A

[^1]: Consider whether this is something that will mean changes to projects that have already been migrated, or to the CDK CLI tool. If changes are required, consider adding a checklist here and/or linking to related PRs.
[^2]: If you are adding a new construct or pattern, has new documentation been added? If you are amending defaults or changing behaviour, are the existing docs still valid?

PR #1031, opened 2022-01-06
test: run project generation after every release

jacobwinch:
## What does this change?

In the past, the project generation tooling (which previously lived in https://github.com/guardian/cdk-cli) frequently became out of date. We recently migrated this code into the same repository as the constructs/patterns (see https://github.com/guardian/cdk/pull/962 and https://github.com/guardian/cdk/pull/971) and started automatically using the latest `cdk` libraries in all newly generated projects (https://github.com/guardian/cdk/pull/1028).

This PR checks that the project generation tooling still works after every release. 

## How to test

I've confirmed that this command will run successfully in GitHub Actions (via a [workflow run](https://github.com/jacobwinch/post-release-testing/runs/4726003628?check_suite_focus=true) in another repository).

By running this command after every release we should notice more quickly if the project generation tooling is broken by changes to the constructs/patterns that it uses.

We will only catch failures related to project generation tooling _after_ the library has been released, so we'll need to be ready to rollback/fix forwards in case of problems. 

If this ends up happening regularly, we may want to consider completing the work to perform this check at CI time (i.e. before release) instead. I've avoided this for now as the work is more complex and it will have an impact on CI performance for every PR.

## Checklist

- [x] I have listed any breaking changes, along with a migration path [^1] N/A
- [x] I have updated the documentation as required for the described changes [^2] N/A

### DIFF
--- a/.github/workflows/cd.yaml
+++ b/.github/workflows/cd.yaml
@@ -32,3 +32,8 @@ jobs:
           # See https://docs.github.com/en/actions/reference/events-that-trigger-workflows#triggering-new-workflows-using-a-personal-access-token
           github-token: ${{ secrets.GU_GUARDIAN_CI_TOKEN }}
           npm-lockfile-version: 2
+
+      - if: ${{ github.ref == 'refs/heads/main' }}
+        name: run-project-generation
+        # This ensures that the newly released version of the library can still be used to generate a new project from scratch
+        run: npx @guardian/cdk@latest new --stack my-stack --app my-app


.github/workflows/cd.yaml
```diff
@@ -32,3 +32,8 @@ jobs:
           # See https://docs.github.com/en/actions/reference/events-that-trigger-workflows#triggering-new-workflows-using-a-personal-access-token
           github-token: ${{ secrets.GU_GUARDIAN_CI_TOKEN }}
           npm-lockfile-version: 2
+
+      - if: ${{ github.ref == 'refs/heads/main' }}
+        name: run-project-generation
+        # This ensures that the newly released version of the library can still be used to generate a new project from scratch
+        run: npx @guardian/cdk@latest new --stack my-stack --app my-app
```
jacobwinch, line 39: The `new` command [installs dependencies](https://github.com/guardian/cdk/blob/bf89943f4ea655994088870d9caa4e08bd5e2089/src/bin/commands/bootstrap.ts#L133-L135) and [runs snapshot tests](https://github.com/guardian/cdk/blob/bf89943f4ea655994088870d9caa4e08bd5e2089/src/bin/commands/bootstrap.ts#L147-L149), so if the command exits successfully we should have a good level of confidence that the project generation tooling is working correctly.

akash1810:
I wonder if it would be better to have this as it's own GHA workflow? Maybe even with a different trigger, such as [release creation](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#release) or tag creation?

jacobwinch:
> Maybe even with a different trigger, such as [release creation](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#release) or tag creation?

I considered the former. I was concerned that it wouldn't work because I believe that [our releases are created using the `GITHUB_TOKEN`](https://github.com/semantic-release/semantic-release/blob/master/docs/usage/ci-configuration.md#authentication)(?) and [this restriction](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow) states that:

> When you use the repository's `GITHUB_TOKEN` to perform tasks on behalf of the GitHub Actions app, events triggered by the `GITHUB_TOKEN` will not create a new workflow run.

nicl (approved):
Great improvement to test the command in CI.

.github/workflows/cd.yaml
jacobwinch: Because we have chosen to run this command post-release, it's happening as part of a workflow that is only triggered for `main` and `beta` branches anyway: https://github.com/guardian/cdk/blob/9918d03f7118d354d2accb0ebeb3bdef63c63180/.github/workflows/cd.yaml#L4-L6

This new step is using the latest _released_ version of the `cli` (via `npx`). The cli command itself is installing the latest _released_ version of the library from `npm` (i.e. the version that will have been released by earlier steps from this workflow) in order to generate a new `cdk` project. If we were running this step against other branches it wouldn't really make sense IMO, because we would be testing the version of the library that's been released, rather than the version that exists on the feature branch. 

As @akash1810 correctly points out:

> not every change to `main` produces a new release.

so technically running it against _every_ build of `main` will still mean that we will sometimes execute this command superfluously. I don't think the cost of that is too high because this build doesn't block anyone/anything.

PR #1047, opened 2022-01-14
refactor: Make default values for EC2 pattern more obvious

akash1810:
## What does this change?

Using destructuring with default values makes things a little clearer as we can co-locate things. This follows a pattern introduced in https://github.com/guardian/cdk/pull/1019.

This is a no-op as demonstrated by the existing tests continuing to pass.

This is actually part of a wider change to allow the EC2 pattern to support the `INFRA` stage - it currently supports `CODE` and `PROD` only.


nicl (approved):
Really like this pattern :)

PR #1048, opened 2022-01-14
fix: Support for `GuStackForInfrastructure` within `GuAutoScalingGroup`

akash1810:
## What does this change?

The `GuAutoScalingGroup` construct defaults its `min` and `max` values, but only for the `CODE` and `PROD` stages. It lacks support for the recently added `INFRA` stage (https://github.com/guardian/cdk/pull/948).

In this change we fix this 🎉 !

All constructs within this library take a `GuStack` as an argument in their constructor. As `GuStackForInfrastructure` extends `GuStack`, we need a way to determine which one we have in order to set an appropriate `min` and `max`. To achieve this, we use a [`Symbol`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol), allowing us to check for it's presence within a [user defined type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards). This is inspired by [AWS CDK](https://github.com/aws/aws-cdk/blob/3721358fa1501e42b3514b8a8f15f05c9615f149/packages/%40aws-cdk/core/lib/construct-compat.ts#L65-L67).

## How to test

See added test. The test was added TDD style, so reviewing the commits in order might be easier?

It is possible to use the EC2 pattern for infrastructure based stacks, for example: Security HQ, Janus or even CDK Playground. These stacks would need redefining/migrating.

Locally, I've observed the execution time of the test suite increase slightly. I'm not sure if this is due to a new test being added, or because the test does snapshot checking. It's not a dramatic increase, but one to keep an eye on if we find our build times become too long.

## Checklist

### DIFF
--- a/src/constructs/autoscaling/asg.ts
+++ b/src/constructs/autoscaling/asg.ts
@@ -3,11 +3,11 @@ import type { AutoScalingGroupProps, CfnAutoScalingGroup } from "@aws-cdk/aws-au
 import { OperatingSystemType, UserData } from "@aws-cdk/aws-ec2";
 import type { ISecurityGroup, MachineImage, MachineImageConfig } from "@aws-cdk/aws-ec2";
 import type { ApplicationTargetGroup } from "@aws-cdk/aws-elasticloadbalancingv2";
-import { Stage } from "../../constants";
+import { Stage, StageForInfrastructure } from "../../constants";
 import { StageAwareValue } from "../../types/stage";
 import { GuStatefulMigratableConstruct } from "../../utils/mixin";
 import { GuAppAwareConstruct } from "../../utils/mixin/app-aware-construct";
-import { GuAmiParameter } from "../core";
+import { GuAmiParameter, GuStackForInfrastructure } from "../core";
 import type { GuStack } from "../core";
 import type { AppIdentity } from "../core/identity";
 import type { GuMigratingResource } from "../core/migrating";
@@ -125,10 +125,16 @@ export class GuAutoScalingGroup extends GuStatefulMigratableConstruct(GuAppAware
       };
     }
 
-    const defaultStageDependentProps = {
-      [Stage.CODE]: { minimumInstances: 1 },
-      [Stage.PROD]: { minimumInstances: 3 },
-    };
+    const defaultStageDependentProps: GuStageDependentAsgProps = GuStackForInfrastructure.isGuStackForInfrastructure(
+      scope
+    )
+      ? {
+          [StageForInfrastructure]: { minimumInstances: 1 },
+        }
+      : {
+          [Stage.CODE]: { minimumInstances: 1 },
+          [Stage.PROD]: { minimumInstances: 3 },
+        };
 
     const mergedProps = {
       ...props,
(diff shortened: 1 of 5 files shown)


src/constructs/autoscaling/asg.ts
```diff
@@ -125,10 +125,16 @@ export class GuAutoScalingGroup extends GuStatefulMigratableConstruct(GuAppAware
       };
     }
 
-    const defaultStageDependentProps = {
-      [Stage.CODE]: { minimumInstances: 1 },
-      [Stage.PROD]: { minimumInstances: 3 },
-    };
+    const defaultStageDependentProps: GuStageDependentAsgProps = GuStackForInfrastructure.isGuStackForInfrastructure(
+      scope
+    )
```
akash1810, line 130: Formatting curtesy of the linter 😅 .

nicl (commented):
I'm a bit nervous about this change (but also our whole approach to stages). We're introducing complex code, for what is quite a brittle solution. At the moment, supporting additional stages requires quite a lot of work and our code is tightly coupled to existing stacks. What about, PROD-ZEBRA, etc.?

I would prefer a solution that makes our constructs/patterns completely ignorant of stages and simply relies on props to configure things.

akash1810:
> I'm a bit nervous about this change ... We're introducing complex code, for what is quite a brittle solution.

I don't quite agree re brittleness, I think our tests just were not covering this scenario previously.

> At the moment, supporting additional stages requires quite a lot of work and our code is tightly coupled to existing stacks. What about, PROD-ZEBRA, etc.?
> 
> I would prefer a solution that makes our constructs/patterns completely ignorant of stages and simply relies on props to configure things.

Agreed! However it is quite tricky to achieve this. I wouldn't be in favour of having `Stage` be a `string` as a lot of type-safety is lost.

I think the ultimate solution would be for Riff-Raff to support CDK as a first class citizen. This will allow us to remove the complexities of `Stage` and `CFN Mappings` entirely, in favour of looking up values from the [context](https://docs.aws.amazon.com/cdk/v2/guide/context.html).

For example, `cdk.json` could be something like this:

```json
{
  "app": "npx ts-node bin/cdk.ts",
  "context": {
    "@aws-cdk/core:enableStackNameDuplicates": "true",
    "aws-cdk:enableDiffNoFail": "true",
    "@aws-cdk/core:stackRelativeExports": "true",
    "gu:cdk": {
      "PROD": {
        "domain": "app.gutools.co.uk",
        "min": 1
      },
      "PROD-ZEBRA": {
        "domain": "app-2.gutools.co.uk",
        "min": 1
      }
    }
  }
}
```

With Riff-Raff natively supporting CDK and looking up values from the context, we remove all `Fn::Sub` and `Fn::FindInMap` calls in the generated template.

WDYT?

akash1810:
> What about, PROD-ZEBRA, etc.?

There's also a (smaller) argument that GuCDK is an opinionated library. `PROD-ZEBRA` is _not_ a common stage value - do we might want to phase variations in stage? Or at least fully understand the use-cases first?

akash1810:
Superseded by #1092. Closing.

PR #1049, opened 2022-01-18
Add stage refactoring POC

nicl:
(Re. discussion https://github.com/guardian/cdk/pull/1048 and also in chat.)

POC of a new approach to stage-handling for props. To illustrate I have added some basic helper code and an example test stack (with output). The GuCertificate construct was duplicated. You can see the existing construct [here](https://github.com/guardian/cdk/blob/ce51d80e5f294c1df70acc81ca1cb537bbb44666/src/constructs/acm/certificate.ts) though and note, in particular, the stage handling, which is quite verbose and complicated. E.g. things like:

```
    const maybeHostedZone = !hasHostedZoneId
      ? undefined
      : HostedZone.fromHostedZoneId(
          scope,
          /* eslint-disable @typescript-eslint/no-non-null-assertion -- `hasHostedZoneId` is true, so we know `hostedZoneId` is present here */
          AppIdentity.suffixText({ app: props.app }, "HostedZone"),
          StageAwareValue.isStageValue(props)
            ? scope.withStageDependentValue({
                app: props.app,
                variableName: "hostedZoneId",
                stageValues: {
                  [Stage.CODE]: props.CODE.hostedZoneId!,
                  [Stage.PROD]: props.PROD.hostedZoneId!,
                },
              })
            : props.INFRA.hostedZoneId!
          /* eslint-enable @typescript-eslint/no-non-null-assertion */
        );
```

which now becomes simply:

```
    const maybeHostedZone = !props.hostedZoneId
      ? undefined
      : HostedZone.fromHostedZoneId(
          scope,
          AppIdentity.suffixText({ app: props.app }, "HostedZone"),
          props.hostedZoneId
        );
```

(Certificate has just 2 of these kinds of stage-aware props, but you can imagine that our more complicated constructs have several.)

For rationale see below. 

---

The main aim was to make constructs and patterns ignorant of stages. This achieves a few things:

* keeps them simple (LOC and also readability/complexity)
* ensures our codebase can handle unusual stage requirements  without significant refactoring
* simplifies the INFRA case (it requires no special code at all)
* makes it easier to e.g. use `cdk deploy` in the future

The downside is that constructs become less opinionated on which props should be stage-specific. Generally, I think this is fine, but there might be cases where e.g. if one prop is customised by stage we would like to ensure another (related) prop is too.

### DIFF
--- a/src/constructs/core/stage.test.ts
+++ b/src/constructs/core/stage.test.ts
@@ -1,0 +1,101 @@
+import "@aws-cdk/assert/jest";
+
+import "../../utils/test/jest";
+import { SynthUtils } from "@aws-cdk/assert";
+import { App, CfnMapping } from "@aws-cdk/core";
+import { GuStack } from "./stack";
+import type { Stage } from "./stage";
+import { GuCertificateExample, NewStageMapping } from "./stage";
+
+describe("Stage refactor (POC)", () => {
+  it("should output a mapping as expected", () => {
+    const stack = new GuStack(new App(), "Test", { stack: "test" });
+
+    const mappings = NewStageMapping(new CfnMapping(stack, "foo"));
+    const domainName = mappings.set("domainName", stack.stage as Stage, {
+      CODE: "foo.example.com",
+      PROD: "prod.example.com",
+    });
+
+    new GuCertificateExample(stack, {
+      app: "foo",
+      domainName: domainName,
+    });
+
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment -- ignore
+    const json = SynthUtils.toCloudFormation(stack);
+
+    console.log(JSON.stringify(json, null, 4));
+
+    /** Yields...
+
+    {
+        "Parameters": {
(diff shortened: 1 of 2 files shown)


src/constructs/core/stage.test.ts
```diff
@@ -0,0 +1,101 @@
...
+
+    const mappings = NewStageMapping(new CfnMapping(stack, "foo"));
+    const domainName = mappings.set("domainName", stack.stage as Stage, {
+      CODE: "foo.example.com",
+      PROD: "prod.example.com",
+    });
+
+    new GuCertificateExample(stack, {
+      app: "foo",
+      domainName: domainName,
```
nicl, line 22: (Construct just receives a string but actually it is a stage-dependant mapping ref.)

nicl:
Update, after some discussion, we're going to go ahead with this! Trello card here: https://trello.com/c/pp6ttXdI/1100-modify-stage-handling-in-guardian-cdk.

Closing this PR now.

PR #1063, opened 2022-01-27
fix: Align UnhealthyInstance alarm period with underlying metric period

jorgeazevedo:
`GuUnhealthyInstancesAlarm` is built on top of the `UnHealthyHostCount` metric, which is measured by the Load Balancer and posted in 60-second intervals. The alarm however, has a period of 5 minutes, which can have unintended consequences.

Let's consider an example data set, where the number of unhealthy instances is always zero except at `10:07` and `10:08`

```
___________
|10:00 | 0 |
|10:01 | 0 |
|10:02 | 0 |
|10:03 | 0 |
|10:04 | 0 |
|10:05 | 0 |
|10:06 | 0 |
|10:07 | 1 |
|10:08 | 1 |
|10:09 | 0 |
(etc.)
-----------
```

`GuUnhealthyInstancesAlarm` has a period of 5 minutes and evaluates 12 datapoints, which means it looks back at the past hour of data and aggregates it into 12 equally spaced buckets. When the alarm is evaluated at `11:02`, the resulting set of datapoints is

```
___________
|10:02 | 0 |
|10:07 | 1 |
|10:12 | 0 |
|10:17 | 0 |
(etc.)
-----------
```
Where only 1 datapoint is above 0. The two values of 1 fell into the same bucket. However, when it is evaluates at `11:03`, the result is

```
___________
|10:03 | 1 |
|10:08 | 1 |
|10:12 | 0 |
|10:17 | 0 |
(etc.)
-----------
```

Where now 2 datapoints are above 0. The two values of 1 fell into different buckets.

The rolling window nature of the evaluation makes the results of this aggregation operation unstable. Out in the field, the behaviour is that an alarm can go into the ALARM state one minute, only to then go into an OK state the next even though it hasn't really recovered yet.

Untitled_document_-_Google_Docs

 This is such counter intuitive behaviour, AWS added a little question mark to this UI that explains this

Untitled_document_-_Google_Docs

## What does this change?
This changes the period of `GuUnhealthyInstancesAlarm` to be one minute instead. This aligns the alarm with the metric, and creates more deterministic alarm conditions. The number of datapoints above or below the threshold is always the same from the point it it first enters the system, until it leaves the time window considered by the alarm.

CloudWatch_Management_Console

This does however change the alarm condition substantially. The direct mathematical equivalent is 30 datapoints out of 60, which means 30 minutes of unhealthy instances in an hour. With the period of 5 minutes, 6 minutes of unhealthy instances if spread out right could be sufficient to trigger an alarm.

@jacobwinch this alarm was introduced in #594, can you see a sensible alarm threshold based on a 1 minute period? I'm not sure what to propose here, I'm lacking the sensibility for what an episode of instances cycling looks like. I was able to trigger the alarm before by deploying prism 4 or 5 times in a row, now it needs to be like 9 or 10.

## How to test
I've been testing these alarm parameters in prism CODE.

No more false alarms and false alarm recoveries!

Cloudwatch alarms are very easy to get wrong and might have unintended consequences. The alarm can arrive/earlier later than expected, generate noise or false alarms.

## Checklist

### DIFF
--- a/src/constructs/cloudwatch/__snapshots__/ec2-alarms.test.ts.snap
+++ b/src/constructs/cloudwatch/__snapshots__/ec2-alarms.test.ts.snap
@@ -495,7 +495,7 @@ Object {
           ],
         },
         "ComparisonOperator": "GreaterThanOrEqualToThreshold",
-        "DatapointsToAlarm": 6,
+        "DatapointsToAlarm": 30,
         "Dimensions": Array [
           Object {
             "Name": "LoadBalancer",
@@ -558,10 +558,10 @@ Object {
             },
           },
         ],
-        "EvaluationPeriods": 12,
+        "EvaluationPeriods": 60,
         "MetricName": "UnHealthyHostCount",
         "Namespace": "AWS/ApplicationELB",
-        "Period": 300,
+        "Period": 60,
         "Statistic": "Maximum",
         "Threshold": 1,
         "TreatMissingData": "notBreaching",
--- a/src/constructs/cloudwatch/ec2-alarms.ts
+++ b/src/constructs/cloudwatch/ec2-alarms.ts
@@ -63,8 +63,8 @@ export class GuUnhealthyInstancesAlarm extends GuAlarm {
   constructor(scope: GuStack, props: GuUnhealthyInstancesAlarmProps) {
     const alarmName = `Unhealthy instances for ${props.app} in ${scope.stage}`;
 
-    const period = Duration.minutes(5);
-    const evaluationPeriods = 12;
+    const period = Duration.minutes(1);
+    const evaluationPeriods = 60;
     const evaluationInterval = Duration.minutes(period.toMinutes() * evaluationPeriods).toHumanString();
 
     const alarmDescription = `${props.app}'s instances have failed healthchecks several times over the last ${evaluationInterval}.
@@ -80,7 +80,7 @@ export class GuUnhealthyInstancesAlarm extends GuAlarm {
       treatMissingData: TreatMissingData.NOT_BREACHING,
       threshold: 1,
       comparisonOperator: ComparisonOperator.GREATER_THAN_OR_EQUAL_TO_THRESHOLD,
-      datapointsToAlarm: 6,
+      datapointsToAlarm: 30,
       evaluationPeriods,
     };
     super(scope, AppIdentity.suffixText(props, "UnhealthyInstancesAlarm"), alarmProps);


jacobwinch (approved):
This is a fantastic PR description - I finally understand the underlying issue 😅 

> @jacobwinch this alarm was introduced in #594, can you see a sensible alarm threshold based on a 1 minute period? I'm not sure what to propose here, I'm lacking the sensibility for what an episode of instances cycling looks like.

I worry that 30 will be slightly too high, but I think we will need to set up a test scenario to know for sure. I am happy to pair on that sometime (please only volunteer for this if you are especially interested!) or pick it up alone as part of this sprint.

I don't think that experiment needs to hold up merging this though - overall it's certainly an improvement and will help us to cut down on unnecessary noise. Thanks for picking this up and articulating the problem (and solution) so clearly 👍

jorgeazevedo:
Thanks Jacob I'm glad it tracks!

yeah, it's difficult to predict the behaviour! For example, if the asg reports 0 and 1 unhealthy instances in an alternating fashion like `01010101`, it will take the full 60 minutes for the alarm to trigger. And if there's two zeros in a row the alarm will immediately recover.

I wonder if 15 out of 60 would be a good balance? If the metric is just continuously at 1, it will take 15 minutes to alert. If it alternates, 30. I've tested it now for prism and it took 5 deployments, about 30 minutes.

CloudWatch_Management_Console

jacobwinch:
One of the scenarios that I'm particularly interested in catching with this alarm is the case where we cannot replace an instance (or cannot scale up). I ran a small experiment to test how @jorgeazevedo's version of the alarm behaves in this scenario:

1. [Configured a test environment (based on this branch)](https://github.com/guardian/cdk-playground/pull/154/commits/467d8f6b57139e25d43ebb7aca87d2168b1e9091).
2. Uploaded a [broken version of the application](https://github.com/guardian/cdk-playground/pull/154/commits/4587bbb82eb7c708945f0a5df47a81594bca6c20) code to S3.
3. Manually terminated an instance (at ~13:03). 
   i. When a new instance is launched it picks up the broken application code, which means that it will never pass the healthchecks. As a result the ASG will constantly cycle instances (which is something we often see in the field). 
4. Received an alarm notification at 13:41.

image

@jorgeazevedo and I discussed this experiment in detail:

* It's interesting to note that we see small spikes in the unhealthy hosts metric whenever an application is deployed. This is because the metric we are using is based on [the target group's concept of health](https://github.com/guardian/cdk/blob/4a8e7d55998d31115a590c01752e363603cf6728/src/constructs/cloudwatch/ec2-alarms.ts#L79). Unlike an ASG, which [applies a grace period](https://github.com/guardian/cdk/blob/4a8e7d55998d31115a590c01752e363603cf6728/src/patterns/ec2-app.ts#L398) before performing health checks, the target group begins performing health checks almost immediately after a new instance is launched/registered - often long before the application has had chance to start. This isn't a 'real problem' that developers need to worry about, so we don't want this to accidentally trip our alarm.
   * As a result we considered using the healthy hosts metric instead (as the graph above demonstrates, this is far more stable). However, to achieve our goal, we'd need to compare the number of healthy instances to the desired capacity of the ASG. As desired capacity is a moving target, there is no way to encode this using `cdk`.
   * We also considered using an ASG metric instead, but we couldn't find anything which looked appropriate.

As a result of the above, we would be reluctant to reduce `datapointsToAlarm` below 30. For example, if we used 15 (as discussed above) there is a chance that 7 or 8 deployments in quick succession could trip the alarm; this is not inconceivable for a 'busy' project like `dotcom-rendering`.

jacobwinch (approved):
I've tried to summarise our discussion from yesterday above, but the tl;dr is: let's ship this 🚢  🎉

jorgeazevedo:
That's a great summary @jacobwinch thanks for putting that together. It will be a very useful reference if (when? lol) we need to revisit this in the future!

PR #1072, opened 2022-02-03
fix: Enforce a CNAME DNS entry has a single answer

akash1810:
## What does this change?

There are currently two ways to create `Guardian::DNS::RecordSet` resource:
  1. `GuCname`
  2. `GuDnsRecordSet`

A CNAME should not return multiple answers without "the correct filters in place (such as SELECT_FIRST_N 1) to limit them to a single answer at resolution time".

`Guardian::DNS::RecordSet` does not create any filters (and it's unlikely to).

The props of `GuCname` enforce this single answer restriction, however if you're still able to use `GuDnsRecordSet` to create a CNAME that violates this restriction.

In this change we throw if a CNAME is being created with multiple answers.

## How to test

See added test.

We're prevented from deploying a potentially broken DNS entry.

n/a

## Checklist

### DIFF
--- a/src/constructs/dns/dns-records.ts
+++ b/src/constructs/dns/dns-records.ts
@@ -26,16 +26,37 @@ export interface GuDnsRecordSetProps {
  */
 export class GuDnsRecordSet {
   constructor(scope: GuStack, id: string, props: GuDnsRecordSetProps) {
+    const { name, recordType, resourceRecords, ttl } = props;
+    const { stage } = scope;
+
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- more `RecordType`s will be added soon!
+    if (recordType === RecordType.CNAME) {
+      /*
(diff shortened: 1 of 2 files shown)


src/constructs/dns/dns-records.ts
```diff
@@ -26,16 +26,37 @@ export interface GuDnsRecordSetProps {
  */
 export class GuDnsRecordSet {
   constructor(scope: GuStack, id: string, props: GuDnsRecordSetProps) {
+    const { name, recordType, resourceRecords, ttl } = props;
+    const { stage } = scope;
+
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- more `RecordType`s will be added soon!
```
akash1810, line 32: As `RecordType` has a single value (`CNAME`), the linter is suggesting this `if` is not necessary as it'll always be true. We'll be adding `TXT` and other records soon, so ignore the linter.

jacobwinch (approved):

PR #1075, opened 2022-02-07
feat: Make ephemeralStorage configurable via props for ecs task

philmcmahon:
## What does this change?
This makes [ephemeralStorage](https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_aws-ecs.TaskDefinition.html#ephemeralstoragegib) configurable in the ECS Task construct. 

On the investigations team our ECS tasks are running out of disk space due to the amount of data they are processing. We need to configure a larger amount of disk space for the fargate containers.

## How to test
Set the storage prop, verify that it gets applied to the containers created by the task.

## Checklist

### DIFF
--- a/src/constructs/ecs/ecs-task.ts
+++ b/src/constructs/ecs/ecs-task.ts
@@ -110,6 +110,7 @@ export interface GuEcsTaskProps extends AppIdentity {
   securityGroups?: ISecurityGroup[];
   customTaskPolicies?: PolicyStatement[];
   environmentOverrides?: TaskEnvironmentVariable[];
+  storage?: number;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


akash1810 (commented):
LGTM (once the build passes).

philmcmahon:
Thanks @akash1810 ! You were right I needed to `npm install` doh

akash1810 (approved):

src/constructs/ecs/ecs-task.ts
```diff
@@ -144,7 +145,7 @@ export class GuEcsTask {
       // see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-taskdefinition.html#cfn-ecs-taskdefinition-cpu for details
       cpu = 2048, // 2 cores and from 4-16GB memory
       memory = 4096, // 4GB
-
+      storage = 20, //20 GB
```
tapichu, line 148: This is an invalid input. Valid value for `ephemeralStorageGiB` are:

* Undefined (which will default to 20 GiB).
* Between 21 and 200 (inclusive).

PR #1076, opened 2022-02-07
feat: broaden parameter store read permissions

jacobwinch:
## What does this change?

This PR adds permissions for some additional (read-only) parameter store API calls to the `GuParameterStoreReadPolicy`. I've based the change on [an example from AWS](https://docs.aws.amazon.com/systems-manager/latest/userguide/sysman-paramstore-access.html).

The primary motivation for this change was to make it easier for apps to adopt [the `play-secret-rotation` library](https://github.com/guardian/play-secret-rotation/blob/main/aws-parameterstore/README.md#updating-applicationcomponents-with-the-rotating-secret). This library uses a slightly different API call to the one that we currently grant permissions for (it uses `ssm:GetParameters` rather than `ssm:GetParametersByPath`). 

Although the primary motivation only applies to Play apps, this PR grants these additional permissions to all EC2 and Lambda based patterns by default - I think it's fine for these apps to read configuration from parameter store using whichever API calls they/their dependencies prefer.

## How to test

I've published a beta version of the library (based on this branch) and confirmed that it works as expected by [testing with Amigo](https://github.com/guardian/amigo/pull/693/commits/72135395d799e8dfb130a653656703a130d286cd).

There should be less friction when using `play-secret-rotation` with `@guardian/cdk`-defined infrastructure.

I presume that we have only granted `ssm:GetParametersByPath` permissions so far because they are sufficient to meet the [requirements of `simple-configuration`](https://github.com/guardian/simple-configuration#when-loading-the-configuration-from-ssm), rather than because of some particular risk/concern around the other permissions that I've added here. That is, I think we're allowing apps to access the same data in a slightly different way, rather than fundamentally changing the permissions of apps here.

It's worth noting that this PR makes the parameter store read-permissions more closely aligned with the way that we grant other permissions ([e.g. DynamoDB read-permissions](https://github.com/guardian/cdk/blob/7290a8596ea64738b82a908ec76cbfd0b1f6cd76/src/constructs/iam/policies/dynamodb.ts#L23-L27)).

## Checklist

### DIFF
--- a/src/constructs/iam/policies/parameter-store-read.ts
+++ b/src/constructs/iam/policies/parameter-store-read.ts
@@ -4,7 +4,11 @@ import type { GuStack } from "../../core";
 import type { AppIdentity } from "../../core/identity";
 import { GuPolicy } from "./base-policy";
 
-export class GuParameterStoreByPathReadPolicyStatement extends PolicyStatement {
+/**
+ * This is helpful for reading all private configuration for a given app. For example, the
+ * [simple-configuration](https://github.com/guardian/simple-configuration) library requires these permissions.
+ */
+export class ReadParametersByPath extends PolicyStatement {
   constructor(scope: GuStack, props: AppIdentity) {
     super({
       effect: Effect.ALLOW,
@@ -14,7 +18,12 @@ export class GuParameterStoreByPathReadPolicyStatement extends PolicyStatement {
   }
 }
 
-export class GuParameterStoreByNameReadPolicyStatement extends PolicyStatement {
+/**
+ * This is helpful for accessing specific pieces of private configuration. For example, the
+ * [play-secret-rotation](https://github.com/guardian/play-secret-rotation) library requires `ssm:GetParameters`
+ * permissions.
+ */
+export class ReadParametersByName extends PolicyStatement {
   constructor(scope: GuStack, props: AppIdentity) {
     super({
       effect: Effect.ALLOW,
@@ -25,15 +34,15 @@ export class GuParameterStoreByNameReadPolicyStatement extends PolicyStatement {
     });
   }
 }
-
+/**
+ * Grants read-only permissions for Parameter Store. These permissions are typically used for accessing private
+ * configuration. See [[`ReadParametersByPath`]] and [[`ReadParametersByName`]] for more details.
+ */
 export class GuParameterStoreReadPolicy extends GuAppAwareConstruct(GuPolicy) {
   constructor(scope: GuStack, props: AppIdentity) {
     super(scope, "ParameterStoreRead", {
       policyName: "parameter-store-read-policy",
-      statements: [
-        new GuParameterStoreByPathReadPolicyStatement(scope, props),
-        new GuParameterStoreByNameReadPolicyStatement(scope, props),
-      ],
+      statements: [new ReadParametersByPath(scope, props), new ReadParametersByName(scope, props)],
       ...props,
     });
   }
--- a/src/constructs/iam/roles/instance-role.ts
+++ b/src/constructs/iam/roles/instance-role.ts
@@ -26,7 +26,8 @@ export type GuInstanceRolePropsWithApp = GuInstanceRoleProps & AppIdentity;
  * 1. Allows for `ssh` access to an EC2 instance via [ssm-scala](https://github.com/guardian/ssm-scala) (instead of standard `ssh`).
  * 2. Allows EC2 instances to communicate with Wazuh, for security monitoring.
  * 3. Allows EC2 instances to download an artifact from AWS S3, for application deployment.
- * 4. Allows EC2 instances to download private configuration from AWS Parameter Store.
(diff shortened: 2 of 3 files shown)


src/constructs/cloudwatch/__snapshots__/lambda-alarms.test.ts.snap
```diff
@@ -199,7 +199,11 @@ Object {
               ],
             },
             Object {
-              "Action": "ssm:GetParametersByPath",
+              "Action": Array [
+                "ssm:GetParametersByPath",
+                "ssm:GetParameters",
+                "ssm:GetParameter",
```
jacobwinch, line 205: I don't strictly need `ssm:GetParameter` for the change I'm working on, but I didn't really see the harm in including it here.

src/constructs/iam/policies/parameter-store-read.ts
```diff
@@ -8,8 +8,8 @@ export class GuParameterStoreReadPolicyStatement extends PolicyStatement {
   constructor(scope: GuStack, props: AppIdentity) {
     super({
       effect: Effect.ALLOW,
-      actions: ["ssm:GetParametersByPath"],
-      resources: [`arn:aws:ssm:${scope.region}:${scope.account}:parameter/${scope.stage}/${scope.stack}/${props.app}`],
+      actions: ["ssm:GetParametersByPath", "ssm:GetParameters", "ssm:GetParameter"],
+      resources: [`arn:aws:ssm:${scope.region}:${scope.account}:parameter/${scope.stage}/${scope.stack}/${props.app}*`],
```
akash1810, line 12: Should this be `/*`?

akash1810: > presumably changing to `/*` would tighten this up?

Yeah, it adds more namespacing. My assumption is we have parameters like `/PROD/deploy/amigo/playSecret` and not `/PROD/deploy/amigo-playSecret`?

jacobwinch: Yes, I think you're right - I'll update this

akash1810 (approved):

src/constructs/iam/policies/parameter-store-read.ts
jacobwinch: Unfortunately https://github.com/guardian/cdk/pull/1076/commits/1b93f2d491fd032e4c90d7cf8f5035f1de4495c2 does not work as desired. A [deployment of Amigo (based on this change) fails](https://riffraff.gutools.co.uk/deployment/view/696519e8-e935-4d19-a47c-8a4d6bf4aee7) with the following error:

`User: arn:aws:sts::<redacted>:assumed-role/amigo-CODE-InstanceRoleAmigo75944747-1WWFSO5Q5QL8F/i-<redacted> is not authorized to perform: ssm:GetParametersByPath on resource: arn:aws:ssm:eu-west-1:<redacted>:parameter/CODE/deploy/amigo because no identity-based policy allows the ssm:GetParametersByPath action` 

It seems that adding the `/` means that the permissions no longer match the request that `simple-configuration` is making.

I think the only way that we can achieve the level of precision that we want is by using two separate policy statements: https://github.com/guardian/cdk/pull/1076/commits/a9197fccd4ec921568e9086aa6d201c5bda5e1ff

src/constructs/iam/policies/parameter-store-read.ts
```diff
@@ -4,7 +4,11 @@ import type { GuStack } from "../../core";
 import type { AppIdentity } from "../../core/identity";
 import { GuPolicy } from "./base-policy";
 
-export class GuParameterStoreReadPolicyStatement extends PolicyStatement {
+/**
+ * This is helpful for reading all private configuration for a given app. For example, the
+ * [simple-configuration](https://github.com/guardian/simple-configuration) library requires these permissions.
+ */
+export class ReadParametersByPath extends PolicyStatement {
```
jacobwinch, line 11: I'm not sure that prefixing everything with `Gu` is particularly helpful (especially in the case of classes like this which we don't expect users to interact with directly), so I've dropped it here for brevity.

akash1810: Yeah, the `Gu` prefix is interesting. I can't find an ADR to explain our use of it. The closest I can find is https://github.com/guardian/cdk/blob/main/docs/architecture-decision-records/001-constructs-and-patterns.md:

> This library should define a number of Guardian flavoured constructs which extend those provided by the AWS CDK library with Guardian defaults baked in. For example, a `GuAutoScalingGroup` and `GuApplicationLoadBalancer`.

> specially in the case of classes like this which we don't expect users to interact with directly

Do we need to export them then?

akash1810 (approved):

src/constructs/iam/policies/parameter-store-read.ts
jacobwinch: > Do we need to export them then?

Yes, unfortunately we need to use them slightly differently when configuring lambda permissions:

https://github.com/guardian/cdk/blob/3c9db96d2194f5a31e1533aa5f21bbbba345c2e9/src/constructs/lambda/lambda.ts#L122-L127

PR #1092, opened 2022-02-16
feat!: Remove stage awareness from constructs

akash1810:
This is a fairly big change. It is possibly easier to review commit by commit. Happy to discuss ways to break it down - I'd be keen for the changes to yield a single new release.

## What does this change?

Following discussions in #1049, #1048 and IRL, this PR moves to simplify the codebase by removing explicit stage awareness from these constructs:
  - `GuAutoScalingGroup`
  - `GuCertificate`
  - `GuCname`
  - `GuAlarm`

These constructs can still be made stage aware, however it is now the decision of the caller.

This change impacts the `Ec2App` pattern too, but thankfully not in a dramatic way. The prop `certificateProps` has been renamed to `domain`; this is because it is more explicit.

~❓ https://github.com/guardian/cdk-playground/pull/161 demonstrates this branch on a simple stack. If we didn't rename `certificateProps` to `domain` this change would be a no-op. Shall we _not_ rename the prop to increase backwards compatibility?~ reverted.

### Breaking changes
BREAKING CHANGE: The props to `GuAutoScalingGroup` has changed.

Scaling props in `GuAutoScalingGroup` are now of type `number` and no longer need to be decorated with stage details. The caller can optionally provide a Mapping (via `withStageDependentValue`) if necessary. `GuCertificate`, `GuCname`, and `GuAlarm` see a similar change.

From:

```ts
new GuAutoScalingGroup(stack, "AutoscalingGroup", {
   // other required props
  stageDependentProps: {
    [Stage.CODE]: { minimumInstances: 2 },
    [Stage.PROD]: { minimumInstances: 5 },
  },
});
```

To:

```ts
// Stage agnostic
new GuAutoScalingGroup(stack, "AutoscalingGroup", {
   // other required props
  minimumInstances: 2,
  maximumInstances: 11,
});

// Stage aware
const app = "TestApp";

new GuAutoScalingGroup(stack, "AutoScalingGroup", {
   // other required props
  minimumInstances: stack.withStageDependentValue<number>({
    app,
    variableName: "minInstances",
    stageValues: { [Stage.CODE]: 1, [Stage.PROD]: 3 },
  }),
  maximumInstances: stack.withStageDependentValue<number>({
    app,
    variableName: "maxInstances",
    stageValues: { [Stage.CODE]: 2, [Stage.PROD]: 6 },
  }),
});
```

## How to test

See changes to unit tests.

Also see https://github.com/guardian/cdk-playground/pull/161 which tests this branch.

A simpler codebase!

Is the API being presented actually clearer?

## Checklist

  - ℹ️  NOTE: The `GuEc2App` could benefit from a refactor, namely documenting the props more explicitly. This is out of scope for this PR, however.

### DIFF
--- a/src/constructs/acm/certificate.ts
+++ b/src/constructs/acm/certificate.ts
@@ -2,16 +2,14 @@ import { Certificate, CertificateValidation } from "@aws-cdk/aws-certificatemana
 import type { CertificateProps } from "@aws-cdk/aws-certificatemanager/lib/certificate";
 import { HostedZone } from "@aws-cdk/aws-route53";
 import { RemovalPolicy } from "@aws-cdk/core";
-import { Stage } from "../../constants";
-import type { GuDomainNameProps } from "../../types/domain-names";
-import { StageAwareValue } from "../../types/stage";
+import type { GuDomainName } from "../../types/domain-names";
 import { GuStatefulMigratableConstruct } from "../../utils/mixin";
 import { GuAppAwareConstruct } from "../../utils/mixin/app-aware-construct";
 import type { GuStack } from "../core";
 import { AppIdentity } from "../core/identity";
 import type { GuMigratingResource } from "../core/migrating";
 
-export type GuCertificatePropsWithApp = GuDomainNameProps & AppIdentity & GuMigratingResource;
+export type GuCertificatePropsWithApp = GuDomainName & AppIdentity & GuMigratingResource;
 
 /**
  * Construct which creates a DNS-validated ACM Certificate.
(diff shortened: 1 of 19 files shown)


src/constructs/acm/certificate.ts
```diff
@@ -57,43 +55,17 @@ export type GuCertificatePropsWithApp = GuDomainNameProps & AppIdentity & GuMigr
...
-                  [Stage.CODE]: props.CODE.hostedZoneId!,
-                  [Stage.PROD]: props.PROD.hostedZoneId!,
-                },
-              })
-            : props.INFRA.hostedZoneId!
-          /* eslint-enable @typescript-eslint/no-non-null-assertion */
-        );
+    const maybeHostedZone = hostedZoneId
+      ? HostedZone.fromHostedZoneId(scope, AppIdentity.suffixText({ app }, "HostedZone"), hostedZoneId)
+      : undefined;
```
akash1810, line 62:
```ts
const maybeHostedZone = hostedZoneId && HostedZone.fromHostedZoneId(scope, AppIdentity.suffixText({ app }, "HostedZone"), hostedZoneId);
```

Would be preferred, however it's type is `"" | undefined | HostedZone` - I can't quite work out why the empty string is there 🤔 .

src/patterns/ec2-app.ts
```diff
@@ -170,16 +173,13 @@ export interface GuEc2AppProps extends AppIdentity {
   userData: GuUserDataProps | string;
   access: AppAccess;
   applicationPort: number;
-  certificateProps: GuDomainNameProps;
```
akash1810, line 173: Renaming `certificateProps` to `domain` is the only breaking change to the API of this pattern. If we didn't rename it, those that are only using the pattern have a no effort upgrade path. Shall we do that?

jacobwinch: I agree that the new name is better, but perhaps we should make this breaking change when we remove stage-awareness from the patterns to avoid changing the API twice in quick succession?

src/constructs/acm/certificate.ts
nicl: Ah it's because "" is falsy, and `hostedZoneId` itself is `string | undefined`. I can't think of a better way than ternary here either.

nicl (approved):
Looks nice and glad about the code removal.

src/patterns/ec2-app.ts
akash1810: Reverted.

src/types/asg.ts
```diff
@@ -0,0 +1,17 @@
...
+   * i.e. when there are no deployment or scaling events in progress.
+   */
+  minimumInstances: number;
+
+  /**
+   * The maximum number of EC2 instances.
+   * If omitted, this will be set to `minimumInstances * 2`.
+   * This allows us to support Riff-Raff's autoscaling deployment type by default.
+   *
+   * Should only be set if you need to scale beyond the default limit (e.g. due to heavy traffic),
```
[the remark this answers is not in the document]
akash1810: 👍🏽  Added.

src/patterns/ec2-app.ts
```diff
@@ -468,19 +519,47 @@ export class GuEc2App {
...
+        snsTopicName,
+        unhealthyInstancesAlarm,
+        actionsEnabledInCode = false,
+      } = monitoringConfiguration;
+
+      /*
+      Stage is known at runtime if it is a CFN Parameter, e.g. CODE or PROD.
+      Stage is known at compile time for singleton stacks, e.g. INFRA.
+       */
+      const isStageDeterminedAtRuntime = Token.isUnresolved(stage);
```
akash1810, line 533: A type guard could have been used here (a la https://github.com/guardian/cdk/pull/1048), however this is slightly simpler.

nicl (approved):
(👍 )

PR #1101, opened 2022-02-17
fix: ephemeralStorage must be > 20 when configuring

philmcmahon:
## What does this change?
The changes in https://github.com/guardian/cdk/pull/1075 breaks the ecs task construct because `EphemeralStorage size should be at least 21` according to cloudformation - this is suprising as the [default is apparently 20gb](https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_aws-ecs.TaskDefinition.html#ephemeralstoragegib)  I guess if you set it to anything other than the default then it expects you to actually increase the amount of storage from the default amount.

I'm not sure what the best thing to do here is - this PR currently sets the default to 21GB which I've tested solves the problem - but it could also probably be `undefined` so that the default 20gb value is used.

## How to test
I've verified that this solves the problem. 

Previously broken deploy: https://riffraff.gutools.co.uk/deployment/view/f0292e42-69f3-43c3-a2ee-08ce6cbe93ab
Fixed deploy: https://riffraff.gutools.co.uk/deployment/view/12e724bb-52af-4f3e-8fef-4b7b6088936e

### DIFF
--- a/src/constructs/ecs/ecs-task.ts
+++ b/src/constructs/ecs/ecs-task.ts
@@ -145,7 +145,7 @@ export class GuEcsTask {
       // see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-taskdefinition.html#cfn-ecs-taskdefinition-cpu for details
       cpu = 2048, // 2 cores and from 4-16GB memory
       memory = 4096, // 4GB
-      storage = 20, //20 GB
+      storage = 21, //20 GB
       containerConfiguration,
       taskCommand,
       taskTimeoutInMinutes = 15,
--- a/src/constructs/ecs/ecs-task.ts
+++ b/src/constructs/ecs/ecs-task.ts
@@ -145,7 +145,7 @@ export class GuEcsTask {
       // see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-taskdefinition.html#cfn-ecs-taskdefinition-cpu for details
       cpu = 2048, // 2 cores and from 4-16GB memory
       memory = 4096, // 4GB
-      storage = 21, //20 GB
+      storage, // default is 20 GB when not provided in props
       containerConfiguration,
       taskCommand,
       taskTimeoutInMinutes = 15,
@@ -155,6 +155,12 @@ export class GuEcsTask {
       securityGroups = [],
       environmentOverrides,
     } = props;
+
+    if (storage && storage < 21) {
+      throw new Error(
+        "Storage must be at least 21. See https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-taskdefinition-ephemeralstorage.html"
+      );
+    }
 
     const { stack, stage } = scope;
 


jacobwinch (approved):
This seems like a bug (and/or an error with the `cdk` docs to me), as the [CFN documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-taskdefinition-ephemeralstorage.html) is consistent with the error message that you're seeing.

👍

src/constructs/ecs/ecs-task.ts
```diff
@@ -145,7 +145,7 @@ export class GuEcsTask {
       // see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-taskdefinition.html#cfn-ecs-taskdefinition-cpu for details
       cpu = 2048, // 2 cores and from 4-16GB memory
       memory = 4096, // 4GB
-      storage = 20, //20 GB
+      storage = 21, //20 GB
```
akash1810, line 148: The comment is out of sync now.

philmcmahon: > I'd suggest _not_ defaulting `storage`, allowing AWS CDK to handle that. Instead, I'd validate `storage` if set and throw when < 21:

Great idea, have pushed a commit to do this, and tested on Lurch CODE

akash1810 (approved):

nicl (approved):
Weird bug. Thanks for the fix!

akash1810:
It doesn't look like there's any validation of `ephemeralStorageGiB ` being >=21 in the [AWS CDK class](https://github.com/aws/aws-cdk/blob/master/packages/%40aws-cdk/aws-ecs/lib/base/task-definition.ts). There doesn't seem to be an open PR for it either - want to raise one?

Interestingly https://github.com/aws/aws-cdk/pull/18106 sets defaults to `20`, which contradicts the CFN docs.

src/constructs/ecs/ecs-task.ts
akash1810: For completeness, probably want to throw when >200 too.

tapichu: My recommendation would be to leave the validation to aws-ecs' `TaskDefinition`:

https://github.com/aws/aws-cdk/blob/master/packages/%40aws-cdk/aws-ecs/lib/fargate/fargate-task-definition.ts#L143-L145

If those ranges change over time, this code will be out of sync. You should be able to unit test that passing invalid values like 20 or 201 already throws an error when creating the `TaskDefinition`.

tapichu:
> It doesn't look like there's any validation of `ephemeralStorageGiB ` being between 21 and 200 in the [AWS CDK class](https://github.com/aws/aws-cdk/blob/master/packages/%40aws-cdk/aws-ecs/lib/base/task-definition.ts). There doesn't seem to be an open PR for it either - want to raise one?
> 
> Interestingly [aws/aws-cdk#18106](https://github.com/aws/aws-cdk/pull/18106) sets defaults to `20`, which contradicts the CFN docs.

Here's the validation: https://github.com/aws/aws-cdk/blob/master/packages/%40aws-cdk/aws-ecs/lib/fargate/fargate-task-definition.ts#L143-L145

aws/aws-cdk#18106 does not set the default to 20, it leaves the value undefined which results in the task getting 20 GiB.

The confusion might come from the documentation. `aws-ecs` chose to document the 20 GiB that users get by default when `ephemeralStorageGiB` is not set, but it doesn't actually set its default value to 20, it leaves it undefined.

PR #1108, opened 2022-02-18
refactor: Organise EC2 pattern for clarity

akash1810:
## What does this change?

The [`ec2-app.ts`](https://github.com/guardian/cdk/blob/main/src/patterns/ec2-app.ts) file is pretty hard to reason about, partly due to its length (>500 lines).

In this change we:
  - Move the `GuPlayApp` and `GuNodeApp` classes into their own file
  - Remove the `GuApplicationPorts` enum as it doesn't provide much value
  - Update the integration test project to use a pattern, mainly to check the import path is still `import { GuPlayApp } from @guardian/cdk`. That is, this change is a no-op.

## How to test

CI should continue to pass.

Navigating through the repo is easier.

n/a

## Checklist


src/patterns/ec2-app/base.test.ts
```diff
@@ -31,7 +31,7 @@ describe("the GuEC2App pattern", function () {
   it("should produce a functional EC2 app with minimal arguments", function () {
     const stack = simpleGuStackForTesting();
     new GuEc2App(stack, {
-      applicationPort: GuApplicationPorts.Node,
+      applicationPort: 3000,
```
akash1810, line 34: I actually think this is more explicit than before!

src/patterns/ec2-app/base.test.ts
```diff
@@ -817,48 +817,4 @@ describe("the GuEC2App pattern", function () {
       ],
     });
   });
-
-  describe("GuNodeApp", () => {
```
akash1810, line 821: These get moved into their own dedicated test file.

jacobwinch (approved):
👍

src/patterns/ec2-app/base.test.ts
jacobwinch: 👍  - I like the `framework` naming too!

akash1810: > 👍 - I like the `framework` naming too!

Phew! I spent more time worrying about the clarity of this than I should have! Sure `Play` is a framework, but `Node`? That's a runtime, no? Couldn't think of anything better, and we might actually remove the classes later as they don't do much anyway 😂 .

PR #1110, opened 2022-02-21
refactor: Move `AppAccess` validation out of EC2 pattern file

akash1810:
## What does this change?

Part of an ongoing effort to reduce the LOC in `ec2-app` (see #1105 and #1108 for previous work).

This is a no-op in terms of functionality. Hopefully it's as clear, if not more clear, to understand as before.

This change moves some CIDR validation logic away into it's own file for increased readability and testing.

## How to test

CI should continue to pass.

A simpler codebase.

n/a

## Checklist

### DIFF
--- a/src/patterns/ec2-app/base.test.ts
+++ b/src/patterns/ec2-app/base.test.ts
@@ -389,7 +389,9 @@ describe("the GuEC2App pattern", function () {
           monitoringConfiguration: { noMonitoring: true },
           userData: "",
         })
-    ).toThrowError();
+    ).toThrowError(
+      "Restricted apps cannot be globally accessible. Adjust CIDR ranges (0.0.0.0/0, 1.2.3.4/32) or use Public."
+    );
   });
 
   it("errors if specifying public CIDR ranges with internal access scope", function () {
@@ -410,7 +412,9 @@ describe("the GuEC2App pattern", function () {
           monitoringConfiguration: { noMonitoring: true },
           userData: "",
         })
-    ).toThrowError();
+    ).toThrowError(
+      "Internal apps should only be accessible on 10. ranges. Adjust CIDR ranges (93.1.2.3/12) or use Restricted."
+    );
   });
 
   it("can configure ASG scaling by stage if desired", function () {
--- a/src/patterns/ec2-app/base.ts
+++ b/src/patterns/ec2-app/base.ts
@@ -25,7 +25,7 @@ import {
   GuApplicationTargetGroup,
   GuHttpsApplicationListener,
 } from "../../constructs/loadbalancing";
-import type { AppAccess, InternalAccess, RestrictedAccess } from "../../types/access";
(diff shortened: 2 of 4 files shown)


src/patterns/ec2-app/base.test.ts
```diff
@@ -389,7 +389,9 @@ describe("the GuEC2App pattern", function () {
           monitoringConfiguration: { noMonitoring: true },
           userData: "",
         })
-    ).toThrowError();
```
akash1810, line 392: Should always test the message being thrown, else something else could have been throwing!

src/patterns/ec2-app/base.ts
```diff
@@ -128,25 +128,6 @@ export interface GuEc2AppProps extends AppIdentity {
...
-  Please either remove the public CIDR range or use the RESTRICTED access scope.`);
-
-function validateRestrictedCidrRanges(access: RestrictedAccess) {
-  const openToWorld = access.cidrRanges.map((range) => range.uniqueId).includes("0.0.0.0/0");
-  if (openToWorld) throw OpenApplicationCidrError;
-}
-
-function validateInternalCidrRanges(access: InternalAccess) {
-  const publicCidrRanges = access.cidrRanges.filter((range) => !range.uniqueId.startsWith("10."));
-  const containsPublicIps = publicCidrRanges.length > 0;
```
akash1810, line 146: re-written as [`some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) for brevity

src/types/access.ts
```diff
@@ -56,3 +57,44 @@ export interface InternalAccess extends Access {
...
+   *
+   * @throws `Error` when CIDR rules are invalid
+   */
+  validate: (access: AppAccess) => {
+    switch (access.scope) {
+      case AccessScope.RESTRICTED: {
+        const ranges = access.cidrRanges.map((_) => _.uniqueId);
+        const rangesString = ranges.join(", ");
+
+        if (ranges.includes(Peer.anyIpv4().uniqueId) || ranges.includes(Peer.anyIpv6().uniqueId)) {
```
jacobwinch, line 77: I think this PR is also adding IPv6 support to this check? 💯

jacobwinch (approved):

src/types/access.ts
akash1810: Oh. Yes. I guess this is _not_ a no-op after all then! Happy to remove IPv6 stuff or update the PR as `fix:` to create a release.

jacobwinch: I think it's fine to include it here, I don't think many people will be specifying IPv6 ranges yet! Feel free to update to `fix` if you want to generate a release though!

akash1810: A new release for this is most likely to just be noise. Will leave as is.

PR #1114, opened 2022-02-22
test: Simplify integration project

akash1810:
## What does this change?

#1113 is failing CI for two reasons:
  - AWS CDK library versions are too old (#1111 is raised to fix this)
  - The snapshot test in the integration project does not mock the library version tag, so with a new release, the value in the snapshot is out of date

This PR addresses the latter by simplifying the integration test. It reverts to the simple version that just added a parameter to a stack, rather than an entire pattern.

This should unblock #1113.

Note:
  - The base branch for this PR is #1113 rather than `main` as I don't quite know how sem-rel will react otherwise
  - We'd need to relax the `check-aws-cdk` rule for this and #1113 to get a green build.


jacobwinch:
👍

PR #1122, opened 2022-02-23
feat: Refactor account-readiness

nicl:
This got a bit bigger than I planned(!) but basically, it does:

* remove /account/vpc/default/* check as no longer required
* move default-VPC-reference-in-parameters check to SSM readiness check
* remove 3-params-per-VPC check as noisy
* remove default-VPC-in-use check as noisy

By 'noisy', I mean that, while useful, these are not actually blockers to getting started with CDK and most accounts fail on one or both. For example, `deployTools` fails on these even though we are happily using CDK to deploy our core apps.

I've also introduced a 'Report' type to make standard reporting a bit easier.

SSM readiness output has been modified slightly as a result, but not in a major way.

e.g. before:

Screenshot 2022-02-23 at 17 24 29

after:

Screenshot 2022-02-23 at 17 24 55

We've obviously lost some VPC-related info here. I think probably it's  best  to document this in  patterns, with notes there on how to use  a non-primary VPC/subnets, but please do shout if you're keen to keep this info.


src/bin/commands/account-readiness/index.ts
```diff
@@ -1,13 +1,28 @@
...
+  name: string;
+  isPass: boolean;
+  msg?: string;
+  errors: Map<string, string>;
+}
 
 export const accountReadinessCommand = async (props: AwsAccountReadiness): CliCommandResponse => {
   // Got a new AWS account readiness command? Add it to this list and ✨
-  const commandResponses: number[] = await Promise.all([ssmParamReadiness(props), vpcReadiness(props)]);
+  const reports = await Promise.all([reportSSM(props)]);
```
jacobwinch, line 14: If there are no immediate plans to add more reports here, then we could simplify the code a bit by working with a single report rather than an array?

src/bin/commands/account-readiness/ssm.ts
```diff
@@ -1,55 +1,96 @@
...
+export const report = async (props: AwsAccountReadiness): Promise<Report> => {
+  const output = await ssmParamReadiness(props);
+  const isPass = output.missingParameters.length === 0 && output.defaultVPCReferences.length === 0;
 
-    notFoundInDetail.forEach(({ path, description }) => {
-      console.log(` ❌ ${chalk.yellowBright(path)} ${chalk.dim(`(${description})`)}`);
-    });
+  const errs = new Map<string, string>();
+  const parametersMsg =
+    "CDK uses predefined SSM parameters within patterns and constructs. For example, the EC2 APP pattern uses a parameter to determine which Kinesis stream to forward logs to. Your CDK app may fail at runtime if these parameters do not exist.";
```
jacobwinch, line 70: This is a really helpful message 👍

src/constants/ssm-parameter-paths.ts
```diff
@@ -29,7 +29,8 @@ export const SSM_PARAMETER_PATHS: Record<string, SsmParameterPath> = {
   },
   PrimaryVpcId: {
     path: `${VPC_SSM_PARAMETER_PREFIX}/primary/id`,
-    description: "Virtual Private Cloud to run EC2 instances within",
+    description:
+      "Virtual Private Cloud to run EC2 instances within. Should NOT be the account default VPC. Declare additional VPCs under e.g./secondary/id and /tertiary/id and similar for subnet-related paths.",
```
jacobwinch, line 33: I'm not opposed to this suggestion, but I don't think it's always what we've done in accounts with more than one VPC so far. My memory is a little hazy, but I seem to recall that some accounts have two VPCs which are roughly equivalent in terms of importance (e.g. a `subscriptions` and `membership` VPC) and in this context it didn't feel appropriate to call one primary and the other secondary.

nicl: That's helpful thanks. I'll remove that bit for now.

src/bin/commands/account-readiness/ssm.ts
```diff
@@ -1,55 +1,96 @@
...
-    return 0;
-  } else {
-    console.log(
-      chalk.red(
-        `\nAWS account requires the above ${notFoundInDetail.length} SSM parameters to work with @guardian/cdk v${LibraryInfo.VERSION}`
-      )
+  if (output.defaultVPCReferences.length > 0) {
+    errs.set(
+      "Default VPC referenced in VPC parameters",
+      `it is recommended to use a custom VPC instead. Found in paths: ${output.defaultVPCReferences.join(", ")}.`
```
akash1810, line 82: Should we be more opinionated here? Rather than recommend the use of a custom VPC, require it?

src/bin/commands/account-readiness/ssm.ts
```diff
@@ -1,55 +1,96 @@
...
 
-  if (notFoundInDetail.length === 0) {
-    // no action needed
-    return 0;
-  } else {
-    console.log(
-      chalk.red(
-        `\nAWS account requires the above ${notFoundInDetail.length} SSM parameters to work with @guardian/cdk v${LibraryInfo.VERSION}`
-      )
+  if (output.defaultVPCReferences.length > 0) {
```
jacobwinch, line 79: This check is a neat way of enforcing best practice without creating a noisy output 💯

jacobwinch (approved):

src/bin/commands/account-readiness/ssm.ts
akash1810: Ah, looking more closely, the rest of this change does move to requiring a custom VPC. Maybe we can change "it is recommended to" to "you should"?

PR #1123, opened 2022-02-23
docs: Add ADR for stage awareness

akash1810:
## What does this change?

Converting [our discussions](https://docs.google.com/document/d/1YCuOYsBVdO1eB80t6j0ElTCtoRc3doqmH-v4v6sC9B4/edit#) on stage awareness into an ADR for prosperity.

I've found this fairly hard to write, please don't hold back on the review!

## How to test

View it [here](https://github.com/guardian/cdk/blob/aa-stage-adr/docs/architecture-decision-records/006-stage-awareness.md).

Does it read well? Is it clear?

### DIFF
--- a/docs/architecture-decision-records/006-stage-awareness.md
+++ b/docs/architecture-decision-records/006-stage-awareness.md
@@ -1,0 +1,238 @@
+# Stage Awareness
+
+## Status
+Proposed.
+
+## Context
+It is best practise to use to produce a single artifact for use on all stages of an application.
+This provides confidence that, should the artifact be deployed to PROD, it'll behave the same as it did when tested on CODE.
+
+It is also best practise to employ continuous delivery of infrastructure. We use Riff-Raff.
+
+It is common to vary infrastructure based on the stage. For example:
+  - An ASG in PROD typically has a higher capacity than CODE as it receives more traffic
+    - PROD typically also uses more powerful instance types for the same reason
+  - The domain name for PROD will be different from CODE
+
+Let's call these "stage aware values".
+
+How should `@guardian/cdk` (GuCDK) support stage aware values?
+
+## Status Quo
+Within YAML CloudFormation templates, we achieve stage awareness using:
+  - A `Stage` [parameter][cfn-param]
+  - A [`Mapping`][cfn-mapping]
+  - The [`FindInMap`][cfn-find] intrinsic function
+
+We have a single YAML template artifact for use in the CODE and PROD stage, and it contains fragments such as:
+
+```yaml
+Parameters:
+  Stage:
+    Type: String
+    Default: CODE
+    AllowedValues:
+      - CODE
+      - PROD
+    Description: Stage name
+Mappings:
+  myapp:
+    CODE:
+      minInstances: 1
+      maxInstances: 2
+    PROD:
+      minInstances: 3
+      maxInstances: 6
+Resources:
+  AutoScalingGroupMyApp:
+  Type: AWS::AutoScaling::AutoScalingGroup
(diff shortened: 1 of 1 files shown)


nicl (commented):
I think there are two ways to go here:

1) don't attempt to list out the various alternatives, just the problem and the proposed solution
2) do as here and fully explain things

If (2), then I would consider including the status quo in the solutions, and perhaps breaking things down as follows:

a) a single-stack across stages (the status quo)

Using intrinsic functions and stage-specific parameters. This allows us to use CDK in a similar way to our existing Cloudformation templates but also the cons you note (discouraged by AWS, mappings limit types to primitives, etc.).

And then approaches - status quo, and also the alternative of requiring users to handle mappings at the top-level.

b) stack-per-stage

List the benefits - simpler constructs, richer types, keep stage-changes together at top-level so easier to  understand, ...-   but also cons.

And then discussion of approachese to doing this - e.g. simple RR change and also CDK deploy (which is a larger change).

akash1810:
@nicl how does [this edit](https://github.com/guardian/cdk/blob/aa-stage-adr/docs/architecture-decision-records/006-stage-awareness.md) look?

docs/architecture-decision-records/006-stage-awareness.md
```diff
@@ -0,0 +1,237 @@
...
+  - Riff-Raff supports the deployment of a single CloudFormation template file
+- :white_check_mark: Support stacks that do not use `CODE` or `PROD`
+  - Users would be able to specify their own stages
+- :x: The simplicity of the API presented to users
+  - Users would have to create a Mapping themselves each time
+    - Creating [mappings within CDK][cdk-mapping] is quite verbose
+    - Goes against some aims of GuCDK which is to simplify infrastructure, making it easy to follow best practices
+  - Stronger, more type safe types such as [`InstanceType`][cdk-instance-type], will not be supported
+- :x: How short the feedback loop is
+  - Likely to be no faster than hand-writing a YAML template
```
nicl, line 132: I'm not sure it's quite as bad as this suggests. A bit of work to create mappings yes, but with helpers it is pretty straightforward.

docs/architecture-decision-records/006-stage-awareness.md
```diff
@@ -0,0 +1,237 @@
...
+stacks:
+  - playground
+regions:
+  - eu-west-1
+deployments:
+  cloudformation:
+    type: cloud-formation
+    app: cdk-playground
+    parameters:
+      templateStagePath: # <- NEW!
```
nicl, line 207: Might be worth citing the existing stage-specific parameter that already exists here. I.e. this kind of approach is not entirely new.

nicl (approved):
This is great. Really clear breakdown of the pros/cons of each and provides helpful context in terms of the status quo too.

docs/architecture-decision-records/006-stage-awareness.md
akash1810: I think we're suggesting GuCDK does not provide any helpers and user's would have to bring their own solution?

jfsoul:
Great work 👏 

Is there another location we could additionally summarise this decision for users of GuCDK at-a-glance?

akash1810:
> Great work 👏
> 
> Is there another location we could additionally summarise this decision for users of GuCDK at-a-glance?

Yes! I think we should add this to a "CDK best practice" style doc, which'll include use of SSM, VPC setup etc. A bit of work on Riff-Raff is needed before this proposal/ADR moves to "accepted", so I'd be tempted to wait until that's done.

Issue #1032, opened 2022-01-07
Do we really need GuSSMParameter?


nicl:
As part of some JSII testing, I was looking at the GuSSMParameter and also the original PR:

https://github.com/guardian/cdk/pull/336

GuSSMParameter is a custom resource that allows using SSM parameters *within* Cloudformation templates. To recap the motivations given at the time:

* support defaults
* not require manual/console changes
* explicitness over what is a secret
* reduce cfn parameters as more parameters is 'unwieldy'

However, I can't see any usage of GuSSMParameter in our codebase.

Is GuSSMParameter really an improvement over [SSM parameters](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html) or should we just delete it?

To take some of the motivations above:

* ssm paths can be changed in CDK of course so no need for console changes (other than in Parameter Store)
* cfn parameters support `NoEcho` to indicate sensitive data

Defaults are tricky, but it's not obvious why data that needs to be in Parameter Store would have a sensible default value *within @guardian/cdk itself*; typically parameter store is used for application-specific toggles or secrets. Indeed, the lack of usage seems to confirm this.

And, the 'con' of increased Cfn parameters seems to me to actually be a win. It is much clearer what the inputs to a stack are if they are actual parameters. The stack then becomes closer to a function conceptually, rather than magically relying on all kinds of global state deep in the stack body (which is what this custom resource seems to achieve).

So, to summarise, it isn't clear that this feature meets the problems described any better than native SSM Cfn parameters and, in some respects, is actually worse.

@guardian/cdk is already pretty large. Let's delete this complex and low-value feature.

nicl:
Update: okay we've agreed to remove this. In general, turns out there has been a bit of a reversal on custom resources anyway due to the random lambdas they create in accounts.

akash1810:
Fixed in #1157 .

PR #1134, opened 2022-03-02
feat!: Remove `Stage` CFN Parameter in favour of property

akash1810:
The diff on this PR is pretty big! Sadly it's unavoidable as we're removing a property from a class that's used throughout the repository.

It might be easier to review commit by commit? I plan to squash the commits at merge time to keep the release notes clean.

## What does this change?

Following https://github.com/guardian/cdk/pull/1123, this change removes all stage awareness from GuCDK in favour of producing a template per stage. This means we no longer produce [Mappings](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/mappings-section-structure.html) in the template, and users can make _any_ property stage aware.

This change works in conjunction with https://github.com/guardian/riff-raff/pull/681.

BREAKING CHANGE: Removal of explicit stage awareness in favour of users defined properties.

From:

```ts
// -- lib/cdk-playground.ts
export class CdkPlayground extends GuStack {
  private static app = "cdk-playground";

    super(scope, id, props);

    new GuPlayApp(this, {
      app: CdkPlayground.app,
      instanceType: InstanceType.of(InstanceClass.T4G, InstanceSize.MICRO),
      access: { scope: AccessScope.PUBLIC },
      userData: {
        distributable: {
          fileName: `${CdkPlayground.app}.deb`,
          executionStatement: `dpkg -i /${CdkPlayground.app}/${CdkPlayground.app}.deb`,
        },
      },
      certificateProps: {
        [Stage.CODE]: {
          domainName: "cdk-playground.code.dev-gutools.co.uk",
        },
        [Stage.PROD]: {
          domainName: "cdk-playground.prod.gutools.co.uk",
        },
      },
      scaling: {
        [Stage.CODE]: {
          minimumInstances: 1,
        },
        [Stage.PROD]: {
          minimumInstances: 3,
        },
      },
    });
  }
}

// -- bin/cdk.ts
const app = new App();

// produces one output file `cdk.out/CdkPlayground.template.json` with `Mappings` and `FindInMap` intrinsic functions
new CdkPlayground(app, "CdkPlayground", { stack: "deploy" });
```

To:

```ts
// -- lib/cdk-playground.ts
export interface CdkPlaygroundProps {
  stage: string;
  domainName: string;
  scaling: GuAsgCapacity;
}

export class CdkPlayground extends GuStack {
  private static app = "cdk-playground";

  constructor(scope: App, id: string, props: CdkPlaygroundProps) {
    const { stage, domainName, scaling } = props;

    super(scope, id, {
      stack: "deploy",
      stage,
    });

    new GuPlayApp(this, {
      app: CdkPlayground.app,
      instanceType: InstanceType.of(InstanceClass.T4G, InstanceSize.MICRO),
      access: { scope: AccessScope.PUBLIC },
      userData: {
        distributable: {
          fileName: `${CdkPlayground.app}.deb`,
          executionStatement: `dpkg -i /${CdkPlayground.app}/${CdkPlayground.app}.deb`,
        },
      },
      certificateProps: {
        domainName,
      },
      scaling,
    });
  }
}

// -- bin/cdk.ts
const app = new App();

// produces output file `cdk.out/CdkPlayground-CODE.template.json`
new CdkPlayground(app, "CdkPlayground-CODE", {
  stage: "CODE",
  domainName: "cdk-playground.code.dev-gutools.co.uk",
  scaling: { minimumInstances: 1 },
});

// produces output file `cdk.out/CdkPlayground-PROD.template.json`
new CdkPlayground(app, "CdkPlayground-PROD", {
  stage: "PROD",
  domainName: "cdk-playground.prod.gutools.co.uk",
  scaling: { minimumInstances: 3 },
});
```

## How to test

- Install GuCDK from this branch into a project
- Update the stack definition of the project
- Update the `riff-raff.yaml` of the project, adding [`templateStagePaths`](https://github.com/guardian/riff-raff/pull/681)
- Deploy the project, observe the correct template file being used 

Indeed this has been done on cdk-playground [here](https://github.com/guardian/cdk-playground/pull/173).

- Any property of a stack can be made stage aware at the user's choosing
- A much simplified code base

This is quite the breaking change! The release notes should be simple and easy to follow!

❓ The migration guides will be out of date after this change. Can we update them _after_ this has gone out and we've migrated one of our YAML defined stacks?

## Checklist


src/constructs/core/parameters/s3.test.ts
```diff
@@ -7,10 +7,10 @@ describe("GuDistributionBucketParameter", () => {
   it("can exist in multiple GuStacks within the same App", () => {
```
akash1810, line 7: "App" here is a CDK [`App`](https://docs.aws.amazon.com/cdk/v2/guide/apps.html#apps_construct) (aka the root of the tree), not a Guardian app (aka a string identifier).

src/constructs/core/stack.ts
```diff
@@ -179,48 +153,12 @@ export class GuStack extends Stack implements StackStageIdentity, GuMigratingSta
   }
 }
 
-/**
- * A GuStack but designed for infrastructure as `Stage` will always be `INFRA`.
- */
-export class GuStackForInfrastructure extends GuStack {
```
akash1810, line 185: This just goes away entirely in favour of `new GuStack({stage: "INFRA"})`.

src/constructs/loadbalancing/elb.test.ts
```diff
@@ -182,7 +182,7 @@ describe("The GuHttpsClassicLoadBalancer class", () => {
 
     const json = SynthUtils.toCloudFormation(stack) as SynthedStack;
 
-    expect(Object.keys(json.Parameters)).not.toContain("CertificateARN");
+    expect(json.Parameters).toBeUndefined();
```
akash1810, line 185: Previously, Parameters would be of length 1 with "Stage". The "Stage" parameter has now gone, so Parameters are undefined instead at this point now.

src/utils/test/simple-gu-stack.ts
```diff
@@ -1,21 +1,11 @@
...
-
-export const simpleGuStackForTesting: (props?: SimpleGuStackProps) => GuStack = (props?: SimpleGuStackProps) =>
-  new GuStack(new App(), "Test", {
-    stack: props?.stack ?? "test-stack",
-    ...props,
-  });
-
-export const simpleInfraStackForTesting: (props?: SimpleGuStackProps) => GuStack = (props?: SimpleGuStackProps) =>
-  new GuStackForInfrastructure(new App(), "Test", {
+export function simpleGuStackForTesting(props?: Partial<GuStackProps>) {
```
akash1810, line 5: Switching the a `function` rather than a `const function` as the type annotations are simpler.

src/constructs/core/parameters/s3.test.ts
jacobwinch: Thanks for disambiguating; perhaps we should update the test name to `"can exist in multiple GuStacks within the same CDK App"`?

src/constructs/core/ssm.test.ts
```diff
@@ -18,9 +18,7 @@ describe("SSM:", () => {
       const stack = simpleGuStackForTesting({ stack: "some-stack" });
```
jacobwinch, line 18: Minor: I don't quite understand the purpose of these first two tests. The assertions are confusing to me. But based on the naming (`requires the scope and parameter name`) it sounds like the compiler is doing the work here and unit tests are superfluous?

jacobwinch:
A minor point on the release notes... I would probably not include differing instance type by stage in the example. Because: a) I think it adds complexity (users were not able to do this before, so they don't need to do it as part of an upgrade) and b) I don't think we should encourage users to vary this property based on stage!

src/constructs/loadbalancing/elb.test.ts
jacobwinch: This is one for a different PR, but I wonder if we should revisit this construct to remove some of this complexity. I'm struggling to think of a scenario where we _want_ users to define the certificate in a different stack to the load balancer and pass the certificate ARN in via a parameter. I think this is a convention that a lot of people used to follow, rather than something that we want to encourage going forwards.

src/patterns/ec2-app/base.ts
```diff
@@ -159,14 +136,7 @@ function restrictedCidrRanges(ranges: IPeer[]) {
...
  *   certificateProps:{
- *     [Stage.CODE]: {
- *       domainName: "code-guardian.com",
- *       hostedZoneId: "id123",
- *     },
- *     [Stage.PROD]: {
- *       domainName: "prod-guardian.com",
- *       hostedZoneId: "id124",
- *     },
+ *     domainName: "prod-guardian.com",
```
jacobwinch, line 139: As this section of the Typedocs will be the entrypoint for many users, I wonder if we should demonstrate how to instantiate this pattern in a stage-aware fashion. I think the usage example from the release notes (and PR description) is extremely helpful - so it would be great to find a way to surface it for all users, rather than just users who are upgrading.

Although it's not technically part of the pattern anymore, it's still probably the most typical use-case and it would be good to give some guidance on the type of properties that we expect users to vary based on stage (since this is no longer encoded).

Alternatively we could keep it simple here and hardcode a string, but link to a more general 'Here's how to set up a stage aware pattern' that we can point to from all of the different patterns?

src/constructs/loadbalancing/elb.test.ts
akash1810: Agreed!

src/constructs/core/ssm.test.ts
akash1810: Agreed. I think addressing https://github.com/guardian/cdk/issues/1032 would remove these tests?

src/constructs/core/parameters/s3.test.ts
nicl: Yeah, I wonder if we should think about better ways to disambiguate 'app' and 'stack' in CDK somehow, as the terms are overloaded. (Not as part of this PR though.)

src/constructs/core/ssm.test.ts
nicl: Ah yep, let me raise a PR to remove that!

nicl (approved):
This is great! Removes a lot of code, and lifting stage-specific values to the top-level stack props really improves clarity over what is actually different across stages (whereas previously things were buried throughout the code).

Jacob has left loads of great comments that I don't have much to add too - agree especially on finding ways to document how to handle stages - though doesn't have to be part of this PR necessarily.

I think there might be a few more stage-mapping references in some of the doc strings for constructs - e.g. like here https://github.com/guardian/cdk/blob/aa-rm-stage/src/constructs/cloudwatch/alarm.ts#L26.

src/constructs/core/ssm.test.ts
nicl: cc @akash1810 : https://github.com/guardian/cdk/pull/1157 PR now raised :)

src/constructs/core/parameters/s3.test.ts
akash1810: Will leave this for a future PR (not least because this will change again when we upgrade to CDK v2).

src/patterns/ec2-app/base.ts
akash1810: Added.

src/constructs/cloudwatch/alarm.ts
```diff
@@ -15,18 +15,11 @@ export interface GuAlarmProps extends AlarmProps, AppIdentity {
...
  *   // other required props
- *   actionsEnabled: stack.withStageDependentValue({
- *     app: "my-app",
- *     variableName: "alarmActionsEnabled",
- *     stageValues: {
- *       [Stage.CODE]: false,
- *       [Stage.PROD]: true,
- *     },
- *   }),
+ *   actionsEnabled: this.stage !== "PROD",
```
jacobwinch, line 22: I know it's only example code, but I think we want `===` here?

src/patterns/ec2-app/base.ts
```diff
@@ -220,6 +181,74 @@ function restrictedCidrRanges(ranges: IPeer[]) {
...
+ *   // other required properties
+ *   scaling: this.stage === "PROD" ? { minimumInstances: 6 } : { minimumInstances: 1 }
+ * });
+ * ```
+ *
+ * If multiple properties need to be stage aware, consider encoding this into the `props` of the calling class:
+ *
+ * ```typescript
+ * export interface MyAppProps extends GuStackProps {
+ *   stage: string;
```
jacobwinch, line 201: Minor: I don't think we need `stage` in this example?

jacobwinch (approved):
Great work 💯

docs/migration-guide.md
```diff
@@ -22,10 +22,9 @@ Then follow the instructions for any errors.
 Firstly, ensure you are running a recent version of Node and then initialise a
 new CDK app from within your repo:
 
-    npx @guardian/cdk new --app [app] --stack [stack] --yaml-template-location cfn.yaml
+    npx @guardian/cdk@latest new --app [app] --stack [stack] --stage [stage] --yaml-template-location /path/to/cfn.yaml
```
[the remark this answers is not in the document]
akash1810: Tempted to leave this as is so that it matches the other parameters. Maybe we can provide an example?

README.md
```diff
@@ -29,18 +29,15 @@ Jump to:
 example, VPC IDs, and the location of dist buckets. To check for account
 readiness and fix any issues, run:
 
-    npx @guardian/cdk account-readiness --profile [profile]
+    npx @guardian/cdk@latest account-readiness --profile [profile]
 
 Then, instantiate a new CDK app:
 
-    npx @guardian/cdk new --app [app] --stack [stack]
+    npx @guardian/cdk@latest new --app [app] --stack [stack] --stage [stage]
```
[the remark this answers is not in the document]
akash1810: > I was a little surprised to find that passing `--stage` more than once was the recommended way to configure multiple stages.

This is `yargs` magic.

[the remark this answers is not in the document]
akash1810: Added an example to further demonstrate usage.

PR #1116, opened 2022-02-23
feat: add --fix flag to account-readiness

nicl:
## What does this change?

This flag outputs a simple Cloudformation template that can be used to bootstrap an account for CDK. It also prompts for parameter values as part of this.

This also provides a starting point for any fuller account bootstrap process we might introduce in the future.

Note, I plan on doing a follow-up PR to alter the output of normal account-readiness for failure cases to point people towards the `--fix` flag.

## How to test

    npm run cli:dev account-readiness -- --profile deployTools --fix

## How can we measure success?

If it gets used! (We will need feedback on this.)

## Example stack:

```
Description: >-
  CDK Bootstrap template used to manage required SSM parameters and other
  resources.
Parameters:
  Stage:
    Type: String
    Default: CODE
    AllowedValues:
      - CODE
      - PROD
    Description: Stage name
Resources:
  accountservicesanghammaradtopicarnEBB061EC:
    Type: AWS::SSM::Parameter
    Properties:
      Type: String
      Value: sdf
      Description: SSM parameter containing the ARN of the Anghammarad SNS topic
      Name: /account/services/anghammarad.topic.arn
      Tags:
        gu:cdk:version: 35.0.1
        gu:repo: guardian/cdk
        Stack: sdflkj
        Stage:
          Ref: Stage
  accountservicesloggingstreamname33701367:
    Type: AWS::SSM::Parameter
    Properties:
      Type: String
      Value: sdflkj
      Description: SSM parameter containing the Name (not ARN) on the kinesis stream
      Name: /account/services/logging.stream.name
      Tags:
        gu:cdk:version: 35.0.1
        gu:repo: guardian/cdk
        Stack: sdflkj
        Stage:
          Ref: Stage
  accountservicesartifactbucket2583022C:
    Type: AWS::SSM::Parameter
    Properties:
      Type: String
      Value: sdlkj
      Description: >-
        SSM parameter containing the S3 bucket name holding distribution
        artifacts
      Name: /account/services/artifact.bucket
      Tags:
        gu:cdk:version: 35.0.1
        gu:repo: guardian/cdk
        Stack: sdflkj
        Stage:
          Ref: Stage
  accountservicesaccessloggingbucket03D60DEC:
    Type: AWS::SSM::Parameter
    Properties:
      Type: String
      Value: sdflkj
      Description: S3 bucket to store your access logs.
      Name: /account/services/access-logging/bucket
      Tags:
        gu:cdk:version: 35.0.1
        gu:repo: guardian/cdk
[...]

PR #1135, opened 2022-03-02
feat: add bootstrap command

nicl:
## What does this change?

(Replaces https://github.com/guardian/cdk/pull/1116.)

Adds a `bootstrap` command:

    @npx @guardian/cdk bootstrap --profile [profile] --dryRun [true|false]

which steps through the creation of a `cdk-bootstrap-INFRA` stack in the target account.

e.g. output like:

Kapture 2022-03-03 at 11 03 38

`account-readiness` will recommend this command in cases where the target account has none of the expected parameters present:

Screenshot 2022-03-02 at 10 47 14

## Improvements

It would be nice to suggest possible parameter values. E.g. vpc IDs could be looked up and presented as choices (Inquirer provides in-built UX for this kind of thing). And the same for subnets. Keen to do this separately though as may not be necessary and also increases PR size.

## Should these parameters be defined elsewhere?

Possibly. At the moment teams seem to manually create the CDK required parameters, which we are keen to avoid. Parameters could be defined in stacks along with associated resources - e.g. the kinesis parameter defined alongside the kinesis stream itself. However, there are some issues with this:

* there may not be cfn stacks for the associated resources
* we are keen on reducing friction to get started with CDK, and this approach would potentially require changes to multiple stacks depending on how accounts have been set up

Ultimately, the `bootstrap` script is simple and probably not right for everyone, but is:

* quick+easy
* ensures new teams have CDK bootstrap resources cloudformed
* mirrors the AWS CDK bootstrap command (which may help understanding here, just as we also mirror the `new` command)

## How to test

Run with the `dryRun` flag against a target account.

Feedback on usage and also presence of the bootstrap stack in Guardian AWS accounts.

At the moment the bootstrap stack will fail if run when some of the SSM parameters already exist in the account. Unfortunately, there doesn't seem to be an easy way to migrate existing parameters.

## Checklist


src/bin/index.ts
```diff
@@ -6,14 +6,16 @@ import type { CliCommandResponse } from "../types/cli";
...
+import { bootstrapCommand } from "./commands/bootstrap";
 import { checkPackageJson } from "./commands/check-package-json";
+import { newCdk } from "./commands/newCdk";
 
 const Commands = {
   AwsCdkVersion: "aws-cdk-version",
   AccountReadiness: "account-readiness",
   CheckPackageJson: "check-package-json",
   New: "new",
+  Bootstrap: "bootstrap",
```
jacobwinch, line 18: Minor: I would probably name this `bootstrap-account` or `prepare-account`, just to avoid people confusing it with the `new` command, which also uses 'bootstrap' terminology: https://github.com/guardian/cdk/blob/129aa9e37c6b8b46855b5b5f775368d476052dec/src/bin/commands/bootstrap.ts#L17-L23

src/bin/commands/bootstrap/bootstrap-stack.ts
```diff
@@ -0,0 +1,115 @@
...
+        parameterName: p.path,
+        stringValue: p.value,
+      });
+    });
+  }
+}
+
+export const accountBootstrapCfn = async (parameters: SsmParameterPath[]): Promise<string> => {
+  const stack = await ask(
+    "Enter a Guardian stack, used to tag resources (e.g. 'frontend' - typically the same as your Janus profile)\n >> "
```
akash1810, line 42: More argument to rename "stack" as account name 😄

src/bin/commands/bootstrap/bootstrap-stack.ts
```diff
@@ -0,0 +1,115 @@
...
+      if (res.$response.error) {
+        return `Stack creation failed: ${res.$response.error.message}`;
+      }
+
+      return res.StackId ?? "STACK ID UNDEFINED";
+    });
+};
+
+const ask = async (question: string): Promise<string> => {
+  const rl = createInterface(stdin, stdout);
```
akash1810, line 93: We already [depend](https://github.com/guardian/cdk/blob/129aa9e37c6b8b46855b5b5f775368d476052dec/package.json#L76) on [`cli-ux`](https://github.com/oclif/cli-ux). Ignoring that it's deprecated, should we use one or the other?

akash1810: [`yargs`](https://github.com/yargs/yargs) also powers the main CLI, that likely has support for accepting input.

src/bin/index.ts
nicl: Ah yep, my strategy was to rename the bootstrap terminology within the new command but this obviously got missed! Will update this too. (to `NewCdkProps`).

src/bin/commands/bootstrap/bootstrap-stack.ts
```diff
@@ -0,0 +1,115 @@
...
+  });
+
+  const input: CreateStackInput = {
+    StackName: "cdk-bootstrap-INFRA",
+    TemplateBody: template,
+    EnableTerminationProtection: true,
+  };
+
+  return cfn
+    .createStack(input)
```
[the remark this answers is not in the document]
nicl: It's a good point. I'll take a look at how this might work.

src/bin/commands/bootstrap/bootstrap-stack.ts
nicl: It doesn't look like ordinary `yargs` supports this, but there's https://www.npmjs.com/package/yargs-interactive which looks really interesting so will try that.

nicl: Update: in the end I used https://www.npmjs.com/package/inquirer directly, as it involved fewer changes to the main command definitions.

src/bin/commands/bootstrap/bootstrap-stack.ts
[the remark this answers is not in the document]
nicl: @akash1810 I can't see a way to do this officially - though there is interesting discussion of workarounds [here](https://github.com/aws/aws-cdk-rfcs/issues/300). I've tidied up the template extraction (to get by name rather than array positional lookup) but not sure there's a simpler solution at present unfortunately.

src/bin/commands/account-readiness/index.ts
```diff
@@ -7,13 +7,25 @@ export interface Report {
...
   const report = await reportSSM(props);
 
   console.log(chalk.bold(report.name + ":") + " " + (report.isPass ? "✅ Pass" : "❌ Fail"));
 
+  if (report.parametersFound == 0) {
+    console.log(`It looks like this account has not been set up for @guardian/cdk yet. Run:
+
+    @npx @guardian/cdk bootstrap
+
+to bootstrap the account.`);
```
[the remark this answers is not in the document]
akash1810: I don't think `@npx` is valid, should it be `npx`?

src/bin/commands/bootstrap/bootstrap-stack.ts
```diff
@@ -0,0 +1,115 @@
...
+  // Inquirer interprets '.' in question names as a special character, so let's
+  // replace it.
+  const safeName = (name: string): string => name.replaceAll(".", "-");
+
+  const stackQ: Question = {
+    type: "string",
+    name: "stack",
+    message:
+      "Enter a Guardian stack, used to tag resources (e.g. 'frontend' - typically the same as your Janus profile)",
+  };
```
akash1810, line 50: This isn't always true (e.g. profile=deployTools, stack=deploy) and might actually be thought of as redundant as we're already required to pass a `--profile` flag to this command.

akash1810:
Not necessarily for this PR...

https://docs.aws.amazon.com/cdk/v2/guide/bootstrapping.html is a great read. Could we create a [custom synthesizer](https://docs.aws.amazon.com/cdk/v2/guide/bootstrapping.html#bootstrapping-custom-synth)?

I've a branch upgrading to AWS CDK v2, by [default](https://github.com/aws/aws-cdk/blob/b1fce4f1641c90a4b7d1d33139453260b452d5cd/packages/%40aws-cdk/core/lib/stack.ts#L401-L403) it uses a `DefaultStackSynthesizer` which [adds a parameter](https://github.com/aws/aws-cdk/blob/b1fce4f1641c90a4b7d1d33139453260b452d5cd/packages/%40aws-cdk/core/lib/stack-synthesizers/default-synthesizer.ts#L451-L481) to your template with the value of CDK your account has been bootstrapped with. Could we create a [custom synthesizer](https://docs.aws.amazon.com/cdk/v2/guide/bootstrapping.html#bootstrapping-custom-synth) for similar purposes? I don't think we'd use the AWS CDK [bootstrap](https://github.com/aws/aws-cdk/blob/7588b517eb17bb5198f91056113eb79a34830867/packages/aws-cdk/lib/api/bootstrap/bootstrap-template.yaml) stack, so a custom synthesizer shouldn't cause issues with it 🤞🏽.

src/types/cli.ts
```diff
@@ -1,11 +1,17 @@
 import type { CredentialProviderChain } from "aws-sdk";
 import type { SubnetList, Vpc } from "aws-sdk/clients/ec2";
 
-export interface AwsAccountReadiness {
+interface AwsConfig {
   credentialProvider: CredentialProviderChain;
   region: string;
 }
 
+export type AwsAccountReadiness = AwsConfig;
```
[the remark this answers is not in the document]
nicl: Ah originally, I thought there was a chance ths interface might expand, but you're right we don't need it now.

src/bin/commands/bootstrap/bootstrap-stack.ts
nicl: Yeah, it would be nice if it was always the same as profile but unfortunately not!

nicl:
@akash1810 thanks for the comments. Have updated with fixes.

Re: synthesizers, that's really interesting. Not as part of this PR, but I wonder if we could use a custom synthesizer to do the following:

* set our @guardian/cdk SSM parameters to default values in SSM
* allow overrides through customising a custom serialiser -  so overrides of  defaults, e.g. for  S3 dist bucket, are at the app level in CDK code rather than a global SSM override.

Does that make sense? Might allow us to provide flexibility while doing more to  promote standardisation of, e.g. dist bucket names.

src/bin/commands/bootstrap/bootstrap-stack.ts
```diff
@@ -0,0 +1,112 @@
+import { readFileSync } from "fs";
+import { StringParameter } from "@aws-cdk/aws-ssm";
+import { App } from "@aws-cdk/core";
+import AWS from "aws-sdk";
+import type { CreateStackInput } from "aws-sdk/clients/cloudformation";
+import chalk from "chalk";
+import { prompt } from "inquirer";
+import type { Question } from "inquirer";
+import { TrackingTag } from "../../../constants/__mocks__/tracking-tag";
```
akash1810, line 9: Don't think we want the mock here?

src/bin/index.ts
akash1810: Heads up: I'd like to further rename `NewCdkProps` (and all related things) to `NewCdkProjectProps`. This should be a different PR and will include other changes to the `new` command such as removing the `multi-app` and `init` flags that (IMO) are not providing much value.

In short, I think the `new` should be simplified by assuming one hasn't setup `cdk` in their repository, and the repository holds a single stack.

src/bin/commands/account-readiness/index.ts
```diff
@@ -7,13 +7,25 @@ export interface Report {
...
+  parametersFound: number;
+  parametersMissing: number;
 }
 
 export const accountReadinessCommand = async (props: AwsAccountReadiness): CliCommandResponse => {
   const report = await reportSSM(props);
 
   console.log(chalk.bold(report.name + ":") + " " + (report.isPass ? "✅ Pass" : "❌ Fail"));
 
+  if (report.parametersFound == 0) {
```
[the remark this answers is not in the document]
nicl: Doh, you'd thought I'd have learnt not to do this by now!

PR #1148, opened 2022-03-07
ci: Perform project generation in CI

akash1810:
This is related to https://github.com/guardian/cdk/pull/1031.

## What does this change?

In this change, we run the new project generator CLI command (`npx @guardian/cdk new`) during CI. This should help us spot any regressions early as CI would fail.

Examples of regressions include:
  - Renaming of properties to a `GuStack`
  - Addition of mandatory properties to a `GuStack` (such as in https://github.com/guardian/cdk/pull/1134 which adds a new `stage` prop)
  - Upgrading to AWS CDK v2 (#1150)

The process is:
  1. CI checkouts the branch
  2. [`npm link`](https://docs.npmjs.com/cli/v8/commands/npm-link) the branch being built
  3. Create and move into a new random folder in `/tmp`
  4. Install `@guardian/cdk` via the `npm link` (i.e. from the current branch)
  5. Use the GuCDK CLI to generate a new project
    - The CLI has been updated to omit `@guardian/cdk` from the generated `package.json`, to ensure we use version from the current branch, and not a published version from `npm`
  6. Run the generated `./script/ci` bash script

The generated `ci` bash script [includes](https://github.com/guardian/cdk/blob/main/src/bin/commands/template/script/ci) a load of `yarn` commands. This has potential for confusion as this repo [explicitly chooses](https://github.com/guardian/cdk/blob/main/docs/architecture-decision-records/003-package-manager.md) to use npm.

## How to test

We want to test the generated project installs `@guardian/cdk` from the current branch and not NPM.

A commit has been added to this branch to `throw` within a `GuStack`. If we can see this exception during CI, then we know we've installed from the branch.

We do ([build log](https://github.com/guardian/cdk/runs/5508025429?check_suite_focus=true)):
image

Spotting regressions early, and before they are shipped.

Build times increase that little bit more 😢 .

## Checklist


package.json
```diff
@@ -19,8 +19,7 @@
     "test": "jest --detectOpenHandles --runInBand",
     "test:custom-lint-rule": "eslint tools/eslint/rules/valid-constructors.test.ts",
     "test:dev": "jest --detectOpenHandles --runInBand --watch",
-    "prepare": "tsc",
-    "prepack": "cp -r src/bin/commands/template lib/bin/commands/template",
+    "prepare": "mkdir -p lib/bin/commands && cp -r src/bin/commands/template lib/bin/commands/template && tsc",
```
akash1810, line 22: Moving the `prepack` steps to `prepare` which is run when `npm link`ing and `npm pack`ing. (`prepare` is the DRYest solution?)

script/ci-project-generation
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+set -e
+
+npm link
```
akash1810, line 5: Installs GuCDK directly from the current branch that's checked out.

akash1810:
Testing this on #1134 is showing that this isn't quite working - I'd expect #1134 to fail as the project generator hasn't been updated following the removal of `GuStageParameter`, however it does not. This suggests this branch is installing GuCDK from npm an not from the local clone.

src/bin/commands/utils/init.ts
```diff
@@ -102,7 +105,7 @@ function createPackageJson(outputDirectory: string): void {
     dependencies: {
       "@aws-cdk/cloudformation-include": LibraryInfo.AWS_CDK_VERSION,
       "@aws-cdk/core": LibraryInfo.AWS_CDK_VERSION,
-      "@guardian/cdk": LibraryInfo.VERSION,
+      ...(!isTest && { "@guardian/cdk": LibraryInfo.VERSION }),
```
akash1810, line 108: Only add `@guardian/cdk` to `package.json` when in not in test as we will `npm link` it instead.

From https://docs.npmjs.com/cli/v8/commands/npm-link#caveat

> Note that package dependencies linked in this way are not saved to package.json by default, on the assumption that the intention is to have a link stand in for a regular non-link dependency. Otherwise, for example, if you depend on redis@^3.0.1, and ran npm link redis, it would replace the ^3.0.1 dependency with file:../path/to/node-redis, which you probably don't want! Additionally, other users or developers on your project would run into issues if they do not have their folders set up exactly the same as yours.
> 
> If you are adding a new dependency as a link, you should add it to the relevant metadata by running npm install <dep> --package-lock-only.
> 
> If you want to save the file: reference in your package.json and package-lock.json files, you can use npm link <dep> --save to do so.

This suggests we can `npm link @guardian/cdk --save`, however the `new` cli command will create `package.json` _and_ install deps in one step; we'd need to opt out of the installation in order to `npm link` at the right time as the order should be:
  - create `package.json`
  - mutate `package.json`
  - install deps

akash1810: The plot thickens... The ideal process would be:
  1. Generate `package.json` and other files, but don't install anything
  2. Mutate `package.json` to install `@guardian/cdk` from local clone
  3. Run `./script/ci` that was generated in 1

However, `./script/ci` fails with [linting errors](https://github.com/guardian/cdk/blob/f224a9517a8270c7107ea140ab131a28105e507b/src/bin/commands/bootstrap.ts#L138). That is, `./script/ci` will always fail.

Whilst the code change in this PR isn't ideal, I think any improvements would need a bigger change, in it's own PR.

jacobwinch: This is a neat solution; thanks for explaining it 👍

akash1810: The branch has been updated to `npm pack` and install from the resulting tarball as it is more representative of real world usage. The effect/need of this change is still the same however.

package.json
akash1810: Fixed this undoing the changes to the `prepare` command in favour of `npm pack`.

jacobwinch (approved):

PR #1151, opened 2022-03-08
docs: Add best practices doc

samanthagottlieb:
## What does this change?

This PR adds a best practices document. It outlines how GuCDK builds in best practices and then lists best practice recommendations to follow when using CDK.

### DIFF
--- a/docs/best-practices.md
+++ b/docs/best-practices.md
@@ -10,7 +10,9 @@ deploying cloud infrastructure with CDK.
 ## Guardian CDK
 Here are some best practices for using the Guardian CDK library:
 - Use [snapshot testing with Jest](https://jestjs.io/docs/snapshot-testing). Snapshot testing makes it easy to review
-  any stack changes when developing locally and also in PRs.
+  any stack changes when developing locally and in PRs. This enables us to catch things early, creating a shorter
+  feedback loop. See the [AWS docs](https://docs.aws.amazon.com/cdk/v2/guide/testing.html#testing_snapshot) on snapshot
+  testing.
 
 - Uploading the json output file produced by running `cdk synth` is preferable to uploading a `cloudformation.yaml`
   file. Using the `cloudformation.yaml` file doesn't guarantee the file reflects the current state of your CDK
--- a/docs/best-practices.md
+++ b/docs/best-practices.md
@@ -1,26 +1,14 @@
 # Best Practices
 
-## Guardian CDK builds in best practices:
-- [Blocks SSH port](https://github.com/guardian/cdk/blob/5a2b429704e324a298e5a348aadf9ab311858bec/src/constructs/ec2/security-groups/base.ts#L58-L60) - uses [SSM-Scala](https://github.com/guardian/ssm-scala) instead
-- Limits the public subnet to things that need to be communicated with (e.g. [load balancers](https://github.com/guardian/cdk/blob/4a8e7d55998d31115a590c01752e363603cf6728/src/patterns/ec2-app.ts#L415-L420))
-- Everything else is in the [private subnet](https://github.com/guardian/cdk/blob/4a8e7d55998d31115a590c01752e363603cf6728/src/patterns/ec2-app.ts#L400) (e.g. apps, servers, databases)
-- [Tags](https://github.com/guardian/cdk/blob/4a8e7d55998d31115a590c01752e363603cf6728/src/constructs/core/stack.ts#L149-L154) resources with [stack, stage and app](https://github.com/guardian/recommendations/blob/ddd2fbd50554cd5d07e42c925e29e5c539807082/AWS.md#general)
+Guardian CDK codifies best practices as defined in the [recommendations](https://github.com/guardian/recommendations)
+repo.
 
-## Best practice recommendations:
-- Test using jest and snapshots
-- Upload the json output file produced by running `cdk synth` is preferable to uploading a `cloudformation.yaml` file
+Here are some best practices for using this library:
+- Use [snapshot testing with Jest](https://jestjs.io/docs/snapshot-testing). Snapshot testing makes it easy to review
+  any stack changes when developing locally and also in PRs.
 
-For instance, this is preferable:
-```
-"synth": "mkdir -p ./build/cdk.out && cdk synth --output ./build/cdk.out > ./build/cdk.out/<project-name>.json"
-```
-to this:
-```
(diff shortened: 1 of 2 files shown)


docs/best-practices.md
```diff
@@ -0,0 +1,26 @@
+# Best Practices
+
+## Guardian CDK builds in best practices:
+- [Blocks SSH port](https://github.com/guardian/cdk/blob/5a2b429704e324a298e5a348aadf9ab311858bec/src/constructs/ec2/security-groups/base.ts#L58-L60) - uses [SSM-Scala](https://github.com/guardian/ssm-scala) instead
+- Limits the public subnet to things that need to be communicated with (e.g. [load balancers](https://github.com/guardian/cdk/blob/4a8e7d55998d31115a590c01752e363603cf6728/src/patterns/ec2-app.ts#L415-L420))
+- Everything else is in the [private subnet](https://github.com/guardian/cdk/blob/4a8e7d55998d31115a590c01752e363603cf6728/src/patterns/ec2-app.ts#L400) (e.g. apps, servers, databases)
+- [Tags](https://github.com/guardian/cdk/blob/4a8e7d55998d31115a590c01752e363603cf6728/src/constructs/core/stack.ts#L149-L154) resources with [stack, stage and app](https://github.com/guardian/recommendations/blob/ddd2fbd50554cd5d07e42c925e29e5c539807082/AWS.md#general)
+
+## Best practice recommendations:
+- Test using jest and snapshots
```
nicl, line 10: It might be worth adding motivation here. E.g. something like 'Snapshot testing makes is easy to review any stack changes when developing locally and also in PRs.'

docs/best-practices.md
```diff
@@ -0,0 +1,26 @@
+# Best Practices
+
+## Guardian CDK builds in best practices:
```
nicl, line 3: I'm not sure about this section. It's great to flag up that CDK aims to establish best practice(!) but this is potentially going to be quite a long and changing list. It might be better to document key points in the patterns and constructs themselves (which we could definitely do more of).

nicl (commented):
Thanks for raising this! And linking it to the readme too.

docs/best-practices.md
samanthagottlieb: Good point, I chatted about this with @akash1810 yesterday and he suggested instead of this section, having a link to the recommendations repo - will update this!

docs/best-practices.md
```diff
@@ -0,0 +1,26 @@
...
+
+For instance, this is preferable:
+```
+"synth": "mkdir -p ./build/cdk.out && cdk synth --output ./build/cdk.out > ./build/cdk.out/<project-name>.json"
+```
+to this:
+```
+"synth": "mkdir -p ./build/cdk.out && cdk synth --output ./build/cdk.out > ./build/cloudformation.yaml"
+```
+Uploading the `cloudformation.yaml` file doesn't guarantee the file will contain yaml. This wouldn't be picked up at the
```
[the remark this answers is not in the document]
samanthagottlieb: Yes, this is perhaps more to the point than what I had written here. My reasoning behind this bit was I had a situation where we were getting failures after CI and we eventually realised that it was because the `cloudformation.yaml` file was empty.

docs/best-practices.md
samanthagottlieb: > Thanks for raising this! And linking it to the readme too.

Thanks so much for the review! I probably should have added the context that this came out of chatting with Akash about Guardian CDK after using it in my team. Hope this doc is helpful!

docs/best-practices.md
```diff
@@ -1,26 +1,14 @@
...
+  any stack changes when developing locally and also in PRs.
 
-For instance, this is preferable:
-```
-"synth": "mkdir -p ./build/cdk.out && cdk synth --output ./build/cdk.out > ./build/cdk.out/<project-name>.json"
-```
-to this:
-```
-"synth": "mkdir -p ./build/cdk.out && cdk synth --output ./build/cdk.out > ./build/cloudformation.yaml"
-```
```
samanthagottlieb: I thought this might be unnecessary extra info, but happy to add it back in if you think it would be useful @nicl.

docs/best-practices.md
```diff
@@ -0,0 +1,14 @@
+# Best Practices
+
+Guardian CDK codifies best practices as defined in the [recommendations](https://github.com/guardian/recommendations)
+repo.
+
+Here are some best practices for using this library:
+- Use [snapshot testing with Jest](https://jestjs.io/docs/snapshot-testing). Snapshot testing makes it easy to review
+  any stack changes when developing locally and also in PRs.
```
akash1810, line 8: Should we expand this, describing why this is preferred compared to the alternative(s)? For example, snapshots allow us to catch things early, therefore it creates a shorter feedback loop.

docs/best-practices.md
```diff
@@ -0,0 +1,14 @@
...
+
+- Uploading the json output file produced by running `cdk synth` is preferable to uploading a `cloudformation.yaml`
+  file. Using the `cloudformation.yaml` file doesn't guarantee the file reflects the current state of your CDK
+  application. However, using the json output file has the benefit that it would alert at CI if the json file hadn't
+  been created properly, alerting about failure sooner.
```
akash1810, line 13: I wonder if there are two recommendations here?
  1. Upload generated files from `cdk.out`
  2. Use CI to generate files (and also run tests in CI)

docs/best-practices.md
akash1810: Worth linking to AWS's docs too? https://docs.aws.amazon.com/cdk/v2/guide/testing.html#testing_snapshot

docs/best-practices.md
```diff
@@ -0,0 +1,14 @@
+# Best Practices
+
+Guardian CDK codifies best practices as defined in the [recommendations](https://github.com/guardian/recommendations)
+repo.
+
+Here are some best practices for using this library:
```
akash1810, line 6: Worth linking to AWS's best practice doc too? https://docs.aws.amazon.com/cdk/v2/guide/best-practices.html

akash1810 (commented):
Great start, thanks for adding this! Added a couple of comments.

docs/best-practices.md
samanthagottlieb: Thanks for this suggestion - I have split the recommendations up, hope this is clearer now.

akash1810:
This is awesome! Thanks for adding it! 🎉

PR #1157, opened 2022-03-11
feat!: remove GuSsmParameter

nicl:
## What does this change?

`GuSSMParameter` is a custom resource that allows using SSM parameters within Cloudformation templates. 

It is quite complex and is (as best we know) unused. See: https://github.com/guardian/cdk/issues/1032 for a wider discussion.

The new preferred approach is simply to either:

* hardcode stage-specific values in your CDK code (if not secrets/sensitive)
* use Cloudformation (SSM-type) parameters if values are sensitive

## How to test

n/a

(Reduced complexity and smaller codebase.)

(Breaking change for anyone using this.)

## Checklist

(I don't think we need to make documentation changes here.)


akash1810 (approved):
image

Better than any Wordle score!

PR #1158, opened 2022-03-11
feat: Return created LambdaRestApis when creating a GuApiLambda

georgeblahblah:
## What does this change?

* Returns the created `LambdaRestApi`s to the caller
* Tests adding a domain name to a `LambdaRestApi` and a `GuCname`

## Checklist

### DIFF
--- a/src/patterns/api-lambda.test.ts
+++ b/src/patterns/api-lambda.test.ts
@@ -1,6 +1,5 @@
 import "@aws-cdk/assert/jest";
 import { SynthUtils } from "@aws-cdk/assert";
-import { DomainName } from "@aws-cdk/aws-apigateway/lib/domain-name";
 import { Runtime } from "@aws-cdk/aws-lambda";
 import { Duration } from "@aws-cdk/core";
 import { GuCertificate } from "../constructs/acm";
@@ -80,6 +79,7 @@ describe("The GuApiLambda pattern", () => {
     });
     const api = apiLambda.apis.get("api");
     expect(api).toBeDefined();
+    /* eslint-disable @typescript-eslint/no-non-null-assertion -- We expect api to be defined beyond this point */
     api!.addDomainName("domain", {
       domainName: "code.theguardian.com",
       certificate: new GuCertificate(stack, {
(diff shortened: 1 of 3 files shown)


src/patterns/api-lambda.test.ts
```diff
@@ -53,4 +56,46 @@ describe("The GuApiLambda pattern", () => {
...
+        domainName: "code.theguardian.com",
+        hostedZoneId: "id123",
+      }),
+    });
+
+    new GuCname(stack, "DNS", {
+      app: "testing",
+      ttl: Duration.days(1),
+      domainName: "code.theguardian.com",
+      resourceRecord: api!.domainName!.domainNameAliasDomainName,
```
akash1810, line 96: Curious why `domainName` is nullable here, I guess that's just how AWS have typed it?

src/patterns/api-lambda.test.ts
```diff
@@ -53,4 +56,46 @@ describe("The GuApiLambda pattern", () => {
...
+        },
+        {
+          id: "api2",
+          description: "this is a test2",
+        },
+      ],
+    });
+    const api = apiLambda.apis.get("api");
+    expect(api).toBeDefined();
+    /* eslint-disable @typescript-eslint/no-non-null-assertion -- We expect api to be defined beyond this point */
```
akash1810, line 82: I'm not a fan of disabling linting to huge blocks as this is also affecting `domainName` on `api`.

Something like this would be preferred:

```ts
const maybeApi = apiLambda.apis.get("api");
expect (api).toBeDefined();

// eslint-disable @typescript-eslint/no-non-null-assertion ......
const api = maybeApi!;
```

As it is more tightly scopes the lint exclusion.

src/patterns/api-lambda.test.ts
georgeblahblah: I think that's correct - it looks like it's an optional value https://github.com/aws/aws-cdk/blob/v1.137.0/packages/@aws-cdk/aws-apigateway/lib/restapi.ts#L150

src/patterns/api-lambda.test.ts
georgeblahblah: That's much clearer, thanks for the suggestion

akash1810 (approved):

PR #1174, opened 2022-03-30
chore(deps)!: Update to AWS CDK v2

akash1810:
Another, unavoidably, massive change! Easiest to review commit by commit.

## What does this change?

Updates to AWS CDK v2, largely by following https://docs.aws.amazon.com/cdk/v2/guide/migrating-v2.html. The main benefit gained is the simpler dependency management in v2's monolithic package. More information [here](https://github.com/aws/aws-cdk-rfcs/blob/master/text/0079-cdk-2.0.md).

The project generator has been updated, so new projects will be created correctly.

To update existing projects:
  - Update `package.json`:
    - Replace all `@aws-cdk/*` packages with the single `aws-cdk-lib` package
    - Add `constructs` dependency
    - See `peerDependencies` for version numbers
  - Fix imports `npx -p aws-cdk-migration rewrite-imports-v2 lib/**/*.ts`
  - Update `cdk.json`, removing deprecated feature flags such as `@aws-cdk/core:enableStackNameDuplicates`

## How to test

CI should continue to pass. As CI is testing synthing a stack (via [`integration-test`](https://github.com/guardian/cdk/tree/main/tools/integration-test)) and [new project generation](https://github.com/guardian/cdk/blob/main/.github/workflows/ci-integration-test.yaml), a green build should hopefully be enough.

Dramatic increase in developer experience from use of a monolithic package, offering a simpler upgrade path for us and downstream users of GuCDK.

Yet another breaking change! See https://github.com/guardian/cdk/discussions/910.

## Checklist


package.json
```diff
@@ -53,25 +55,6 @@
...
-    "@aws-cdk/aws-events-targets": "1.147.0",
-    "@aws-cdk/aws-iam": "1.147.0",
-    "@aws-cdk/aws-kinesis": "1.147.0",
-    "@aws-cdk/aws-lambda": "1.147.0",
-    "@aws-cdk/aws-lambda-event-sources": "1.147.0",
-    "@aws-cdk/aws-rds": "1.147.0",
-    "@aws-cdk/aws-s3": "1.147.0",
-    "@aws-cdk/aws-stepfunctions": "1.147.0",
-    "@aws-cdk/aws-stepfunctions-tasks": "1.147.0",
-    "@aws-cdk/core": "1.147.0",
```
akash1810, line 74: It's wonderful, isn't it?!

src/constructs/autoscaling/asg.test.ts
```diff
@@ -133,22 +128,22 @@ describe("The GuAutoScalingGroup", () => {
...
-          "Fn::GetAtt": ["SecurityGroupTestingA32D34F9", "GroupId"], // auto-generated logicalId
+          "Fn::GetAtt": [Match.stringLikeRegexp("SecurityGroupTesting"), "GroupId"],
         },
         {
-          "Fn::GetAtt": ["SecurityGroup1TestingCA3A17A4", "GroupId"], // auto-generated logicalId
+          "Fn::GetAtt": [Match.stringLikeRegexp("SecurityGroup1Testing"), "GroupId"],
         },
         {
-          "Fn::GetAtt": ["SecurityGroup2Testing6436C75B", "GroupId"], // auto-generated logicalId
+          "Fn::GetAtt": [Match.stringLikeRegexp("SecurityGroup2Testing"), "GroupId"],
```
akash1810, line 146: This is a lot nicer, IMO, as we're not matching the exact GUID, thus it becomes more flexible.

src/utils/test/assertions.ts
```diff
@@ -0,0 +1,103 @@
...
+      return -1;
+    }
+    if (first.Key.toLowerCase() > second.Key.toLowerCase()) {
+      return 1;
+    }
+    return 0;
+  });
+}
+
+export class GuTemplate {
```
akash1810, line 30: This class is a replacement for `test/jest.ts` and `test/synthed-stack.ts`. The aim is to provide a similar API to the [assertions model offered by AWS CDK v2](https://docs.aws.amazon.com/cdk/v2/guide/testing.html).

The API of `GuTemplate` isn't quite the same as `Template`: `new. GuTemplate(stack)` vs `Template.fromStack(stack)`. Happy to match the API if desired.

src/patterns/ec2-app/base.test.ts
```diff
@@ -649,18 +640,14 @@ describe("the GuEC2App pattern", function () {
...
-    expect(lbKey).toBeTruthy();
-
-    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- We assert above that this is truthy
-    const loadBalancer = json.Resources[lbKey!];
-
-    expect(loadBalancer.Properties.LoadBalancerAttributes).toEqual(
-      expect.not.arrayContaining([{ Key: "access_logs.s3.enabled", Value: "true" }])
-    );
+    Template.fromStack(stack).hasResourceProperties("AWS::ElasticLoadBalancingV2::LoadBalancer", {
+      LoadBalancerAttributes: Match.arrayEquals([
```
akash1810, line 645: I think this is a simpler, and more complete, test than before. Previously we were checking that `access_logs.s3.enabled` was not present in `LoadBalancerAttributes`; other properties could have been set and changed between changes and we wouldn't know. Now, we're checking the exact contents of `LoadBalancerAttributes`.

.github/dependabot.yml
```diff
@@ -18,11 +18,10 @@ updates:
       prefix: "chore(deps): "
     # force the lockfile and package.json to be updated
     versioning-strategy: increase
-    # We update @aws-cdk/* libraries ourselves (via `script/update-aws-cdk`)
+    # We update AWS CDK libraries ourselves (via `script/update-aws-cdk`)
     # Dependabot wants to update them individually, but they should all be done at once.
     ignore:
-      - dependency-name: "aws-cdk"
-      - dependency-name: "@aws-cdk/*"
+      - dependency-name: "aws-cdk-lib"
```
jacobwinch, line 24: Now that this is a monolithic package, can we allow dependabot to update it for us?

If so, I think we could also remove our update script and [some CI stuff](https://github.com/guardian/cdk/blob/5b42e24646dbc176826fc16999dc821783225726/.github/workflows/ci.yaml#L20-L24) which aims to stop us from forgetting to keep this library up to date?

src/constructs/core/stack.ts
```diff
@@ -114,7 +114,14 @@ export class GuStack extends Stack implements StackStageIdentity, GuMigratingSta
...
     } = props;
 
-    super(app, id, { ...props, stackName: cloudFormationStackName });
+    super(app, id, {
+      ...props,
+      stackName: cloudFormationStackName,
+
+      // TODO Use `DefaultStackSynthesizer` or create own synthesizer?
+      //  see https://docs.aws.amazon.com/cdk/v2/guide/bootstrapping.html#bootstrapping-custom-synth
+      synthesizer: new LegacyStackSynthesizer(),
```
jacobwinch, line 123: Thinks for figuring this out! Maybe our `synthesizer` can produce YAML?!

jacobwinch (approved):
Great work 👍

.github/dependabot.yml
akash1810: In theory, yes we can get Dependabot to help here. In practice, I'm not sure if we'd want to? Dependabot runs every week, and bumps to `aws-cdk-lib` will incur a new release - do we want to create a new major release so frequently?

jacobwinch: I think you're right that weekly updates to this library would be too frequent.

I think that we have a few options here:

1. Keep things as they are but update the comment here which is now out of date. Suggestion for this:

```
# We update AWS CDK libraries ourselves (via `script/update-aws-cdk`) as this is a breaking change for users and it results in a new major version being released
```

2. Allow dependabot to update this library too; update the schedule so that dependabot checks all dependencies less frequently (e.g. fortnightly or monthly).

3. Create a dedicated update schedule for this library only (using [allow](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#allow)) - I'm not 100% sure if this would work but in theory it would allow us to continue checking/updating most dependencies weekly _and_ automatically update `aws-cdk-lib` less frequently.

Personally I'd prefer to remove the customised automation around this dependency (and I don't see any huge risk presented by option 2) but I'm happy to go with 1 if you just want to get this merged.

akash1810: > Personally I'd prefer to remove the customised automation around this dependency

+1

Let's decouple it from this (already massive) change and do 1. for now.

PR #1191, opened 2022-04-08
fix: add withoutImdvs2 to props of GuAutoScalingGroup and GuEc2App 

TomPretty:
## What does this change?

We're adding this prop to allow consumers of the GuAutoScalingGroup construct and GuEc2App pattern to be able to explicitly opt out of IMDSv2. Previously, requireImdsv2 was hardcoded to true inside of GuAutoScalingGroup. It's best practice to use IMDSv2, but it's good to provide an escape hatch for applications that aren't ready to adopt it.

NB: GuAutoScalingGroup defaults the value to false, so existing consumers will be unaffected by this change.

## How to test

N/A

support-frontend can log again!

N/A

## Checklist

### DIFF
--- a/src/constructs/autoscaling/asg.ts
+++ b/src/constructs/autoscaling/asg.ts
@@ -37,6 +37,7 @@ export interface GuAutoScalingGroupProps
   userData: UserData | string;
   additionalSecurityGroups?: ISecurityGroup[];
   targetGroup?: ApplicationTargetGroup;
+  withoutImdsv2?: boolean;
 }
 
 /**
@@ -71,6 +72,7 @@ export class GuAutoScalingGroup extends GuStatefulMigratableConstruct(GuAppAware
       targetGroup,
       userData: userDataLike,
       vpc,
+      withoutImdsv2 = false,
     } = props;
 
     // Ensure min and max are defined in the same way. Throwing an `Error` when necessary. For example when min is defined via a Mapping, but max is not.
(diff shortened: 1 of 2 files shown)


akash1810:
Looks like CI is failing, could you run `./script/update-aws-cdk` as [described](https://github.com/guardian/cdk/runs/5882194084?check_suite_focus=true) and rebase?

src/constructs/autoscaling/asg.ts
```diff
@@ -67,6 +66,7 @@ export class GuAutoScalingGroup extends GuStatefulMigratableConstruct(GuAppAware
       imageId = new GuAmiParameter(scope, { app }),
       minimumInstances,
       maximumInstances,
+      requireImdsv2 = true,
```
akash1810, line 69: Having given it some thought, I don't think we should call this `requireImdsv2` as the AWS default is `false` and this will appear in intellisense, even though GuCDK will default it to `true`, which will be confusing.

`withoutImdsv2` might be better? We've precedent for `without*` in the library already [here](https://github.com/guardian/cdk/blob/7b3299e2cb98f64ac23018441c29f825d2f1dbcc/src/constructs/core/stack.ts#L29) and [here](https://github.com/guardian/cdk/blob/7b3299e2cb98f64ac23018441c29f825d2f1dbcc/src/constructs/iam/roles/instance-role.ts#L16).

TomPretty:
> Looks like CI is failing, could you run [`./script/update-aws-cdk`](https://github.com/guardian/cdk/blob/main/script/update-aws-cdk) as [described](https://github.com/guardian/cdk/runs/5882194084?check_suite_focus=true) and rebase?

Yeah I've done that! New PR: #1192

src/constructs/autoscaling/asg.ts
TomPretty: Yeah makes sense - I've pushed that up!

akash1810 (approved):

PR #1201, opened 2022-04-12
chore(deps): Remove deprecated dependency `cli-ux`

akash1810:
## What does this change?

Replace `cli-ux` with `@oclif/core` per https://github.com/oclif/core/blob/main/src/cli-ux/README.md. It could also be argued that we should drop this dependency entirely and simply use `console.log`...

## How to test

As this is a no-op, CI should continue to pass.

Removal of a deprecated dependency.

n/a

## Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,9 +8,9 @@
       "name": "@guardian/cdk",
       "version": "41.0.0",
       "dependencies": {
+        "@oclif/core": "1.7.0",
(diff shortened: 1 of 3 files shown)


akash1810:
Closes #1065.

jacobwinch (approved):
> It could also be argued that we should drop this dependency entirely and simply use console.log...

I'd be happy with the above suggestion too, but feel free to merge this!

PR #1211, opened 2022-04-13
feat: Add support for automatic log shipping

nicl:
(Applies to the Ec2App patterns.)

A LogKinesisStreamName tag is added to the ASG and nested resources (including instances). These are read to enable automatic log shipping, initially of Cloud Init Logs.

For more info see:

* https://github.com/guardian/amigo/blob/46725214bac8d4ced5f3b2df27bc49ec6c37bd20/roles/cdk-base/README.md
* https://github.com/guardian/amigo/pull/706

And also the recent Amiable PR that would have not been necessary with this change:

https://github.com/guardian/amiable/pull/122

### DIFF
--- a/src/constructs/iam/roles/instance-role.ts
+++ b/src/constructs/iam/roles/instance-role.ts
@@ -12,7 +12,14 @@ import type { GuPolicy } from "../policies";
 import { GuRole } from "./roles";
 
 export interface GuInstanceRoleProps {
+  /**
+   * By default, instances are given permissions to write to Kinesis. Set to
+   * 'true' to prevent this. Note, disabling will prevent not just application
+   * logs being shipped but also anything else - for example, automatic log
+   * shipping of Cloud Init and other logs by the cdk-base role in your AMI.
+   */
   withoutLogShipping?: boolean;
+
   additionalPolicies?: GuPolicy[];
 }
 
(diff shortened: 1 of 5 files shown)


src/constructs/iam/roles/instance-role.ts
```diff
@@ -12,7 +12,12 @@ import type { GuPolicy } from "../policies";
 import { GuRole } from "./roles";
 
 export interface GuInstanceRoleProps {
+  // By default, instances are given permissions to write to Kinesis. Set to
+  // 'true' to prevent this. Note, disabling will prevent not just application
+  // logs being shipped but also anything else - for example, automatic log
+  // shipping of Cloud Init and other logs by the cdk-base role in your AMI.
```
akash1810, line 18: Should we make this a typedoc comment to make it visible in intellisense and the documentation site? Should we also link to the AMIgo recipe?

src/patterns/ec2-app/base.ts
```diff
@@ -348,6 +348,13 @@ export class GuEc2App {
     // TODO: Do we need to tag all resources with this value? What would the use-cases be?
     Tags.of(autoScalingGroup).add(TagKeys.PATTERN_NAME, this.constructor.name, { applyToLaunchedInstances: true });
 
+    // This allows automatic shipping of instance Cloud Init logs when using the
+    // `cdk-base` Amigo role on your AMI.
+    Tags.of(autoScalingGroup).add(
+      "LogKinesisStreamName",
```
akash1810, line 354: Could we move this string into [`TagKeys`](https://github.com/guardian/cdk/blob/main/src/constants/tag-keys.ts)?

akash1810 (approved):

PR #1212, opened 2022-04-13
chore: add VS Code extension recommendations

nicl:
Automatically recommend key extensions when project is opened in VS Code.


jacobwinch (approved):

akash1810 (approved):
nit: `feat` will create a [new minor release](https://www.conventionalcommits.org/en/v1.0.0/), which I don't think we want here? `chore` might be better?

PR #1244, opened 2022-05-05
feat(tagging): Tag resources with build number

akash1810:
## What does this change?

The build number is one (/the only?) way to know which version of the infrastructure definition is running, and is useful information when triaging incidents.

In this change we look up the build number from environment variables:
  - `BUILD_NUMBER` ([TeamCity](https://www.jetbrains.com/help/teamcity/predefined-build-parameters.html#Predefined+Server+Build+Parameters))
  - `GITHUB_RUN_NUMBER` ([GitHub Actions](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables))
  - If not found, we fall back to `"unknown"`.

The build number is then applied as a `gu:build-number` tag to all resources in the stack.

Coupled with https://github.com/guardian/devx-logs and https://github.com/guardian/cloudwatch-logs-management, this tag will get automatically added as markers to logs shipped to the ELK stack.

The project generator has been updated to configure a mock of this new tag. Existing projects would need to add `jest.mock("@guardian/cdk/lib/constants/build-number-tag");` to `jest.setup.js`.

## How to test

CI should continue to pass as tests have been updated.

More information is available when triaging incidents, reducing uncertainty.

### Accuracy
With this change, the marker on the log lines is determined by the AWS tag. If one performs a custom deploy in Riff-Raff, excluding the cloud-formation deployment, the value of the log line marker will not be accurate:

- Deploy all from build 1
- `gu:build-number` tag is "1"
- Log lines have marker `gu:build-number` with value of "1"
- Perform an application only deploy of build 2
- `gu:build-number` tag does not change
- Log lines continue to have marker `gu:build-number` with value of "1"

Maybe a more accurate tag name is `gu:infrastructure-build-number`?

### Increased deployment times
Today, if the IaC hasn't changed between builds, the cost of a "deploy all" is just that of the autoscaling deployment as the cloud-formation task is a no-op.

With this change, every build will produce CloudFormation changes and so the cost of "deploy all" will increase. This might be of particular concern when performing rollback deployments.

## Checklist

### DIFF
--- a/docs/best-practices.md
+++ b/docs/best-practices.md
@@ -19,44 +19,39 @@ Here are some recommendations for using the Guardian CDK library:
 developing locally and in PRs. This enables us to catch things early, creating a shorter feedback loop. See the
 [AWS docs](https://docs.aws.amazon.com/cdk/v2/guide/testing.html#testing_snapshot) on snapshot testing.
 
-#### Mocking `gu:cdk:version`
-The `@guardian/cdk` library follows [Semantic Versioning](https://semver.org/).
+#### Mocking
+GuCDK will add tags to resources:
+  - `gu:cdk:version` represents the version of GuCDK being used. This helps with usage tracking.
+  - `gu:build-number` represents the build number from CI. This helps when triaging incidents.
 
-GuCDK will add the `gu:cdk:version` tag to resources to aid usage tracking.
-This number changes with every version and as a result automatic version upgrade PRs (e.g. via Dependabot) will fail their snapshot tests.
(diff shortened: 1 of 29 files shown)


akash1810:
Alternatives to this, that will remove the noted risks could be to mutate the user data of an ASG, writing the build number to a known location for devx-logs to pick up. For lambdas, we could expose the build as an environment variable.

akash1810:
This needs more thinking. Closing.

PR #1250, opened 2022-05-09
feat: add new pattern for API Gateway with routing to multiple Lambdas

jacobwinch:
## What does this change?

This PR introduces a new pattern for an [API Gateway instance backed by multiple lambdas](https://docs.google.com/document/d/1e_7JaOf4gHwcmZy9g_1ZlIGVJBdbdyf3RjeDv2fXfJE/edit#bookmark=id.wcb1lwyg1stn).

In order to accommodate the new pattern, I've technically made some breaking changes. These changes only affect constructs/types, so it is unlikely that many users will be interacting with them directly.

BREAKING CHANGE: A small number of types/classes have been renamed or relocated.

1. The `Http5xxAlarmProps` type should now be imported from "@guardian/cdk/lib/constructs/cloudwatch"
2. The `Gu5xxPercentageAlarm` construct is renamed: `GuAlb5xxPercentageAlarm`
3. The `Gu5xxPercentageAlarmProps` type is renamed: `GuAlb5xxPercentageAlarmProps`

## How to test

I've deployed an example project (using this new pattern) to the Developer Playground account and confirmed that it works as expected.

Another fairly common infrastructure pattern is supported by the GuCDK.

This introduces a second way for users to configure a serverless API using API Gateway. I think both use-cases are valid, but having this choice adds some complexity for users. I've tried to explain when each pattern should be used via the docs.

Technically this PR introduces breaking changes to some constructs/types, but this is low risk as the changes are superficial (renaming/moving imports) and they do not affect the underlying infrastructure.

## Checklist

### DIFF
--- a/src/constructs/cloudwatch/alarm.ts
+++ b/src/constructs/cloudwatch/alarm.ts
@@ -32,7 +32,7 @@ export interface Http5xxAlarmProps
  * ```
  *
  * This library provides an implementation of some commonly used alarms, which require less boilerplate than this construct,
- * for example [[`Gu5xxPercentageAlarm`]]. Prefer using these more specific implementations where possible.
+ * for example [[`GuAlb5xxPercentageAlarm`]]. Prefer using these more specific implementations where possible.
  */
 export class GuAlarm extends Alarm {
   constructor(scope: GuStack, id: string, props: GuAlarmProps) {
--- a/src/constructs/cloudwatch/alarm.ts
+++ b/src/constructs/cloudwatch/alarm.ts
@@ -9,6 +9,15 @@ export interface GuAlarmProps extends AlarmProps, AppIdentity {
   snsTopicName: string;
 }
 
+export interface Http5xxAlarmProps
+  extends Omit<
+    GuAlarmProps,
+    "snsTopicName" | "evaluationPeriods" | "metric" | "period" | "threshold" | "treatMissingData" | "app"
+  > {
+  tolerated5xxPercentage: number;
+  numberOfMinutesAboveThresholdBeforeAlarm?: number;
+}
+
 /**
  * Creates a CloudWatch alarm which sends notifications to the specified SNS topic.
  *
--- a/src/constructs/cloudwatch/api-gateway-alarms.ts
+++ b/src/constructs/cloudwatch/api-gateway-alarms.ts
@@ -1,0 +1,45 @@
+import { Duration } from "aws-cdk-lib";
+import type { RestApi } from "aws-cdk-lib/aws-apigateway";
+import { ComparisonOperator, MathExpression, TreatMissingData } from "aws-cdk-lib/aws-cloudwatch";
+import type { GuStack } from "../core";
+import { AppIdentity } from "../core";
+import type { GuAlarmProps, Http5xxAlarmProps } from "./alarm";
+import { GuAlarm } from "./alarm";
+
+interface GuApiGateway5xxPercentageAlarmProps
+  extends Pick<GuAlarmProps, "snsTopicName">,
+    Http5xxAlarmProps,
+    AppIdentity {
+  apiGatewayInstance: RestApi;
+}
+
+/**
+ * Creates an alarm which is triggered whenever the percentage of requests with a 5xx response code exceeds
+ * the specified threshold.
+ */
+export class GuApiGateway5xxPercentageAlarm extends GuAlarm {
+  constructor(scope: GuStack, props: GuApiGateway5xxPercentageAlarmProps) {
+    const mathExpression = new MathExpression({
+      expression: "100*m1/m2",
+      usingMetrics: {
+        m1: props.apiGatewayInstance.metricServerError(),
+        m2: props.apiGatewayInstance.metricCount(),
+      },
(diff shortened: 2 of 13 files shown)


src/constructs/cloudwatch/alarm.ts
```diff
@@ -9,6 +9,15 @@ export interface GuAlarmProps extends AlarmProps, AppIdentity {
   snsTopicName: string;
 }
 
+export interface Http5xxAlarmProps
```
jacobwinch, line 12: Users now use this same interface to configure a 5XX alarm, regardless of whether they are using EC2 or Lambda for compute.

src/constructs/cloudwatch/ec2-alarms.ts
```diff
@@ -5,18 +5,9 @@ import type { GuStack } from "../core";
 import { AppIdentity } from "../core";
 import type { GuApplicationLoadBalancer, GuApplicationTargetGroup } from "../loadbalancing";
 import { GuAlarm } from "./alarm";
-import type { GuAlarmProps } from "./alarm";
+import type { GuAlarmProps, Http5xxAlarmProps } from "./alarm";
 
-export interface Http5xxAlarmProps
```
jacobwinch, line 10: This interface is now shared, see https://github.com/guardian/cdk/pull/1250/files#r867990801

src/constructs/cloudwatch/api-gateway-alarms.ts
```diff
@@ -0,0 +1,45 @@
...
+    });
+    const defaultAlarmName = `High 5XX error % from ${props.app} (ApiGateway) in ${scope.stage}`;
+    const defaultDescription = `${props.app} exceeded ${props.tolerated5xxPercentage}% error rate`;
+    const alarmProps = {
+      ...props,
+      metric: mathExpression,
+      treatMissingData: TreatMissingData.NOT_BREACHING,
+      threshold: props.tolerated5xxPercentage,
+      comparisonOperator: ComparisonOperator.GREATER_THAN_THRESHOLD,
+      alarmName: props.alarmName ?? defaultAlarmName,
```
akash1810, line 39: minor: Could we use destructuring with default values (e.g [here](https://github.com/guardian/cdk/blob/b140d9f4025ebf56eb6b0cd2279e6e18a551eed1/src/constructs/ecs/ecs-task.ts#L142-L157)) to improve readability a little?

src/patterns/api-multiple-lambdas.ts
```diff
@@ -0,0 +1,135 @@
...
+
+export interface ApiTarget {
+  /**
+   * The path for the request (e.g. /test).
+   */
+  path: string;
+  /**
+   * The HTTP method for the target (e.g. GET, POST, PUT).
+   */
+  httpMethod: string;
```
akash1810, line 16: Could we use the [`HttpMethod` enum](https://docs.aws.amazon.com/cdk/api/v2/docs/@aws-cdk_aws-apigatewayv2-alpha.HttpMethod.html) here?

src/patterns/api-multiple-lambdas.ts
```diff
@@ -0,0 +1,135 @@
...
+ *     }
+ *   }
+ * });
+ * ```
+ *
+ * For all API configuration options, see [[`GuApiMultipleLambdasProps`]].
+ *
+ * For details on configuring the individual Lambda functions, see [[`GuLambdaFunction`]].
+ */
+export class GuApiMultipleLambdas {
```
akash1810, line 119: minor: The name of this pattern isn't immediately obvious. Not sure what an alternative could be; maybe `GuApiGatewayWithLambdaByPath`, which is... snappy 🤣 .

src/patterns/api-multiple-lambdas.ts
akash1810: Just noticed this enum is in `@aws-cdk/aws-apigatewayv2-alpha`, which we're not currently dependant on and would be overkill to add just for this. A [union type](https://www.typescriptlang.org/docs/handbook/typescript-in-5-minutes-func.html#unions) would work.

src/patterns/api-multiple-lambdas.ts
jacobwinch: Yeah, it's hard to name these patterns without becoming very verbose! I think `GuApiGatewayWithLambdaByPath` is an improvement though!

src/patterns/api-multiple-lambdas.ts
jacobwinch: Good idea 👍

src/constructs/cloudwatch/api-gateway-alarms.ts
jacobwinch: I had a quick look at this but destructuring makes things awkward here as we just want to pass the whole object into the [superclass](https://github.com/guardian/cdk/blob/8b434a5d728cb9fbcd73dbf3645927b50e4afb93/src/constructs/cloudwatch/api-gateway-alarms.ts#L43).

michaelwmcnamara (approved):
Fantastic work!
LGTM and looks like all comments from others have been addressed.

PR #1258, opened 2022-05-11
feat: allow users to opt-in to application log shipping via GuEc2App

jacobwinch:
## What does this change?

This PR helps users to configure application log shipping via the `GuEc2App` pattern (or one of its subclasses).

**By default, application logging is switched off.** This is because the majority of users of the pattern will be migrating (or have already migrated) an app which already ships application logs via some other means. We do not want to accidentally ship duplicate logs to ELK and we do not want to increase friction by forcing users to switch their approach to logging at the point of adoption.

The downside of this is that brand new applications will not automatically pick up this sensible default (i.e. they'll need to remember to opt-in via the pattern props).

I suggest that we update the default behaviour (via a breaking change) when a significant number of apps have started using `devx-logs`.

## How to test

Unit testing should be sufficient here. Once merged, I'll upgrade and slightly [simplify Amiable's `cdk` definition](https://github.com/guardian/amiable/blob/c7062e5fdc6ad0be3aa24ee3bb92289a93877f27/cdk/lib/amiable/amiable.ts#L56).

There's a little less effort involved in making the infrastructure changes required to ship application logs via `devx-logs`.

I think this is pretty low risk as it doesn't modify existing behaviour.

## Checklist

### DIFF
--- a/src/patterns/ec2-app/base.test.ts
+++ b/src/patterns/ec2-app/base.test.ts
@@ -653,6 +653,96 @@ describe("the GuEC2App pattern", function () {
     });
   });
 
+  it("can be configured with application logging", function () {
+    const stack = simpleGuStackForTesting({ env: { region: "eu-west-1" } });
+    const app = "test-gu-ec2-app";
+    new GuEc2App(stack, {
+      applicationPort: 3000,
+      access: { scope: AccessScope.PUBLIC },
+      app,
+      instanceType: InstanceType.of(InstanceClass.T4G, InstanceSize.MEDIUM),
+      certificateProps: {
+        domainName: "code-guardian.com",
+      },
+      scaling: {
+        minimumInstances: 1,
+      },
+      monitoringConfiguration: { noMonitoring: true },
+      userData: "",
+      applicationLogging: { enabled: true },
+    });
+
+    GuTemplate.fromStack(stack).hasGuTaggedResource("AWS::AutoScaling::AutoScalingGroup", {
+      appIdentity: { app },
+      propagateAtLaunch: true,
+      additionalTags: [
+        { Key: "Name", Value: "Test/AutoScalingGroupTestguec2app" },
+        { Key: TagKeys.PATTERN_NAME, Value: "GuEc2App" },
+        { Key: TagKeys.LOG_KINESIS_STREAM_NAME, Value: { Ref: "LoggingStreamName" } },
+        { Key: TagKeys.SYSTEMD_UNIT, Value: "test-gu-ec2-app.service" },
+      ],
+    });
+  });
+
+  it("can be configured with application logging using a custom systemd unit name", function () {
+    const stack = simpleGuStackForTesting({ env: { region: "eu-west-1" } });
+    const app = "test-gu-ec2-app";
+    new GuEc2App(stack, {
+      applicationPort: 3000,
+      access: { scope: AccessScope.PUBLIC },
+      app,
+      instanceType: InstanceType.of(InstanceClass.T4G, InstanceSize.MEDIUM),
+      certificateProps: {
+        domainName: "code-guardian.com",
+      },
+      scaling: {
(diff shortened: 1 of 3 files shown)


src/patterns/ec2-app/base.ts
```diff
@@ -307,6 +340,15 @@ export class GuEc2App {
       withoutImdsv2,
     } = props;
 
+    // We should really prevent users from doing this via the type system,
```
jacobwinch, line 343: We have quite a few props related to logging now:

`accessLogging`
`applicationLogging`
`roleConfiguration.withoutLogShipping`

We are also configuring the tag for `cloud-init-output.log` shipping automatically: https://github.com/guardian/cdk/pull/1211.

I was tempted to change the API to introduce a single `loggingConfiguration` interface, which explained all of this and tried to make the configuration a bit more consistent. However, we have been making a lot of breaking changes recently and this feels like something that can wait.

src/patterns/ec2-app/base.test.ts
```diff
@@ -653,6 +653,93 @@ describe("the GuEC2App pattern", function () {
...
+          minimumInstances: 1,
+        },
+        monitoringConfiguration: { noMonitoring: true },
+        userData: "",
+        applicationLogging: { enabled: true },
+        roleConfiguration: {
+          withoutLogShipping: true,
+        },
+      });
+    }).toThrowError();
```
nicl, line 740: I wonder if we should be more explicit here about the error that should be thrown?

src/patterns/ec2-app/base.test.ts
```diff
@@ -653,6 +653,93 @@ describe("the GuEC2App pattern", function () {
...
+          domainName: "code-guardian.com",
+        },
+        scaling: {
+          minimumInstances: 1,
+        },
+        monitoringConfiguration: { noMonitoring: true },
+        userData: "",
+        applicationLogging: { enabled: true },
+        roleConfiguration: {
+          withoutLogShipping: true,
```
nicl, line 737: I agree this is a bit confusing (re you thoughts about whether we should group the logging props) but happy to wait til a later release to figure out a cleaner interface here. (Hopefully we can provide an automatic rewrite/migration tool at that point to help with things.)

nicl (approved):
🚢 nice! Looking forward to seeing this in action :)

src/patterns/ec2-app/base.ts
```diff
@@ -29,6 +29,25 @@ export interface AccessLoggingProps {
   prefix?: string;
 }
 
+/**
+ * To ship your application logs to ELK automatically, you must:
+ *
+ * 1. Set the `enabled` flag to true
+ * 2. Include the `cdk-base` Amigo role in your AMI
+ * 3. Log to `journald`. We recommend doing this by logging to `stdout` and
+ * using `systemd` to start your app
```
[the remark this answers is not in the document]
jacobwinch: Good point, I will update this.

src/patterns/ec2-app/base.test.ts
jacobwinch: Yes, I think you're right - I will update this test.

src/patterns/ec2-app/base.test.ts
jacobwinch: > Hopefully we can provide an automatic rewrite/migration tool at that point to help with things.

I think this is a great idea for changes like this where the `GuCDK` API will change but the actual template that gets generated remains the same!

src/patterns/ec2-app/base.ts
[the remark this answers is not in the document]
akash1810: I wonder if it's worth us doing a scan of all launch configurations and see how many will meet these requirements today. If it's a high enough percentage, maybe that warrants us having a construct that takes minimal props (probably just app name?) and creates the user data.

PR #1259, opened 2022-05-11
feat: Add `overrideLogicalId` function to `GuStack`

akash1810:
## Checklist


src/constructs/core/__data__/cfn.yaml
```diff
@@ -0,0 +1,10 @@
+AWSTemplateFormatVersion: '2010-09-09'
```
akash1810, line 1: `CfnInclude` will throw if it cannot find the YAML file to include in a CDK stack.

This file won't make it into the published module.

src/constructs/core/migrating.ts
```diff
@@ -17,55 +23,5 @@ export interface GuMigratingResource {
...
-        );
-      }
-
-      if (isStateful) {
-        Annotations.of(construct).addInfo(
-          `GuStack has 'migratedFromCloudFormation' set to false. ${id} is a stateful construct, it's logicalId will be auto-generated and AWS will create a new resource.`
-        );
-      }
-    }
-  },
```
akash1810, line 70: A _lot_ of complexity is being removed!

src/constructs/core/stack.ts
```diff
@@ -138,6 +130,24 @@ export class GuStack extends Stack implements StackStageIdentity, GuMigratingSta
 
       this.tryAddRepositoryTag();
     }
+
+    const { node } = this;
+
+    node.addValidation({
```
akash1810, line 136: The validation step happens just before synthesis:

image

See:
  - [App lifecycle](https://docs.aws.amazon.com/cdk/v2/guide/apps.html#lifecycle)
  - The [`constructs` docs](https://github.com/aws/constructs/blob/bd0d7c30c794c9198bd7e866362cd4bc32eeed91/API.md#addvalidationvalidation-)

akash1810:
This is a preparation step for (an attempt) to auto-generate a `riff-raff.yaml` file. The next step is to remove the [`GuAppAwareConstruct` mixin](https://github.com/guardian/cdk/blob/a8f7692928ba7a1691aabb8260ef3777adcae16c/src/utils/mixin/app-aware-construct.ts) with an actual `Construct` on the [tree](https://docs.aws.amazon.com/cdk/v2/guide/constructs.html#constructs_tree).

src/constructs/core/stack.ts
```diff
@@ -138,6 +130,24 @@ export class GuStack extends Stack implements StackStageIdentity, GuMigratingSta
...
+      validate(): string[] {
+        const NO_ERRORS: string[] = [];
+
+        const isMigrating = !!node.findAll().find((construct: IConstruct) => construct instanceof CfnInclude);
+
+        if (isMigrating) {
+          Annotations.of(node.root).addWarning(
```
nicl, line 143: How does this warning get surfaced? Is it to the console?

nicl (commented):
I really like the improved simplicity here! I wonder if we can take it even further?

* remove even the `WithStaticLogicalId` mixin
* only have the `overrideLogicalId` helper
* only test the helper once in a single unit test

This would allow us to remove code from most of these 40 odd files.

The helper function is not cumbersome to use:

```
 this.overrideLogicalId(sqs, {
      logicalId: "Notifications",
      reason:
        "Persisting this logicalId from the YAML template as the resource is stateful and used by multiple downstream systems that we don't own.",
    });
```

especially given that only a few resources typically need their IDs overridden.

So I'm not sure the mixin and resulting code across so many files is worth it.

The many unit tests here (pre-existing I know) also feel redundant; it's not clear they add value over a single unit test of a simple construct.

akash1810:
> So I'm not sure the mixin and resulting code across so many files is worth it.

We've discussed this IRL and have agreement! Having the user call `overrideLogicalId` is a lot more explicit than passing in some props, and it also means less code for GuCDK.

There is an unknown, however - patterns create multiple resources; we'd need a way to expose these for use with `overrideLogicalId`. But how?

Regarding less code for GuCDK, I suspect it'll be a pretty dramatic reduction as we'd be able to remove all the constructs that simply add the stateful mixins to the AWS CDK construct, for example [`GuKinesisStream`](https://github.com/guardian/cdk/blob/a8f7692928ba7a1691aabb8260ef3777adcae16c/src/constructs/kinesis/kinesis-stream.ts). I'd propose we make these changes separately as the diff on this PR would become unwieldy; maybe we could use stacked PRs so that we only publish a update.

akash1810:
Closing and replacing with:
  - #1266
  - #1278
  - #1286

src/constructs/core/stack.ts
akash1810: > How does this warning get surfaced? Is it to the console?

Yes. Screenshot added in #1286.

PR #1260, opened 2022-05-11
feat: Enable OIDC on load balancer

andrew-nowak:
Co-authored-by: @tjmw
Co-authored-by: @philmcmahon 

## What does this change?

Adds OIDC configuration as an option to EC2 apps. This allows us to avoid implementing the OIDC flow, and making that an AWS concern.

Only currently developed and tested against a development Okta account.

## Checklist

### DIFF
--- a/src/patterns/ec2-app/base.ts
+++ b/src/patterns/ec2-app/base.ts
@@ -1,9 +1,9 @@
-import { Duration, Tags } from "aws-cdk-lib";
+import { Duration, SecretValue, Tags } from "aws-cdk-lib";
 import type { BlockDevice } from "aws-cdk-lib/aws-autoscaling";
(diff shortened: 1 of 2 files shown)


src/patterns/ec2-app/base.ts
```diff
@@ -104,6 +104,13 @@ export interface GuEc2AppProps extends AppIdentity {
   scaling: GuAsgCapacity;
   certificateProps: GuDomainName;
   withoutImdsv2?: boolean;
+  auth?: GuAuth;
+}
+
+// TODO find better home
+interface GuAuth {
+  oktaServer: string;
+  oktaClientId: string;
```
nicl, line 113: Is this stuff specific to the app or universal?

PR #1265, opened 2022-05-16
feat: change the interface for existingKinesisStream

marjisound:
## What does this change?
This PR is a change on top of the branch `feat-cross-account-kinesis-lambda` and after merge will be part of https://github.com/guardian/cdk/pull/1264

### DIFF
--- a/src/patterns/kinesis-lambda.ts
+++ b/src/patterns/kinesis-lambda.ts
@@ -31,16 +31,11 @@ export interface CrossAccountKinesisStream {
 /**
  * Used to provide information about an existing Kinesis stream to the [[`GuKinesisLambda`]] pattern.
  *
- * Specify a `existingLogicalId` to inherit a Kinesis stream which has already
- * been created via a CloudFormation stack. This is necessary to avoid data loss and interruptions of
- * service when migrating stacks from CloudFormation to `cdk`.
  *
- * Specify an `externalKinesisStreamName` to link the lambda to a Kinesis stream owned by a different stack
- * on the same aws account (or created outside of version control).
+ * Specify an `streamArn` to link the lambda to an existing Kinesis stream owned by a different stack
+ * (or created outside of version control).
  *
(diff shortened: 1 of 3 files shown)


src/patterns/kinesis-lambda.ts
```diff
@@ -49,29 +44,29 @@ export interface CrossAccountKinesisStream {
...
  * }
  * ```
  */
 export interface ExistingKinesisStream extends GuMigratingResource {
-  externalKinesisStreamName?: string;
-  crossAccountKinesisStream?: CrossAccountKinesisStream;
+  /**
+   * kinesis stream Arn that is owned by a different stack
+   */
+  streamArn: string;
```
akash1810, line 65: Following @jacobwinch 's [comments](https://github.com/guardian/cdk/pull/1264#discussion_r872379592), could this be a [`GuArnParameter`](https://github.com/guardian/cdk/blob/a8f7692928ba7a1691aabb8260ef3777adcae16c/src/constructs/core/parameters/base.ts#L35-L43) type? We'll then be using the type system to avoid adding private/secret information into VCS.

Parameter's have a [`valueAsString` function](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.CfnParameter.html#valueasstring) that can be used to read the value when needed.

I think similar applies to `roleArn` below too?

marjisound: Sorry @akash1810 yesterday when @jacobwinch and I were trying to deploy the lambda that was using the cdk changes for kinesis lambda, we realised that  there's a limitation in AWS that doesn't allow a lambda to be triggered by a cross account kinesis stream 😢 

So we have to put the lambda in the same account as the kinesis stream meaning all of these changes we were making on the cdk project are not needed at all. But thanks a lot for looking into the changes and helping out 🙏

marjisound:
This PR and it's parent PR https://github.com/guardian/cdk/pull/1264 are not needed anymore because there's a limitation in AWS that doesn't allow lambda being triggered by a cross account kinesis stream

PR #1266, opened 2022-05-16
feat: Add `overrideLogicalId` function to `GuStack`

akash1810:
## What does this change?

This change includes the "public facing" parts of #1259 as they are generally useful. The internal parts of the library remain unchanged as we continue to discuss approaches on this.

In this change we add a new `overrideLogicalId` function to `GuStack` which can be used to simplify the mechanism of retaining stateful resources when moving from a JSON/YAML template into GuCDK.

This change also moves to deprecate the `GuMigratableConstruct` and `GuStatefulMigratableConstruct` mixins, in favour of this new function. These mixins aren't used outside of the repository, so maybe we should make this change when updating the library internals?

## How to test

See added tests.

An improved DX? See the updated docs for specific examples of how.

## Checklist


nicl (approved):
💯

PR #1275, opened 2022-05-16
feat(project-generator): Allow user to set package manager

akash1810:
## What does this change?

For better or for worse, we do not have a departmental standard when it comes to Node package managers. In this change, we add a mandatory `package-manager` flag to the `new` command of the CLI with options of `npm` or `yarn`. This is after seeing some friction in teams efforts to adopt GuCDK.

We currently use `yarn`. For repositories using `npm`, teams either commit to a somewhat Frankenstein build process, or spend time unpicking unwanted mess.

The addition of a `package-manager` flag should reduce this friction.

The consequence of this change is a step away from the `script/task` pattern.

## How to test

Our integration test, that runs in CI, testing the `new` CLI command has been updated. That is, a successful build should be enough.

Less friction when adding GuCDK to a repository.

- For those unfamiliar with Node package managers, the choice of which to use is no longer obvious?
- What about `pnpm` and others?

## Checklist


src/bin/index.ts
```diff
@@ -76,6 +77,11 @@ const parseCommandLineArguments = () => {
               type: "array",
               demandOption: true,
             })
+            .option("package-manager", {
+              description: "The Node package manager to use. Match this to your repository. If uncertain, prefer npm.",
```
akash1810, line 81: Opinion purely based on a [previous experience](https://github.com/guardian/cdk-playground/pull/1).

src/bin/commands/new-project/index.ts
```diff
@@ -156,11 +170,13 @@ export const newCdkProject = async (props: NewProjectProps): CliCommandResponse
 
   console.log("Project successfully created! Next steps:");
```
jacobwinch, line 171: I know that this PR isn't updating the logic, but I think the next steps here should only be logged for migrations i.e. `if (config.yamlTemplateLocation)`?

src/bin/index.ts
jacobwinch: I wonder if we should tell people how to check which package manager is in use (i.e. search for `package-lock.json` vs `yarn.lock`), just to help those who are completely unfamiliar with the Node ecosystem?

jacobwinch: Very minor point, but I also wonder if we should slightly rephrase `If uncertain, prefer npm`.

How about something like: `If the repository is not currently using either package manager and there is no strong convention in your team, we recommend `npm`?

src/bin/commands/new-project/index.ts
akash1810: I think we can simplify this even further by simply saying "Project successfully created! For next steps, see the docs", as it keeps the code super simple, and allows us to update the next steps w/out a new release.

jacobwinch (dismissed):
Great work 💯

src/bin/index.ts
akash1810: Yes, great improvement!

jacobwinch:
> * What about `pnpm` and others?

😱

src/bin/commands/new-project/index.ts
akash1810: Will make these changes on this PR.

jacobwinch:
cc @JamieB-gu - I think you've requested this feature before 😀

src/bin/index.ts
akash1810: Done.

src/bin/commands/new-project/index.ts
akash1810: @jacobwinch I've added a new commit for this now.

docs/setting-up-a-gucdk-project.md
```diff
@@ -0,0 +1,55 @@
+# Setting up a GuCDK project
```
jacobwinch, line 1: These docs are great!

src/bin/commands/new-project/index.ts
```diff
@@ -144,23 +150,26 @@ export const newCdkProject = async (props: NewProjectProps): CliCommandResponse
...
 
-  console.log("Project successfully created! Next steps:");
-
-  [
-    "Run ./script/diff to confirm there are no destructive changes (there should only be tag additions)",
-    "Update the repository's CI configuration to run ./script/ci",
-    "Raise a PR with these changes",
-  ].map((step) => console.log(`  - ${step}`));
+  const docsUrl = "https://github.com/guardian/cdk/tree/main/docs";
+  CliUx.ux.url(chalk.green(`Please see the docs (${docsUrl}) for next steps.`), docsUrl);
```
akash1810, line 172: This is a bit fancy 💅🏽 

image

src/bin/commands/new-project/utils/init.ts
```diff
@@ -79,7 +78,7 @@ function createPackageJson(outputDirectory: string): void {
     "jest",
     "prettier",
     "ts-jest",
-    "ts-mode",
+    "ts-node",
```
akash1810, line 81: Correcting a typo.

src/bin/index.ts
JamieB-gu: 👍 great suggestions Jacob!

Perhaps another reason for suggesting `npm` is that there's probably a greater chance of it being installed in various environments if you already have Node installed? I think they're often bundled together 🤔. This may be useful for projects where there aren't existing conventions or infrastructure, e.g. Scala projects?

src/bin/index.ts
```diff
@@ -106,9 +113,16 @@ parseCommandLineArguments()
...
+        const { init, app, stack, yamlTemplateLocation, stage, packageManager } = argv;
         const stages = stage.map((_) => (_ as string).toUpperCase());
-        return newCdkProject({ init, app, stack, yamlTemplateLocation, stages });
+        return newCdkProject({
+          init,
+          app,
+          stack,
+          yamlTemplateLocation,
+          stages,
+          packageManager: packageManager as PackageManager,
```
JamieB-gu, line 124: Am I right in thinking that this is a type-cast and there's no validation for this field like there is for, e.g., some of the paths? So if someone mis-spelled the package manager, or actually even typed `pnpm`, that string would make its way through to the `runTask` function?

JamieB-gu (commented):
This is great, thanks for doing this! 🚀

I have a couple of questions but feel free to ignore, I'm not very familiar with this project 🙈

src/bin/index.ts
akash1810: Yep you're correct. We're using [yargs's choices](https://yargs.js.org/docs/#api-reference-choiceskey-choices) to ensure the value of `packageManager` is correct by this point.

docs/setting-up-a-gucdk-project.md
```diff
@@ -0,0 +1,55 @@
...
+This guide describes the steps to set up a GuCDK project in your repository, integrating it with CI and CD.
+
+Requirements:
+  - A choice of Node package manager. We support `npm` and `yarn`.
+    Match this to the repository. A `package-lock.json` file means `npm` is used, a `yarn.lock` file means `yarn` is used.
+    If the repository has neither file, and there is no strong convention in your team, we recommend npm.
+  - Familiarity with the repository's CI process.
+  - Use of a recent version of Node. We recommend the latest LTS version.
```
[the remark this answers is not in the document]
akash1810: > Would it be useful to link to this page for people unfamiliar with Node's LTS releases? https://nodejs.org/en/about/releases/

Yes, great point.

> More broadly, do we want to provide instructions to help people set up Node/npm in their projects, or is that beyond the scope of these docs? How to install locally and how to get it set up in CI and that kind of thing?

I think this is out of scope for this doc.

src/bin/index.ts
JamieB-gu: Oh that's a nice feature. Ok so that'll catch it before it gets to this point 👍

jacobwinch (approved):

docs/migration-guide.md
```diff
@@ -1,60 +1,36 @@
...
+```sh
+npm run diff
+```
 
 You might find that the diff contains a lot of noise due to tags that are automatically added by `@guardian/cdk`.
-Tag changes will not cause disruption to your infrastructure. 
+Tag changes will not cause disruption to your infrastructure.
 If desired, you can view a diff which excludes the tagging changes by temporarily [amending your stack's props](https://github.com/guardian/cdk-playground/blob/26af40f6432cb94b3e5ef34648d422a7a6ee6b33/cdk/bin/cdk.ts#L6).
 
 Once you are happy with the diff, we recommend running CDK synth as part of your CI process.
```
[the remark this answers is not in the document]
akash1810: Yep! Done.

PR #1276, opened 2022-05-17
chore(deps): Update aws-cdk to 2.23.0

jacobwinch:
Small follow up to https://github.com/guardian/cdk/pull/1251.

Whilst working on https://github.com/guardian/amiable/pull/138, I copy & pasted the versions from peer dependencies without thinking about it. This mismatch leads to the following error when running `cdk synth`:

`This CDK CLI is not compatible with the CDK library used by your application. Please upgrade the CLI to the latest version.`

This PR should help others to avoid this problem.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -51,7 +51,7 @@
         "typescript": "~4.6.4"
       },
       "peerDependencies": {
-        "aws-cdk": "2.20.0",
+        "aws-cdk": "2.23.0",
         "aws-cdk-lib": "2.23.0",
         "constructs": "10.1.5"
       }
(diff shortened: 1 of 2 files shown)


jacobwinch:
> suggesting the script broke when we moved to V2?

Yes, I think you're right.

The command looks OK: https://github.com/guardian/cdk/blob/6299a8a7c03edf0453519f4c3ab8187fe68628ce/package.json#L27

My best guess is that it doesn't work correctly because `aws-cdk` is a peer dependency only: https://github.com/guardian/cdk/blob/6299a8a7c03edf0453519f4c3ab8187fe68628ce/package.json#L71

Because it's [working OK with the integration tests, where it's a `devDependency`](https://github.com/guardian/cdk/pull/1251/files#diff-500a7d76ad7140b46017620ad0b0bb2d0b7a3bf6886090bfc0029629f9e7e581).

jacobwinch:
> maybe add `aws-cdk` as a dev dependency to trick `ncu`?

Yeah it's a bit of a hack, but I think it's fairly harmless!

PR #1278, opened 2022-05-18
fix!: Remove `migratedFromCloudFormation` prop on `GuStack`

akash1810:
This is a replacement of #1259 following #1266, specifically addressing https://github.com/guardian/cdk/pull/1259#issuecomment-1126307740.

## What does this change?

In this change the following (internal parts of the repository) are removed, having been previously marked as deprecated in #1266:
  - `GuMigratableConstruct` mixin
  - `GuStatefulMigratableConstruct` mixin

This is to aid the removal of the redundant `migratedFromCloudFormation` prop on `GuStack`.

This also means the following (internal parts of the repository) are unused and therefore removed:
  - `GuMigratingResource.setLogicalId` function
  - `GuStatefulConstruct` interface
  - `GuMigratingResource` interface

As these changes are internal facing, and to previously deprecated areas of the repository, I do _not_ think this is a breaking change. Whilst there are some changes made to the patterns, these patterns are still in "alpha"(#1293).

### Out of scope for this PR
  - Removal of constructs now simply acting as aliases for their AWS CDK equivalent, such as `GuKinesisStream`. These will be removed in a future PR.

## How to test

The existing (/remaining) unit tests should suffice.

A reduction of complexity in the repository and for consumers of the library.

The main risk is those resources that are keeping their logical ID via the behaviour offered by the mixins and `GuMigratingResource` interface will stop doing so. This risk will be caught by the compiler, so I'm not that concerned.

## Checklist


nicl (approved):
🥳 so much code gone! :) great work!

src/constructs/kinesis/kinesis-stream.ts
```diff
@@ -22,7 +21,7 @@ export interface GuKinesisStreamProps extends StreamProps, GuMigratingResource {
  * new GuKinesisStream(stack, "LoggingStream", { existingLogicalId: "MyCloudFormedKinesisStream" });
  * ```
  */
-export class GuKinesisStream extends GuStatefulMigratableConstruct(Stream) {
+export class GuKinesisStream extends Stream {
```
akash1810, line 24: This construct is now a pure alias to the AWS one and isn't needed. I'd suggest we remove it, but not sure if it's correct to do here or in it's own PR?

src/constructs/iam/roles/roles.ts
```diff
@@ -1,11 +1,10 @@
...
 import type { RoleProps } from "aws-cdk-lib/aws-iam";
-import { GuMigratableConstruct } from "../../../utils/mixin";
-import type { GuMigratingResource, GuStack } from "../../core";
+import type { GuStack } from "../../core";
 
-export interface GuRoleProps extends RoleProps, GuMigratingResource {}
+export type GuRoleProps = RoleProps;
 
-export class GuRole extends GuMigratableConstruct(Role) {
+export class GuRole extends Role {
```
akash1810, line 7: This construct is now a pure alias to the AWS one and isn't needed. I'd suggest we remove it, but not sure if it's correct to do here or in it's own PR?

src/constructs/sns/sns-topic.ts
```diff
@@ -21,7 +20,7 @@ interface GuSnsTopicProps extends TopicProps, GuMigratingResource {}
  *  new GuSnsTopic(stack, "SnsTopic", { existingLogicalId: "MyCloudFormedSnsTopic" });
  * ```
  */
-export class GuSnsTopic extends GuStatefulMigratableConstruct(Topic) {
+export class GuSnsTopic extends Topic {
```
akash1810, line 23: This construct is now a pure alias to the AWS one and isn't needed. I'd suggest we remove it, but not sure if it's correct to do here or in it's own PR?

src/patterns/kinesis-lambda.ts
```diff
@@ -40,8 +39,8 @@ import type { StreamErrorHandlingProps, StreamProcessingProps } from "../utils/l
  * existingKinesisStream: { externalKinesisStreamName: "KinesisStreamFromAnotherStack" }
  * ```
  */
-export interface ExistingKinesisStream extends GuMigratingResource {
-  externalKinesisStreamName?: string;
+export interface ExistingKinesisStream {
```
akash1810, line 42: By dropping `GuMigratingResource` we're pushing the responsibility of persisting a logical ID for the kinesis stream to the user. In order to achieve this, we need to make the stream available (see the `public readonly` change below).

src/patterns/sns-lambda.ts
```diff
@@ -35,8 +36,8 @@ import { GuSnsTopic } from "../constructs/sns";
  *  existingSnsTopic: { externalTopicName: "MySnsTopicNameFromAnotherStack" }
  * ```
  */
-export interface ExistingSnsTopic extends GuMigratingResource {
-  externalTopicName?: string;
+export interface ExistingSnsTopic {
```
akash1810, line 39: By dropping `GuMigratingResource` we're pushing the responsibility of persisting a logical ID for the SNS topic to the user. In order to achieve this, we need to make the stream available (see the `public readonly` change below).

src/constructs/iam/roles/roles.ts
jacobwinch: I think removing all constructs in this state in a single follow-up PR is a good way to break this change down

src/constructs/sns/sns-topic.ts
```diff
@@ -18,10 +16,16 @@ interface GuSnsTopicProps extends TopicProps, GuMigratingResource {}
  * ```
  * Inherit the SNS topic (rather than creating a new one) using:
  * ```typescript
- *  new GuSnsTopic(stack, "SnsTopic", { existingLogicalId: "MyCloudFormedSnsTopic" });
+ * class MyStack extends GuStack {
+ *   constructor(app: App, id: string, props: GuStackProps) {
+ *     super(app, id, props);
+ *     const topic = new GuSnsTopic(stack, "SnsTopic");
+ *     this.overrideLogicalId(topic, { logicalId: "MyCloudFormedKinesisStream", reason: "Retaining a resource from a YAML template" })
```
jacobwinch, line 23:
```suggestion
 *     this.overrideLogicalId(topic, { logicalId: "MyCloudFormedSnsTopic", reason: "Retaining a resource from a YAML template" })
```

src/patterns/kinesis-lambda.ts
jacobwinch: This interface was previously dealing with 2 cases (but I think I did a bad job with the semantics!):

1. `existingKinesisStream` i.e. it belongs to _this_ CFN stack, but you had created it before you started using this pattern
2. `externalKinesisStream` i.e. it belongs to _another_ CFN stack 

Now that 1 has been removed, it feels a bit confusing to have the prop called `externalKinesisStreamName` (since it might not be external).

The good news here is that I don't think this pattern has many users yet (I think Journalism Dev were the first people to pick this up and that was only in the last fortnight or so), so I think we can break this API without causing too many problems.

Happy to chat about this one in person as I feel like I didn't get the API quite right in the first instance and it'd be nice to make this clearer at the second attempt!

src/patterns/sns-lambda.test.ts
```diff
@@ -21,19 +21,18 @@ describe("The GuSnsLambda pattern", () => {
   });
 
   it("should inherit an existing SNS topic correctly if an existingLogicalId is passed via existingSnsTopic in a migrating stack", () => {
```
jacobwinch, line 23: Minor: I think we just need to update this test name

src/patterns/sns-lambda.ts
jacobwinch: I think this is following the similar (slightly questionable) API that the `GuKinesisLambda` was using, so we might want to try to improve both simultaneously?

jacobwinch:
I think this new approach is a huge improvement 💯 

Overall this looks great, but I think it might be worth a quick review of the `GuSnsLambda` and `GuKinesisLambda` patterns, as this feels like a good opportunity to make them a bit more intuitive!

src/patterns/kinesis-lambda.ts
jacobwinch: I think we can defer this in response to https://github.com/guardian/cdk/pull/1293, so will resolve this thread (i.e. no action needed here, IMO!).

src/patterns/sns-lambda.ts
jacobwinch: I think we can defer this in response to https://github.com/guardian/cdk/pull/1293, so will resolve this thread (i.e. no action needed here, IMO!).

src/constructs/loadbalancing/elb.ts
```diff
@@ -44,8 +40,11 @@ interface GuClassicLoadBalancerProps extends Omit<LoadBalancerProps, "healthChec
  * ```
  * If you are running an application which only accepts traffic over HTTPs, consider using [[`GuHttpsClassicLoadBalancer`]]
  * to reduce the amount of boilerplate needed when configuring your load balancer.
+ *
+ * This resource is stateful.
+ * @see https://github.com/guardian/cdk/blob/main/docs/stateful-resources.md
  */
-export class GuClassicLoadBalancer extends GuStatefulMigratableConstruct(GuAppAwareConstruct(LoadBalancer)) {
+export class GuClassicLoadBalancer extends GuAppAwareConstruct(LoadBalancer) {
```
jacobwinch, line 47: Also one for another PR... I wonder if we should remove this construct entirely?

jacobwinch:
> I do not think this is a breaking change

Hmm, I'm not sure that I agree with this. I thought that upgrading to this version without following the advice in https://guardian.github.io/cdk/classes/constructs_core.GuStack.html#overrideLogicalId, could cause some logical ids to change? 

At best, I thought this would cause snapshot tests to fail and at worst it could lead to accidental resource replacement.

> The main risk is those resources that are keeping their logical ID via the behaviour offered by the mixins and GuMigratingResource interface will stop doing so. This risk will be caught by the compiler, so I'm not that concerned.

This implies that I might have misunderstood though (since I thought it would compile but fail the snapshot tests) - would you mind elaborating a little?

akash1810:
> > The main risk is those resources that are keeping their logical ID via the behaviour offered by the mixins and GuMigratingResource interface will stop doing so. This risk will be caught by the compiler, so I'm not that concerned.
> 
> This implies that I might have misunderstood though (since I thought it would compile but fail the snapshot tests) - would you mind elaborating a little?

(This applies to all constructs/patterns) We currently have a `GuKinesisStream` that accepts props of this shape:

```ts
interface GuKinesisStreamProps extends StreamProps, GuMigratingResource { }
```

If we expand(?) `GuMigratingResource`, this results in something like this:

```ts
interface GuKinesisStreamProps extends StreamProps {
  existingLogicalId?: {
    logicalId: string;
    reason: string;
  }
}
```

The changes in this PR mean `GuKinesisStreamProps` no longer extends `GuMigratingResource`, thus no longer has `existingLogicalId`.

This means whilst this is valid on `main` today:

```ts
const props: GuKinesisStreamProps = {
  existingLogicalId: {
    logicalId: "MyStream",
    reason: "demo"
  }
}
```

It is no longer valid on this branch ([demo](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgLLAOZTmUHkDeAsAFDLIQAewAzriBgDID2GwCcANgJIAmA-AC5CpcuU6t2XPsLpQ8AblFioEODWYhZYeQyVlkAX1LGSpUJFiIUAYU1yArgjAAFKMwAONCpUghe3uhYOHgiBiBwALYQ2roY+qbm4NDwSMgAchAA7nYgjq7uXmHkEdGxiiakpAj2YMiaMci5js5unt4AvMXIpY0ARHB9ADTKVLT0TJIcPLzCxAbiU9KzyH0AjMPK5KrqmsJ9AEx9yqaJJDV5dWBZzMKZObVQDgXtyF3zJVH9gyMGY3R4FhsaYybqLYHLfYbX5iZA7DRaVZHE6VMznWrIMAAC1UKHeyl6+x+o2oAIYQKkMzmW2QEghVNW0Jp8L2SOOBlOQA)). However, if one is not explicit about the type of `props`, it will be valid, but ignored when used. In this case, we'll rely on snapshot tests being run in CI to catch issues.

jacobwinch:
> It is no longer valid (fails to compile) on this branch ([demo](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgLLAOZTmUHkDeAsAFDLIQAewAzriBgDID2GwCcANgJIAmA-AC5CpcuU6t2XPsLpQ8AblFioEODWYhZYeQyVlkAX1LGSpUJFiIUAYU1yArgjAAFKMwAONCpUghe3uhYOHgiBiBwALYQ2roY+qbm4NDwSMgAchAA7nYgjq7uXmHkEdGxiiakpAj2YMiaMci5js5unt4AvMXIpY0ARHB9ADTKVLT0TJIcPLzCxAbiU9KzyH0AjMPK5KrqmsJ9AEx9yqaJJDV5dWBZzMKZObVQDgXtyF3zJVH9gyMGY3R4FhsaYybqLYHLfYbX5iZA7DRaVZHE6VMznWrIMAAC1UKHeyl6+x+o2oAIYQKkMzmW2QEghVNW0Jp8L2SOOBlOQA)). However, if one is not explicit about the type of `props`, it will be valid, but ignored when used. In this case, we'll rely on snapshot tests being run in CI to catch issues.

Thanks for explaining this so thoroughly, I found it very helpful! 

I would argue that this is a breaking change either way then, because a) your code that used to compile won't compile anymore or b) your tests that used to pass will now fail! I guess we can get a better agreement on exactly what we consider to be a breaking change when we play: https://trello.com/c/SVmKOUpI/1281-draft-a-proposal-for-our-approach-to-making-breaking-changes-to-gucdk.

Anyway, the change itself looks great, so don't let this digression prevent you from merging!

akash1810:
> I would argue that this is a breaking change either way then, because a) your code that used to compile won't compile anymore or b) your tests that used to pass will now fail! I guess we can get a better agreement on exactly what we consider to be a breaking change when we play: https://trello.com/c/SVmKOUpI/1281-draft-a-proposal-for-our-approach-to-making-breaking-changes-to-gucdk.

Fair. It's definitely safest to release a new major, and I agree we can be more regimented after playing that card.

I'll opt to squash this branch with the commit of:

```
fix!: Remove `migratedFromCloudFormation` prop on `GuStack`


BREAKING CHANGE: The `migratedFromCloudFormation` prop on `GuStack` has been removed.

The `migratedFromCloudFormation` prop on `GuStack` was used in the logic deciding when a resource should persist it's logical ID.

Since #1266, this is no longer required as we should call the `overrideLogicalId` function on `GuStack` instead.

This change also means we remove some internal parts of the library:
  - The `GuMigratableConstruct` mixin
  - The `GuStatefulMigratableConstruct` mixin
  - The `GuMigratingResource.setLogicalId` function
  - The `GuStatefulConstruct` interface
  - The `GuMigratingResource` interface

The removal of `GuMigratingResource` means the props for a few constructs have changed shape, notably the removal of the optional `existingLogicalId` property.

Please refer to [the docs](https://github.com/guardian/cdk/blob/main/docs/stateful-resources.md).
```

PR #1286, opened 2022-05-19
feat: Add warning for stateful resources when migrating to GuCDK

akash1810:
## What does this change?

A `CfnInclude` construct is added to a stack when [migrating from a YAML/JSON template](https://docs.aws.amazon.com/cdk/v2/guide/use_cfn_template.html).

In this change, we produce a warning to take care over any stateful resources, and suggest use of the `overrideLogicalId` function to ensure such resources are not removed unintentionally.

This becomes useful for the proposed changes in #1278 which seeks to drop our [current messaging](https://github.com/guardian/cdk/blob/abc237fff54a0efbffa5a3733182c1cb738fff4a/src/constructs/core/migrating.ts#L35-L73). This change feels useful in isolation of #1278, hence it being it's own PR.

## How to test

This has been beta released to 43.4.0-beta.2, so to test:
  - Use GuCDK at version 43.4.0-beta.2 in a project that has a `CfnInclude` construct
  - Synth the stack and observe the output on the console

I've done this on [RIff-Raff](https://github.com/guardian/deploy-tools-platform/blob/abcd7c6049e2f5870c7910266a4aeb008aca6b21/cdk/lib/riff-raff.ts) and the result is seen in yellow below:

image

More warnings to take care of stateful resources.

Is the warning visible/obvious enough to be seen?

## Checklist

### DIFF
--- a/src/constructs/core/__data__/cfn.yaml
+++ b/src/constructs/core/__data__/cfn.yaml
@@ -1,0 +1,10 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Description: A CloudFormation template for testing purposes
+Parameters:
+  BucketName:
+    Type: String
+Resources:
+  ImportantBucket:
+    Type: AWS::S3::Bucket
+    Properties:
+      BucketName: !Ref BucketName
--- a/src/constructs/core/stack.ts
+++ b/src/constructs/core/stack.ts
@@ -1,9 +1,10 @@
 import { Annotations, App, LegacyStackSynthesizer, Stack, Tags } from "aws-cdk-lib";
 import type { CfnElement, StackProps } from "aws-cdk-lib";
 import { Template } from "aws-cdk-lib/assertions";
+import { CfnInclude } from "aws-cdk-lib/cloudformation-include";
 import type { IConstruct } from "constructs";
 import gitUrlParse from "git-url-parse";
-import { ContextKeys, TagKeys, TrackingTag } from "../../constants";
+import { ContextKeys, LibraryInfo, TagKeys, TrackingTag } from "../../constants";
 import type { GuMigratingStack } from "../../types";
 import { gitRemoteOriginUrl } from "../../utils/git";
 import type { StackStageIdentity } from "./identity";
@@ -140,6 +141,25 @@ export class GuStack extends Stack implements StackStageIdentity, GuMigratingSta
 
       this.tryAddRepositoryTag();
     }
+
+    const { node } = this;
+
+    node.addValidation({
+      validate(): string[] {
+        const NO_ERRORS: string[] = [];
+
+        node
+          .findAll()
+          .filter((construct) => construct instanceof CfnInclude)
+          .map((cfnInclude) => {
+            Annotations.of(cfnInclude).addWarning(
+              `As you're migrating a YAML/JSON template to ${LibraryInfo.NAME}, be sure to check for any stateful resources! See https://github.com/guardian/cdk/blob/main/docs/stateful-resources.md.`
+            );
+          });
+
+        return NO_ERRORS;
+      },
+    });
   }
 
   /**
--- a/src/constructs/core/stack.test.ts
+++ b/src/constructs/core/stack.test.ts
@@ -1,8 +1,10 @@
-import { App } from "aws-cdk-lib";
+import { join } from "path";
+import { Annotations, App } from "aws-cdk-lib";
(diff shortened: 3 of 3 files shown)


src/constructs/core/__data__/cfn.yaml
```diff
@@ -0,0 +1,10 @@
+AWSTemplateFormatVersion: '2010-09-09'
```
akash1810, line 1: `CfnInclude` will throw if it cannot find the YAML file to include in a CDK stack.

This file won't make it into the published module.

src/constructs/core/stack.ts
```diff
@@ -140,6 +141,25 @@ export class GuStack extends Stack implements StackStageIdentity, GuMigratingSta
 
       this.tryAddRepositoryTag();
     }
+
+    const { node } = this;
+
+    node.addValidation({
```
akash1810, line 147: The validation step happens just before synthesis:

image

See:
  - [App lifecycle](https://docs.aws.amazon.com/cdk/v2/guide/apps.html#lifecycle)
  - The [`constructs` docs](https://github.com/aws/constructs/blob/bd0d7c30c794c9198bd7e866362cd4bc32eeed91/API.md#addvalidationvalidation-)

src/constructs/core/stack.ts
```diff
@@ -140,6 +141,25 @@ export class GuStack extends Stack implements StackStageIdentity, GuMigratingSta
...
+      validate(): string[] {
+        const NO_ERRORS: string[] = [];
+
+        node
+          .findAll()
+          .filter((construct) => construct instanceof CfnInclude)
+          .map((cfnInclude) => {
+            Annotations.of(cfnInclude).addWarning(
+              `As you're migrating a YAML/JSON template to ${LibraryInfo.NAME}, be sure to check for any stateful resources! Use 'overrideLogicalId' to preserve a resource's logical ID, ensuring it's not removed unintentionally.`
```
jacobwinch, line 156: I wonder if we can be a little clearer here about exactly when this is required? In an ideal world we would scan the list of resources and give the logical ids that we think might be at risk, but I suspect that will be too much effort vs value provided. Perhaps a good compromise would be to provide some examples of stateful resources that people should look to retain (e.g. databases, DNS records, S3 buckets).

On the other hand, we don't want to make people feel nervous unnecessarily! For example, an ALB could be considered a stateful resource but our migration guides actually suggest replacing this (once the new one is serving all traffic).

It's not easy to capture all of this in a short warning message and I think there is also some jargon here that could benefit from further explanation e.g. I suspect that many users will not know what a [logical id](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resources-section-structure.html) is in this context. Consequently, it might be better to link out to a markdown file where we can: explain these nuances, provide a full example of retaining a stateful resource with `overrideLogicalId`, link to relevant AWS docs etc. 

The other advantage of pulling some of this detail into markdown is that we could also link out to it from the migration guide (perhaps from [this section](https://github.com/guardian/cdk/blob/main/docs/migration-guide.md#migrating-resources)), to give users the best possible chance of reading it!

Sorry for the massive comment - what do you think?

src/constructs/core/stack.ts
akash1810: > I think it could also be very useful for ensuring that key best practices are met

I think we could use a combination of this and [`Aspects`](https://aws.amazon.com/blogs/devops/align-with-best-practices-while-creating-infrastructure-using-cdk-aspects/)? I'd like to experiment with placing the [Security Group port 22 check](https://github.com/guardian/cdk/blob/00dfd9d791dba6be96995643ecd439c3441c3095/src/constructs/ec2/security-groups/base.ts#L56-L58) here (in a later PR!) so that all SGs are checked, not just those that extend `GuBaseSecurityGroup`.

src/constructs/core/stack.ts
akash1810: I think we're both on the same page! I'm in the middle of writing a `stateful-resources.md` file! I'll add that to this PR and then update all the doc changes in #1278 (well https://github.com/guardian/cdk/pull/1278/commits/c7f037ad84f3be52cc463673214ffd6ea6819c82) to link to it, just to keep things DRY.

akash1810: Updated this change, linking to the stateful resources doc from #1291.

jacobwinch (approved):
Great work 👍

PR #1290, opened 2022-05-20
docs: Update migration guide with a workflow for stateful resources

akash1810:
## What does this change?

Describing the [red/green refactor](https://blog.cleancoder.com/uncle-bob/2014/12/17/TheCyclesOfTDD.html) cycle of a migration.

## How to test

[Read the update](https://github.com/guardian/cdk/blob/aa-migration/docs/migration-guide.md) - does it make sense?

### DIFF
--- a/docs/migration-guide.md
+++ b/docs/migration-guide.md
@@ -16,48 +16,61 @@ To ensure AWS your account is set up correctly, run:
 
 Then follow the instructions for any errors.
 
-## Initial setup
-Create a new GuCDK project in your repository following [this guide](setting-up-a-gucdk-project.md).
-This will create a new CDK (Typescript) app that simply wraps your existing Cloudformation template.
+## Process
+Generally speaking, we've found this to be a good process to follow when migrating a stack to GuCDK (assumes use of `npm`):
 
-Pay attention to the output of the command as it may describe further manual steps at this point.
-Once done, we recommend running `git diff` to see the full set of changes.
+### Introduce CDK to the repository with CI/CD
+1. Create a [new project](setting-up-a-gucdk-project.md) specifying the `--yaml-template-location` flag.
+2. Check to see what changes will be made to the stack:
 
-Finally, confirm that your new CDK stack is not materially different from your existing stack by running a diff:
+   ```shell
+   npm run diff -- --profile <AWS PROFILE NAME> <STACK ID FROM bin/cdk.ts>
+   ```
 
-```sh
-npm run diff
+   This should only show differences in how resources are tagged; GuCDK will add various tags to _all_ resources in the stack.
+   If you find this initial diff too noisy, you could temporarily exclude these tags by amending your [stack's props](https://guardian.github.io/cdk/interfaces/constructs_core.GuStackProps.html#withoutTags),
+   specifically set `withoutTags` to `true`.
+
+4. Configure CI and CD. See [here](setting-up-a-gucdk-project.md) for more detail.
+5. Raise a PR and merge.
+
+### Migration to CDK
+We now have something like this:
+
+```
+CDK(cfn.yaml) -> cfn.json
 ```
 
-You might find that the diff contains a lot of noise due to tags that are automatically added by `@guardian/cdk`.
-Tag changes will not cause disruption to your infrastructure.
-If desired, you can view a diff which excludes the tagging changes by temporarily [amending your stack's props](https://github.com/guardian/cdk-playground/blob/26af40f6432cb94b3e5ef34648d422a7a6ee6b33/cdk/bin/cdk.ts#L6).
+That is, CDK is merely wrapping an existing template.
 
-Once you are happy with the diff, we recommend running CDK synth as part of your CI process.
-See [here](./setting-up-a-gucdk-project.md) for more detail on CI/CD.
+The end goal is for CDK to do all the work, and remove the YAML template:
 
-## Migrating resources
+```
+CDK -> cfn.json
+```
 
-After initial setup, you have something like this:
+In the previous phase, we also created a [snapshot test](https://docs.aws.amazon.com/cdk/v2/guide/testing.html#testing_snapshot) (`cdk/lib/*.test.ts`),
+which we can treat as the source of truth, representing the exact resources in the stack. We can now begin the migration process.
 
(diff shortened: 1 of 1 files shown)


docs/migration-guide.md
```diff
@@ -16,46 +16,54 @@ To ensure AWS your account is set up correctly, run:
...
 
-## Initial setup
-Create a new GuCDK project in your repository following [this guide](setting-up-a-gucdk-project.md).
-This will create a new CDK (Typescript) app that simply wraps your existing Cloudformation template.
+## Process
+Generally speaking, we've found this to be a good process to follow when migrating a stack to GuCDK (assumes use of `npm`):
 
-Pay attention to the output of the command as it may describe further manual steps at this point.
-Once done, we recommend running `git diff` to see the full set of changes.
+### Introduce CDK to the repository with CI/CD
```
jacobwinch, line 22: This section is really helpful!

docs/migration-guide.md
```diff
@@ -16,46 +16,54 @@ To ensure AWS your account is set up correctly, run:
...
+CDK -> cfn.json
+```
 
-    CDK -> cfn.yaml
+In the previous phase, we also created a [snapshot test](https://docs.aws.amazon.com/cdk/v2/guide/testing.html#testing_snapshot) (`cdk/lib/*.test.ts`),
+which we can treat as the source of truth, and we can now begin the migration process.
 
-To get there, we need to migrate resources defined in the old cfn.yaml template
-into CDK (Typescript).
+Generally, this looks like this:
```
jacobwinch, line 53: Hmm, I don't think this describes the process that we typically use. Most of the migrations that we've carried out so far* have dual-stacked resources. We typically recommend introducing almost all resources in one move (via a pattern) and then shipping a PR, before decommissioning resources in another PR.

The approach described here sounds useful for replacing each resource with a construct; this might be helpful for some migrations but it doesn't feel like something that we apply generally to me.

*These have mostly been EC2 but I'm expecting API & Lambda to follow the same approach.

docs/migration-guide.md
```diff
@@ -16,46 +16,54 @@ To ensure AWS your account is set up correctly, run:
...
-starting point. For example, to integrate with CI/CD.
+That is, CDK is merely wrapping an existing template. The end goal is for CDK to do all the work:
 
-The end goal, though, is for CDK to do all of the work:
+```
+CDK -> cfn.json
+```
 
-    CDK -> cfn.yaml
+In the previous phase, we also created a [snapshot test](https://docs.aws.amazon.com/cdk/v2/guide/testing.html#testing_snapshot) (`cdk/lib/*.test.ts`),
```
jacobwinch, line 50: It's definitely worth mentioning this; I have frequently paired with developers who are not aware of this test (or its purpose) after running the `new` command

docs/migration-guide.md
akash1810: You're right. This process is more tied to the stateful side of the stack. It's certainly the approach @dskamiotis and I are taking migrating Dynamo tables in [presence-indicator](https://github.com/guardian/presence-indicator/pull/126).

jacobwinch: Thanks for sharing, I need to migrate a Dynamo table soon so I'll take a look!

akash1810: Updated now, scoping this list to the stateful resources.

docs/migration-guide.md
```diff
@@ -16,46 +16,57 @@ To ensure AWS your account is set up correctly, run:
...
+CDK -> cfn.json
+```
 
-The end goal, though, is for CDK to do all of the work:
+In the previous phase, we also created a [snapshot test](https://docs.aws.amazon.com/cdk/v2/guide/testing.html#testing_snapshot) (`cdk/lib/*.test.ts`),
+which we can treat as the source of truth, representing the exact resources in the stack, and we can now begin the migration process.
 
-    CDK -> cfn.yaml
+As a first step, it is a good idea to read through the template, understanding which resources are [stateful](stateful-resources.md) or not.
+Pro-tip: Annotate the template with inline comments and get it reviewed by your team - it might transpire that a resource is no longer needed as the feature that added it has been removed.
```
jacobwinch, line 54: I like this idea!

docs/migration-guide.md
```diff
@@ -16,46 +16,57 @@ To ensure AWS your account is set up correctly, run:
...
 
-To get there, we need to migrate resources defined in the old cfn.yaml template
-into CDK (Typescript).
+For the stateful resources:
```
jacobwinch, line 56: Thanks for adding these steps, it's very clear!

IMO, the ordering of this guide implies that we want to migrate stateful resources _before_ we migrate the stateless parts (i.e. before we look at the pattern-specific migration guides). Do you think that's the approach we want to take?

Personally I have a few reservations about this because I think you can get most of the benefits of `GuCDK` even if you leave something like a Dynamo table in a CFN file permanently.

I would be tempted to pull the advice re stateless resource migration into its own file and link out to it at the end of the 'pattern' migrations e.g. [here](https://github.com/guardian/cdk/blame/main/docs/migration-guide-ec2.md#L182-L185).

akash1810: How do the latest edits read?

jacobwinch (approved):

docs/migration-guide.md
```diff
@@ -16,48 +16,61 @@ To ensure AWS your account is set up correctly, run:
...
+CDK -> cfn.json
+```
 
-That is, CDK is merely wrapping an existing template. This is useful as a
-starting point. For example, to integrate with CI/CD.
+In the previous phase, we also created a [snapshot test](https://docs.aws.amazon.com/cdk/v2/guide/testing.html#testing_snapshot) (`cdk/lib/*.test.ts`),
+which we can treat as the source of truth, representing the exact resources in the stack. We can now begin the migration process.
 
-The end goal, though, is for CDK to do all of the work:
+**Pro-tip:** Annotate the template with inline comments and get it reviewed by your team - it might transpire that a resource is no longer needed and can be removed, meaning there's less to migrate!
```
kenoir, line 55: This confused me at first, maybe this is a bit clearer.
```md
**Pro-tip:** Annotate the existing YAML template with inline comments as part of the migration and get it reviewed by your team ...
```

PR #1291, opened 2022-05-20
docs: Add documentation on stateful resources

akash1810:
## What does this change?

When migrating a stack's template from YAML to CDK, it's important to take care of stateful resources. This change attempts to document why, impact, and resolutions.

## How to test

Does it [read](https://github.com/guardian/cdk/blob/aa-stateful-resource/docs/stateful-resources.md) well?

### DIFF
--- a/docs/stateful-resources.md
+++ b/docs/stateful-resources.md
@@ -1,0 +1,110 @@
+# Stateful Resources
+We'll define a stateful resource as one that will cause downtime for users of the application if accidentally deleted.
+
+This may include, for example:
+- Databases
+- Messaging streams (Kinesis, SNS, SQS)
+- S3 Buckets
+
+The recommendation is to employ CD via Riff-Raff with your CDK stack. See [here](setting-up-a-gucdk-project.md) for more detail.
+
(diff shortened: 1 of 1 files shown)


docs/stateful-resources.md
```diff
@@ -0,0 +1,104 @@
...
+When wanting to preserve the logical ID for a resource, one can call `overrideLogicalId` on `GuStack`:
+
+```typescript
+class MyStack extends GuStack {
+  constructor(app: App, id: string, props: GuStackProps) {
+    super(app, id, props);
+
+    // resources for my stack
+    const dataBucket = new Bucket(this, "DataBucket");
+    this.overrideLogicalId(dataBucket, { logicalId: "DataBucket", reason: "Retaining a stateful resource previously defined in YAML" });
```
jacobwinch, line 73:
```suggestion
    this.overrideLogicalId(dataBucket, { 
      logicalId: "DataBucket", 
      reason: "Retaining a stateful resource previously defined in YAML" 
   });
```

Sorry to be a linter here, but this just means that it will all display nicely on a standard screen size without needing to scroll.

jacobwinch (approved):
This looks great, it's very clear!

docs/stateful-resources.md
akash1810: Great point! Thanks! I've become so used to the linter running on save 😅 .

PR #1293, opened 2022-05-23
fix: Mark `GuKinesisLambda` and `GuSnsLambda` patterns as alpha

akash1810:
## What does this change?

The shape of the incoming props, specifically around handling of an existing resource, is likely to be changed over time as we understand use-cases. So, annotate these patterns as `alpha`.

See these attempts to update the incoming props:
  - https://github.com/guardian/cdk/pull/1264/files#r872391677
  - https://github.com/guardian/cdk/pull/1265
  - https://github.com/guardian/cdk/pull/1278#discussion_r876660606

Although [TypeDoc](https://typedoc.org/guides/doccomments/) doesn't [support](https://github.com/TypeStrong/typedoc/issues/1381) `@alpha` it does at least standout:

image

[TSDoc](https://tsdoc.org/pages/tags/alpha/) seems to support a wider range of tags, but not sure it's worth moving just yet.

## How to test

n/a

Permission to break the API in future releases 🤞🏽 ?

Across the org, there is only [one use](https://github.com/guardian/google-chat-bots/blob/main/cdk/lib/sns-to-gchat.ts) of `GuSnsLambda`, this is in a DevX owned repo. That is, the risk of having an unstable API isn't really a reality yet.

## Checklist

### DIFF
--- a/src/patterns/kinesis-lambda.ts
+++ b/src/patterns/kinesis-lambda.ts
@@ -95,6 +95,8 @@ export interface GuKinesisLambdaProps extends Omit<GuFunctionProps, "errorPercen
  * This pattern will create a new Kinesis stream by default. If you are migrating a stack from CloudFormation,
  * you will need to opt-out of this behaviour. For information on overriding the default behaviour,
  * see [[`GuKinesisLambdaProps`]].
+ *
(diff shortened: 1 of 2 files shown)


jacobwinch (approved):
👍

PR #1299, opened 2022-05-25
docs: clarify purpose of snapshot tests

johnduffell:
Just to make it clear that step doesn't update the snapshot, but uses it as an expected value


akash1810 (approved):
Thanks!

akash1810:
The `lint-PR` is failing because we want conventional commits and conventional PRs as they're parsed in CD (more detail [here](https://github.com/guardian/cdk/blob/main/docs/contributing.md#releasing)).

If you add the `docs: ` prefix to your PR it should fix it. I can't remember if single commit PRs also need to have a prefix or not...

PR #1313, opened 2022-05-27
feat: Add `parameters` getter on `GuStack`

akash1810:
## What does this change?

This is more correct than `getParam` as it return ALL parameters, not just those of type `GuParameter`. As a result the `getParam` and `setParam` functions on `GuStack` are also removed. (There's an argument that this getter could be internal/private to the library - see below.)

This change also somewhat reduces the utility of the [`GuParameter`](https://github.com/guardian/cdk/blob/90382af4c1939ffa7f5385da2739a0259b5b678b/src/constructs/core/parameters/base.ts#L14-L26) construct. We might be able to simplify (or even remove!) it.

## How to test

CI should continue to pass.

Less code, and a more complete API.

> **Note**
> I propose we ship this as a minor update. The risk is this isn't inline with [semver](https://semver.org/).

Although this change removes the `getParam` and `setParam` functions on `GuStack`:
  - no-one is using them ([`getParam`](https://github.com/search?q=org%3Aguardian+getParam+extension%3Ats&type=Code&ref=advsearch&l=&l=), [`setParam`](https://github.com/search?q=org%3Aguardian+setParam+extension%3Ats&type=Code))
  - they were initially added (#92) to aid the sharing of parameters across constructs. However these kinds of parameters have since become singletons (e.g. [`GuDistributionBucketParameter`](https://github.com/guardian/cdk/blob/2b65959b7aadac5e39f0f691c156146c111d1f55/src/constructs/core/parameters/s3.ts#L6-L28)) which is a much simpler API

Given the above, there might be an argument to make the new `parameters` getter internal to the repository.

## Checklist

### DIFF
--- a/src/constructs/core/stack.ts
+++ b/src/constructs/core/stack.ts
@@ -1,4 +1,4 @@
-import { Annotations, App, LegacyStackSynthesizer, Stack, Tags } from "aws-cdk-lib";
+import { Annotations, App, CfnParameter, LegacyStackSynthesizer, Stack, Tags } from "aws-cdk-lib";
 import type { CfnElement, StackProps } from "aws-cdk-lib";
 import { Template } from "aws-cdk-lib/assertions";
 import { CfnInclude } from "aws-cdk-lib/cloudformation-include";
@@ -8,7 +8,6 @@ import { ContextKeys, LibraryInfo, TagKeys, TrackingTag } from "../../constants"
 import { gitRemoteOriginUrl } from "../../utils/git";
 import type { StackStageIdentity } from "./identity";
 import type { GuStaticLogicalId } from "./migrating";
-import type { GuParameter } from "./parameters";
 
 export interface GuStackProps extends Omit<StackProps, "stackName"> {
   /**
@@ -62,8 +61,6 @@ export class GuStack extends Stack implements StackStageIdentity {
   private readonly _stack: string;
   private readonly _stage: string;
 
-  private params: Map<string, GuParameter>;
-
   get stage(): string {
     return this._stage;
   }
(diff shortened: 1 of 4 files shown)


src/constructs/stack-set/stack-set.ts
```diff
@@ -142,7 +142,9 @@ export class GuStackSet extends CfnStackSet {
     const stackSetInstanceParameters = props.stackSetInstanceParameters ?? {};
 
     const params = Object.keys(stackSetInstanceParameters);
-    const undefinedStackSetParams = props.stackSetInstance.parameterKeys.filter((_) => !params.includes(_));
+    const parameterKeys = Object.keys(props.stackSetInstance.parameters);
+
+    const undefinedStackSetParams = parameterKeys.filter((_) => !params.includes(_));
 
     if (undefinedStackSetParams.length !== 0) {
       throw new Error(`There are undefined stack set parameters: ${undefinedStackSetParams.join(", ")}`);
```
akash1810, line 150: Our current model is to feed stack set instance's parameters from the parent stack, rather than have the stack set instance read from parameter store, as the orchestration of this is tricky.

Here we check that all parameters on the stack set instance also exist on the parent stack, throwing if not.

src/constructs/core/stack.ts
```diff
@@ -88,20 +85,16 @@ export class GuStack extends Stack implements StackStageIdentity {
...
-  }
-
-  get parameterKeys(): string[] {
-    return Array.from(this.params.keys());
+  /**
+   * Returns all the parameters on this stack.
+   *
+   * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html
+   */
+  get parameters(): Record<string, CfnParameter> {
```
akash1810, line 93: Today, this is only used in the `GuStackSet` construct.

jacobwinch:
> The best I can think of is to make it it's own function on an obscure import path, but that'll still be available, albeit annoying.

Sorry if I'm missing something obvious, but could we just move the code into `stack-set.ts`, since that's the only place that it's used at the moment anyway?

> Any objections to leaving as is?

If that's not possible for some reason (or you don't like the idea!) feel free to leave it as is 👍

PR #1315, opened 2022-05-27
feat: Add more hierarchy in the CDK tree with a `GuApp` construct

akash1810:
## What does this change?

At the Guardian, we have a tagging convention as described [here](https://github.com/guardian/recommendations/blob/main/AWS.md). Today, GuCDK adds the `Stack` and `Stage` tags to all resources through the `GuStack` class. Adding the `App` tag is the responsibility of the construct. Usually, this means using the [`GuAppAwareConstrct` mixin](https://github.com/guardian/cdk/blob/2b65959b7aadac5e39f0f691c156146c111d1f55/src/utils/mixin/app-aware-construct.ts), but in some cases [more work](https://github.com/guardian/cdk/blob/2b65959b7aadac5e39f0f691c156146c111d1f55/src/constructs/ecs/ecs-task.ts#L261-L262) is needed.

This is fine, however there are a few limitations:
  - The shape of props incoming to a component needs to extend [`AppIdentity`](https://github.com/guardian/cdk/blob/2b65959b7aadac5e39f0f691c156146c111d1f55/src/constructs/core/identity.ts#L9-L11)
  - It's very much opt in, even though you want it 99% of the time

In this change, we move to solve this with types. Specifically, we:
  - Introduce a `GuApp` class, which:
    - Is a simple `Construct`
    - Applies the `App` tag
  - Alter the linting rule, so that the scope of a construct/class is a `GuApp` rather than a `GuStack`
    - Anything that doesn't belong to an App, should be implemented as a singleton

The new `GuApp` construct also becomes a node on the CDK tree. One can visualise this tree to understand the cause for a resource to be in the template. This is further described in the [ADR](https://github.com/guardian/cdk/blob/aa-app-construct/docs/architecture-decision-records/002U1-component-constructors.md).

## How to test

Current tests should suffice.

A more structured representation of an App.

(This structure will also make it easier to auto-generate a `riff-raff.yaml` file!)

CDK autogenerates resource's logical ID using it's path in the tree as input. This change updates a resource's path in the tree, therefore the logical IDs change (as demonstrated by all the failing tests!). For stateful resources, this is a bit painful and we'd want a better, more stable, solution here.

## Checklist


docs/architecture-decision-records/002U1-component-constructors.md
```diff
@@ -0,0 +1,125 @@
+# Component Constructors - Update 1
```
akash1810, line 1: Might be easier to read [here](https://github.com/guardian/cdk/blob/aa-app-construct/docs/architecture-decision-records/002U1-component-constructors.md)?

The filename was chosen so that it comes under `002-component-constructors.md` in [file browsers](https://github.com/guardian/cdk/tree/aa-app-construct/docs/architecture-decision-records).

akash1810:
It's worth noting that this change will change an nodes path in the CDK tree, which will cause logical ID changes. This would be problematic for stateful resources.

Whilst this feature is valuable overall, it's not a priority to address right now.

Closing.

PR #1327, opened 2022-06-08
feat: proxy new command to projen project

nicl:
To illustrate how things might work with: https://github.com/guardian/cdk-app-ts


nicl:
Closing as we are not really satisfied with the Projen DX so this is not something to take forward.

PR #823, opened 2021-09-27
feat: Create a construct to create IAM resources for GitHub Actions

akash1810:
## What does this change?

With https://github.com/guardian/actions-assume-aws-role we're able to Assume Role within GitHub Actions.

This change introduces a construct to simplify the creation of this role.

The `GuGithubActionsRole` construct creates the three required resources:
  - `AWS::IAM::Policy`
  - `AWS::IAM::Role`
  - `AWS::IAM::OIDCProvider`

This allows teams to do:

```typescript
new GuGithubActionsRole(stack, {
  policies: [
    new GuGetS3ObjectsPolicy(stack, "GetObjects", {
      bucketName: "super-secret-stuff",
    }),
  ],
});
```

Which is smaller than the YAML alternative:

```yaml
GitHubBucketAccessPolicy:
  Type: AWS::IAM::Policy
  Properties:
    PolicyName: GitHubBucketAccessPolicy
    PolicyDocument:
      Statement:
        Action:
          - s3:GetObject
        Effect: Allow
        Resource:
          - arn:aws:s3:::super-secret-stuff/*
    Roles:
      - Ref: GitHubRole

GitHubRole:
  Type: AWS::IAM::Role
  Properties:
    AssumeRolePolicyDocument:
      Statement:
        - Effect: Allow
          Action: sts:AssumeRoleWithWebIdentity
          Principal:
            Federated: !Ref GitHubOidc
          Condition:
            StringLike:
              vstoken.actions.githubusercontent.com:sub: repo:guardian/*

GitHubOidc:
  Type: AWS::IAM::OIDCProvider
  Properties:
    Url: https://vstoken.actions.githubusercontent.com
    ClientIdList: [sigstore]
    ThumbprintList: [a031c46782e6e6c662c2c87c76da9aa62ccabd8e]
```

## Does this change require changes to existing projects or CDK CLI?

No.

## Does this change require changes to the library documentation?

Yes and is done.

## How to test

See added tests.

## How can we measure success?

Less YAML, fewer mistakes 🎉 .

## Have we considered potential risks?

Is the reasoning behind `GitHubOidcProvider` clear enough?

PR #1350, opened 2022-06-23
Fix 'Incorrect token audience' error for GH OIDC

rtyley:
This is an update to the construct that creates IAM resources for GitHub Actions, first introduced with https://github.com/guardian/cdk/pull/823 in early October 2021 - apparently the `ClientIdList` value should no longer be `sigstore`, as of 19th October 2021!

- https://github.com/aws-actions/configure-aws-credentials/issues/291
- https://github.com/aws-actions/configure-aws-credentials/issues/280#issuecomment-946608061
- https://github.com/aws-actions/configure-aws-credentials/pull/284

The new value is `sts.amazonaws.com`, which I think corresponds to this line in the [GitHub docs](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services#adding-the-identity-provider-to-aws):

> For the "Audience": Use sts.amazonaws.com if you are using the official action.

When using the old value of `sigstore` in the `AWS::IAM::OIDCProvider` `ClientIdList` field, running the `aws-actions/configure-aws-credentials` GitHub Action will give you an `"Error: Incorrect token audience"` error:

https://github.com/guardian/facia-scala-client/runs/7025740057?check_suite_focus=true#step:3:6


akash1810 (commented):
Thanks!

I think you'll need to update the [test](https://github.com/guardian/cdk/blob/09e5db014dea8d4d96b6d0d1c1f33c5453884bb2/src/constructs/iam/roles/__snapshots__/github-actions.test.ts.snap#L40) too.

rtyley:
> I think you'll need to update the [test](https://github.com/guardian/cdk/blob/09e5db014dea8d4d96b6d0d1c1f33c5453884bb2/src/constructs/iam/roles/__snapshots__/github-actions.test.ts.snap#L40) too.

Thanks - done with amended commit 36a4bdf!

akash1810:
The `check-aws-cdk` check is failing, however I'm not sure it should block this (or other PRs). The intention of this check is to keep up to date with releases to AWS CDK, which historically received updates very frequently.

Dependabot is the typical solution here, but it wasn't quite working.

I've dropped the requirement to have this check pass, as we should think of ways to automate this.

akash1810 (approved):

PR #1366, opened 2022-07-05
chore: Enable `noUncheckedIndexedAccess`

akash1810:
It might be easier to review this change commit by commit.

## What does this change?

[Introduced](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-1.html#checked-indexed-accesses---nouncheckedindexedaccess) in TS 4.1, [`noUncheckedIndexedAccess`](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess) offers stricter type checking, which is a good thing!

From the release notes:

> That means that in our last example, `opts.yadda` will have the type `string | number | undefined` as opposed to just `string | number`. If you need to access that property, you’ll either have to check for its existence first or use a non-null assertion operator (the postfix `!` character).

## How to test

CI passing should be sufficient.

### DIFF
--- a/src/bin/commands/new-project/utils/imports.ts
+++ b/src/bin/commands/new-project/utils/imports.ts
@@ -10,11 +10,11 @@ interface Import {
 export class Imports {
   imports: Record<string, Import>;
 
-  constructor(imports?: Record<string, Import>) {
-    this.imports = imports ?? {};
+  private constructor(imports: Record<string, Import>) {
+    this.imports = imports;
   }
 
-  addImport(lib: string, components: string[], type = false): void {
+  private addImport(lib: string, components: string[], type = false): void {
     if (!Object.keys(this.imports).includes(lib)) {
       const imports = type ? { types: components, components: [] } : { types: [], components };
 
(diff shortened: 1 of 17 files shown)


src/bin/commands/new-project/utils/imports.ts
```diff
@@ -8,33 +8,66 @@ interface Import {
 }
 
 export class Imports {
-  imports: Record<string, Import>;
+  private readonly imports: Record<string, Import>;
 
-  constructor(imports?: Record<string, Import>) {
-    this.imports = imports ?? {};
+  private constructor(imports: Record<string, Import>) {
```
akash1810, line 13: Although `Imports` is exported, instantiation is only performed by:
  - `newStackImports`
  - `newAppImports`
  - `newTestImports`

In this PR we make `Imports`'s constructor private, and convert the aforementioned functions to static methods on the class to improve namespacing.

src/bin/commands/new-project/utils/imports.ts
```diff
@@ -8,33 +8,66 @@ interface Import {
...
+    this.imports = imports;
   }
 
-  addImport(lib: string, components: string[], type = false): void {
```
akash1810, line 17: Calls to `addImport` always had `type` set to `false`. This meant we could remove the `type` argument. Once that's done, the value of `addImport` quickly disappears, so we can just remove it!

src/constants/ssm-parameter-paths.ts
```diff
@@ -52,3 +64,7 @@ export const SSM_PARAMETER_PATHS: Record<string, SsmParameterPath> = {
     optional: true,
   },
 };
+
+export const ALL_SSM_PARAMETER_PATHS: SsmParameterPath[] = Object.values(
+  NAMED_SSM_PARAMETER_PATHS
+) as SsmParameterPath[];
```
akash1810, line 70: Keeping things DRY here.

src/constants/ssm-parameter-paths.ts
```diff
@@ -8,9 +8,21 @@ export interface SsmParameterPath {
...
+  PrimaryVpcId: SsmParameterPath;
+  PrimaryVpcPrivateSubnets: SsmParameterPath;
+  PrimaryVpcPublicSubnets: SsmParameterPath;
+  FastlyCustomerId: SsmParameterPath;
+}
+
 export const VPC_SSM_PARAMETER_PREFIX = "/account/vpc";
 
-export const SSM_PARAMETER_PATHS: Record<string, SsmParameterPath> = {
+export const NAMED_SSM_PARAMETER_PATHS: NamedSsmParameterPaths = {
```
akash1810, line 25: `NAMED_SSM_PARAMETER_PATHS` is slightly clearer than `SSM_PARAMETER_PATHS`?

jacobwinch (approved):
👍

PR #1372, opened 2022-07-07
feat(riff-raff.yaml): Generate a `riff-raff.yaml` from a CDK app

akash1810:
> **Note**
> Easiest to review [commit by commit](https://github.com/guardian/cdk/pull/1372/commits).
> Requires https://github.com/guardian/riff-raff/pull/704, https://github.com/guardian/riff-raff/pull/788, https://github.com/guardian/riff-raff/pull/790

At the Guardian, we use [Riff-Raff](https://github.com/guardian/riff-raff) for deployment, configured via a `riff-raff.yaml` file. This file describes the steps to take during a deployment. For example, to deploy a lambda, we'll typically have:

```yaml
regions:
  - eu-west-1
stacks:
  - playground
allowedStages:
  - CODE
  - PROD
deployments:
  cloudformation:
    type: cloud-formation
    app: my-application
    parameters:
      templateStagePaths:
        CODE: MyApplication-CODE.template.json
        PROD: MyApplication-PROD.template.json
  my-application:
    type: aws-lambda
    dependencies:
      - cloudformation
    parameters:
      bucketSsmLookup: true
      fileName: my-app.zip
      lookupByTags: true
```

Creating this file can be a little complicated, and when creating a new service we usually copy the file from an existing repository, and then spend a couple of hours massaging it to fit. Even then, it's not uncommon for us to manually provision the infrastructure first, then use Riff-Raff from the second deployment onwards.

## What does this change and how to use?
CDK knows[^3] everything to create this file. In this change, we auto-generate the `riff-raff.yaml` file, saving developer's time. The resulting file will be structured such that it can be used to provision a _new_ stack from scratch via a Riff-Raff deployment, removing the need to manually provision things first.

> **Note**
> These changes are marked as [experimental](https://github.com/guardian/cdk/blob/main/docs/architecture-decision-records/005-x01-package-structure.md).

In order to make this feature easy to adopt, a new `GuRootExperimental` construct has been created. Adoption means going from this:

```ts
// bin/cdk.ts
import { App } from "aws-cdk-lib";

const app = new App();
new MyStack(app, "some-id", {});
```

To this:

```ts
// bin/cdk.ts
import { GuRootExperimental } from "@guardian/cdk/lib/constructs/root";

const app = new GuRootExperimental();
new MyStack(app, "some-id", {});
```

The auto-generated version of the above would be:

```yaml
allowedStages:
  - CODE
  - PROD
deployments:
  lambda-upload-eu-west-1-playground-my-application:
    type: aws-lambda
    stacks:
      - playground
    regions:
      - eu-west-1
    app: my-application
    contentDirectory: my-application
    parameters:
      bucketSsmLookup: true
      lookupByTags: true
      fileName: my-app.zip
    actions:
      - uploadLambda
  cfn-eu-west-1-playground-my-application-stack:
    type: cloud-formation
    regions:
      - eu-west-1
    stacks:
      - playground
    app: my-application-stack
    contentDirectory: cdk.out
    parameters:
      templateStagePaths:
        CODE: MyApplication-CODE.template.json
        PROD: MyApplication-PROD.template.json
    dependencies:
      - lambda-upload-eu-west-1-playground-my-application
  lambda-update-eu-west-1-playground-my-application:
    type: aws-lambda
    stacks:
      - playground
    regions:
      - eu-west-1
    app: my-application
    contentDirectory: my-application
    parameters:
      bucketSsmLookup: true
      lookupByTags: true
      fileName: my-app.zip
    actions:
      - updateLambda
    dependencies:
      - cfn-eu-west-1-playground-my-application-stack
```

Crucially, the dependencies between deployments will be configured such that the order of events is:
  1. Upload artifact to S3
  2. Deploy infrastructure
  3. Update the application

This will allow us to deploy applications from scratch using Riff-Raff - no more confusing first deployments!

### Supported Riff-Raff deployment types
The following deployment types are currently supported:
  - `cloud-formation`
  - `aws-lambda`
  - `autoscaling`

It also supports deploying across multiple regions and multiple stacks.

The added tests can be used to see more examples of the generated file.

#### Directory structure
It's worth noting the addition of `contentDirectory` and `app` to each deployment. This impacts the directory structure of the files uploaded to Riff-Raff during CI. The files should match this structure:

```
.
├── cdk.out
│   └── MyApplication-CODE.template.json
│   └── MyApplication-PROD.template.json
└── my-application
    └── my-app.zip
```

Where all the CloudFormation templates are in `cdk.out` and application artifacts are in `<app>`. The file names are down to the user; a `.zip` file is used for illustration purposes only.

#### Use of tags
The generated `riff-raff.yaml` file will tell Riff-Raff to perform service discovery by tags.

If a service currently hardcodes the CloudFormation stack name, or lambda function names in it's `riff-raff.yaml` file, then it cannot switch to this experimental process. We'd need to think about if it's correct to support these stacks, and if yes, how. I'd suggest this is future work too.

#### Riff-Raff `templates`
It's possible to [snapshot test](https://jestjs.io/docs/snapshot-testing) the generated `riff-raff.yaml` file, however it's worth noting that if you're migrating a service that uses `templates` in `riff-raff.yaml`, the snapshot will not exactly match the manually crafted `riff-raff.yaml` file, as `RiffRaffYamlFileExperimental` doesn't use `templates`, favouring instead to expand all the deployments. However, Riff-Raff will perform the same steps.

## How to test
Some unit tests have been added, but we should also be able to:
  - Produce a [simple file](https://github.com/guardian/cdk-playground/blob/774ba0d57a6d2654b07981bc0d2677fe4a4375ad/riff-raff.yaml)
  - Produce a [slightly more complex file](https://github.com/guardian/elasticsearch-node-rotation/blob/0f3c749dbfad6cacb537e8afaca4cf4e0edf2973/riff-raff.yaml)
  - Produce a [single stack, multi-region file](https://github.com/guardian/consent-management-platform/blob/190511d4e83cea84c10e90b7e6200f0e37109292/riff-raff.yaml)
  - Produce a [single region, multi-stack file](https://github.com/guardian/cfn-private-resource-types/blob/main/dns/guardian-dns-record-set-type/riff-raff.yaml)

It is significantly easier to configure Riff-Raff for services, both new and existing.

- Is the code readable, and maintainable?
- Is the resulting `riff-raff.yaml` actually valid?

## Checklist

## TODO as follow-ups
- [ ] Update project generator to use `GuRootExperimental` (with an opt-out flag?)

[^3]: If it doesn't know, it'll now throw. For example, when `region` is not explicitly set.


src/utils/riff-raff-yaml-file.ts
```diff
@@ -0,0 +1,286 @@
...
+                ),
+              };
+            },
+            {}
+          ),
+        };
+      },
+      {}
+    );
+  }
```
JamieB-gu, line 104:
```suggestion
  private groupByClassNameStackRegionStage(cdkStacks: GuStack[]): GroupedCdkStacks {
    return Object.entries(this.groupByClassName(cdkStacks)).reduce(
      (accClassName, [className, stacksGroupedByClassName]) => ({
        ...accClassName,
        [className]: Object.entries(this.groupByStackTag(stacksGroupedByClassName)).reduce(
          (accStackTag, [stackTag, stacksGroupedByStackTag]) => ({
            ...accStackTag,
            [stackTag]: Object.entries(this.groupByRegion(stacksGroupedByStackTag)).reduce(
              (accRegion, [region, stacksGroupedByRegion]) => ({
                ...accRegion,
                [region]: this.groupByStageTag(stacksGroupedByRegion),
              }),
              {}
            ),
          }),
          {}
        ),
      }),
      {}
    );
  }
```

src/utils/riff-raff-yaml-file.ts
```diff
@@ -0,0 +1,286 @@
...
+                    ...accStageTag,
+                    [stageTag]: this.isCdkStackPresent(name, stackTag, region, stageTag) ? "✅" : "❌",
+                  };
+                }, {}),
+              };
+            }, {}),
+          };
+        }, {}),
+      };
+    }, {});
```
JamieB-gu, line 162:
```suggestion
    const checks: AppValidation = allCdkStacks.reduce((accClassName, { constructor: { name } }) => ({
      ...accClassName,
      [name]: allStackTags.reduce((accStackTag, stackTag) => ({
        ...accStackTag,
        [stackTag]: allRegions.reduce((accRegion, region) => ({
          ...accRegion,
          [region]: allStageTags.reduce((accStageTag, stageTag) => ({
            ...accStageTag,
            [stageTag]: this.isCdkStackPresent(name, stackTag, region, stageTag) ? "✅" : "❌",
          }), {}),
        }), {}),
      }), {}),
    }), {});
```

src/utils/riff-raff-yaml-file.ts
```diff
@@ -0,0 +1,286 @@
...
+
+    const missingDefinitions = Object.values(checks).flatMap((groupedByStackTag) => {
+      return Object.values(groupedByStackTag).flatMap((groupedByRegion) => {
+        return Object.values(groupedByRegion).flatMap((groupedByStage) => {
+          return Object.values(groupedByStage).filter((_) => _ === "❌");
+        });
+      });
+    });
```
JamieB-gu, line 170:
```suggestion
    const missingDefinitions = Object.values(checks).flatMap((groupedByStackTag) =>
      Object.values(groupedByStackTag).flatMap((groupedByRegion) =>
        Object.values(groupedByRegion).flatMap((groupedByStage) =>
          Object.values(groupedByStage).filter((_) => _ === "❌");
        );
      );
    );
```

src/utils/array.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Group an array of T by a field in T.
+ * @param input what to group
+ * @param identityFn what field to group on
```
JamieB-gu, line 4: I think the name `identity` or "identity function" may be commonly associated with the identity function, i.e. the one that returns its argument unchanged?

```ts
const identity = <A>(a: A): A => a
```

JamieB-gu (commented):
One possible suggestion for readability might be to drop the explicit `return` from all single-statement functions, particularly where you've got several of them nested? It doesn't reduce the logical complexity of what these methods are doing, but it may remove some of the syntax/indentation overhead? I've put a few suggestion examples in.

jfsoul (commented):
Exciting!

src/utils/riff-raff-yaml-file.test.ts
```diff
@@ -0,0 +1,323 @@
...
+
+    expect(actual).toMatchInlineSnapshot(`
+      "allowedStages:
+        - PROD
+        - CODE
+      deployments:
+        my-application-stack-cfn-deploy-eu-west-1:
+          type: cloud-formation
+          regions:
+            - eu-west-1
```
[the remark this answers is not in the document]
akash1810: This is to support services like [this](https://github.com/guardian/consent-management-platform/blob/190511d4e83cea84c10e90b7e6200f0e37109292/riff-raff.yaml), where we deploy the same CDK class to different regions - each region has it's own JSON template.

Pretty niche, but the consent-management-platform repository was one of the drivers for this, as it took too long to manually work out the structure of the file.

src/utils/riff-raff-yaml-file.test.ts
```diff
@@ -0,0 +1,323 @@
...
+          stacks:
+            - test
+          app: my-application-stack
+          contentDirectory: /tmp/cdk.out
+          parameters:
+            templateStagePaths:
+              TEST: test-stack.template.json
+        my-app-ami-test-eu-west-1:
+          type: ami-cloudformation-parameter
```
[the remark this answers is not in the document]
akash1810: Possibly, I couldn't work it out! `ami-cloudformation-parameter` is used because GuCDK adds an AMI parameter per app.

If we had 2 ASGs:
  1. `new GuEc2App(..., ..., { app: "frontend" } )`
  2.`new GuEc2App(..., ..., { app: "api" } )` 

The CFN JSON will have 2 parameters:
  1. AMIFrontend
  2. AMIApi

`ami-cloudformation-parameter` was the only way I could get Riff-Raff to update both parameters.

nicl (commented):
Just to record some of our IRL discussions here. This looks great! I am quite torn though on whether, in practice, it will turn out to be a 'leaky abstraction' - it represents a lot of new and complicated code, and there are a lot of future edge cases that I suspect will cause problems.

So my recommendation would be we roll this out as part of some kind of cdk-kitchen of the kind we plan to introduce as part of the Backwards Compat/v2 work (re: https://trello.com/c/RAPEWY0v/1309-backwards-compatibility-v2 and the attached doc which discuss the idea further). We can then get some feedback by trialling it within devx and also across some external projects and review at that point if it makes sense to 'promote' it into @guardian/cdk.

src/utils/riff-raff-yaml-file.ts
```diff
@@ -0,0 +1,528 @@
...
+ *   3. Autoscaling deploy (`autoscaling`)
+ *
+ * It assumes a Riff-Raff bundle structure as follows:
+ *
+ * ```
+ * .
+ * ├── cdk.out
+ * │   └── MyApplication.template.json
+ * ├── my-application
+ * │   └── my-application.deb
```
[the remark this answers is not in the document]
akash1810: > It is worth us discussing packaging/debs here as I'm not convinced this is the right choice - they are complicated to produce. (But not a blocker to this PR.)

Agreed! Worth noting that debs are not required. I've attempted to further describe the directory layout in a [README](https://github.com/guardian/cdk/blob/aa-riff-raff/src/utils/riff-raff-yaml-file/README.md).

nicl (approved):
Looks great! Thanks for rejigging to place under `experimental` and the work around that too.

akash1810:
Force pushed following a rebase of main with #1427 being merged.

nicl (commented):
Some other thoughts, not blockers but suggestions for things that could improve readability:

* consider using pure functions instead of a classs? (better to default to pure functions whenever possible)

* the nested `groupByClassNameStackRegionStage` feels pretty gnarly. I wonder if flattening things can help here? All you need I think is something like:

`GuStack[][]` where stacks are grouped by name/stack/region.

The deeply nested structure that is currently used:

```
type GroupedCdkStacks = Record<
  ClassName,
  Record<StackTag, Record<Region, Record<StageTag, CdkStacksDifferingOnlyByStage>>>
>;
```
 
is very hard to understand/work with and also doesn't quite capture as well the actual relationships here.

* consider introducing a `Generator` type for improved clarity of the helpers that generate riffraff deployments. e.g.

```
type Generator = (info: DeploymentInfo, stacks: GuStack[]) => Deployment[];
const cfnDeployments: Generator = (info: DeploymentInfo, stacks: GuStack[]): Deployment[] => { ..}
const asgDeployemtns: Generator ...
const lambdaDeployments: Generator ...
```

and consolidate the work into these functions.

--
A very (incomplete) exploration of some of the above is here:

https://gist.github.com/nicl/5edc441007e15281fad833c81bf2c564

--

Lastly, I wonder if generating a single riffraff file may prove limiting here in repos where CDK contains multiple stacks (for example, when one app only exists for INFRA and another for CODE/PROD). How could we get around this?

akash1810:
The robots have not created a release for this change - https://github.com/guardian/cdk/runs/7922010296?check_suite_focus=true#step:4:159

image

I think this is because https://github.com/guardian/cdk/pull/1432 was merged whilst CD on this merge was running. Ugh!

Two possible resolutions:
  1. Revert this PR and re-raise/merge it
  2. Bundle this change into another release

2 is easiest, I think? With #1434 being good candidate?

akash1810:
> Lastly, I wonder if generating a single riffraff file may prove limiting here in repos where CDK contains multiple stacks (for example, when one app only exists for INFRA and another for CODE/PROD). How could we get around this?

This is already supported. As a `GuRiffRaffExperimental` is attached to an `App`, one possible solution is to have multiple instances of `App`. Using the `GuRootExperimental` helper, this would look like this:

```ts
const root1 = new GuRoofExperimental();

const stack1 = new GuStack(root1);
const stack 2 = new GuStack(root1);

const root2 = new GuRootExperimental();

const stack3 = new GuStack(root2);
```

By default, the output directory of each `App` is `cdk.out`, and as we're always writing to `riff-raff.yaml` we'll suffer from last write wins. However, we can customise the output directory:

```ts
const root1 = new GuRoofExperimental({outdir: `cdk.out/root1`);

const stack1 = new GuStack(root1);
const stack 2 = new GuStack(root1);

const root2 = new GuRoofExperimental({outdir: `cdk.out/root2`);

const stack3 = new GuStack(root2);
```

Which will yield 2 `riff-raff.yaml` files, one in `cdk.out/root1` and another in `cdk.out/root2`.

Having said that, I can't think of a service that has this requirement?

nicl:
@akash1810 thanks for the responses. Just to expand on some of the points:

* re class - a nice option is to create pure functions for the core behaviour (so making e.g. unit testing easier) and then, if you need a construct, creating a slim class that simply calls the pure functions as required. At the moment the class has a lot of fields:

```
  private readonly allCdkStacks: GuStack[];
  private readonly allStackTags: StackTag[];
  private readonly allStageTags: StageTag[];
  private readonly allRegions: Region[];

  private readonly riffRaffYaml: RiffRaffYaml;
  private readonly outDir: string;
```

This complicates unit testing. At the moment all of the tests are essentially integration tests - requiring full app/stack setup. There are nice things about these tests, but having some simpler unit tests that focus on narrow behaviour (for example, the `getAutoscalingDeployment` function) would be nice too.

* re flattening - I'm not sure how flattening makes the validator more difficult, there is no loss of info, just of nesting if that makes sense? - have a look at the example code to get a sense here (https://gist.github.com/nicl/5edc441007e15281fad833c81bf2c564#file-riff-raff-yaml-file-ts-L93). There I grouped by a string (object keys can't be anything else in JS unfortunately), but you could just use `[][]Deployment` I think.
* agreed, `Generator` is mostly about bringing clarity to an implicit pattern, but also would enforce consistency in some areas where it is missing and guide the reader on which bits of the code are top-level and which are helpers (you'd probably want to unit test the generators for example but not the helpers).

PR #1391, opened 2022-07-19
chore(deps)!: Update aws-cdk-lib to 2.32.1

akash1810:
BREAKING CHANGE: Update aws-cdk-lib to 2.32.1


kenoir:
What does this break?

akash1810:
> What does this break?

Nothing explicitly.

Users of GuCDK need to use the _same_ version of `aws-cdk` and `constructs`. We release bumps of these deps as a breaking change for a couple of reasons:
  - We can't rely on Dependabot to auto bump the version of GuCDK as `@guardian/cdk`, `aws-cdk` and `constructs` all need updating at the same time
  - This version of AWS CDK _might_ include some changes that yield resource replacement, which might not be desired. Releasing a new major version provides a signal that care should be taken

kenoir (approved):

PR #1408, opened 2022-08-08
fix(deps): Update aws-cdk-lib to 2.35.0

akash1810:
Update aws-cdk-lib to 2.35.0. Also update the script to create a patch release, rather than a new major release.


src/patterns/__snapshots__/api-lambda.test.ts.snap
```diff
@@ -523,7 +523,7 @@ Object {
         "lambdaapiANY4EBCADD1",
       ],
       "Properties": Object {
-        "Description": "Automatically created by the RestApi construct",
+        "Description": "this is a test",
```
akash1810, line 526: This change is a welcomed fix from [v2.34.0](https://github.com/aws/aws-cdk/releases/tag/v2.34.0), specifically https://github.com/aws/aws-cdk/pull/21207.

nicl (commented):
👍 though caveat, this (if I recall our discussions correctly) shouldn't actually be a breaking change. Yes users will need to bump some deps, but it shouldn't change the API interface they see (in a breaking way) nor the outputted Cloudformation.

nicl (approved):

PR #1418, opened 2022-08-11
docs(adr): Add ADR for supporting experimental features

akash1810:
## What does this change?

Adds an ADR following a team discussion on how we can add experimental features in a clear and obvious way.

## How to test

- Does it [read](https://github.com/guardian/cdk/blob/aa-experimental-adr/docs/architecture-decision-records/005-x01-package-structure.md) well?
- Does it accurately reflect the discussion?

### DIFF
--- a/docs/architecture-decision-records/005-x01-package-structure.md
+++ b/docs/architecture-decision-records/005-x01-package-structure.md
@@ -1,0 +1,102 @@
+# Package Structure - Extension 01
+
+## Status
+Proposed.
+
+## Context
+This is an extension to the [package structure ADR](./005-package-structure.md)
+and describes how this project can support early and/or experimental features from within the team and from external contributions.
+
+These features should be considered "in development, but with an unstable API".
+Supporting these types of features would allow for their inclusion in a published NPM library,
+simplifying any testing or dog-fooding "in the wild".
+
+There is no commitment to make these features to stable in future. For this reason, we'll term them "experimental" rather than "alpha".
+
+The process of authoring experimental features should offer little friction for contributors, consumers, and maintainers.
+
+## Positions
+Each position will be evaluated on the following criteria:
+- Contributing new experimental features
+- Consuming an experimental feature
+- Long term maintenance
+
+### 1. Create a separate NPM package
+Named `@guardian/cdk-experiments`, or similar, this package would hold these features and would be separate from the core `@guardian/cdk` package.
+
+This additional package would likely require converting the repository to a mono-repo using NPM Workspaces, with at least three workspaces:
+- `common` for shared code (not published)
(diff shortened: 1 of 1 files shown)


nicl (commented):
I think perhaps this oversells the complexities of an external package - for example, it's not clear why releases would need to be synchronised - perhaps just a major version compatibility, which should be easier when we make fewer of those).

And it perhaps underplays the risks of one-package, namely that people treat things as part of the supported API.

But broadly I think this captures our thinking and certainly our decision/preferred solution. 👍 

Would be great to extend the ADR to include the `public`/`internal`/`experimental` structure, and also to include something on also ensuring experimental constructs/patterns also follow:

1. A naming convention to denote that they are experimental (e.g. `MyConstructAlpha` or `MyConstructExperimental` or equivalent)
2. Notes in their docs to also flag this fact

Not directly here, but I wonder if we should expand the Contribution guidelines following this to explain all this as well.

akash1810:
> I think perhaps this oversells the complexities of an external package - for example, it's not clear why releases would need to be synchronised - perhaps just a major version compatibility, which should be easier when we make fewer of those).

Struggling to work this part clearly. I've made a second attempt at this now.

The concern stems from:
  - `@guardian/cdk` v10 requires AWS CDK v2.1
  - My project uses `@guardian/cdk` v8 (i.e. _not_ latest)
  - I author an experiment
  - To use the experiment, I have to first move to `@guardian/cdk` v10, then move to the next version of `@guardian/cdk-experiments`

nicl (approved):

docs/architecture-decision-records/005-x01-package-structure.md
```diff
@@ -0,0 +1,102 @@
...
+  import { EarlyFeature } from "@guardian/cdk/experimental";
+  ```
+
+  As the experiments are included in the `@guardian/cdk` package, compatibility is a given.
+
+  There is a risk that usage of an experimental feature goes unnoticed however, especially as IDEs typically fold import statements.
+
+- :white_check_mark: Long term maintenance
+
+  An additional directory has minimal maintenance cost.
```
kenoir, line 85: How will changes to the experimental folder be reflected in the semantic versioning? For example when you release a change that adds some experiments but does not change anything else how is that reflected in the release version?

akash1810: Releases will be made "as normal": new experiments will create a minor release, and fixes a minor. For those not making use of an experiment, Dependabot raised PRs to bump the version of `@guardian/cdk` being used should still be trusted and merged.

kenoir: Might be worth adding that to this ADR.

PR #1419, opened 2022-08-11
fix(experimental): Move experimental patterns into the correct directory

akash1810:
Easiest to review commit by commit.

## What does this change?

This follows https://github.com/guardian/cdk/pull/1418 where we want to be more explicit about what parts of the library may not have a stable API.

We currently have two experimental patterns in the library - this change updates them to follow the repository's updated practices. As such:
  - `GuKinesisLambda` is renamed to `GuKinesisLambdaExperimental`
  - `GuSnsLambda` is renamed to `GuSnsLambdaExperimental`

A small linting rule is also added to ensure any _exported_ class in `src/experimental` has a name ending in 'Experimental'.

Ideally we'd also lint to encourage the use of the `@experimental` tag in typedocs, but I didn't have much luck implementing this, either manually or with https://www.npmjs.com/package/@es-joy/jsdoccomment.

## How to test

CI passing should suffice.

Providing a more obvious API for consuming projects.

N/A

## Checklist


nicl (approved):
This is great! Thanks @akash1810 🥳

PR #1427, opened 2022-08-17
fix(bootstrap): Remove `bootstrap` CLI command

akash1810:
## What does this change?

Originally introduced in #1135, the bootstrap CLI command provided an interactive way to provision the SSM Parameters required by GuCDK into an AWS Account. Whilst this worked, the CloudFormation stack it created in an account was not tracked in VCS as it was dynamically generated.

We have replaced the process of account setup with https://github.com/guardian/aws-account-setup, which forces use of VCS.

## How to test

CI passing should be enough 🤞🏽 .

Less code!

What happens to all the stacks provisioned by this method?! Do we delete them and recreate with `aws-account-setup`?

It looks like the stack created would always be named `cdk-bootstrap-INFRA`, and via a Prism lookup, it appears there are 0 stacks of that name. That is, risks are minimal.

## Checklist


nicl (approved):
LGTM! The new approach is soo much nicer here :)

PR #1442, opened 2022-08-24
chore: new item test-update in package.json

JuliaBrigitte:
## What does this change?
Added test-update to package.json to make pre commit check easier by including the snapshot ci will expect

## How to test
script/ci-project-generation
will create a test project in 
/private/tmp/cdk-project
where you can check out the generated package.json


nicl (approved):
👍 (we can fix the AWS version issue separately.)

akash1810:
LGTM.

Worth noting that most repositories are configured to only run CI checks on "first party" branches and _not_ forks. Looks like this repository does not follow this pattern (yet!), but it's preferred to branch rather than fork to guarantee CI checks run.

PR #1466, opened 2022-08-31
fix(GuVpc): Configure AZs for eu-west-1 region, so you don't have to

akash1810:
## What does this change?

The number of AZs in eu-west-1 is fairly static (at least it hasn't changed in over a decade).

In this change, we programmatically set the context for the `GuVpc` construct so that users don't have to commit the, slightly magical, `cdk.context.json` file. This has caused some confusion in the past, see https://github.com/guardian/editorial-tools-platform/pull/623#discussion_r928768448.

Note: The [aws-account-setup](https://github.com/guardian/aws-account-setup/blob/e455a0826e4d827ef3ddb840efce280c942f6dff/lib/aws-account-setup.ts#L120-L125) repo is already successfully using this "trick".

## How to test

See CI?

It's worth noting that the snapshot test has been updated quite significantly. This is because the AZs are known at synth time, rather than runtime (i.e. we've shifted left). This is a good thing, as it means the shape of the resulting CFN resources become more deterministic.

Less confusion for the end user.

What happens if we create a VPC outside of eu-west-1? Currently, the `cdk.context.json` file still needs to be added to VCS by users as before.

We're almost exclusively in this region though, so I'm not sure how much of an issue this is. I'd suggest we address it at the time.

## Checklist

### DIFF
--- a/src/constructs/vpc/__snapshots__/vpc.test.ts.snap
+++ b/src/constructs/vpc/__snapshots__/vpc.test.ts.snap
@@ -127,15 +127,8 @@ Object {
     },
     "MyVpcapplicationSubnet1Subnet0A4FCBEC": Object {
       "Properties": Object {
-        "AvailabilityZone": Object {
-          "Fn::Select": Array [
-            0,
-            Object {
-              "Fn::GetAZs": "",
-            },
-          ],
-        },
-        "CidrBlock": "10.0.128.0/18",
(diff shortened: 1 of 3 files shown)


nicl (approved):
Nice! Much better DX for users.

akash1810:
Successfully tested this in the security-platform repo.

PR #1471, opened 2022-09-02
fix(GuLambdaFunction): Tightly scope S3 permission

akash1810:
## What does this change?

Update the policy for `GuLambdaFunction`s, granting S3 read to the exact object of the lambda, rather than the entire bucket.

This follows the principle of least privilege.

s/o @Fweddi for pointing this out!

## How to test

See updated tests.

A close following of the principle of least privilege.

None. Riff-Raff will upload to this path already and a lambda should not need to read from elsewhere in the bucket. In the unlikelihood that it does, an additional explicit permission should be added by end-users.

Any lambda that is relying on this side-effect would have to be updated to regain permissions to any additional objects in the bucket being accessed.

## Checklist

### DIFF
--- a/src/constructs/lambda/lambda.ts
+++ b/src/constructs/lambda/lambda.ts
@@ -108,7 +108,7 @@ export class GuLambdaFunction extends Function {
       });
     }
 
-    bucket.grantRead(this);
+    bucket.grantRead(this, objectKey);
 
     const ssmParamReadPolicies: PolicyStatement[] = [
       new ReadParametersByPath(scope, props),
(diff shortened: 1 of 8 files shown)


src/constructs/lambda/lambda.ts
```diff
@@ -108,7 +108,7 @@ export class GuLambdaFunction extends Function {
       });
     }
 
-    bucket.grantRead(this);
+    bucket.grantRead(this, objectKey);
```
akash1810, line 111: This is the main change (all the others are tests).

Fweddi (approved):

jfsoul (approved):
💯 I think the only contention here is how to ensure consumers can apply this change safely, assuming some may have been relying on the wider permissions.

akash1810:
> 💯 I think the only contention here is how to ensure consumers can apply this change safely, assuming some may have been relying on the wider permissions.

An, admittedly quick, check for usages doesn't suggest there are any instances of this. 🤞🏽

PR #1472, opened 2022-09-02
refactor: Remove unused code

akash1810:
## What does this change?

First introduced in #336 and later removed in #1157, this change removes some code (and a dependency) for an old custom resource.

## How to test

Tests should continue to pass, (which demonstrates the removed code is no longer used?).

Less code to maintain.

n/a

## Checklist


tools/integration-test/src/integration-test-stack.test.ts
```diff
@@ -1,4 +1,3 @@
-import libraryInfo from '@guardian/cdk/lib/constants/library-info';
```
akash1810, line 1: Changes in this file are not related to this PR per se, they're made as they're flagged by the linter.

kenoir (commented):
Mentioned PRs make it clear this is unused anywhere, so looks good.

kenoir (approved):

PR #1483, opened 2022-09-08
refactor: Clearer default value for GuParameter

kenoir:
## What does this change?

This change update the default value for `GuParameter` to describe that the value is used to route to the value in SSM, which is non-obvious.

The intention is to provide slightly better hints to consumers when they use `GuParameter` in their projects in order to make using GuCDK a more pleasant experience.


kenoir:
I considered filling the value with a good expected default from the known stack and stage name, but we are missing the app name in this context.

nicl (approved):
A definite improvement on status quo so 👍 .

One thought: do we have access to the actual valuesof `stack/stage/app/param` here? If so, it is worth actually inlining the values rather than the generic `/$stage/$stack/$app/$param` here?

PR #1394, opened 2022-07-19
feat: add Metadata aspect

nicl:
**Note, this is an initial PR to introduce the aspect, but a subsequent PR (or perhaps commit here) is intended to migrate our existing GuStack and constructs to use this.**

This aspect registers (template) metadata about Guardian constructs. This will be used to understand usage in the wild. The exact format is:

```
"gu:cdk:version": "1234",
"gu:cdk:constructs": ["construct-a", "construct-b", ...],
```

Metadata is polled by [a lambda](https://github.com/guardian/cdk-metadata) and eventually graphed in metrics.gutools.co.uk. 

It will take time to fully roll this out, as @guardian/cdk users will need to upgrade and also deploy for us to begin collecting data for them.

For more on aspects, see:

https://docs.aws.amazon.com/cdk/v2/guide/aspects.html

PR #1484, opened 2022-09-08
docs: add mock for version-info to best practices

georgeblahblah:
## What does this change?

- Adds mock for `version-info` constants to the best practices docs, to ensure snapshot tests don't fail across GuCDK version upgrades following the introduction of the Metadata aspect (8f610647f53c3fe17a3b40f68a66202460281ba7)

### DIFF
--- a/docs/best-practices.md
+++ b/docs/best-practices.md
@@ -30,6 +30,7 @@ To use it, first, create `jest.setup.js` and add the global mock:
 
 ```javascript
 jest.mock("@guardian/cdk/lib/constants/tracking-tag");
+jest.mock("@guardian/cdk/lib/constants/version-info");
 ```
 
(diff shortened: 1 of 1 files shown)


akash1810 (commented):
I think this is a regression from https://github.com/guardian/cdk/pull/1394. I'd be keen for the single `tracking-tag` mock to suffice for all scenarios, else all projects need to update their Jest config.

akash1810:
@georgeblahblah https://github.com/guardian/cdk/pull/1485 is available in v47.3.3 now. Does that work for you, and would there be objections to closing this?

georgeblahblah:
> @georgeblahblah #1485 is available in v47.3.3 now. Does that work for you, and would there be objections to closing this?

Oh amazing, thanks @akash1810! #1485 is a much better solution, I'll close this

PR #1485, opened 2022-09-08
fix(metadata): Use mocked library version in Metadata Aspect

akash1810:
## What does this change?

In https://github.com/guardian/cdk/pull/1394 we introduced a Metadata Aspect to enable us to track which specific constructs and patterns are in use, and which version. This manifests itself as `Metadata` in the CloudFormation template.

The version maps to the NPM package version, as such, it changes over time. This means the value stored within snapshot tests would also change. This will become noisy over time.

In https://github.com/guardian/cdk/pull/448 we published a mock and wired every project up to use it. This results in the library version being static in snapshots.

In this change, we update the Metadata Aspect to retrieve the version number in the same way, making it's value static within snapshots also.

This change is an alternative to https://github.com/guardian/cdk/pull/1484. Whilst #1484 would work, it would also require changes to each cdk project's Jest configuration. In this change, we avoid this as they're already configured to use this mocked value.

## How to test

See CI.

Simplifying the upgrade path, as one can just bump the library version.

n/a

## Checklist

### DIFF
--- a/src/aspects/metadata.ts
+++ b/src/aspects/metadata.ts
@@ -1,6 +1,6 @@
 import type { IAspect, Stack } from "aws-cdk-lib";
 import type { IConstruct } from "constructs";
-import { LibraryInfo, MetadataKeys } from "../constants";
+import { MetadataKeys, TrackingTag } from "../constants";
 
 export class Metadata implements IAspect {
   readonly stack: Stack;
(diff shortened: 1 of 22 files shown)


nicl (approved):

akash1810:
I think this has broken the package, at least I'm getting this in the tracker repository when upgrading to 47.3.3:

```console
    Cannot find module '@guardian/cdk/lib/constants/tracking-tag' from 'jest.setup.cdk.js'

    > 1 | jest.mock('@guardian/cdk/lib/constants/tracking-tag');
        |                          ^
      2 |
```

PR #1497, opened 2022-09-19
feat: Optional appName for tag

tomrf1:
It's useful to have all resources in a stack tagged with `App`, specifically for monitoring costs by app

### DIFF
--- a/src/constructs/core/stack.ts
+++ b/src/constructs/core/stack.ts
@@ -23,7 +23,7 @@ export interface GuStackProps extends Omit<StackProps, "stackName"> {
   /**
    * Optional name of the app. If defined, all resources will have an App tag.
    */
-  appName?: string;
+  app?: string;
 
   /**
    * The AWS CloudFormation stack name (as shown in the AWS CloudFormation UI).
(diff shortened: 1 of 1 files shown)


akash1810 (commented):
A really sensible change.

I wonder if we should rename the property to `app`, and consequently rename `app` in the constructor to `scope`? `scope` is a fairly typical name within the AWS CDK for this type of argument.

IIUC the patterns will add the `app` tag to resources too, so I'd be curious to see what would happen if multiple `GuEc2App` patterns are added to the same `GuStack` and each set the app tag. Maybe that's worth a test and/or some validation logic?

akash1810:
> The docstring for `GuStack` already suggests the app tag should be set based on the constructor `app` arg

I think the docs are out of date; we correct that description!

> cc @akash1810 @tomrf1 I wonder if we should just require `app` as part of `GuStackProps` and then stop adding the app tag via our patterns? It's not clear to me that patterns should be adding app tags to things themselves - feels like the stack should have this responsibilty.

I don't think we can do this. Tags at the `GuStack` level are applied to all resources. If the stack is simple, this is fine. However where we have multiple apps, for example two lambdas, it'll be incorrect to apply `App` at `GuStack` as each lambda will get the same value.

tomrf1:
@akash1810 @nicl I've pushed up the rename from `appName` to `app`

nicl (approved):
Nice :shipit:

akash1810 (approved):
🏷️

akash1810:
I don't think this is working, as we're never assigning `this._app` from the props 😓 .

PR #1504, opened 2022-09-26
feat!: remove support for VPC SSM parameters in GuVPC construct

nicl:
Breaking change: VPC parameter store values are no longer created by the `GuVPC` construct. Instead, they should be specified in the `aws-account-setup` repository: https://github.com/guardian/aws-account-setup. If you are upgrading and have used this construct you should migrate the parameters to aws-account-setup as part of this upgrade.

The motivation for this change is to avoid confusion and consolidate CDK-related parameters into a single place.

The number of `GuVPC` is currently small (security, presence, ...?) so this should not affect too many people.


akash1810:
>  If you are upgrading and have used this construct you should migrate the parameters to aws-account-setup as part of this upgrade.

Is it worth describing this process? I think it would require deleting the SSM params provisioned by this stack, and re-provisioning them via aws-account-setup? If so, any deploys performed between this time would fail as they'll refer to non-existent params.

akash1810 (approved):

PR #1510, opened 2022-09-29
docs: pass `--ci` to `npm test` in example CI script

georgeblahblah:
## What does this change?

- Updates docs to pass the `--ci` flag to `npm test` in the example script. I'm unsure if this is the best place, because it assumes Jest. Perhaps a note near the example would be better


akash1810:
I wonder if it should be in the [`package.json` that gets generated](https://github.com/guardian/cdk/blob/527057a34a0dc13ec73cbe89114fce6b93b97645/src/bin/commands/new-project/utils/init.ts#L111-L122) by `npx @guardian/cdk@latest new` too/instead?

georgeblahblah:
> I wonder if it should be in the [`package.json` that gets generated](https://github.com/guardian/cdk/blob/527057a34a0dc13ec73cbe89114fce6b93b97645/src/bin/commands/new-project/utils/init.ts#L111-L122) by `npx @guardian/cdk@latest new` too/instead?

@akash1810 do you mean a new `script` (like `test:ci`) or for all `test` runs to include `--ci`?

akash1810:
`test:ci` I think, and update the "setting up CI" notes to run that.

I am however more and more curious as to why Jest doesn't detect the environment it's running in - it feels like it should, and we're missing something...

akash1810 (approved):

akash1810:
> I am however more and more curious as to why Jest doesn't detect the environment it's running in - it feels like it should, and we're missing something...

This might be fixed via https://github.com/facebook/jest/pull/12378 (as noted in https://github.com/guardian/cdk/pull/1392/files). Let's update Jest to v29 (latest) first before merging this change in, as it might not be needed.

akash1810:
https://github.com/guardian/aws-account-setup/pull/23 demonstrates that Jest v29 better detects the environment, and we don't need to explicitly pass the `--ci` flag.

georgeblahblah:
> [guardian/aws-account-setup#23](https://github.com/guardian/aws-account-setup/pull/23) demonstrates that Jest v29 better detects the environment, and we don't need to explicitly pass the `--ci` flag.
> 
> (Well, any version of Jest from [v28.0.0](https://github.com/facebook/jest/blob/main/CHANGELOG.md#2800)).

Great! I will close this then!

PR #1511, opened 2022-09-29
test: Purposefully use a different version of `aws-cdk-lib` and `aws-cdk`

akash1810:
This is a demo PR to test the behaviour of using a mismatched version of `aws-cdk` and `aws-cdk-lib`.

See also: https://github.com/guardian/cdk/pull/1512.


akash1810:
Attaching the build logs below for prosperity as GH will only retain logs for 30 days.

[1_ProjectGeneration.txt](https://github.com/guardian/cdk/files/9682847/1_ProjectGeneration.txt)

PR #1512, opened 2022-09-29
ci: Revise the process of updating AWS CDK dependencies

akash1810:
## What does this change?

We want to keep up to date with releases of [AWS CDK](https://github.com/aws/aws-cdk/releases). We currently achieve this with a check in CI (see https://github.com/guardian/cdk/pull/192).

However as we've recently slowed down development in the repository, this requirement is causing friction for contributors. Let's remove it in favour of some automation.

Inspired by https://github.com/cdk-dev/bump-cdk-action, this change removes the CI check and introduces a scheduled job to update aws-cdk, aws-cdk-libs, and constructs on a schedule. The resulting PR will need to be reviewed like any other PR. 

As bump-cdk-action describes:

> Because of the way CDK is compiled, all peer dependencies must be the same version.
> This causes a maintenance headache of not only bumping the versions, but also making sure they go together.

If aws-cdk and aws-cdk-libs are not on the same version, you're greeted with this error upon synth:

```console
This CDK CLI is not compatible with the CDK library used by your application. Please upgrade the CLI to the latest version.
(Cloud assembly schema version mismatch: Maximum schema version supported is 20.0.0, but found 21.0.0)
```

The error from running test is even more cryptic, as seen in https://github.com/guardian/cdk/pull/1511:

```console
Running tests......FAIL lib/integration-test.test.ts
  ● Test suite failed to run

    lib/integration-test.test.ts:9:41 - error TS2345: Argument of type 'IntegrationTest' is not assignable to parameter of type 'Stack'.
      Types of property 'tags' are incompatible.
        Type 'import("/tmp/cdk-project/15331705-5ac0-4f1e-be00-7dbde95012f1/node_modules/aws-cdk-lib/core/lib/tag-manager").TagManager' is not assignable to type 'import("/tmp/cdk-project/15331705-5ac0-4f1e-be00-7dbde95012f1/cdk/node_modules/aws-cdk-lib/core/lib/tag-manager").TagManager'.
          Types have separate declarations of a private property 'tags'.

    9     const template = Template.fromStack(stack);
```

### Dependabot
Ideally we'd use Dependabot to achieve this, however grouped dependencies are not supported. See https://github.com/dependabot/dependabot-core/issues/1296.

### Credentials
The scheduled job is configured to run as the guardian-ci user, for similar reasons that the [CD job](https://github.com/guardian/cdk/blob/527057a34a0dc13ec73cbe89114fce6b93b97645/.github/workflows/cd.yaml) is - PRs raised by the default `secrets.GITHUB_TOKEN` will not trigger CI, as demonstrated in https://github.com/guardian/cdk/pull/1515, whereas a PR raised by the guardian-ci user does trigger the checks, as demonstrated in https://github.com/guardian/cdk/pull/1516.

## How to test
It's a bit tricky as we want to have this job run from main, but the changes will not be in main until we merge.

Maybe one test could be to check [this PR](https://github.com/guardian/cdk/pull/1516) that was raised in testing? However, note it includes extra commits as it was created from this branch and not main.

Less friction for contributors, and more automation.

n/a

## Checklist


.github/workflows/update-aws-cdk.yaml
```diff
@@ -0,0 +1,25 @@
+on:
+  schedule:
+    # At 10:00 on day-of-month 10.
+    # See https://crontab.guru/#0_10_10_*_*
+    - cron: '0 10 10 * *'
```
akash1810, line 5: A totally arbitrary choice.

script/update-aws-cdk
```diff
@@ -1,53 +1,28 @@
...
-    echo 'Error: jq is not installed but required. To install, run `brew install jq` or see https://stedolan.github.io/jq/.' >&2
-    exit 1
-  fi
-}
-
-checkYarnLock() {
-  "${DIR}"/check-yarn-lock
-}
-
-ensureAtHeadOfOrigin() {
```
akash1810, line 24: We checkout from main in the job, so we don't need to form this check.

kenoir (approved):
Definitely better than the check and fail approach.

PR #1518, opened 2022-09-30
ci: Run AWS CDK updater today

akash1810:
Run the AWS CDK updater (see #1512) today at 1PM to test it. Once it runs, we can revert this change.


kenoir (approved):
⏲️

akash1810:
This will run at 13:00 UTC, which is 14:00 BST. See https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule.

PR #1530, opened 2022-10-03
chore(deps): Update Jest to v29

akash1810:
## What does this change?

Bump:
  - jest to 29.1.2
  - @types/jest to 29.1.1
  - ts-jest to 29.0.3

Also included in this change are some Jest configuration changes.

The `transform` change resolves this deprecation warning:

```console
ts-jest[ts-jest-transformer] (WARN) Define `ts-jest` config under `globals` is deprecated. Please do
transform: {
    <transform_regex>: ['ts-jest', { /* ts-jest config goes here in Jest */ }],
},
```

The `snapshotFormat` allows us to retain the snapshot format from Jest 28 (we'll remove in a follow-up PR). See:
  - https://jestjs.io/docs/upgrading-to-jest29#snapshot-format
  - https://jestjs.io/blog/2022/08/25/jest-29

## How to test

See CI.

## How can we measure success?

Keeping up-to-date.

## Have we considered potential risks?

n/a

## Checklist

- [ ] I have listed any breaking changes, along with a migration path [^1]
- [ ] I have updated the documentation as required for the described changes [^2]

[^1]: Consider whether this is something that will mean changes to projects that have already been migrated, or to the CDK CLI tool. If changes are required, consider adding a checklist here and/or linking to related PRs.
[^2]: If you are adding a new construct or pattern, has new documentation been added? If you are amending defaults or changing behaviour, are the existing docs still valid?

PR #1531, opened 2022-10-03
chore: Use default Jest snapshot format

akash1810:
## What does this change?

In #1530 we upgraded to Jest v29. We retained the v28 snapshot format to decrease the diff.

In this change, we adopt the (simpler) snapshot format from v29. See https://jestjs.io/blog/2022/08/25/jest-29.

This should be a no-op.

## How to test

See CI.

Less custom configuration, and simpler snapshot formats.

Other than the big diff for the reviewer, there aren't any.

## Checklist

### DIFF
--- a/src/constructs/acm/__snapshots__/certificate.test.ts.snap
+++ b/src/constructs/acm/__snapshots__/certificate.test.ts.snap
@@ -1,37 +1,37 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`The GuCertificate class should create a new certificate (which requires manual DNS changes) if hosted zone ids are not provided 1`] = `
-Object {
-  "Metadata": Object {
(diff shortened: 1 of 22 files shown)


nicl (approved):

src/constructs/acm/__snapshots__/certificate.test.ts.snap
```diff
@@ -1,37 +1,37 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`The GuCertificate class should create a new certificate (which requires manual DNS changes) if hosted zone ids are not provided 1`] = `
-Object {
-  "Metadata": Object {
-    "gu:cdk:constructs": Array [
+{
+  "Metadata": {
```
akash1810, line 5: These snapshots look more like traditional CloudFormation templates now!

PR #1532, opened 2022-10-04
fix(metadata): Enable reporting on more constructs

akash1810:
## What does this change?

In #1394 we introduced an Aspect to add a Metadata element to the synthesised template. It walks the CDK tree and if it finds a node that starts with "Gu", it adds to the Metadata element.

A node on the CDK tree is a specific type - [`Node`](https://github.com/aws/constructs/blob/4926d6f304442915777db36f039ea7b38c83ad1d/src/construct.ts#L18-L21). In this change, we make some of our custom "constructs" formally extend [`Construct`](https://github.com/aws/constructs/blob/4926d6f304442915777db36f039ea7b38c83ad1d/src/construct.ts#L418-L424). A `Construct` [has a `Node`](https://github.com/aws/constructs/blob/4926d6f304442915777db36f039ea7b38c83ad1d/src/construct.ts#L8-L16), and so appears on the tree.

That is, by extending `Construct` _and_ having a class name prefixed "Gu", we are able to report usage of these constructs.

In this change, we begin to report on the following (and any sub-classes):
- `GuDnsRecordSet `
- `GuEcsTask`
- `GuEc2App`
- `GuApiGatewayWithLambdaByPath`

### Notes
- A `Construct` also has a validate phase, so it might make sense to refactor [some checks](https://github.com/guardian/cdk/blob/59e7ad230158ecf41d96a97790539a404d2e48cd/src/constructs/ecs/ecs-task.ts#L159-L163) to this model, but suggest we do this separately.
- Not everything can be tracked via the Metadata Aspect by simply extending `Construct`, only nodes that sit within/below a `GuStack` can. [`RiffRaffYamlFileExperimental`](https://github.com/guardian/cdk/blob/59e7ad230158ecf41d96a97790539a404d2e48cd/src/experimental/riff-raff-yaml-file/index.ts), for example, sits _above_ a `GuStack` in the tree.
- Is there potential for a lint rule here - any exported class should extend `Construct`? It shouldn't error though!

## How to test

See CI - the snapshots have been updated to reflect the changes.

More usage data.

n/a

## Checklist

### DIFF
--- a/src/constructs/dns/__snapshots__/dns-records.test.ts.snap
+++ b/src/constructs/dns/__snapshots__/dns-records.test.ts.snap
@@ -5,6 +5,7 @@ exports[`The GuCname construct should create the correct resources with minimal 
   "Metadata": {
     "gu:cdk:constructs": [
       "GuStack",
+      "GuCname",
     ],
     "gu:cdk:version": "TEST",
   },
@@ -30,6 +31,7 @@ exports[`The GuDnsRecordSet construct should create the correct resources with m
   "Metadata": {
     "gu:cdk:constructs": [
       "GuStack",
+      "GuDnsRecordSet",
     ],
     "gu:cdk:version": "TEST",
   },
--- a/src/constructs/dns/dns-records.ts
+++ b/src/constructs/dns/dns-records.ts
@@ -1,5 +1,6 @@
 import type { Duration } from "aws-cdk-lib";
 import { CfnResource } from "aws-cdk-lib";
+import { Construct } from "constructs";
 import type { GuDomainName } from "../../types";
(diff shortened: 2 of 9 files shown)


src/constructs/dns/__snapshots__/dns-records.test.ts.snap
```diff
@@ -5,6 +5,7 @@ exports[`The GuCname construct should create the correct resources with minimal
   "Metadata": {
     "gu:cdk:constructs": [
       "GuStack",
+      "GuCname",
```
akash1810, line 8: `GuCname` is a subclass of `GuDnsRecordSet`, so gets tracking "for free".

kenoir (approved):

PR #1539, opened 2022-10-06
feat: Allow overriding of VPC and subnets in the GuEc2App construct

fredex42:
## What does this change?

Adds two new parameters to `GuEc2App` which allow overriding of deployment subnets and VPC:
```
- `vpcOverride` - supply a specific `GuVpc` instance instead of the default one
- `deploymentSubnetsOverride` - supply a list of subnets to deploy into.
```
If you input invalid combinations of vpc and deployment subnets then CDK should complain.

This is needed in order to support CAPI apps in CDK.  CAPI uses a non-standard VPC setup where each of our stacks lives in its own VPC so we need to be able to supply programatically determined, different VPC and subnets groups when building stacks.

- Able to deploy CAPI apps into the existing CAPI networking setup

### Breaking changes
None, the newly introduced properties are optional and the code defaults to the existing behaviour if they are not supplied

## Checklist


src/patterns/ec2-app/base.ts
```diff
@@ -347,6 +349,8 @@ export class GuEc2App extends Construct {
       userData,
       withoutImdsv2,
       imageRecipe,
+      vpcOverride,
```
akash1810, line 352: I think using destructuring with default values here would make it more obvious what the default is, and removes the `??` below. For example:

```ts
const { vpc = THE_DEFAULT } = props;
```

src/patterns/ec2-app/base.ts
```diff
@@ -143,6 +143,8 @@ export interface GuEc2AppProps extends AppIdentity {
   certificateProps: GuDomainName;
   withoutImdsv2?: boolean;
   imageRecipe?: string;
+  vpcOverride?: IVpc;
+  deploymentSubnetsOverride?: ISubnet[];
```
akash1810, line 147: I think `vpc` and `privateSubnets` would be clearer naming here, as the type of `undefined | T` already suggests it's possible to override any defaulted value.

fredex42: OK, fair enough. I shied away from naming it "vpc" due to a naming collision in the constructor function but that is an easy change

src/patterns/ec2-app/base.ts
fredex42: 👍

akash1810 (approved):

akash1810:
Note this didn't trigger a release, with semantic-release [stating](https://github.com/guardian/cdk/actions/runs/3206302178/jobs/5239888973#step:4:234):

> [4:45:14 PM] [semantic-release] › ℹ  There are no relevant changes, so no new version is released.

None of the individual commits followed [conventional commits](https://www.conventionalcommits.org/), but I'd have assumed the merge commit would have satisfied requirements 🤔 . Maybe we should get something like [this](https://github.com/guardian/cdk/pull/1435) working?

PR #1555, opened 2022-10-14
refactor: Convert `CfnInclude` warning from Validation to Aspect

akash1810:
## What does this change?

> Aspects are a way to apply an operation to all constructs in a given scope. The aspect could modify the constructs, such as by adding tags, or it could verify something about the state of the constructs, such as ensuring that all buckets are encrypted.
> – source: https://docs.aws.amazon.com/cdk/v2/guide/aspects.html

An Aspect is more suited for adding this annotation to a stack. See https://github.com/guardian/cdk/pull/1286 for a complete description of this annotation.

## How to test

CI should continue to pass.

More correct use of the CDK framework.

N/A

## Checklist


src/aspects/cfn-include-reporter.ts
```diff
@@ -0,0 +1,18 @@
...
+import { CfnInclude } from "aws-cdk-lib/cloudformation-include";
+import type { IConstruct } from "constructs";
+import { LibraryInfo } from "../constants";
+
+/**
+ * An Aspect to remind users to check for stateful resources when migrating a YAML/JSON template.
+ */
+export class CfnIncludeReporter implements IAspect {
+  public visit(node: IConstruct) {
+    if (node instanceof CfnInclude) {
```
kenoir, line 12: Not blocking but this makes me think, are there any edge cases where `CfnInclude` is used that are not part of a migration?

kenoir (approved):

src/aspects/cfn-include-reporter.ts
akash1810: I'm not familiar with any instance of this. I'd suggest it's bad practice to do so too, as it's mixing two IaC methods.

PR #1566, opened 2022-10-21
feat: Support overriding public subnets in the same way as private ones

fredex42:
## What does this change?

Adds a `publicSubnets` override to the `GuEc2App` constructor.  This is required by the CAPI team, because we must support multiple different VPCs for different stages each of which have their own subnet lists.

## How to test

Add the `publicSubnets` parameter to `GuEc2App` and these should be used in preference to the default VPC's subnets. Normally you would derive these from a StringList SSM parameter

CAPI can support their existing network topology

If incorrectly used, this can result in an un-deployable stack (e.g. the supplied subnets do not correspond to the supplied VPC).  The docs make clear that this feature should only be used if absolutely required though.

## Checklist


src/patterns/ec2-app/base.ts
```diff
@@ -290,7 +291,8 @@ function restrictedCidrRanges(ranges: IPeer[]) {
  * ```
  *
  * If you _must_ support a non-standard VPC setup, it's possible to override the VPC and subnet selection into which
- * instances will be started. This is done by setting the `vpc` and `privateSubnets` props.
+ * instances will be started. This is done by setting the `vpc` and `privateSubnets` props.  You can also override
```
nicl, line 294: I wonder if we could be a bit clearer here that if you set one of these you need to set all of them?

nicl (approved):
Thanks!

src/patterns/ec2-app/base.ts
fredex42: 👍 Updated

PR #1576, opened 2022-10-31
feat: add first attempt at experimental error budget alarm

jacobwinch:
## What does this change?

This PR adds a new experimental construct which aims to create alarms that will notify teams when they are at risk of missing their SLO targets.

The alarm configuration that we create follows the recommendations in the [SRE Workbook](https://sre.google/workbook/alerting-on-slos/) to implement `Multiwindow, Multi-Burn-Rate Alerts`. There is also another nice explanation of this alerting strategy [here](https://developers.soundcloud.com/blog/alerting-on-slos).
 
The CloudWatch implementation of this strategy creates nine separate CloudWatch alarms. The main alarms that users need to care about are the three [composite alarms](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/Create_Composite_Alarm.html). These track different burn rates of error budget consumption: 

* fast (e.g. complete outages)
* medium (e.g. moderate error percentage over hours, rather than minutes) 
* slow (e.g. a small percentage of errors sustained over a much longer period)

Multiple burn rates are used to ensure that we are able to pick up all scenarios which could threaten the SLO, without interrupting developers with an unnecessary level of urgency:

image
_The image above is taken from Google's `Alerting on SLOs` chapter._

Each composite alarm creates two child alarms; these are used to support multiple windows (or periods in CloudWatch terminology). A composite alarm is configured to send notifications if both of its child alarms are in an `ALARM` state. The child alarms themselves have no actions/notifications configured and users should not need to interact with them directly.  Here's an example of how that looks in the AWS console:

image

Multiple windows are helpful for accuracy and reset time. The long window helps us to avoid sending an alert in scenarios where no intervention is required, for example a very brief spike in errors that is resolved automatically (e.g. when AWS replaces a 'bad' EC2 instance). The short window helps to ensure that the alert is reset (or moved back into `OK` state, in CloudWatch terminology) promptly once a problem has been resolved:

image
_The image above is taken from Google's `Alerting on SLOs` chapter._

The composite alarms have an awareness of priority, meaning that low priority alarm notifications are [suppressed](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/Create_Composite_Alarm.html#Create_Composite_Alarm_Suppression) if a higher priority alarm is already firing. That is, if you receive an alert about fast error budget consumption, you should not receive an alert from the medium or slow versions of the alarm. Here's an example of how that looks in the AWS console:

image

Although the generated alarm configuration is quite complex most of this is abstracted away. Developers working on `cdk` code need only supply high level configuration like this:

```typescript
new GuErrorBudgetAlarmExperimental(stack, {
    sloName: "MapiFrontsAvailability",
    sloTarget: 0.999,
    badEvents: new Metric({ metricName: "HttpErrors", namespace: "TestLoadBalancerMetrics" }),
    validEvents: new Metric({ metricName: "HttpRequests", namespace: "TestLoadBalancerMetrics" }),
    snsTopicNameForAlerts: "test-sns-topic",
});
```

## How to test

I have (manually) tested this alert configuration using the `cdk-playground` project to confirm that it works as expected. I've added a couple of snapshot tests to document how this works and keep a record of a working version of the configuration. We'll need to continue to test this with real services so that we can see how it behaves with real traffic patterns and real SLOs etc.

We will attempt to use this construct with the MSS team over the next couple of months. We hope that this approach to alerting will prove to be useful. If we're correct about that (and we decide that CloudWatch is an appropriate tool for this task) then we will look to 'productionise' this construct so that it can be used more widely. If it is not helpful (or if CloudWatch doesn't seem like the best tool for the job), we'll just delete this construct and help MSS to pick up the preferred solution instead.

I think this is pretty low risk. We are experimenting with a new technique and are therefore using an experimental construct, as per [this ADR](https://github.com/guardian/cdk/blob/main/docs/architecture-decision-records/005-x01-package-structure.md).

## Checklist

### DIFF
--- a/src/experimental/constructs/error-budget-alarm.test.ts
+++ b/src/experimental/constructs/error-budget-alarm.test.ts
@@ -1,0 +1,42 @@
+import { Template } from "aws-cdk-lib/assertions";
+import { MathExpression, Metric } from "aws-cdk-lib/aws-cloudwatch";
+import { simpleGuStackForTesting } from "../../utils/test";
+import { GuErrorBudgetAlarmExperimental } from "./error-budget-alarm";
+
+describe("The ErrorBudgetAlarmExperimental construct", () => {
+  it("should create the correct resources with basic config", () => {
+    const stack = simpleGuStackForTesting();
+    new GuErrorBudgetAlarmExperimental(stack, {
+      sloName: "MapiFrontsAvailability",
+      sloTarget: 0.999,
+      badEvents: new Metric({ metricName: "HttpErrors", namespace: "TestLoadBalancerMetrics" }),
+      validEvents: new Metric({ metricName: "HttpRequests", namespace: "TestLoadBalancerMetrics" }),
+      snsTopicNameForAlerts: "test-sns-topic",
+    });
+    expect(Template.fromStack(stack).toJSON()).toMatchSnapshot();
+  });
+
+  it("should accept math expressions for more complicated definitions of bad/valid events", () => {
+    const stack = simpleGuStackForTesting();
+    new GuErrorBudgetAlarmExperimental(stack, {
+      sloName: "MapiFrontsAvailability",
+      sloTarget: 0.999,
+      badEvents: new MathExpression({
+        expression: "loadBalancerErrors + targetErrors",
+        usingMetrics: {
+          loadBalancerErrors: new Metric({ metricName: "LbHttpErrors", namespace: "TestLoadBalancerMetrics" }),
+          targetErrors: new Metric({ metricName: "TargetHttpErrors", namespace: "TestTargetMetrics" }),
+        },
+      }),
+      validEvents: new MathExpression({
+        expression: "allRequests - invalidRequests",
+        usingMetrics: {
+          allRequests: new Metric({ metricName: "HttpRequests", namespace: "TestLoadBalancerMetrics" }),
+          invalidRequests: new Metric({ metricName: "BadRequests", namespace: "TestLoadBalancerMetrics" }),
+        },
+      }),
+      snsTopicNameForAlerts: "test-sns-topic",
+    });
+    expect(Template.fromStack(stack).toJSON()).toMatchSnapshot();
+  });
+});
--- a/src/experimental/constructs/error-budget-alarm.ts
+++ b/src/experimental/constructs/error-budget-alarm.ts
@@ -1,0 +1,208 @@
+import { Duration } from "aws-cdk-lib";
+import type { CfnCompositeAlarm, IMetric } from "aws-cdk-lib/aws-cloudwatch";
+import {
+  Alarm,
+  AlarmRule,
+  AlarmState,
+  ComparisonOperator,
+  CompositeAlarm,
+  MathExpression,
+  TreatMissingData,
+} from "aws-cdk-lib/aws-cloudwatch";
+import { SnsAction } from "aws-cdk-lib/aws-cloudwatch-actions";
+import type { ITopic } from "aws-cdk-lib/aws-sns";
+import { Topic } from "aws-cdk-lib/aws-sns";
+import { Construct } from "constructs";
+import type { GuStack } from "../../constructs/core";
+
+export interface ErrorBudgetAlarmProps {
+  sloName: string;
+  sloTarget: number;
+  badEvents: IMetric;
+  validEvents: IMetric;
+  snsTopicNameForAlerts: string;
+}
+
+interface BurnRate {
+  speed: "Fast" | "Medium" | "Slow";
+  burnRate: number;
+}
+
+interface BurnRateMonitoring extends BurnRate {
+  longPeriod: Duration;
+  shortPeriod: Duration;
+}
+
+interface MonitorBurnRateForPeriodProps {
+  alarmName: string;
+  burnRateSpeed: BurnRate;
+  period: Duration;
+  errorBudget: number;
+  badEvents: IMetric;
+  validEvents: IMetric;
+}
+
+interface CompositeBurnRateAlarmProps {
+  sloName: string;
+  burnRateMonitoring: BurnRateMonitoring;
+  errorBudget: number;
+  badEvents: IMetric;
+  validEvents: IMetric;
+  snsTopic: ITopic;
+  suppressorAlarm?: CompositeBurnRateAlarm;
+}
+
+/**
+ * Use this construct to create alarms that will notify you when you are at risk of missing your SLO targets.
+ *
+ * These alarms follow the recommendations in the [SRE Workbook](https://sre.google/workbook/alerting-on-slos/) to
+ * implement Multiwindow, Multi-Burn-Rate Alerts.
+ *
+ * The CloudWatch implementation of this strategy creates nine separate CloudWatch alarms. The main alarms that users
+ * need to care about are the three composite alarms. These track different burn rates of error budget consumption:
+ * fast (e.g. complete outages), medium (e.g. moderate error percentage over hours, rather than minutes) and slow (e.g.
+ * a small percentage of errors sustained over a much longer period). Multiple burn rates are used to ensure that
+ * we are able to pick up all scenarios which could threaten the SLO, without interrupting developers with an
+ * unnecessary level of urgency.
+ *
+ * Each composite alarm creates two child alarms; these are used to support multiple windows (or periods in CloudWatch
+ * terminology). A composite alarm is configured to send notifications if both of its child alarms are in an alarm state.
+ * The child alarms themselves have no actions/notifications configured and users should not need to interact with them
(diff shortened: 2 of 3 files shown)


src/experimental/constructs/error-budget-alarm.ts
```diff
@@ -0,0 +1,207 @@
...
+ * alert in scenarios where no intervention is required, for example a very brief spike in errors that is resolved
+ * automatically (e.g. when AWS replaces a 'bad' EC2 instance). The short window helps to ensure that the alert is reset
+ * (or moved back into OK state, in CloudWatch terminology) promptly once a problem has been resolved.
+ *
+ * The composite alarms have an awareness of priority, meaning that low priority alarm notifications are suppressed if
+ * a higher priority alarm is already firing. That is, if you receive an alert about fast error budget consumption, you
+ * should not receive an alert from the medium or slow versions of the alarm.
+ */
+export class ErrorBudgetAlarmExperimental {
+  // TODO: Should this extend Construct?
```
jacobwinch, line 80: I noticed that a couple of our other classes (which are not subclasses of an AWS construct) extend `Construct` e.g. [here](https://github.com/guardian/cdk/blob/10762e709cbf9686364ece07b1738b8a493a0b0b/src/patterns/ec2-app/base.ts#L331) and [here](https://github.com/guardian/cdk/blob/043953d45c9337935f5c8ab89fbb072905d3068b/src/constructs/dns/dns-records.ts#L24).

Do I need to do something similar? cc @akash1810 @nicl

kenoir: I believe this enables better tracking of GuCDK usage: https://github.com/guardian/cdk/pull/1532

src/experimental/constructs/error-budget-alarm.ts
```diff
@@ -0,0 +1,207 @@
...
+
+    const snsTopic: ITopic = Topic.fromTopicArn(
+      scope,
+      `SnsSloAlarmsFor${props.sloName}`,
+      `arn:aws:sns:${scope.region}:${scope.account}:${props.snsTopicNameForAlerts}`
+    );
+
+    const fastBurnRate: BurnRateMonitoring = {
+      speed: "Fast",
+      burnRate: 14.4,
```
kenoir, line 93: How have these burn rates been arrive at?

kenoir:
This is pretty nice, good explanation of what's going on too. You might be able to represent the `AND` nature of the composite alarms and the cascading priorities of long < medium < fast neatly with a diagram to get the point across even quicker.

src/experimental/constructs/error-budget-alarm.ts
jacobwinch: Ah, thanks for the link / explanation! I'm not sure that we strictly need this yet but I've added it anyway as it was trivial to change this (and it saves me from forgetting about it later!)

src/experimental/constructs/error-budget-alarm.ts
jacobwinch: Good question - I should have mentioned this in the PR description. Fast/medium are just based on [Google's recommendation](https://sre.google/workbook/alerting-on-slos/#recommended_parameters_for_an_slo_based_a), as that seems like a sensible place to start.

Unfortunately we cannot use their recommendation for slow burn rate due to a CloudWatch limitation (the maximum period that you can monitor with a CloudWatch alarm is 1 day), so I have used the other example that they give for a slow burn alarm instead for now.

jacobwinch:
> You might be able to represent the `AND` nature of the composite alarms and the cascading priorities of long < medium < fast neatly with a diagram to get the point across even quicker.

Thanks for the suggestion. I've added a couple of screenshots from the AWS console as that helped me to visualise what was going on when I was first configuring this!

michaelwmcnamara:
This might be something for a future pr, but I think we should have some validation or tests to check that the alarm a user has defined will actually be able to trigger - ie if the error budget they define is too tolerant or the assessment period they define is too short such that the triggering condition could not be reached with a 100% outage, we should have something to let them know - given this is a new concept and all.

michaelwmcnamara (approved):
Fantastic work! 
From an SLO monitoring point of view this looks great! My only suggestion is mentioned above and could be addressed in a separate PR. Provided the Ops point of view is happy with this, it has my approval!

src/experimental/constructs/error-budget-alarm.test.ts
```diff
@@ -0,0 +1,42 @@
...
+      sloName: "MapiFrontsAvailability",
+      sloTarget: 0.999,
+      badEvents: new Metric({ metricName: "HttpErrors", namespace: "TestLoadBalancerMetrics" }),
+      validEvents: new Metric({ metricName: "HttpRequests", namespace: "TestLoadBalancerMetrics" }),
+      snsTopicNameForAlerts: "test-sns-topic",
+    });
+    expect(Template.fromStack(stack).toJSON()).toMatchSnapshot();
+  });
+
+  it("should accept math expressions for more complicated definitions of bad/valid events", () => {
```
jorgeazevedo, line 19: Great test, it documents an important aspect of this.

jorgeazevedo:
Thanks for adding the screenhots from AWS Jacob. I was having troubling mapping this in my mind, but with the screenshots I found it much easier to follow!

I like the constructor interface, where the clients passes the metrics that will form the basis for the ratio. This mimics the [SLI definition interface in GCP](https://youtu.be/uhpAScSerec?t=130).
Screenshot 2022-10-31 at 13 29 32

jorgeazevedo (approved):

jacobwinch:
> This might be something for a future pr, but I think we should have some validation or tests to check that the alarm a user has defined will actually be able to trigger - ie if the error budget they define is too tolerant or the assessment period they define is too short such that the triggering condition could not be reached with a 100% outage, we should have something to let them know - given this is a new concept and all.

Thanks for this suggestion @michaelwmcnamara. I agree that we should add this; I also think that this could be introduced via a follow-up PR, so I've created a card for this: https://trello.com/c/5bgpvcVP/1549-error-budget-alerts-add-validation-for-very-low-or-very-high-slo-targets.

src/experimental/constructs/error-budget-alarm.ts
michaelwmcnamara: There is a strong scaling argument for keeping these burn-rates standardised as much as possible - as the number of alerts increases it becomes more and more confusing to manage different burn rates. 
One approach could be to have a standard construct which we treat as the recommended approach and a custom construct where burn rates could be manually set.
Another approach ,suggested by Google, is to class error budget alerts into different families or buckets and set standard burn specific to each family's needs. See [here](https://sre.google/workbook/alerting-on-slos/#alerting_at_scale) for details. Recommend we look at these approaches in future PR's though.

src/experimental/constructs/error-budget-alarm.ts
akash1810: There might be a need to signpost this more clearly, either in the contributing docs or in a features page.

src/experimental/constructs/error-budget-alarm.ts
```diff
@@ -0,0 +1,208 @@
...
+ * terminology). A composite alarm is configured to send notifications if both of its child alarms are in an alarm state.
+ * The child alarms themselves have no actions/notifications configured and users should not need to interact with them
+ * directly. Multiple windows are helpful for accuracy and reset time. The long window helps us to avoid sending an
+ * alert in scenarios where no intervention is required, for example a very brief spike in errors that is resolved
+ * automatically (e.g. when AWS replaces a 'bad' EC2 instance). The short window helps to ensure that the alert is reset
+ * (or moved back into OK state, in CloudWatch terminology) promptly once a problem has been resolved.
+ *
+ * The composite alarms have an awareness of priority, meaning that low priority alarm notifications are suppressed if
+ * a higher priority alarm is already firing. That is, if you receive an alert about fast error budget consumption, you
+ * should not receive an alert from the medium or slow versions of the alarm.
```
nicl, line 78: This is really nice. Alarm overload is a real thing and can get quite confusing during an incident.

src/experimental/constructs/error-budget-alarm.ts
```diff
@@ -0,0 +1,208 @@
...
+ */
+export class GuErrorBudgetAlarmExperimental extends Construct {
+  constructor(scope: GuStack, props: ErrorBudgetAlarmProps) {
+    super(scope, props.sloName); // The assumption here is that `sloName` will be unique (per stack)
```
nicl, line 82: Is it worth prefixing with something to guarantee this is unique? - not sure if the ID gets used for much, but if not then this could be a sensible small improvement here.

nicl (commented):
Haven't gone through line by line, but the approach sounds really interesting - can see a lot of value in the long/short alarm combo here, and the linked article helps explain the benefits.

PR #1580, opened 2022-11-01
fix: Add restriction to the slo target value

jorgeazevedo:
## What does this change?

This makes `GuErrorBudgetAlarmExperimental` throw an error if the client passes an SLO target outside the valid range, which is 95% to 99.95%.

This is because outside of these bounds, the way the alarms are configured do not produce useful results. For example, with an SLO target of 90%, the fast alarm will never trigger (this is the one that work on an hour-long time window).

If the SLO target is overambitious, for example 99.999%, then the alarm will fire but only after the SLO target has been missed, and its error budget consumed (in an hour long window, this would happen in 26 seconds). Because this alert wouldn't be actionable, we don't recommend people use it. For this level of availability, a totally different approach needs to be taken.

## How to test

We have added unit tests to cover the different boundary conditions.

This is low risk because it's an experimental construct.

## Checklist

### DIFF
--- a/src/experimental/constructs/error-budget-alarm.test.ts
+++ b/src/experimental/constructs/error-budget-alarm.test.ts
@@ -26,7 +26,7 @@ describe("The ErrorBudgetAlarmExperimental construct", () => {
         validEvents: new Metric({ metricName: "HttpRequests", namespace: "TestLoadBalancerMetrics" }),
         snsTopicNameForAlerts: "test-sns-topic",
       });
-    }).toThrowError();
+    }).toThrowError("ErrorBudgetAlarm only works with SLO targets between 0.95 and 0.9995");
   });
 
   it("should fail to create Alarm when slo target is below 95%", () => {
@@ -39,7 +39,7 @@ describe("The ErrorBudgetAlarmExperimental construct", () => {
         validEvents: new Metric({ metricName: "HttpRequests", namespace: "TestLoadBalancerMetrics" }),
         snsTopicNameForAlerts: "test-sns-topic",
       });
-    }).toThrowError();
+    }).toThrowError("ErrorBudgetAlarm only works with SLO targets between 0.95 and 0.9995");
   });
 
   it("should accept math expressions for more complicated definitions of bad/valid events", () => {
--- a/src/experimental/constructs/error-budget-alarm.test.ts
+++ b/src/experimental/constructs/error-budget-alarm.test.ts
@@ -16,6 +16,32 @@ describe("The ErrorBudgetAlarmExperimental construct", () => {
     expect(Template.fromStack(stack).toJSON()).toMatchSnapshot();
   });
 
+  it("should fail to create Alarm when slo target is above 99.95%", () => {
+    const stack = simpleGuStackForTesting();
+    expect(() => {
+      new GuErrorBudgetAlarmExperimental(stack, {
+        sloName: "MapiFrontsAvailability",
+        sloTarget: 0.9996,
+        badEvents: new Metric({ metricName: "HttpErrors", namespace: "TestLoadBalancerMetrics" }),
+        validEvents: new Metric({ metricName: "HttpRequests", namespace: "TestLoadBalancerMetrics" }),
+        snsTopicNameForAlerts: "test-sns-topic",
+      });
+    }).toThrowError();
+  });
+
+  it("should fail to create Alarm when slo target is below 95%", () => {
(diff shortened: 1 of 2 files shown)


src/experimental/constructs/error-budget-alarm.ts
```diff
@@ -15,6 +15,16 @@ import { Topic } from "aws-cdk-lib/aws-sns";
...
 import type { GuStack } from "../../constructs/core";
 
+/**
+ * Props for creating an ErrorBudgetAlarm
+ *
+ * - sloName: The name of the SLO
+ * - sloTarget: The percentange target for the SLO, expressed as a decimal floating point number. Must be between 0.95 and 0.995
+ * - badEvents: the metric representing the bad events. Can be any IMetric type, like Metric or MathExpression
+ * - validEvents: the metric representing the events to be considered. Can be any IMetric type, like Metric or MathExpression
+ * - snsTopicNameForAlerts: the sns topic used by the alarm to send email alerts
```
jacobwinch, line 25: Minor suggestion:

```suggestion
 * - snsTopicNameForAlerts: the sns topic name (not the full arn) used by the alarm to send email alerts
```

src/experimental/constructs/error-budget-alarm.ts
```diff
@@ -15,6 +15,16 @@ import { Topic } from "aws-cdk-lib/aws-sns";
 import { Construct } from "constructs";
 import type { GuStack } from "../../constructs/core";
 
+/**
+ * Props for creating an ErrorBudgetAlarm
```
jacobwinch, line 19: Thanks for adding these docs @jorgeazevedo & @michaelwmcnamara 💯

jacobwinch (approved):

src/experimental/constructs/error-budget-alarm.ts
akash1810: Could we move these docs to be inlined to help [the TypeDoc site](https://guardian.github.io/cdk/)?

```ts
export interface ErrorBudgetAlarmProps {
  /**
   * The name of the SLO. Assumed to be unique per stack.
   */
  sloName: string;

  /**
   * The percentange target for the SLO, expressed as a decimal floating point number. Must be between 0.95 and 0.995.
   */
  sloTarget: number;

  badEvents: IMetric;
  validEvents: IMetric;
  snsTopicNameForAlerts: string;
}
```

src/experimental/constructs/error-budget-alarm.test.ts
```diff
@@ -16,6 +16,32 @@ describe("The ErrorBudgetAlarmExperimental construct", () => {
...
+    const stack = simpleGuStackForTesting();
+    expect(() => {
+      new GuErrorBudgetAlarmExperimental(stack, {
+        sloName: "MapiFrontsAvailability",
+        sloTarget: 0.9996,
+        badEvents: new Metric({ metricName: "HttpErrors", namespace: "TestLoadBalancerMetrics" }),
+        validEvents: new Metric({ metricName: "HttpRequests", namespace: "TestLoadBalancerMetrics" }),
+        snsTopicNameForAlerts: "test-sns-topic",
+      });
+    }).toThrowError();
```
akash1810, line 29: Can we add the error message that's thrown to the test, to be sure only the expected error is seen?

src/experimental/constructs/error-budget-alarm.test.ts
```diff
@@ -16,6 +16,32 @@ describe("The ErrorBudgetAlarmExperimental construct", () => {
...
+    const stack = simpleGuStackForTesting();
+    expect(() => {
+      new GuErrorBudgetAlarmExperimental(stack, {
+        sloName: "MapiFrontsAvailability",
+        sloTarget: 0.9499,
+        badEvents: new Metric({ metricName: "HttpErrors", namespace: "TestLoadBalancerMetrics" }),
+        validEvents: new Metric({ metricName: "HttpRequests", namespace: "TestLoadBalancerMetrics" }),
+        snsTopicNameForAlerts: "test-sns-topic",
+      });
+    }).toThrowError();
```
akash1810, line 42: Can we add the error message that's thrown to the test, to be sure only the expected error is seen?

src/experimental/constructs/error-budget-alarm.ts
jacobwinch: Tangentially related PR: https://github.com/guardian/cdk/pull/1587!

PR #1591, opened 2022-11-04
feat: Allow users of GuApiGatewayWithLambdaByPathProps construct to require an apiKey on resources

rupertbates:
## What does this change?
The `GuApiGatewayWithLambdaByPathProps` construct provides an easy way to set up an Api Gateway endpoint backed by a lambda, however there was previously no way to specify that this endpoint should require an api key. 

This PR adds that functionality by adding an optional `apiKeyRequired` property to the `ApiTarget` interface used to specify endpoints.

## How to test
There is a Jest test provided

Users of the `GuApiGatewayWithLambdaByPathProps` construct should be able to create endpoints secured by an api key without additional code.

None that I can see, the property is optional so behaviour for existing clients is unchanged

## Checklist


nicl (approved):
Looks good!

PR #1606, opened 2022-11-11
fix(pattern-tracking): Remove `gu:cdk:pattern-name` tag

akash1810:
## What does this change?

The `gu:cdk:pattern-name` tag was added (#603) to patterns to aid usage tracking[^1].

In #1394 more exhaustive tracking was added via CFN Metadata. That is, we no longer need to apply the `gu:cdk:pattern-name` tag.

❓ Should we remove this field from Prism too? See https://github.com/guardian/prism/pull/224.

[^1]: It was only added to the EC2 pattern, so tracking was only partial.

## How to test

Observe CI.

Less code!

n/a

## Checklist

- [ ] I have listed any breaking changes, along with a migration path [^2]
- [ ] I have updated the documentation as required for the described changes [^3]

### DIFF
--- a/src/constants/metadata-keys.ts
+++ b/src/constants/metadata-keys.ts
@@ -2,7 +2,6 @@ export const MetadataKeys = {
   VERSION: "gu:cdk:version",
   CONSTRUCTS_KEY: "gu:cdk:constructs",
   REPOSITORY_NAME: "gu:repo",
-  PATTERN_NAME: "gu:cdk:pattern-name",
   LOG_KINESIS_STREAM_NAME: "LogKinesisStreamName",
   SYSTEMD_UNIT: "SystemdUnit",
 };
--- a/src/patterns/ec2-app/__snapshots__/base.test.ts.snap
+++ b/src/patterns/ec2-app/__snapshots__/base.test.ts.snap
@@ -88,11 +88,6 @@ exports[`the GuEC2App pattern can produce a restricted EC2 app locked to specifi
             "Value": "test-gu-ec2-app",
           },
           {
-            "Key": "gu:cdk:pattern-name",
-            "PropagateAtLaunch": true,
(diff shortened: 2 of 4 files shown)


kenoir:
I think by removing this we loose the ability to identify whether a particular ASG or Lambda is using a `Gu` pattern? The tracking added in [cdk-metadata](https://github.com/guardian/cdk/pull/1394) is good for tracking in aggregate but can't tell us if a particular resource is part of a GuCDK construct.

Is my understanding here correct and do we mind loosing the ability to identify a particular resource as using a GuCDK construct?

akash1810:
> I think by removing this we loose the ability to identify whether a particular ASG or Lambda is using a `Gu` pattern? The tracking added in [cdk-metadata](https://github.com/guardian/cdk/pull/1394) is good for tracking in aggregate but can't tell us if a particular resource is part of a GuCDK construct.

I don't think we have this today. The most we have is knowledge if an ASG originates from the `GuEc2App` pattern. Every other resource in a stack is anonymous.

I think getting this information is possible. CDK holds an internal [tree representation](https://docs.aws.amazon.com/cdk/v2/guide/constructs.html#constructs_tree) of all the constructs in a stack. If we included path metadata (or analysed the `tree.json` files generated alongside CFN templates), we should gain this insight.

However, a [previous decision](https://github.com/guardian/cdk/blob/main/docs/architecture-decision-records/002-component-constuctors.md#decision) currently prevents this as we're adding all constructs to the _same_ node in the tree - `GuStack`.

Revising this decision would be a breaking change, as a construct's logicalId is derived from it's path in the tree. We do have [ways to protect stateful resources](https://github.com/guardian/cdk/blob/3fb2df0592e0499cd77f26ce46c11479b9db5fff/src/constructs/core/stack.ts#L176-L238) though.

kenoir (approved):

PR #1619, opened 2022-11-16
docs: add example alerts for Availability and Latency SLOs

jacobwinch:
## What does this change?

We've recently been doing some discovery work to ascertain the correct CloudWatch metrics/statistics for typical Availability or Latency SLOs.

This PR adds tests so that we keep working examples which demonstrate how to configure error budget alarms for both of these common SLI types.

## How to test

This PR only adds tests 😀 

I've deployed the CloudFormation that gets generated by these examples for `cdk-playground` and confirmed that the alerts fire under expected scenarios (e.g. deliberately serving a percentage of errors or slow requests).

We have working examples of alerts for these SLI types documented via unit tests.

The risk here is minor, but documenting these via tests will slow down CI slightly (IMO this is worth it because the tests can't go out of date whereas standard documentation is likely to!).

## Checklist

### DIFF
--- a/src/experimental/constructs/error-budget-alarm.test.ts
+++ b/src/experimental/constructs/error-budget-alarm.test.ts
@@ -1,5 +1,9 @@
 import { Template } from "aws-cdk-lib/assertions";
 import { MathExpression, Metric } from "aws-cdk-lib/aws-cloudwatch";
+import { InstanceClass, InstanceSize, InstanceType } from "aws-cdk-lib/aws-ec2";
+import { HttpCodeElb, HttpCodeTarget } from "aws-cdk-lib/aws-elasticloadbalancingv2";
+import { AccessScope } from "../../constants";
+import { GuEc2App } from "../../patterns";
 import { simpleGuStackForTesting } from "../../utils/test";
 import { GuErrorBudgetAlarmExperimental } from "./error-budget-alarm";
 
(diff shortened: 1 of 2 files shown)


src/experimental/constructs/error-budget-alarm.test.ts
```diff
@@ -16,6 +20,66 @@ describe("The ErrorBudgetAlarmExperimental construct", () => {
...
+      },
+      scaling: {
+        minimumInstances: 1,
+      },
+    });
+    new GuErrorBudgetAlarmExperimental(stack, {
+      sloName: "MapiFrontsLatency",
+      sloTarget: 0.999,
+      badEvents: ec2App.loadBalancer.metricTargetResponseTime({
+        statistic: "TC(0.5:)" // This gets a count of slow requests i.e. the number of requests that completed in 0.5 seconds or more
```
jacobwinch, line 75: N.B. it's possible to perform the entire calculation `(badEvents/validEvents) * 100` using the [PR statistic](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/Statistics-definitions.html) `PR(0.5:)`.

Using `TC` allows us to use a standard construct regardless of the SLI type and also means that users can provide a more complex definition of valid events (e.g. using metric math to exclude certain response codes).

src/experimental/constructs/error-budget-alarm.test.ts
```diff
@@ -16,6 +20,66 @@ describe("The ErrorBudgetAlarmExperimental construct", () => {
...
+      monitoringConfiguration: { noMonitoring: true },
+      userData: "#!/bin/dev foobarbaz",
+      certificateProps: {
+        domainName: "code-guardian.com",
+      },
+      scaling: {
+        minimumInstances: 1,
+      },
+    });
+    new GuErrorBudgetAlarmExperimental(stack, {
```
jacobwinch, line 39: If desired, we could extend this class in the future (e.g. `GuErrorBudgetAvailabilityAlarmExperimental` and `GuErrorBudgetLatencyAlarmExperimental`) to pre-populate these metrics for these common SLI types.

What do you think @michaelwmcnamara and @jorgeazevedo?

src/experimental/constructs/error-budget-alarm.test.ts
jacobwinch, line 75: We might want to provide a nicer interface for users here. For example, the interface could ask users to input the `Duration` for a good event (i.e. should it be 200ms, 300ms, 1000ms etc.) and the construct could write out the statistic part for them

src/experimental/constructs/error-budget-alarm.test.ts
```diff
@@ -16,6 +20,66 @@ describe("The ErrorBudgetAlarmExperimental construct", () => {
...
+      sloName: "MapiFrontsLatency",
+      sloTarget: 0.999,
+      badEvents: ec2App.loadBalancer.metricTargetResponseTime({
+        statistic: "TC(0.5:)", // This gets a count of slow requests i.e. the number of requests that completed in 0.5 seconds or more
```
jacobwinch, line 75: N.B. it's possible to perform the entire calculation `(badEvents/validEvents) * 100` using the [PR statistic](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/Statistics-definitions.html) `PR(0.5:)`.

Using `TC` allows us to use a standard construct regardless of the SLI type and also means that users can provide a more complex definition of valid events (e.g. using metric math to exclude certain response codes).

src/experimental/constructs/error-budget-alarm.test.ts
```diff
@@ -16,6 +20,66 @@ describe("The ErrorBudgetAlarmExperimental construct", () => {
...
+      userData: "#!/bin/dev foobarbaz",
+      certificateProps: {
+        domainName: "code-guardian.com",
+      },
+      scaling: {
+        minimumInstances: 1,
+      },
+    });
+    new GuErrorBudgetAlarmExperimental(stack, {
+      sloName: "MapiFrontsLatency",
```
jacobwinch, line 72: We might want to provide a nicer interface for users here. For example, the interface could ask users to input the `Duration` for a good event (i.e. should it be 200ms, 300ms, 1000ms etc.) and the construct could write out the statistic part for them

src/experimental/constructs/error-budget-alarm.test.ts
jorgeazevedo: Yes! We can essentially document/standardise how to compute the bad events/valid events for these known cases like elb availability and latency by providing those classes

jorgeazevedo (approved):
Nice one! Good to write these down while they're still fresh in our minds

PR #1621, opened 2022-11-19
ci: Remove unnecessary Snyk schedule

kenoir:
## What does this change?

Removes the Snyk schedule, which is unnecessary and will cause the workflow to become suspended after 60 days of inactivity on the repo!


akash1810:
CI is failing; https://github.com/guardian/cdk/pull/1620 fixes it.

nicl (approved):

PR #1649, opened 2022-12-12
refactor(GuSnsTopic)!: Remove `GuSnsTopic` construct

akash1810:
## What does this change?

The `GuSnsTopic` simply extends AWS's `Topic`, and doesn't add any additional functionality.

In this change, we remove this construct in favour of users using the AWS construct directly.

BREAKING CHANGE: `GuSnsTopic` is removed.

Use AWS's `Topic` directly instead.

```ts
// before
import { GuSnsTopic } from "../sns";
new GuSnSTopic();

// after
import { Topic } from "aws-cdk-lib/aws-sns";
new Topic();
```

## How to test

CI passing should be enough here.

Less code to maintain.

n/a

## Checklist


kenoir (approved):
LGTM 👍

PR #1657, opened 2022-12-16
feat: Support `aws-s3` Riff-Raff deployments

akash1810:
## What does this change?

In https://github.com/guardian/riff-raff/pull/948 Riff-Raff gained the ability to reference an S3 bucket using an SSM Parameter path 🎉 .

In this change we introduce a new construct: `GuS3OriginBucket`, creating:
  - [`AWS::S3::Bucket`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket.html) - a bucket destined to be an origin for a static site, whether that be a Fastly, CloudFront, etc.
  - [`AWS::SSM::Parameter`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ssm-parameter.html) - a Parameter, whose value is the bucket name. Downstream services should reference the bucket using this.

We also add [`RiffRaffYamlFileExperimental`](https://github.com/guardian/cdk/tree/e06109412032bcb8eca51a753a3e4d0bcce83c46/src/experimental/riff-raff-yaml-file) support for the [`aws-s3` Riff-Raff deployment type](https://riffraff.gutools.co.uk/docs/magenta-lib/types#awss3), creating a `riff-raff.yaml` file that (in it's simplest form) contains (annotations added for illustration only):

```yaml
allowedStages:
  - CODE
  - PROD
deployments:
  cfn-eu-west-1-test-my-application-stack:
    type: cloud-formation
    regions:
      - eu-west-1
    stacks:
      - test
    app: my-application-stack
    contentDirectory: /tmp/cdk.out
    parameters:
      templateStagePaths:
        CODE: my-stack-CODE.template.json
        PROD: my-stack-PROD.template.json
  upload-static-files-eu-west-1-test-my-app:
    type: aws-s3 # <-- NEW!
    actions:
      - uploadStaticFiles
    dependencies:
      - cfn-eu-west-1-test-my-application-stack
    stacks:
      - test
    regions:
      - eu-west-1
    app: my-app
    contentDirectory: my-app # <-- Files in this directory within the Riff-Raff bundle will be uploaded to S3.
    parameters:
      bucketSsmKeyStageParam:
        CODE: /CODE/test/my-app/my-app-origin-bucket # <-- Stage aware!
        PROD: /PROD/test/my-app/my-app-origin-bucket
      publicReadAcl: false # <-- Never create per object ACLs. If the bucket content needs to be public, it’s better to control this via the bucket policy.
      cacheControl: private # <-- Default.
```

The above `riff-raff.yaml` is generated from:

```ts
// lib/my-application-stack.ts
class MyApplicationStack extends GuStack {
  // eslint-disable-next-line custom-rules/valid-constructors -- unit testing
  constructor(app: App, id: string, props: GuStackProps) {
    super(app, id, props);

    new GuS3OriginBucket(this, "my-static-site", {
      app: "my-app",
    });
  }
}

// bin/cdk.ts
const app = new GuRootExperimental();

new MyApplicationStack(app, "my-stack-CODE", { stack: "test", stage: "CODE", env: { region: "eu-west-1" } });
new MyApplicationStack(app, "my-stack-PROD", { stack: "test", stage: "PROD", env: { region: "eu-west-1" } });
```

You can optionally also set the following [`aws-s3` deployment parameters](https://riffraff.gutools.co.uk/docs/magenta-lib/types#awss3):
  - cacheControl
  - surrogateControl
  - mimeTypes

## How to test

- See the added tests.
- [Validate](https://riffraff.gutools.co.uk/configuration/validation) the generated `riff-raff.yaml`

More generation. Less frustration.

- ❓ Should an `AWS::SSM::Parameter` be created as standard in `GuS3Bucket`?

- ❓ Is `GuS3OriginBucket` an appropriate name?

  The name `GuS3OriginBucket` suggests this will only work for static site type content. However, it can technically also be used for those cases where we use the `aws-s3` deployment type for configuration, for example the [DevX ELK stack](https://github.com/guardian/deploy-tools-platform/blob/main/elk/riff-raff.yaml).

## Checklist

### DIFF
--- a/src/constructs/s3/index.ts
+++ b/src/constructs/s3/index.ts
@@ -1,6 +1,7 @@
 import { RemovalPolicy } from "aws-cdk-lib";
 import type { BucketProps } from "aws-cdk-lib/aws-s3";
 import { BlockPublicAccess, Bucket } from "aws-cdk-lib/aws-s3";
+import { StringParameter } from "aws-cdk-lib/aws-ssm";
 import { GuAppAwareConstruct } from "../../utils/mixin/app-aware-construct";
 import type { AppIdentity, GuStack } from "../core";
 
@@ -16,6 +17,50 @@ export class GuS3Bucket extends GuAppAwareConstruct(Bucket) {
       publicReadAccess: false,
       blockPublicAccess: BlockPublicAccess.BLOCK_ALL,
       ...props,
+    });
+  }
+}
+
+export interface GuS3OriginBucketProps extends GuS3BucketProps {
+  /**
+   * Customises the SSM Parameter path, making it stage agnostic.
+   * Useful when a single bucket is used across multiple stages.
+   *
+   * When `false` the path will be `/<STAGE>/<STACK>/<APP>/<APP>-origin-bucket`.
+   * When `true` the path will be `/<STACK>/<APP>/<APP>-origin-bucket`.
+   *
+   * @default false
+   */
+  withoutStageAwareness?: boolean;
+}
+
+/**
+ * An S3 bucket for use as an origin for a static site.
+ * The bucket name will be stored in an SSM Parameter.
+ * Other infrastructure, or applications, that use this bucket should discover its name by the Parameter.
+ */
+export class GuS3OriginBucket extends GuS3Bucket {
+  /**
+   * The name of the created SSM Parameter.
+   */
+  public readonly parameterName: string;
+
+  constructor(scope: GuStack, id: string, props: GuS3OriginBucketProps) {
+    const { stage, stack } = scope;
+    const { app, withoutStageAwareness = false } = props;
+
+    super(scope, id, props);
+
+    const stageAgnosticParameterParts = [stack, app, `${app}-origin-bucket`];
+    const parameterParts = withoutStageAwareness
+      ? stageAgnosticParameterParts
+      : [stage, ...stageAgnosticParameterParts];
+
+    this.parameterName = `/${parameterParts.join("/")}`;
+
+    new StringParameter(scope, `${id}-ssm-param`, {
+      parameterName: this.parameterName,
+      stringValue: this.bucketName,
     });
   }
 }
--- a/src/constructs/s3/index.ts
+++ b/src/constructs/s3/index.ts
@@ -1,7 +1,9 @@
+import type { Duration } from "aws-cdk-lib";
 import { RemovalPolicy } from "aws-cdk-lib";
 import type { BucketProps } from "aws-cdk-lib/aws-s3";
 import { BlockPublicAccess, Bucket } from "aws-cdk-lib/aws-s3";
 import { StringParameter } from "aws-cdk-lib/aws-ssm";
+import type { IConstruct } from "constructs";
 import { GuAppAwareConstruct } from "../../utils/mixin/app-aware-construct";
 import type { AppIdentity, GuStack } from "../core";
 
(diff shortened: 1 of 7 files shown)


src/constructs/s3/index.ts
```diff
@@ -19,3 +22,130 @@ export class GuS3Bucket extends GuAppAwareConstruct(Bucket) {
     });
   }
 }
+
+interface GuS3OriginBucketDeploymentProps {
```
akash1810, line 26: Only a subset of the available [parameters](https://riffraff.gutools.co.uk/docs/magenta-lib/types#awss3) are configurable, on the basis that the rest have sensible defaults and only impact the directory structure of the bucket's content.

src/constructs/s3/index.ts
```diff
@@ -19,3 +22,130 @@ export class GuS3Bucket extends GuAppAwareConstruct(Bucket) {
...
+  /**
+   * Customises the SSM Parameter path, making it stage agnostic.
+   * Useful when a single bucket is used across multiple stages.
+   *
+   * When `false` the path will be `/<STAGE>/<STACK>/<APP>/<APP>-origin-bucket`.
+   * When `true` the path will be `/<STACK>/<APP>/<APP>-origin-bucket`.
+   *
+   * @default false
+   */
+  withoutStageAwareness: boolean;
```
akash1810, line 36: Also the [recommendation](https://github.com/guardian/recommendations/discussions/74) is to have a bucket per stage, the `aws-s3` deployment type has a `prefixStack` parameter, defined as:

```
Prefix the S3 bucket key with the target stack. Default true.
```

That is, I suspect historic usage of `aws-s3` would indicate that we have a shared bucket.

src/experimental/riff-raff-yaml-file/deployments/s3.ts
```diff
@@ -0,0 +1,61 @@
...
+    props: {
+      type: "aws-s3",
+      actions: ["uploadStaticFiles"],
+      dependencies: [cfnDeployName],
+      stacks: new Set([stack]),
+      regions: new Set([region]),
+      app,
+      contentDirectory: app,
+      parameters: {
+        bucketSsmKeyStageParam: stages.reduce(
```
akash1810, line 39: For ease, always set `bucketSsmKeyStageParam`, even if the same bucket is across stages, for example:

Stage aware:
```yaml
bucketSsmKeyStageParam:
  CODE: /CODE/test/my-bucket/shared-bucket-origin-bucket
  PROD: /PROD/test/my-bucket/shared-bucket-origin-bucket
```

Shared:
```yaml
bucketSsmKeyStageParam:
  CODE: /test/shared-bucket/shared-bucket-origin-bucket
  PROD: /test/shared-bucket/shared-bucket-origin-bucket
```

akash1810:
cc @Fweddi

src/experimental/riff-raff-yaml-file/deployments/s3.ts
akash1810: Sharing wouldn't work as we'd deploy CFN in both stages, and both will want to create the Parameter on `/test/shared-bucket/shared-bucket-origin-bucket`, which would fail!

bryophyta: > Sharing wouldn't work as we'd deploy CFN in both stages, and both will want to create the Parameter on `/test/shared-bucket/shared-bucket-origin-bucket`, which would fail!

Given this, it seems like we should remove the `withoutStageAwareness` prop, and the ternary expression here which checks it?

```
[stage]: withoutStageAwareness ? baseParameterName : `/${stage}${baseParameterName}`,
```

Or am I misunderstanding the above comment?

PR #1674, opened 2023-01-03
feat: add Google Auth to EC2 patterns

nicl:
*Resurrection of https://github.com/guardian/cdk/pull/1608. Now that we know it is possible to limit access to just `guardian.co.uk` emails at the Google level, I am less concerned that teams will fail to validate the token correctly. Nonetheless, this is still a risk here if teams do not effectively limit network access to their instances.*

*See also https://github.com/guardian/deploy-tools-platform/pull/634 - an example of us applying this recently to Riffraff.*

Introduces support for Google Auth as part of our GuEc2App patterns. Auth is handled at the ALB layer via AWS's OIDC support.

Note, to secure access, in addition to using this new feature, teams MUST:

* ensure that User Access is set to 'Internal' in the Google Console (under Consent Screen)
* ensure that network access to EC2 instances is limited to the ALB (via security groups)

This pattern, by default, provides the second out of the box, but cannot do the first.

For more information, see:
https://docs.aws.amazon.com/elasticloadbalancing/latest/application/listener-authenticate-users.html.

### DIFF
--- a/src/patterns/ec2-app/base.ts
+++ b/src/patterns/ec2-app/base.ts
@@ -1,10 +1,11 @@
-import { Duration, Tags } from "aws-cdk-lib";
+import { Duration, SecretValue, Tags } from "aws-cdk-lib";
 import type { BlockDevice } from "aws-cdk-lib/aws-autoscaling";
 import { HealthCheck } from "aws-cdk-lib/aws-autoscaling";
 import type { InstanceType, IPeer, ISubnet, IVpc } from "aws-cdk-lib/aws-ec2";
 import { Port } from "aws-cdk-lib/aws-ec2";
-import { ApplicationProtocol } from "aws-cdk-lib/aws-elasticloadbalancingv2";
+import { ApplicationProtocol, ListenerAction } from "aws-cdk-lib/aws-elasticloadbalancingv2";
 import { Bucket } from "aws-cdk-lib/aws-s3";
+import { StringParameter } from "aws-cdk-lib/aws-ssm";
 import { Construct } from "constructs";
 import { AccessScope, MetadataKeys, NAMED_SSM_PARAMETER_PATHS } from "../../constants";
 import { GuCertificate } from "../../constructs/acm";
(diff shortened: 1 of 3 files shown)


src/patterns/ec2-app/base.ts
```diff
@@ -470,6 +511,45 @@ export class GuEc2App extends Construct {
...
+
+        authorizationEndpoint: "https://accounts.google.com/o/oauth2/v2/auth",
+        issuer: "https://accounts.google.com",
+        tokenEndpoint: "https://oauth2.googleapis.com/token",
+        userInfoEndpoint: "https://openidconnect.googleapis.com/v1/userinfo",
+      });
+
+      listener.addAction("auth", { action: authAction });
+
+      Tags.of(loadBalancer).add(MetadataKeys.CDK_FEATURE, "google-auth");
```
akash1810, line 550: What's the purpose of this tag?

src/patterns/ec2-app/base.ts
```diff
@@ -470,6 +511,45 @@ export class GuEc2App extends Construct {
       open: access.scope === AccessScope.PUBLIC,
     });
 
+    if (props.googleAuth?.enabled) {
+      const configPrefix = `${scope.stage}/${scope.stack}/${app}`;
+      const clientIdPath = props.googleAuth.clientIdPath ?? `/${configPrefix}/googleClientID`;
```
akash1810, line 516: I think [destructuring with default values](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#default_value) would be clearer here, as we can group things together more easily:

```ts
const { googleAuth } = props;
const {
  clientIdPath = `/${configPrefix}/googleClientID`,
  clientSecretPath = `${configPrefix}/clientSecret`
} = googleAuth;
```

src/patterns/ec2-app/base.ts
```diff
@@ -470,6 +511,45 @@ export class GuEc2App extends Construct {
...
+      const clientId = StringParameter.fromStringParameterAttributes(this, "clientID", {
+        parameterName: clientIdPath,
+      }).stringValue;
+
+      // Unfortunately, Cloudformation doesn't support directly using secret
+      // Parameter Store values. But it is possible to use Secrets Manager.
+      const secretPath = props.googleAuth.clientSecretPath ?? `${configPrefix}/clientSecret`;
+      const clientSecret = SecretValue.secretsManager(secretPath);
+
+      const authAction = ListenerAction.authenticateOidc({
```
akash1810, line 527: Should we add this listener action at the [construct level](https://github.com/guardian/cdk/tree/main/src/constructs/loadbalancing/alb) so that non-pattern users also benefit?

src/patterns/ec2-app/base.ts
nicl: To support possible tracking in future. It's not a construct so our generic metadating tracking doesn't work unfortunately, but it is still something that would be good to track in the future.

src/patterns/ec2-app/base.ts
nicl: Yeah, that's a nice idea. I'd hope most adoption would be via the pattern, but could see it being used standalone too.

nicl: Having looked a bit more closely, I'm not sure this is straightforward after all. The ALB type is separate from the Listener and Target types, and the auth functionality interacts with them all, so think it is best to keep this at the pattern level for this PR at least.

akash1810:
The load balancer's current egress rules are pretty tight, only allowing connection to the EC2 instances. We'd need to allow it to connect to the Google endpoints too.

src/patterns/ec2-app/base.ts
akash1810: > To support possible tracking in future.

IIUC we don't have mechanisms to make use of this tag right now? That is, should we decouple this change from the OIDC part, whilst tracking is undefined?

src/patterns/ec2-app/base.ts
```diff
@@ -62,6 +63,45 @@ export interface Alarms {
...
+
+  // Parameter Store path containing your Google Client ID. Defaults to
+  // `/:STAGE/:stack/:app/googleClientId`.
+  clientIdPath?: string;
+
+  // *Secrets Manager* (NOT Parameter Store) path containing your Google Client
+  // Secret. Defaults to `/:STAGE/:stack/:app/googleClientSecret`. Nb. you
+  // should store your secret as plaintext rather than key-value structured data
+  // when creating it.
+  clientSecretPath?: string;
```
akash1810, line 102: With our current restrictions in Google Cloud, creation of a new project is manual and non-trivial. Additionally, creating this Secret Manager value is manual.

I think we can significantly improve the DX here with a combination of [terraform](https://github.com/hashicorp/terraform-provider-google/issues/6074) and [custom types](https://github.com/iann0036/cfn-tf-custom-types) to create the Google project with CloudFormation. It should also be possible to make the Google Project `internal` too.

src/patterns/ec2-app/base.ts
```diff
@@ -470,6 +511,46 @@ export class GuEc2App extends Construct {
...
+        scope: "openid email",
+
+        // See the `hd` section of
+        // https://developers.google.com/identity/protocols/oauth2/openid-connect#authenticationuriparameters.
+        // Note, this is NOT sufficient to ensure access is limited to Guardian
+        // emails. Users should also ensure set 'User Type' to 'Internal' in the
+        // Google Console Consent Screen to ensure only @guardian.co.uk members
+        // can access your app and ensure (through Security Groups) that only
+        // the ALB is able to access the EC2 instances.
+        authenticationRequestExtraParams: { hd: "guardian.co.uk" },
```
akash1810, line 541: This allows anyone with a Guardian email address to authenticate. This is not enough to satisfy the requirements from @guardian/infosec, as we also need to validate one's membership in a Google Group.

akash1810:
Converting this to draft to remove it from the daily [prnouncer](https://github.com/guardian/actions-prnouncer) message.

PR #1694, opened 2023-01-19
fix: Switch from our home baked SSM policy to the current managed policy

sihil:
## What does this change?

This PR effectively adds some extra SSM permissions required for AWS Inspector V2.

Once upon a time the recommended managed policy allowed full access to all S3 buckets but this craziness [is no longer the case and the managed policy is treated with care](https://aws.amazon.com/blogs/mt/applying-managed-instance-policy-best-practices/). We should switch to this, at least in part, because it has some extra permissions needed for AWS Inspector (v2) and new required SSM permissions will be added for us in the future.

## How to test

This has been and can be tested by doing the following:
 - in the CDK playground project:
   - create a new branch
   - switch to using this CDK branch with `script/switch-cdk sihil/switch-to-managed-ssm-policy`
   - update the test snapshots
   - commit, push and open a draft PR to trigger a build
 - deploy your branch of the CDK playground
 - in to the developer playground account:
   - locate the CDK playground EC2 instance
   - open the IAM role
   - check that the managed role is there and no inline policy relating to SSM exists
 - check you can `ssm ssh` to the EC2 instance with developer playground CLI creds

We'll be able to roll out AWS Inspector v2.

It's possible that this is incorrect and SSM access will break for instances. This would likely break a few things like ES node rotation but is unlikely to cause user facing issues.

## Checklist

- [ ] ~I have listed any breaking changes, along with a migration path~ [^1]
- [ ] ~I have updated the documentation as required for the described changes~ [^2]


sihil:
This has been tested. I've left a cdk-playground branch and build around should you want to check it out when deployed. See [this draft cdk-playground PR](https://github.com/guardian/cdk-playground/pull/323) and [this deploy in riff-raff](https://public.riffraff.gutools.co.uk/deployment/view/3ceaab01-42e8-4ebe-b3b0-5f0870a3d7cc?verbose=1).

akash1810 (approved):

PR #1725, opened 2023-02-14
feat(riff-raff.yaml): Add support for encrypted AMIs

akash1810:
## What does this change?
Following https://github.com/guardian/riff-raff/pull/991, add support for encrypted AMIs. By default AMIs are assumed to be encrypted, as this is best practice. Not sure if this would be surprising?

## How to test
See updated tests.

It's possible to auto-generate a more complete `riff-raff.yaml` file.

If an AMIgo recipe is not configured to create an encrypted copy, and one hasn't explcitly said `Encrypted: false`, then the deployment will fail with Riff-Raff unable to locate a relevant AMI.

## Checklist

### DIFF
--- a/src/constructs/autoscaling/asg.ts
+++ b/src/constructs/autoscaling/asg.ts
@@ -15,6 +15,7 @@ import { GuInstanceRole } from "../iam";
 // Since we want to override the types of what gets passed in for the below props,
 // we need to use Omit<T, U> to remove them from the interface this extends
 // https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys
+// TODO avoid prop nesting to improve discoverability. See https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md#flat.
 export interface GuAutoScalingGroupProps
   extends Omit<
       AutoScalingGroupProps,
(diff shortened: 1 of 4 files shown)


twrichards (approved):
Very nice, glad to see this follow-up PR so soon after our initial foray into this. 

How many projects are currently using this experimental riff-raff.yaml generation? Will this new default of `Encrypted: true` cause any  problems for teams? (if no encrypted bake exists for their recipes)

akash1810:
> How many projects are currently using this experimental riff-raff.yaml generation? Will this new default of Encrypted: true cause any problems for teams? (if no encrypted bake exists for their recipes)

There are currently [13 usages](https://github.com/search?q=org%3Aguardian+GuRootExperimental++NOT+is%3Aarchived+NOT+repo%3Aguardian%2Fcdk+&type=code). I wonder if it's better to remove the default here, and add it in a future, breaking change.

twrichards:
> > How many projects are currently using this experimental riff-raff.yaml generation? Will this new default of Encrypted: true cause any problems for teams? (if no encrypted bake exists for their recipes)
> 
> There are currently [13 usages](https://github.com/search?q=org%3Aguardian+GuRootExperimental++NOT+is%3Aarchived+NOT+repo%3Aguardian%2Fcdk+&type=code). I wonder if it's better to remove the default here, and add it in a future, breaking change.

Hmmm there's fewer than 13 cases where `imageRecipe` is used (which is where this new default could have impact) https://github.com/search?q=org%3Aguardian+imageRecipe&type=code

... I've manually checked and all the recipes used have encryption turned on ([with one exception](https://github.com/guardian/backend-101-scala-starter-akash1810/blob/92ca45aef067ab8acf50dcba3be6ebf4eb6b3420/cdk/lib/backend-101-scala-starter-akash-1810.ts#L39), but its archived) 🎉 

... so go ahead and 🚢  (better to get this default in now before there's wider adoption of this feature).

PR #1731, opened 2023-02-20
feat: Use LaunchTemplates instead of LaunchConfiguration

kenoir:
## What does this change?

This change moves `GuAutoScalingGroup` to use `LaunchTemplates` as `LaunchConfigurations` [are deprecated](https://aws.amazon.com/blogs/compute/amazon-ec2-auto-scaling-will-no-longer-add-support-for-new-ec2-features-to-launch-configurations/).

The API to the consumer should remain unchanged, but this will results in EC2 instances getting a different `Name` tag in its current implementation. See https://github.com/guardian/cdk/pull/1731/files#diff-b219fec113d015feaadba12ad710f805f4ba8fc1fcf3c270be34c8fd31b67c0fR101 for the logic that generates the `Name`.

> **Note**
> This change removes tests for `Name` tag on an `AutoScalingGroup` as these are not generated or set by GuCDK and moves from the ASG as Tags to `TagSpecifications` on `LaunchTemplate`.

## How to test

See https://github.com/guardian/cdk-playground/pull/344 for a test on a real-world application.

## Checklist

~- [ ] I have listed any breaking changes, along with a migration path [^1]~
~- [ ] I have updated the documentation as required for the described changes [^2]~

### DIFF
--- a/src/constructs/autoscaling/asg.test.ts
+++ b/src/constructs/autoscaling/asg.test.ts
@@ -62,9 +62,11 @@ describe("The GuAutoScalingGroup", () => {
       Type: "AWS::EC2::Image::Id",
     });
 
-    template.hasResourceProperties("AWS::AutoScaling::LaunchConfiguration", {
-      ImageId: {
-        Ref: "AMITesting",
+    template.hasResourceProperties("AWS::EC2::LaunchTemplate", {
+      LaunchTemplateData: {
+        ImageId: {
+          Ref: "AMITesting",
+        },
       },
     });
   });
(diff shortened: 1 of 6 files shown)


nicl (approved):

nicl:
(p.s. I'm assuming this doesn't result in a replacement of the ASG?)

kenoir:
> (p.s. I'm assuming this doesn't result in a replacement of the ASG?)

It does not, i've tested this by deploying cdk-playground in the dev-playground account.

davidfurey:
This introduces a breaking change for apps that use the method `app.autoScalinGroup.addSecurityGroup`. The error is "You cannot add security groups when the Auto Scaling Group is created from a Launch Template.".

I **think** switching to `app.autoScalingGroup.connections.addSecurityGroup` is a straightforward fix.

PR #1732, opened 2023-02-21
feat: New GuPlayWorkerApp pattern

davidfurey:
## What does this change?

This adds a new EC2 pattern - `GuPlayWorkerApp`. A few teams run Play apps which do useful data processing tasks like consuming queues and pushing data to other queues, or polling web services, but notably do not provide much of a HTTP API themselves apart from a health check.  For these apps, having an ALB is required for healthchecking but the https certificate / domain are not useful.  None of the existing patterns support this requirement.

This change ensures that certificates are still required for regular Play and Node apps.  For services without a certificate the the ALB port to 8080 instead of 80 and sets the ALB as non-open to discourage abusing the change to host browser facing HTTP services.

If this pattern becomes widely used, it might be worth exploring if the ALB could be replaced by a lighter weight (and cheaper) healthchecking lambda. It has never seemed worth the saving to do this for just one project but GuCDK might make it easy to roll out these kinds of cost savings.

## Checklist

### DIFF
--- a/src/constructs/loadbalancing/alb/application-listener.ts
+++ b/src/constructs/loadbalancing/alb/application-listener.ts
@@ -45,8 +45,8 @@ export class GuHttpsApplicationListener extends GuAppAwareConstruct(ApplicationL
     const { certificate, targetGroup } = props;
 
     const mergedProps: GuApplicationListenerProps = {
-      port: typeof certificate === undefined ? 8080 : 443,
-      protocol: typeof certificate === undefined ? ApplicationProtocol.HTTP : ApplicationProtocol.HTTPS,
+      port: typeof certificate === "undefined" ? 8080 : 443,
+      protocol: typeof certificate === "undefined" ? ApplicationProtocol.HTTP : ApplicationProtocol.HTTPS,
       ...props,
       certificates: typeof certificate === "undefined" ? [] : [
         {
--- a/src/constructs/loadbalancing/alb/application-listener.ts
+++ b/src/constructs/loadbalancing/alb/application-listener.ts
@@ -46,7 +46,7 @@ export class GuHttpsApplicationListener extends GuAppAwareConstruct(ApplicationL
 
     const mergedProps: GuApplicationListenerProps = {
       port: typeof certificate === undefined ? 8080 : 443,
-      protocol: ApplicationProtocol.HTTPS,
+      protocol: typeof certificate === undefined ? ApplicationProtocol.HTTP : ApplicationProtocol.HTTPS,
       ...props,
       certificates: typeof certificate === "undefined" ? [] : [
         {
--- a/src/constructs/loadbalancing/alb/application-listener.ts
+++ b/src/constructs/loadbalancing/alb/application-listener.ts
@@ -5,7 +5,7 @@ import type { GuCertificate } from "../../acm";
 import type { AppIdentity, GuStack } from "../../core";
 import type { GuApplicationTargetGroup } from "./application-target-group";
 
-export interface GuApplicationListenerProps extends ApplicationListenerProps, AppIdentity { }
+export interface GuApplicationListenerProps extends ApplicationListenerProps, AppIdentity {}
 
 /**
  * Construct which creates a Listener.
@@ -28,7 +28,7 @@ export class GuApplicationListener extends GuAppAwareConstruct(ApplicationListen
 
 export interface GuHttpsApplicationListenerProps
   extends Omit<GuApplicationListenerProps, "defaultAction" | "certificates">,
-  AppIdentity {
+    AppIdentity {
   targetGroup: GuApplicationTargetGroup;
   certificate?: GuCertificate;
 }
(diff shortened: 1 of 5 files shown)


src/constructs/loadbalancing/alb/application-listener.ts
```diff
@@ -45,14 +45,17 @@ export class GuHttpsApplicationListener extends GuAppAwareConstruct(ApplicationL
     const { certificate, targetGroup } = props;
 
     const mergedProps: GuApplicationListenerProps = {
-      port: 443,
-      protocol: ApplicationProtocol.HTTPS,
+      port: typeof certificate === "undefined" ? 8080 : 443,
```
akash1810, line 48: I think we can be a bit terser here, removing the `typeof`:

```ts
const port = certificate ? 443 : 8080;
```

Same applies for `protocol` and `certificates` below.

src/patterns/ec2-app/framework.ts
```diff
@@ -4,11 +4,31 @@ The plan is to sprinkle some framework specific tooling into them.
...
+ * Creates an instance of [[`GuEc2App`]], with an application port of 9000 and no access to the load balancer.
+ * This is useful for applications that only have an LB for health check purposes
+ */
+export class GuPlayWorkerApp extends GuEc2App {
+  static readonly PORT: number = 9000;
+
+  constructor(scope: GuStack, props: GuEc2WorkerProps) {
+    super(scope, {
+      ...props,
+      applicationPort: GuPlayApp.PORT,
```
akash1810, line 27:
```suggestion
      applicationPort: GuPlayWorkerApp.PORT,
```

akash1810 (commented):
Thanks for this, looks good!

Added a couple of style comments. For the app you're building, is it worth considering if you'd want to add [prout](https://github.com/guardian/prout) to it? I'd imagine that might change things here?

davidfurey:
> Thanks for this, looks good!
> 
> Added a couple of style comments. For the app you're building, is it worth considering if you'd want to add [prout](https://github.com/guardian/prout) to it? I'd imagine that might change things here?

Thanks for the feedback!

Good point about prout. There is already an established pattern of running these sorts of services with internal load balancers beyond the reach of prout, e.g. Ophan's bigloader. Creating a CDK pattern to make these migrations more straightforward might make it easier in the future for us to roll out a feature that can determine if private apps are properly deployed.

src/patterns/ec2-app/framework.ts
```diff
@@ -4,11 +4,31 @@ The plan is to sprinkle some framework specific tooling into them.
...
 import { GuEc2App } from "./base";
 
-type GuEc2FrameworkAppProps = Omit<GuEc2AppProps, "applicationPort">;
+type GuEc2FrameworkAppProps = Omit<GuEc2AppProps, "applicationPort"> & { certificateProps: GuDomainName };
+
+type GuEc2WorkerProps = Omit<GuEc2AppProps, "applicationPort" | "access">;
+
+/**
+ * Creates an instance of [[`GuEc2App`]], with an application port of 9000 and no access to the load balancer.
+ * This is useful for applications that only have an LB for health check purposes
```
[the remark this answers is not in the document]
davidfurey: I agree that in many cases Lambda is the way to go.  But this service is http long polling an external API, so needs to be running continually. I believe this will be cheaper as an instance rather than a lambda.  I suspect that services like big loader are also processing enough data that it is also more cost effective to to run a worker.

Perhaps we just need a rule of thumb of when it is more cost effective to use this pattern. And to create a nice pattern to make the Lambda path appealing.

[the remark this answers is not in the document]
davidfurey: If the concern is about docs rather than the pattern itself, can we merge this as is and follow up with docs? This is currently blocking another piece of work.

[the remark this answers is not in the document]
jacobwinch: > I agree that in many cases Lambda is the way to go. But this service is http long polling an external API, so needs to be running continually.

Thanks for explaining, that makes sense 👍 

> can we merge this as is and follow up with docs?

Yes, I don't consider this question about docs to be blocking (sorry I should have made that clearer in my initial comment). I will let someone from the @guardian/devx-operations team provide the official review (I'm not in that team any more), but 👍  from my perspective.

src/patterns/ec2-app/framework.ts
```diff
@@ -4,11 +4,31 @@ The plan is to sprinkle some framework specific tooling into them.
...
+ */
+export class GuPlayWorkerApp extends GuEc2App {
```
akash1810, line 21: I'd be tempted to not restrict this to Play, and name it `GuEc2WorkerApp` instead. However, we can make this change later once we understand if there are non Play use-cases.

akash1810 (approved):

src/patterns/ec2-app/framework.ts
davidfurey: I share your temptation, but also think we should only change it if there turn out to be non-Play use-cases

PR #1749, opened 2023-03-02
docs: enforce tsdocs for exported members (illustration)

nicl:
## Overview

Leverages https://github.com/guardian/eslint-plugin-tsdoc-required to enforce certain documentation requirements.

The aim is to provide more localised documentation - focusing on the props - and to mirror AWS's own approach here.

For now, this is only applied to a single file. If people approve the approach, we can gradually extend it across the codebase.

## Details

Requires the following for exported members:

* exported members (classes, interfaces, ...) must be documented with a valid TSDoc
* for exported 'props' interfaces, no top-level comment is required, but properties must be documented with a valid TSDoc

E.g. valid:

```
export interface MyProps {
  /**
   * Description of someProperty.
   */
  someProperty: string;
  /**
   * Description of anotherProperty.
   */
  anotherProperty: number;
}
```

(It is completely fine, though not required, to document the interface itself at the top.)

vs no longer valid:

```
/*
 * Interface MyProps does ...
 *
 * It has the following requirements...
 *
 * Prop A does X, use it .... prop B does y, use it as follows...
 */
export interface MyProps {
  someProperty: string;
  anotherProperty: number;
}
```

Note, initially this is only enabled for our base EC2 file. This PR is really for discussion and illustration purposes.

## Future (suggested) steps:

* extend to our other patterns
* add @deprecated and later @internal tags to non patterns that we don't want people to use directly (the majority of them - though will need to discuss) - some could be simply removed at this point
* switch this lint rule on as an error for the entire codebase + add any missing comments/fix invalid comments

### DIFF
--- a/src/patterns/ec2-app/base.ts
+++ b/src/patterns/ec2-app/base.ts
@@ -1,3 +1,4 @@
+/* eslint "@guardian/tsdoc-required/tsdoc-required": 2 -- to begin rolling this out for public APIs. */
 import { Duration, Tags } from "aws-cdk-lib";
 import type { BlockDevice } from "aws-cdk-lib/aws-autoscaling";
 import { HealthCheck } from "aws-cdk-lib/aws-autoscaling";
@@ -27,7 +28,19 @@ import type { GuAsgCapacity, GuDomainName } from "../../types";
 import type { AmigoProps } from "../../types/amigo";
 
 export interface AccessLoggingProps {
+  /**
+   * Enable (load balancer) access logs.
+   *
+   * Note, you will need to specify a region in your stack declaration to use
+   * this.
+   * See`https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-elasticloadbalancingv2.ApplicationLoadBalancer.html#logwbraccesswbrlogsbucket-prefix`
+   */
   enabled: boolean;
+  /**
+   * S3 prefix for the logs.
+   *
+   * @defaultValue no prefix
+   */
   prefix?: string;
 }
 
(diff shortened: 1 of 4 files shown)


src/patterns/ec2-app/base.ts
```diff
@@ -158,176 +194,11 @@ function restrictedCidrRanges(ranges: IPeer[]) {
 }
 
 /**
- * Pattern which creates the resources needed to run an application on EC2 instances.
- * For convenience, you may wish to use [[`GuPlayApp`]] or [[`GuNodeApp`]], which extend this class.
- *
- * This pattern will grant your EC2 instances a number of commonly needed IAM permissions. For more information on
- * this, see [[`GuInstanceRole`]]. To add additional permissions to your EC2 instances, see [[`GuEc2AppProps`]].
- *
```
kenoir, line 166: This change removes some debatably useful documentation here, (though I agree this might not be the best place for these examples). Where is the best place for examples of usage of these patterns?

kenoir (commented):
📜 🥇

src/patterns/ec2-app/base.ts
```diff
@@ -158,176 +194,11 @@ function restrictedCidrRanges(ranges: IPeer[]) {
...
- *   userData: {
- *     distributable: {
- *       fileName: "app-name.deb",
- *       executionStatement: `dpkg -i /app-name/app-name.deb`,
- *     }
- *   },
- * });
- * ```
- *
- * ### Varying values by stage
```
nicl, line 228: This is really useful information, but it is not related to the pattern so I wondered if it was better provided elsewhere?

src/patterns/ec2-app/base.ts
```diff
@@ -158,176 +194,11 @@ function restrictedCidrRanges(ranges: IPeer[]) {
...
- *  privateSubnets: a list of ISubnet indicating the subnets for the autoscaling group to deploy instances into (usually
- *    obtained from `GuVpc.subnets`)
- *  publicSubnets: a list of ISubnet indicating the subnets for the loadbalancer to use, if the loadbalancer
- *    is intended for public access (usually obtained from `GuVpc.subnets`)
- *    On the other hand, if the access is private, then an "internal" loadbalancer is used in the `privateSubnets` list.
- * Please note that no checking is applied to the subnet lists to ensure that they correspond to the vpc given; as
- * an advanced feature it is assumed that the caller ensures this.  If the subnets are mismatched then the resulting
- * Cloudformation will fail to deploy.
- *
- * ### Customising the infrastructure generated by this pattern
```
nicl, line 306: Again, this is useful information, but more general than this pattern, so I wonder if we should lift it elsewhere?

nicl: (i.e. we wouldn't want to repeat it on all of our patterns - it's more some general info on how to use CDK.)

akash1810:
Love this! 💎

kenoir (approved):

src/patterns/ec2-app/base.ts
nicl: I'm not sure - I wonder if we should add some methods on patterns too, a la CDK itself, to support this kind of thing. E.g. something like this (for a function example):

```
public addToRolePolicy(statement: PolicyStatement): void {
		this.function.addToRolePolicy(statement);
	}
```

I experimented a bit with this here anyway, as it felt nice:

https://github.com/guardian/cdk-v2/blob/main/lib/lambda-task.ts#L159

PR #1750, opened 2023-03-03
feat: Add `GuFastlyKinesisLogRole` role

marsavar:
## What does this change?
Adds a new `GuFastlyKinesisLogRole` construct, which lets Fastly write to a specific Kinesis stream. 
Fastly allows log streaming to AWS using either S3  or Kinesis. There is currently a similar `GuRole` which lets Fastly write logs to S3, but the one for Kinesis was missing, hence this PR adds it.

In order to do so, the role needs to have a specific policy. The existing `GuLogShipping` policy didn't quite fit our needs because it assumes a single stream per account. I therefore created a new policy the lets consumers specify the stream that Fastly needs to write into.

Since the new role needs to use Fastly's AWS account number which already existed in the `GuFastlyLogsIamRole` construct, I lifted it out of there into a const, so we can reuse it elsewhere in the codebase (including tests).

## How to test
Usual tests! 

Fewer lines of code to cloudform this role.

## Checklist

- [x] I have listed any breaking changes, along with a migration path [^1] I don't believe this breaks existing constructs.
- [x] I have updated the documentation as required for the described changes [^2] (TS doc)


src/constructs/iam/roles/fastly-kinesis-log.ts
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * A construct to create an IAM Role for Fastly to assume in order to write to a
+ * specific Kinesis stream.
+ *
+ * In order to use this construct, an SSM parameter named `/account/external/fastly/customer.id`
+ * needs to exist in the AWS account's parameter store, and the value should be
+ * the Guardian's Fastly customer id.
+ *
+ */
+export class GuFastlyKinesisLogRole extends GuRole {
```
kenoir, line 29: Nit: As this is a sibling of the `GuFastlyLogsIamRole` should this be one level up in the `constructs/iam` folder, or is the other role in the wrong place?

marsavar: I thought about this when I created the files and concluded that the other one (`GuFastlyLogsIamRole`) is possibly in the wrong place.
Since this PR adds a new `AWS::IAM::Role` I think the logical place for this construct to live is the `iam/roles` folder, in the same way that `AWS::IAM::Policy` constructs live in the `iam/policies` folder.

nicl:
@marsavar hey great work! Can we move this under `experimental`? Our fault we haven't documented this yet (will raise a PR) but that's our preferred route for new constructs - similar to Source Kitchen, the aim is to introduce things without a clear guarantee of future support, monitor usage, and then promote (or not) depending on adoption and other feedback.

akash1810 (approved):

PR #1751, opened 2023-03-03
feat: Cognito for Google auth

nicl:
Introduce Google auth to EC2 patterns

The aim is to support:

* human authentication
* for EC2-pattern users
* that is *secure* and *easy to adopt*
    
This is achieved by integrating Cognito at the ALB level.
 
In addition to a workspace check - i.e. the user is part of @guardian.co.uk - a group check is also required. When multiple groups are specified a user need only be a member of one to gain access.

It is recommended that people create a specific Google group for their application and use this for the membership check. If broader P+E dev access is required, use `engineering@guardian.co.uk`.

The overall architecture is as follows:

```mermaid
sequenceDiagram
    Note right of ALB: Shaded region is defined in CDK pattern.
    rect rgba(0, 0, 255, .1)
      ALB->>Cognito: Handle authentication for me please...
      Cognito->>Google: Are they part of our org?
      Google-->>Cognito: Looks good!
      Cognito->>Cognito Lambda: Any other checks?
    end
    Cognito Lambda->>Gatekeeper Lambda: Can you check group membership for me?
    Gatekeeper Lambda-->>Cognito Lambda: Looks good!
    rect rgba(0, 0, 255, .1)
      Cognito Lambda-->>Cognito: Looks good!
      Cognito-->>ALB: Looks good!
    end
```

The shaded region is defined in this PR, though both lambdas are defined [here](https://github.com/guardian/cognito-gatekeeper/pull/2).

The Cognito Lambda is triggered by Cognito as at the 'pre-auth' and also 'pre-signup' stages. For more on Cognito Lambda triggers see [here](https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-identity-pools-working-with-aws-lambda-triggers.html).

For context see:

* [Auth lambda and architecture](https://github.com/guardian/cognito-gatekeeper/pull/2)
* [Teamcity example (inlined)](https://github.com/guardian/deploy-tools-platform/pull/670)
* [Trello card](https://trello.com/c/XiOqsax2/1733-aws-cognito-auth-in-guardian-cdk)
* [Guide to setup Google](https://docs.google.com/document/d/1_k1FSE52AZHXufWLTiKTI3xy5cGpziyHazSHTKrYfco/edit#)
* [Proposal doc for CDK integration](https://docs.google.com/document/d/1SfvjNRIzv1bNYRho7s5i_YZv6YyWadp4j9rNOqvpnJg/edit#)

Note, ideally we will automate the Google side, but that is a follow up.

**TODO: publish to beta and test!**

## Questions

At the moment everything is defined in the EC2 pattern rather than the `GuApplicationLoadBalancer` construct. This is because a lot of the related constructs (`GuApplicationListener` and `GuApplicationTargetGroup`) are *not* defined in the ALB construct. I wonder if medium-term we should integrate these into the `GuApplicationLoadBalancer` pattern? This is probably something to do in any `v2` of the library though. I.e.

### DIFF
--- a/src/patterns/ec2-app/base.test.ts
+++ b/src/patterns/ec2-app/base.test.ts
@@ -758,4 +758,103 @@ describe("the GuEC2App pattern", function () {
       additionalTags: [{ Key: MetadataKeys.LOG_KINESIS_STREAM_NAME, Value: { Ref: "LoggingStreamName" } }],
     });
   });
+
+  it("supports Google auth at the ALB", () => {
+    const stack = simpleGuStackForTesting({ env: { region: "eu-west-1" } });
+    const app = "app";
+    const domain = "domain-name-for-your-application.example";
+
+    new GuEc2App(stack, {
+      applicationPort: 3000,
+      access: { scope: AccessScope.PUBLIC },
+      app,
+      instanceType: InstanceType.of(InstanceClass.T4G, InstanceSize.MEDIUM),
+      certificateProps: {
+        domainName: domain,
+      },
+      scaling: {
+        minimumInstances: 1,
+      },
+      monitoringConfiguration: { noMonitoring: true },
+      userData: "",
+      accessLogging: { enabled: false },
+      googleAuth: {
+        enabled: true,
+        domain,
+      },
+    });
+
+    Template.fromStack(stack).hasResourceProperties("AWS::ElasticLoadBalancingV2::Listener", {
+      DefaultActions: [
+        {
+          Type: "authenticate-cognito",
+          Order: 1,
+        },
+        {
+          Type: "forward",
+          Order: 2,
+        },
+      ],
+    });
+  });
+
+  it("throws when googleAuth.allowedGroups is empty", () => {
+    const stack = simpleGuStackForTesting({ env: { region: "eu-west-1" } });
+    const app = "app";
+    const domain = "domain-name-for-your-application.example";
+
+    expect(
+      () =>
+        new GuEc2App(stack, {
+          applicationPort: 3000,
+          access: { scope: AccessScope.PUBLIC },
+          app,
+          instanceType: InstanceType.of(InstanceClass.T4G, InstanceSize.MEDIUM),
+          certificateProps: {
+            domainName: domain,
+          },
+          scaling: {
+            minimumInstances: 1,
+          },
+          monitoringConfiguration: { noMonitoring: true },
+          userData: "",
+          accessLogging: { enabled: false },
+          googleAuth: {
+            enabled: true,
+            domain,
+            allowedGroups: [],
+          },
+        })
+    ).toThrowError("googleAuth.allowedGroups cannot be empty!");
+  });
+
+  it("throws when googleAuth.allowedGroups contains groups using non-standard domains", () => {
+    const stack = simpleGuStackForTesting({ env: { region: "eu-west-1" } });
+    const app = "app";
+    const domain = "domain-name-for-your-application.example";
+
+    expect(
+      () =>
+        new GuEc2App(stack, {
+          applicationPort: 3000,
+          access: { scope: AccessScope.PUBLIC },
+          app,
+          instanceType: InstanceType.of(InstanceClass.T4G, InstanceSize.MEDIUM),
+          certificateProps: {
+            domainName: domain,
+          },
+          scaling: {
+            minimumInstances: 1,
+          },
+          monitoringConfiguration: { noMonitoring: true },
+          userData: "",
+          accessLogging: { enabled: false },
+          googleAuth: {
+            enabled: true,
+            domain,
+            allowedGroups: ["apple@guardian.co.uk", "engineering@theguardian.com"],
+          },
+        })
+    ).toThrowError("googleAuth.allowedGroups must use the @guardian.co.uk domain.");
+  });
 });
--- a/src/patterns/ec2-app/base.ts
+++ b/src/patterns/ec2-app/base.ts
@@ -1,10 +1,22 @@
-import { Duration, Tags } from "aws-cdk-lib";
+/* eslint "@guardian/tsdoc-required/tsdoc-required": 2 -- to begin rolling this out for public APIs. */
+import crypto from "crypto";
+import { Duration, SecretValue, Tags } from "aws-cdk-lib";
 import type { BlockDevice } from "aws-cdk-lib/aws-autoscaling";
 import { HealthCheck } from "aws-cdk-lib/aws-autoscaling";
+import {
+  ProviderAttribute,
+  UserPool,
+  UserPoolClientIdentityProvider,
+  UserPoolIdentityProviderGoogle,
+} from "aws-cdk-lib/aws-cognito";
 import type { InstanceType, IPeer, ISubnet, IVpc } from "aws-cdk-lib/aws-ec2";
 import { Port } from "aws-cdk-lib/aws-ec2";
-import { ApplicationProtocol } from "aws-cdk-lib/aws-elasticloadbalancingv2";
+import { ApplicationProtocol, ListenerAction } from "aws-cdk-lib/aws-elasticloadbalancingv2";
+import { AuthenticateCognitoAction } from "aws-cdk-lib/aws-elasticloadbalancingv2-actions";
+import { Effect, PolicyStatement } from "aws-cdk-lib/aws-iam";
+import { Architecture, Runtime } from "aws-cdk-lib/aws-lambda";
 import { Bucket } from "aws-cdk-lib/aws-s3";
+import { StringParameter } from "aws-cdk-lib/aws-ssm";
 import { Construct } from "constructs";
 import { AccessScope, MetadataKeys, NAMED_SSM_PARAMETER_PATHS } from "../../constants";
 import { GuCertificate } from "../../constructs/acm";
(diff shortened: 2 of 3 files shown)


src/patterns/ec2-app/base.ts
```diff
@@ -378,6 +436,94 @@ export class GuEc2App extends Construct {
...
+        cognitoDomain: {
+          // Must be unique (as global across AWS)
+          domainPrefix: `com-gu-${app}-${scope.stage.toLowerCase()}`,
+        },
+      });
+
+      const userPoolIdp = new UserPoolIdentityProviderGoogle(scope, "google-idp", {
+        userPool: userPool,
+        clientId: clientId.valueAsString,
+        clientSecret: SecretValue.secretsManager(`/${scope.stage}/${scope.stack}/${app}/client-secret`).toString(),
```
nicl, line 486: TODO - can we read JSON here instead?

src/patterns/ec2-app/base.ts
```diff
@@ -182,6 +194,66 @@ export interface GuEc2AppProps extends AppIdentity {
...
+     * Set this to the same as for certificateProps.
+     */
+    domain: string;
+    /**
+     * Groups used for membership checks.
+     *
+     * If specified, cannot be empty. Users must be a member of at least one
+     * group to gain access.
+     *
+     * @defaultValue [`engineering@theguardian.com`]
```
nicl, line 232: TODO - I think these actually need to use the `guardian.co.uk` variant so let's update the default value and docs here to make this clear.

src/patterns/ec2-app/base.ts
```diff
@@ -377,6 +449,115 @@ export class GuEc2App extends Construct {
       }
     }
 
+    if (props.googleAuth?.enabled) {
+      const prefix = `/${scope.stage}/${scope.stack}/${app}`;
```
nicl, line 453: I feel like this is probably already defined in CDK somewhere. @akash1810 any thoughts? Or if not, perhaps we should define it and make it easy to use both internally and externally.

src/patterns/ec2-app/base.ts
```diff
@@ -377,6 +449,115 @@ export class GuEc2App extends Construct {
...
+
+      const {
+        allowedGroups = ["engineering@theguardian.com"],
+        clientIDSSMPath = `${prefix}/google-auth-client-id`,
+        clientSecretSecretsManagerPath = `${prefix}/google-auth-client-secret`,
```
nicl, line 458: I wonder if we should include the pattern here - i.e. `ec2-google-auth-client-id` - just in case we introduce Google Auth support for e.g. API Gateway too in future.

src/patterns/ec2-app/base.ts
```diff
@@ -377,6 +449,115 @@ export class GuEc2App extends Construct {
...
+        oAuth: {
+          callbackUrls: [`https://${props.googleAuth.domain}/oauth2/idpresponse`],
+        },
+
+        // Note: id and access validity token validity cannot be less than one
+        // hour (this is the cognito cookie duration). To quickly invalidate
+        // credentials, disable the user in Cognito.
+        idTokenValidity: Duration.hours(1),
+        accessTokenValidity: Duration.hours(1),
+        refreshTokenValidity: Duration.days(7),
```
nicl, line 545: TODO - we should probably support overriding these as some projects may want to be more aggressive on durations. Though note, allowing customisation allows less secure configurations here - there is an argument to just use the same settings everywhere.

src/patterns/ec2-app/base.ts
```diff
@@ -377,6 +449,115 @@ export class GuEc2App extends Construct {
...
+        lambdaTriggers: {
+          preAuthentication: authLambda,
+          preSignUp: authLambda,
+        },
+      });
+
+      const userPoolDomain = userPool.addDomain("domain", {
+        cognitoDomain: {
+          // Must be unique (as global across AWS)
+          domainPrefix: `com-gu-${app.toLowerCase()}-${scope.stage.toLowerCase()}`,
```
akash1810, line 507: Should we add a GUID to the end of this to further ensure uniqueness?

src/patterns/ec2-app/base.ts
akash1810: I don't think it is centrally defined. The closest I can think of is within [the SSM policy](https://github.com/guardian/cdk/blob/0110e9786aa0d1db49a6569a18eb082585dd8eef/src/constructs/iam/policies/parameter-store-read.ts#L15). 👍🏽 to DRYing it out.

akash1810 (commented):
Wonder if we should survey (or similar if) this would be helpful at the L2 `GuLoadBalancer` construct level, rather than just the pattern?

src/patterns/ec2-app/base.test.ts
```diff
@@ -758,4 +758,43 @@ describe("the GuEC2App pattern", function () {
...
+          Type: "authenticate-cognito",
+          Order: 1,
+        },
+        {
+          Type: "forward",
+          Order: 2,
+        },
+      ],
+    });
+  });
```
akash1810, line 799: 👍🏽

akash1810 (commented):
Summarising our IRL conversation...
- (blocking) Let's place the client ID and client secret in one SSM Secrets Manager item, as JSON. The aim here is to simplify the DX as there's one place to look (especially helpful for rotation)
- (non-blocking) Worth validating the `allowedGroups` prop, checking the email domain is `guardian.co.uk`?
- (non-blocking) Pushing this feature to the L2 load balancer construct is a little tricky at the moment as we have [seperated the ALB, listener, target group into different constructs](https://github.com/guardian/cdk/tree/main/src/constructs/loadbalancing/alb). If we had a single construct for all three elements, it'll be easier. This isn't a blocker here, but we should create an issue for this[^1]
- (non-blocking) This'll be a really nice feature to integration test. Alas, we never got round to implementing this.


[^1]: Might be a nice `good-first-issue` task?

src/patterns/ec2-app/base.ts
```diff
@@ -377,6 +452,121 @@ export class GuEc2App extends Construct {
...
+        throw new Error("googleAuth.allowedGroups must use the @guardian.co.uk domain.");
+      }
+
+      const deployToolsAccountId = StringParameter.fromStringParameterName(
+        scope,
+        "deploy-tools-account-id-parameter",
+        NAMED_SSM_PARAMETER_PATHS.DeployToolsAccountId.path
+      );
+
+      // See https://github.com/guardian/cognito-gatekeeper for the source code
```
akash1810, line 477: Looks like the repo has been renamed? Should we also consider making the repo public?

src/patterns/ec2-app/base.ts
akash1810: Worth moving this TODO into the file?

src/patterns/ec2-app/base.ts
```diff
@@ -377,6 +452,121 @@ export class GuEc2App extends Construct {
...
+        environment: {
+          ALLOWED_GROUPS: allowedGroups.join(","),
+          GATEKEEPER_FUNCTION_ARN: gatekeeperFunctionArn,
+        },
+      });
+
+      authLambda.addToRolePolicy(
+        new PolicyStatement({
+          effect: Effect.ALLOW,
+          actions: ["lambda:InvokeFunction"],
```
akash1810, line 501: Just double checking the singleton gateway lambda allows any org account to invoke it?

akash1810 (approved):
Couple of minor, non-blocking comments. Looks great.

src/patterns/ec2-app/base.ts
nicl: (will do as a follow-up depending on requests from teams)

src/patterns/ec2-app/base.ts
nicl: (In fact, this is not relevant, as we already include the app name here which should be unique.)

src/patterns/ec2-app/base.ts
nicl: (Will do in a follow-up as feels like a general thing.)

src/patterns/ec2-app/base.ts
nicl: Thanks - yep renamed for clarity. Will update this. I kept it private as it gives details of key security behaviour, but we can raise at Infosec catchup perhaps.

src/patterns/ec2-app/base.ts
nicl: I think you're right - we need a resource policy on the Gatekeeper lambda that allows invocation from our accounts. Will add it there.

nicl: @akash1810 fixed this here! https://github.com/guardian/cognito-auth-lambdas/pull/3

nicl:
@akash1810 heads up have tested this now successfully on Grafana CODE. To get it working I had to add:

https://github.com/guardian/cdk/pull/1751/commits/9669b9de7cb1920ee354709e8f490c48fab6dfac

let me know if you think this approach works or we should consider an alternative. I designed it to be consistent with the `bucketNamePath` prop which aims at the same kind of thing - just turns out it isn't sufficient to achieve what I'd hoped.

akash1810:
> To get it working I had to add:
> 
> [9669b9d](https://github.com/guardian/cdk/commit/9669b9de7cb1920ee354709e8f490c48fab6dfac)
> 
> let me know if you think this approach works or we should consider an alternative. I designed it to be consistent with the `bucketNamePath` prop which aims at the same kind of thing - just turns out it isn't sufficient to achieve what I'd hoped.

Ah! Great spot! This change feels pragmatic 👍🏽 .

PR #1752, opened 2023-03-08
feat: add optional `okAction` boolean prop to `GuAlarm` 

twrichards:
## What does this change?
Add optional `okAction` boolean prop to `GuAlarm`, which if true adds the same SNS action to `OKActions` as is added to `AlarmActions`.

## Checklist


src/constructs/cloudwatch/alarm.ts
```diff
@@ -7,6 +7,7 @@ import type { AppIdentity, GuStack } from "../core";
 
 export interface GuAlarmProps extends AlarmProps, AppIdentity {
   snsTopicName: string;
+  okAction?: boolean;
```
akash1810, line 10: I wonder if this should be a required property? This would mean teams have to actively consider if their alarm should auto-resolve or not. cc @guardian/devx-reliability for their opinion.

twrichards: @guardian/devx-reliability any thoughts? I'm keen to get this PR merged (I'm personally happy with it being optional, could make it mandatory in a future dedicated PR to unlock this PR if needs be)

jacobwinch: Sorry for the delay @twrichards. I don't have a strong opinion on auto-resolving alarms (or not!) and making it mandatory will increase friction for users trying to upgrade, so I'd support leaving it as optional for now.

PR #1769, opened 2023-03-15
feat: support automatic configuration of lambda functions

nicl:
This PR introduces support for automatic configuration of lambdas via:

via https://github.com/guardian/ssm-to-env.

When`config.enabled = true` in `GuLambdaFunction` props configuration for the lambda is automatically fetched and provided to the lambda as env vars.

Some details:

* configuration is sourced from *all* parameters with a matching `/:STAGE/:stack/:app` prefix
* some transformation of parameter names is done - see the code docs for info here

Note, this is opt-in initially. It is applied to our `GuLambdaFunction` construct, which is used by the following patterns:

* api-lambda
* scheduled-lambda
* sns-lambda (experimental)
* kinesis-lambda (experimental)

TODO:

* [x] write some tests!
* [ ] check with @kenoir for thoughts on if I'm using ssm-via-env right

## For discussion

Should we enable this by default? At the moment it is opt-in but if this is our recommended way to configure Lambdas, which I am keen to argue for, then defaulting it to on would make sense. One caveat is that it will break any users of Lambda Layer wrapper scripts. But these must be quite (very?)  rare. And those impacted can simply use the props to disable this feature if required.

### DIFF
--- a/src/constructs/lambda/lambda.ts
+++ b/src/constructs/lambda/lambda.ts
@@ -4,7 +4,8 @@ import type { PolicyStatement } from "aws-cdk-lib/aws-iam";
 import { Code, Function, LayerVersion, RuntimeFamily } from "aws-cdk-lib/aws-lambda";
 import type { FunctionProps, Runtime } from "aws-cdk-lib/aws-lambda";
 import { Bucket } from "aws-cdk-lib/aws-s3";
-import { StringParameter } from "aws-cdk-lib/aws-ssm";
+import { CfnParameter, StringParameter } from "aws-cdk-lib/aws-ssm";
+import { NAMED_SSM_PARAMETER_PATHS } from "../../constants";
 import { GuDistributable } from "../../types";
 import { GuLambdaErrorPercentageAlarm, GuLambdaThrottlingAlarm } from "../cloudwatch";
 import type { GuLambdaErrorPercentageMonitoringProps, GuLambdaThrottlingMonitoringProps } from "../cloudwatch";
@@ -58,7 +59,7 @@ export interface GuFunctionProps extends GuDistributable, Omit<FunctionProps, "c
      *
      * Use https://github.com/guardian/devx-config to manage config values.
      *
-     * @defaultValue true
+     * @defaultValue false
      */
     enabled: boolean;
     /**
@@ -121,7 +122,7 @@ export class GuLambdaFunction extends Function {
   public readonly fileName: string;
 
   constructor(scope: GuStack, id: string, props: GuFunctionProps) {
-    const { app, fileName, runtime, memorySize, timeout, bucketNamePath, config = { enabled: true } } = props;
+    const { app, fileName, runtime, memorySize, timeout, bucketNamePath, config = { enabled: false } } = props;
 
     const bucketName = bucketNamePath
       ? StringParameter.fromStringParameterName(scope, "bucketoverride", bucketNamePath).stringValue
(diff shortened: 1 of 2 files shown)


src/constructs/lambda/lambda.ts
```diff
@@ -108,6 +139,21 @@ export class GuLambdaFunction extends Function {
...
+      // For details, see https://github.com/guardian/ssm-to-env.
+
+      const configPrefix = props.config.ssmPrefix ?? `/${scope.stage}/${scope.stack}/${this.app}`;
+      const configLayer = LayerVersion.fromLayerVersionArn(
+        scope,
+        "config-layer",
+        "arn:aws:lambda:eu-west-1:ACCOUNT-INFO-TODO:layer:ssm-to-env-layer:6" // todo how to include account number here?
+      );
+
+      this.addLayers(configLayer);
```
kenoir, line 152: I think it would be useful to understand how this layer will play with other layers that a consumer might add. I don't think I've yet thought that through in the `ssm-to-env` implementation (though it's unlikely to a problem quickly as mutliple layers is probably in our infra is so rare as to not be a concern).

src/constructs/lambda/lambda.ts
```diff
@@ -108,6 +139,21 @@ export class GuLambdaFunction extends Function {
       });
     }
 
+    if (props.config) {
+      // For details, see https://github.com/guardian/ssm-to-env.
```
kenoir, line 143: I think that the ssm-to-env project should probably be in the same repo as devx-config which would simplify packaging.

src/constructs/lambda/lambda.ts
```diff
@@ -108,6 +139,21 @@ export class GuLambdaFunction extends Function {
...
+
+      const configPrefix = props.config.ssmPrefix ?? `/${scope.stage}/${scope.stack}/${this.app}`;
+      const configLayer = LayerVersion.fromLayerVersionArn(
+        scope,
+        "config-layer",
+        "arn:aws:lambda:eu-west-1:ACCOUNT-INFO-TODO:layer:ssm-to-env-layer:6" // todo how to include account number here?
```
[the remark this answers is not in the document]
nicl: Yep.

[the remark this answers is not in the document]
kenoir: > we don't want the account ID hard coded here?
I wonder whether it could be part of https://github.com/guardian/private-infrastructure-config ... but that would add an unpleasant dependency that would be annoying for consumers.

src/constructs/lambda/lambda.ts
nicl: Agree - this makes a lot of sense. Hoping to do this in a follow-up but in a way that is transparent to users of @guardian/cdk.

src/constructs/lambda/lambda.ts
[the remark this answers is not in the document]
nicl: (Resolved via https://github.com/guardian/aws-account-setup/pull/61.)

src/constructs/lambda/lambda.ts
nicl: I've added a comment to warn users that this will not work if they have their own wrapper script defined. I agree that these cases are likely to be very rare though so not too concerned.

src/constructs/lambda/lambda.ts
```diff
@@ -129,6 +165,31 @@ export class GuLambdaFunction extends Function {
...
+        NAMED_SSM_PARAMETER_PATHS.DeployToolsAccountId.path
+      );
+
+      const configPrefix = config.ssmPrefix ?? `/${scope.stage}/${scope.stack}/${this.app}`;
+
+      // TODO - guessing this needs to be parameterised by arch?
+      const configLayer = LayerVersion.fromLayerVersionArn(
+        scope,
+        "config-layer",
+        `arn:aws:lambda:eu-west-1:${deployToolsAccountId.stringValue}:layer:ssm-to-env-layer:6`
```
akash1810, line 185: Should this be region agnostic? Is `6` the version number? How do we update this over time?

nicl: Region agnostic 👍 . Suggesting that version bumps are a PR change here for now and then a CDK version bump for the library user. But it might make sense to parameterise this in the pattern in the future - once we get more adoption perhaps and/or more versions.

akash1810: What about pulling the layer from S3 via [`code`](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda.LayerVersion.html#code)?

kenoir: The [lambda layer](https://github.com/guardian/ssm-to-env/tree/main) can be deployed via riff-raff which was updated to support deploying lambda layers, that will bump the version number when successfully deployed (the layer is configured to available organisation wide).  You could potentially store the targeted layer version in SSM and configure that here as with the deploy tools account number, or rely on a GuCDK version update to change the version used.

PR #1789, opened 2023-03-28
feat: Add devx global dist bucket

nicl:
**Note, interested in what the actual path here should be. Perhaps `/account/external...`? instead of under services?**

Adds global dist bucket so that we can reference it. This bucket contains artifacts that need to be accessilble to all of our AWS accounts.

Once merged, I will raise a PR on aws-account-setup to add this parameter to all of our accounts. There is somewhat of a chicken and egg issue here, but the window of time between being present in CDK and in accounts (via aws-account-setup) is likely to be small and the only users, initially at least, Devx.

### DIFF
--- a/src/constants/ssm-parameter-paths.ts
+++ b/src/constants/ssm-parameter-paths.ts
@@ -18,6 +18,7 @@ interface NamedSsmParameterPaths {
   PrimaryVpcPrivateSubnets: SsmParameterPath;
   PrimaryVpcPublicSubnets: SsmParameterPath;
   FastlyCustomerId: SsmParameterPath;
+  DevxGlobalDistBucket: SsmParameterPath;
 }
 
 export const VPC_SSM_PARAMETER_PREFIX = "/account/vpc";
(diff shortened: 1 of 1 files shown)


src/constants/ssm-parameter-paths.ts
```diff
@@ -58,6 +59,10 @@ export const NAMED_SSM_PARAMETER_PATHS: NamedSsmParameterPaths = {
     path: `${VPC_SSM_PARAMETER_PREFIX}/primary/subnets/public`,
     description: "A comma-separated list of public subnets",
   },
+  DevxGlobalDistBucket: {
+    path: "/account/services/devx.global.bucket",
```
akash1810, line 63: Does `global` imply the bucket is reachable outside of the org? How about `/organisation/services/artifact.bucket`?

akash1810 (approved):
(non-blocking) I've offered an alternative path, but not sure if it's any better/worse.

src/constants/ssm-parameter-paths.ts
nicl: this feels a lot better to me :)

PR #1838, opened 2023-04-28
feat: Allow riff-raff.yaml auto-generation to work with managed lambda deploys

kenoir:
## What does this change?

Updates CDK to allow riff-raff.yaml auto-generation to work with managed (not uploaded by the consuming stack) lambda deploys. This is to allow CDK to reference a lambda S3 sources that are managed and uploaded by DevX (for example the [cognito auth lambda](https://github.com/guardian/cdk/blob/main/src/patterns/ec2-app/base.ts#L485)), to deploy successfully from a  generated riff-raff.yaml.

RiffRaff changes are also required to support this: https://github.com/guardian/riff-raff/pull/1129

## How to test

This is being tested by the deployment of https://github.com/guardian/birthdays/pull/75

The `googleAuth` option in the `guEc2App` pattern can be used successfully by non-DevX teams.

### DIFF
--- a/src/constructs/lambda/lambda.ts
+++ b/src/constructs/lambda/lambda.ts
@@ -92,6 +92,7 @@ function defaultMemorySize(runtime: Runtime, memorySize?: number): number {
 export class GuLambdaFunction extends Function {
   public readonly app: string;
   public readonly fileName: string;
+  public readonly bucketNamePath: string | undefined;
 
   constructor(scope: GuStack, id: string, props: GuFunctionProps) {
     const { app, fileName, runtime, memorySize, timeout, bucketNamePath, withoutFilePrefix } = props;
@@ -122,6 +123,7 @@ export class GuLambdaFunction extends Function {
 
     this.app = app;
     this.fileName = fileName;
+    this.bucketNamePath = bucketNamePath;
 
     if (props.errorPercentageMonitoring) {
       new GuLambdaErrorPercentageAlarm(scope, `${id}-ErrorPercentageAlarmForLambda`, {
--- a/src/constructs/lambda/lambda.ts
+++ b/src/constructs/lambda/lambda.ts
@@ -46,7 +46,7 @@ export interface GuFunctionProps extends GuDistributable, Omit<FunctionProps, "c
   /**
    * TODO
    */
-  managedDeployment?: boolean
+  managedDeployment?: boolean;
 }
 
 function defaultMemorySize(runtime: Runtime, memorySize?: number): number {
@@ -98,6 +98,7 @@ export class GuLambdaFunction extends Function {
   public readonly fileName: string;
   public readonly bucketNamePath: string | undefined;
   public readonly managedDeployment: boolean;
+  public readonly withoutFilePrefix: boolean;
 
   constructor(scope: GuStack, id: string, props: GuFunctionProps) {
     const { app, fileName, runtime, memorySize, timeout, bucketNamePath, withoutFilePrefix, managedDeployment } = props;
@@ -130,6 +131,7 @@ export class GuLambdaFunction extends Function {
     this.fileName = fileName;
     this.bucketNamePath = bucketNamePath;
     this.managedDeployment = managedDeployment ?? false;
+    this.withoutFilePrefix = withoutFilePrefix ?? false;
 
     if (props.errorPercentageMonitoring) {
       new GuLambdaErrorPercentageAlarm(scope, `${id}-ErrorPercentageAlarmForLambda`, {
--- a/src/constructs/lambda/lambda.ts
+++ b/src/constructs/lambda/lambda.ts
@@ -43,6 +43,10 @@ export interface GuFunctionProps extends GuDistributable, Omit<FunctionProps, "c
    * accounts.
    */
   withoutFilePrefix?: boolean;
+  /**
+   * TODO
+   */
+  managedDeployment?: boolean
 }
 
 function defaultMemorySize(runtime: Runtime, memorySize?: number): number {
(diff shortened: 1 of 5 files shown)


src/constructs/lambda/lambda.ts
```diff
@@ -92,9 +96,12 @@ function defaultMemorySize(runtime: Runtime, memorySize?: number): number {
 export class GuLambdaFunction extends Function {
   public readonly app: string;
   public readonly fileName: string;
+  public readonly bucketNamePath: string | undefined;
+  public readonly managedDeployment: boolean;
+  public readonly withoutFilePrefix: boolean;
 
   constructor(scope: GuStack, id: string, props: GuFunctionProps) {
-    const { app, fileName, runtime, memorySize, timeout, bucketNamePath, withoutFilePrefix } = props;
+    const { app, fileName, runtime, memorySize, timeout, bucketNamePath, withoutFilePrefix, managedDeployment } = props;
```
akash1810, line 104: I think destructuring with default values would be simpler here, making the value more immediately obvious, and removes the need to nullish coalesce:

```ts
const {
  app,
  fileName,
  runtime,
  memorySize,
  timeout,
  buckNamePath,
  withoutFilePrefix = false,
  withoutFilePrefix = false
} = props;
```

src/constructs/lambda/lambda.ts
```diff
@@ -43,6 +43,10 @@ export interface GuFunctionProps extends GuDistributable, Omit<FunctionProps, "c
    * accounts.
    */
   withoutFilePrefix?: boolean;
+  /**
+   * TODO
+   */
+  managedDeployment?: boolean;
```
akash1810, line 49: 👀

src/experimental/riff-raff-yaml-file/deployments/lambda.ts
```diff
@@ -2,6 +2,32 @@ import type { GuStack } from "../../../constructs/core";
...
 
+const locationProps = (
+  lambda: GuLambdaFunction
+): {
+  bucketSsmLookup?: boolean;
+  bucketSsmKey?: string;
+  prefixStackToKey?: boolean;
+  prefixAppToKey?: boolean;
+  prefixStageToKey?: boolean;
+} => {
```
akash1810, line 13: Wonder if it's worth pulling this out into a discrete `interface` to improve readability with commented fields? Either in this same file, or within the [types.ts](https://github.com/guardian/cdk/blob/main/src/experimental/riff-raff-yaml-file/types.ts)?

akash1810 (commented):
Could we add an explicit test for the use-case mentioned? IIUC the expectation is the resulting `riff-raff.yaml` file should not include an `upload-lambda` deployment, it's a little difficult to say definitively w/out a test, IMO.

src/constructs/lambda/lambda.ts
kenoir: @akash1810 I've added some documentation here, but I wonder if this parameter should more explicitly be called something like `skipUpload`? What do you think? I'm slightly uncomfortable coupling the name to the functionality coded elsewhere, especially as the `riff-raff.yaml` autogeneration is still "experimental" (IMO it no longer is!).

akash1810: > I wonder if this parameter should more explicitly be called something like `skipUpload`?

👍🏽 this feels better, yes! We've a few `withoutX` props in the library, it's not a formal pattern, but wonder if `withoutArtifactUpload` is an alternative name?

> especially as the `riff-raff.yaml` autogeneration is still "experimental" (IMO it no longer is!).

We've yet to define a process to promote experimental features. Maybe we just move it and create a new major version?

kenoir: Will rename to `withoutArtifactUpload`. 

> Maybe we just move it and create a new major version? This wasn't covered by the [initial ADR](https://github.com/guardian/cdk/blob/main/docs/architecture-decision-records/005-x01-package-structure.md).

This sounds good to me. I think the feature now has enough use for us to say it's not experimental anymore.

src/experimental/riff-raff-yaml-file/deployments/lambda.ts
```diff
@@ -2,6 +2,32 @@ import type { GuStack } from "../../../constructs/core";
...
+
+const locationProps = (lambda: GuLambdaFunction): S3LocationProps => {
+  const bucketProp =
+    lambda.bucketNamePath === undefined
+      ? {
+          bucketSsmLookup: true,
+        }
+      : {
+          bucketSsmKey: lambda.bucketNamePath,
+        };
```
nicl, line 21: I wasn't sure if `bucketSsmLookup: true` is required in both cases. But looking at RR it looks not the case - though this serves as a reminder that we should remove the `bucketSsmLookup` parameter fullstop and just use whichever of bucketName or bucketSsmKey is set (or the default value for the latter if neither).

src/constructs/lambda/lambda.ts
nicl: @kenoir yep! We discussed this recently and will hopefully promote GuRoot to non-experimental soon, as it now has wide usage and proven value.

nicl (approved):
Thanks for this - nice solution to a complex problem!

src/constructs/lambda/lambda.ts
```diff
@@ -92,9 +100,21 @@ function defaultMemorySize(runtime: Runtime, memorySize?: number): number {
...
-    const { app, fileName, runtime, memorySize, timeout, bucketNamePath, withoutFilePrefix } = props;
+    const {
+      app,
+      fileName,
+      runtime,
+      memorySize,
+      timeout,
+      bucketNamePath,
+      withoutFilePrefix = false,
+      withoutArtifactUpload: withoutArtifactUpload = false,
```
akash1810, line 116: This syntax is confusing me a little. Does this not work?

```
const { withoutArtifactUpload = false } = props;
```

src/experimental/riff-raff-yaml-file/deployments/lambda.ts
kenoir: I think this change should be made in a separate PR. Updating in this branch modifies existing behaviour and causes the `riff-raff.yaml` generation tests outside of the scope of this change to fail, (although I do not expect the change to be impactful i'd rather be cautious).

PR #1849, opened 2023-05-05
Default to NPM since this is the recommendation

davidfurey:
## What does this change?

Sets a default value for the argument `package-manager`.  This matches the recommendation and makes the default workflow less verbose.  

If this isn't desired, we should update the riff-raff example to include an explicit `--package-manager` (https://github.com/guardian/cdk/blob/main/docs/setting-up-a-gucdk-project.md?plain=1#L24)


akash1810:
> we should update the riff-raff example to include an explicit `--package-manager`

I'd opt for this, only because Yarn is popular across the department[^1].

[^1]: Personally, I prefer NPM after encountering too many subtle issues with Yarn over the years.

PR #1850, opened 2023-05-05
docs: Note about test-update

davidfurey:
## What does this change?

Documentation change to show how to update snapshot tests.


akash1810 (approved):
Should the yarn equivalent be listed too?

davidfurey:
> Should the yarn equivalent be listed too?

Possibly, although maybe it is reasonable to expect `yarn` users to be power users (since that isn't the default path) and know how to translate from npm to yarn terminology? Is it `yarn run test-update`?

akash1810:
> Is it `yarn run test-update`?

👍🏽  Unlike npm, yarn doesn't require the explicit `run` for custom scripts, so `yarn test-update` also works.

PR #1853, opened 2023-05-10
feat: GuEc2App custom healthcheck

davidfurey:
## What does this change?

Allows you to customise the `GuEc2App` healthcheck with a prop, because sometimes `/healthcheck` isn't the path that you want for your healthcheck, and passing a prop is nicer than calling `app.targetGroup.configureHealthCheck`


src/patterns/ec2-app/base.ts
```diff
@@ -11,6 +11,7 @@ import {
 } from "aws-cdk-lib/aws-cognito";
 import type { InstanceType, IPeer, ISubnet, IVpc } from "aws-cdk-lib/aws-ec2";
 import { Port } from "aws-cdk-lib/aws-ec2";
+import type { HealthCheck as ELBHealthCheck } from "aws-cdk-lib/aws-elasticloadbalancingv2";
```
akash1810, line 14: minor: ELBv2 is an Application Load Balancer, should the import alias be `ALBHealthCheck`?

akash1810 (approved):

src/patterns/ec2-app/base.ts
akash1810: Agreed, the naming is not helpful. I think this could either be `ALBHealthCheck` or `ELBv2HealthCheck`, but not `ELBHealthCheck` which is a misleading mix of both, with overlaps with the classic load balancers (ELBs).

PR #1866, opened 2023-05-23
feat(GuStack): Add a getter for the repository name

akash1810:
## What does this change?
Adds a new `repositoryName` field to a `GuStack`, returning the value of the GitHub repository that is already added as a `gu:repo` tag to the stack (see #537).

Initially, this is to address [a TODO in service-catalogue](https://github.com/guardian/service-catalogue/blob/baff2dc75744e847ae3ff437e78c2f5e2f30cfcb/packages/cdk/lib/ecs/task.ts#L90); it feels useful outside of that service, so adding it to GuCDK.

## How to test

See added test.

Less work for consuming projects to do?

N/A

## Checklist

### DIFF
--- a/src/constructs/core/stack.test.ts
+++ b/src/constructs/core/stack.test.ts
@@ -54,6 +54,19 @@ describe("The GuStack construct", () => {
     });
   });
 
+  it("should expose the repository information as a property", () => {
+    const stack = new GuStack(
+      new App({ context: { [ContextKeys.REPOSITORY_URL]: "https://github.com/guardian/my-repository" } }),
+      "Test",
+      {
+        stack: "test",
+        stage: "TEST",
+      }
+    );
+
+    expect(stack.repositoryName).toEqual("guardian/my-repository");
+  });
+
   it("can persist the logicalId for any resource (consumer's POV)", () => {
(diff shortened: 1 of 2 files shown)


src/constructs/core/stack.test.ts
```diff
@@ -54,6 +54,19 @@ describe("The GuStack construct", () => {
     });
   });
 
+  it("should expose the repository information as a property", () => {
```
NovemberTang, line 57: might be good to test the negative test too, but not a blocker

NovemberTang (approved):

src/constructs/core/stack.test.ts
akash1810: Good idea. Alas, I'm not sure this is easily done. The value is obtained from the context, or by [calling `git config --get remote.origin.url`](https://github.com/guardian/cdk/blob/001912ce5d93a90c6ab5bba9828ce9352ef92e85/src/utils/git.ts#L3-L7). The latter will always resolve to `guardian/cdk`. I'm not sure how best to fall into the `undefined` path in a test. We could mock this, but not sure its worth it. WDYT?

PR #1874, opened 2023-05-26
fix: Add explicit permissions to cd.yaml

kenoir:
## What does this change?

This change adds explicit permissions to the CD GitHub workflow in order to fix an issue with the build where these are required (see https://github.com/guardian/cdk/actions/runs/5090361143/jobs/9149327417#step:4:226).

`contents: write`: required to allow pushing the release tag
`id-token: write`: To allow authentication with NPM

## How to test

Merge this to main and the build should be able to release a new version.

People can continue to release and use new versions of `guardian/cdk`.

### DIFF
--- a/.github/workflows/cd.yaml
+++ b/.github/workflows/cd.yaml
@@ -7,6 +7,9 @@ on:
 jobs:
   CD:
     runs-on: ubuntu-latest
+    permissions:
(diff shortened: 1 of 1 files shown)


akash1810 (approved):
💎 thanks!

PR #1882, opened 2023-06-02
Add kinesis stream arn ssm param constant

vlbee:
## What does this change?

In order to set up cloudwatch-log-management to set retention and ship to the ELK stack, we need to have the kinesis stream arn set up by aws-account-set up (which was only setting up the Name). aws-account-setup has a dependency on cdk for the ssm parameter path and description. 

Required for these two PRs:
https://github.com/guardian/aws-account-setup/pull/69
https://github.com/guardian/cloudwatch-logs-management/pull/245


akash1810 (commented):
I think we should consider this parameter as legacy, and thus I'm not sure we should encode it in GuCDK.

PR #1891, opened 2023-06-13
fix: add `subnets` prop to `GuScheduledEcsTask`

marsavar:
## What does this change?
Adds a new optional prop to this pattern.

When attempting to synth a template, the command fails with this error message:

```
 There are no 'Private' subnet groups in this VPC.
```

This is expected - it's not possible to know what kinds of subnets are in a given VPC at compile time if a context file isn't present. Passing in this prop should address the issue and make synth work.

## How to test
I made this change to my local version of GuCDK and updated `package.json` to point to the version of GuCDK in my filesystem rather than the one from npm, so that I was able to test this change in my stack. After passing in private subnets obtained with `GuVpc.subnetsFromParameter` I was able to `synth` without errors. 

We don't have to commit a context file to version control when using this pattern.

## Checklist

### DIFF
--- a/src/constructs/ecs/__snapshots__/ecs-task.test.ts.snap
+++ b/src/constructs/ecs/__snapshots__/ecs-task.test.ts.snap
@@ -209,8 +209,8 @@ exports[`The GuEcsTask pattern should create the correct resources with lots of 
   "Metadata": {
     "gu:cdk:constructs": [
       "GuStack",
-      "GuSubnetListParameter",
       "GuEcsTask",
+      "GuSubnetListParameter",
       "GuDistributionBucketParameter",
     ],
     "gu:cdk:version": "TEST",
@@ -325,7 +325,7 @@ exports[`The GuEcsTask pattern should create the correct resources with lots of 
       },
       "Type": "AWS::IAM::Role",
     },
-    "CdkJsonStringify1": {
+    "CdkJsonStringify2": {
       "DeletionPolicy": "Delete",
       "Properties": {
         "ServiceToken": {
@@ -459,7 +459,7 @@ exports[`The GuEcsTask pattern should create the correct resources with lots of 
               "","TaskDefinition":"test-stack-TEST-ecs-test","NetworkConfiguration":{"AwsvpcConfiguration":{"Subnets":",
               {
                 "Fn::GetAtt": [
-                  "CdkJsonStringify1",
+                  "CdkJsonStringify2",
                   "Value",
                 ],
               },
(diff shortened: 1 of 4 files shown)


src/constructs/ecs/ecs-task.ts
```diff
@@ -200,6 +202,7 @@ export class GuEcsTask extends Construct {
 
     const task = new EcsRunTask(scope, `${id}-task`, {
       cluster,
+      subnets: { subnets },
```
akash1810, line 205: I think we can be more opinionated here, and always place the task in the private subnet (similar to how we implement EC2 apps). WDYT?

src/constructs/ecs/__snapshots__/ecs-task.test.ts.snap
```diff
@@ -3,6 +3,107 @@
 exports[`The GuEcsTask pattern should create the correct resources with lots of config 1`] = `
 {
   "Mappings": {
+    "DefaultCrNodeVersionMap": {
```
akash1810, line 6: Do we know where these new mappings originate from? Feels unrelated to the change we're making in this PR, so it's surprising to see.

src/constructs/ecs/ecs-task.test.ts
```diff
@@ -22,6 +23,8 @@ const testPolicy = new PolicyStatement({
   resources: ["databaseSecretArn"],
 });
 
+const subnets = (stack: GuStack, type: SubnetType, app?: string) => GuVpc.subnetsFromParameter(stack, { type, app });
```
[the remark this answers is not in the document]
marsavar: Oops, you're right - I wrote that function with grandiose plans for testing but it's fairly redundant :) will remove

akash1810 (commented):
Worth noting that `service-catalogue` takes a slightly different approach to ECS than the GuCDK library, namely it:
  - Places all tasks in one cluster (GuCDK creates a cluster per task)
  - Implements a [VPC work-around](https://github.com/guardian/service-catalogue/blob/f5c89b18aca523e1704fc2395688f304cacb45ed/packages/cdk/lib/cloudquery.ts#L54-L68)

Appreciate this isn't strictly related to this PR, but worth considering.

src/constructs/ecs/ecs-task.ts
marsavar: Were you imagining something like this?
 
```typescript
const privateSubnets = GuVpc.subnetsFromParameter(scope, { type: SubnetType.PRIVATE, app });

const task = new EcsRunTask(scope, `${id}-task`, {
    cluster,
    subnets: { subnets: privateSubnets },
    ...
}
```
This would effectively make the `subnets` prop redundant, since we're defining the subnets in the pattern constructor.

marsavar: All done, thanks. Have also added a couple of tests.
Let me know if I need to do anything else before I merge! :-)

akash1810 (approved):

PR #1898, opened 2023-06-15
chore: Allow CD to comment on PRs

akash1810:
## What does this change?
Extend the [permissions](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) for the CD workflow to allow the ability to comment on PRs.

We use the `semantic-release` tool to publish to NPM. One feature of this tool is it will comment and label PRs to shwo which version it's in. For example:

image

We've stopped seeing this recently. Checking the CD logs, it appears to related to a lack of permissions.

image

Extend the permissions to allow for writing PRs (via https://github.com/orgs/community/discussions/26644).

## How to test
Merge the PR, and check the CD logs.

We reinstate previously helpful features.

N/A

## Checklist

### DIFF
--- a/.github/workflows/cd.yaml
+++ b/.github/workflows/cd.yaml
@@ -10,6 +10,7 @@ jobs:
     permissions:
       contents: write
       id-token: write
+      pull-requests: write
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3


akash1810:
> Merge the PR, and check the CD logs.

Alas, it appears that a release needs to be made for us to see things working!

PR #1899, opened 2023-06-15
fix(email): Remove sender restriction

akash1810:
## What does this change?

Removes the restriction for an email to be send from the domain `theguardian.com`.

There are a few services in the Journalism stream using the `gutools.co.uk` domain, so this restriction was overly tight.

Additionally, "shift left" by making the email sender a compile time prop, rather than a CFN Parameter, which is only known at runtime. Migration involves setting this new prop, which the compiler will help with.

## How to test

See the updated tests.

Supporting more teams.

This change removes the `GuGuardianEmailSenderParameter` parameter. There [doesn't appear to be extensive usage of it](https://github.com/search?q=org%3Aguardian+GuGuardianEmailSenderParameter&type=code), so there is little risk here.

## Checklist


akash1810:
This PR also allows us to test https://github.com/guardian/cdk/pull/1898.

src/constructs/iam/policies/ses.ts
```diff
@@ -1,14 +1,20 @@
-import { GuGuardianEmailSenderParameter } from "../../core";
 import type { GuStack } from "../../core";
 import { GuAllowPolicy } from "./base-policy";
 
-export class GuSESSenderPolicy extends GuAllowPolicy {
-  constructor(scope: GuStack) {
-    const emailSenderParam = new GuGuardianEmailSenderParameter(scope);
+interface GuSESSenderPolicyProps {
+  /**
+   * The email address to allow sending from.
```
kenoir, line 6:
```suggestion
   * The email address or domain name to allow sending from.
```

kenoir (approved):
👌 one comment, not blocking.

PR #1906, opened 2023-06-23
feat!: Update GuLambdaErrorPercentageAlarm

kenoir:
## What does this change?

Adds props for `lengthOfEvaluationPeriod` and replaces `numberOfMinutesAboveThresholdBeforeAlarm` with `numberOfEvaluationPeriodsAboveThresholdBeforeAlarm`.

This allows `GuLambdaErrorPercentageAlarm` to monitor scheduled lambdas more effectively by taking into account the cadence with which the lambda runs.

For example, in this case for a lambda that runs every 15 minutes we want to know if it fails on twice concurrent runs, we would set:

```ts
const alarmProps = {
        toleratedErrorPercentage: 1,
	lengthOfEvaluationPeriod: Duration.minutes(15),
        numberOfEvaluationPeriodsAboveThresholdBeforeAlarm: 2,
}
```

See https://github.com/guardian/pressreader/pull/42 for an example of where overriding the current behaviour was required to monitor effectively.

We may also be able to do this specifically for scheduled lambdas, however my attempts to make that change in CDK result in having to reach from constructs into patterns, and required (for me) tricky fiddling with inheritance which arguably adds unnecessary complexity. Changing the prop name instead flags to consumers more clearly how the alarm is functioning.

## How to test

- [X] Run the tests. 
- [x] Include this version of the CDK via this branch and test locally the correct CF is provisioned. 

See https://github.com/guardian/pressreader/pull/45 for an example of use locally before merge.

Better monitoring of scheduled lambdas.

This is a breaking change as it renames an existing prop to be clearer about its purpose.

The migration path is to update references to `numberOfMinutesAboveThresholdBeforeAlarm` -> `numberOfEvaluationPeriodsAboveThresholdBeforeAlarm`.

## Checklist

- [x] ~I have updated the documentation as required for the described changes [^2]~

### DIFF
--- a/src/constructs/cloudwatch/lambda-alarms.test.ts
+++ b/src/constructs/cloudwatch/lambda-alarms.test.ts
@@ -1,3 +1,4 @@
+import { Duration } from "aws-cdk-lib";
 import { Template } from "aws-cdk-lib/assertions";
 import { Runtime } from "aws-cdk-lib/aws-lambda";
 import { simpleGuStackForTesting } from "../../utils/test";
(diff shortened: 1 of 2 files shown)


kenoir:
Question: What further API / migration path documentation changes need to be made?

akash1810:
I've yet to properly think about this PR, but:

> This is a breaking change as it renames an existing prop to be clearer about its purpose.

This prop is [used in a few places](https://github.com/search?q=org%3Aguardian+numberOfMinutesAboveThresholdBeforeAlarm++NOT+is%3Aarchived+NOT+repo%3Aguardian%2Fcdk+&type=code), so releasing as a breaking change is sensible. We'd want to [have a `!` in the commit and PR title](https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with--to-draw-attention-to-breaking-change) to create a breaking change.

akash1810:
This change looks reasonable to me 👍🏽. Let's [update](https://github.com/guardian/cdk/pull/1906#issuecomment-1604051252) the commit/PR to denote the breaking change, and ship it.

akash1810 (approved):

PR #1915, opened 2023-06-30
feat: Add GuEmailIdentity

kenoir:
## What does this change?

This change adds a `GuEmailIdentity` construct to encode how to set up a [SES verified identity](https://docs.aws.amazon.com/ses/latest/dg/verify-addresses-and-domains.html) for a `gutools.co.uk` domain.

## How to test

- [X] Run the tests. 
- [X] Include this version of the CDK via this branch and test locally the correct CF is provisioned. 

See https://github.com/guardian/image-syndication/pull/19/commits/0093e299b8c9d5c514c529e487792085342597fd for testing before merge.

This is usable in our projects, simplifying configuration where consumers need to send emails via SNS.

## Checklist

### DIFF
--- a/src/constructs/ses/index.ts
+++ b/src/constructs/ses/index.ts
@@ -1,0 +1,28 @@
+import { Duration } from "aws-cdk-lib";
+import type { EmailIdentityProps } from "aws-cdk-lib/aws-ses";
+import { EmailIdentity, Identity } from "aws-cdk-lib/aws-ses";
+import type { AppIdentity, GuStack } from "../core";
+import { GuCname } from "../dns";
+
+export interface GuEmailIdentityProps extends Omit<EmailIdentityProps, "identity">, AppIdentity {
+  guToolsSubdomain: string;
+}
+
+export class GuEmailIdentity extends EmailIdentity {
+  constructor(scope: GuStack, id: string, props: GuEmailIdentityProps) {
+    super(scope, id, {
+      // Auto verification is only supported for gutools.co.uk domains
+      identity: Identity.domain(`${props.guToolsSubdomain}.gutools.co.uk`),
+      ...props,
+    });
+
+    this.dkimRecords.forEach(({ name, value }, index) => {
+      new GuCname(scope, `EmailIdentityDkim${index}`, {
+        app: props.app,
+        domainName: name,
+        resourceRecord: value,
+        ttl: Duration.hours(1),
+      });
+    });
+  }
+}
--- a/src/constructs/ses/index.ts
+++ b/src/constructs/ses/index.ts
@@ -1,6 +1,7 @@
 import { Duration } from "aws-cdk-lib";
 import type { EmailIdentityProps } from "aws-cdk-lib/aws-ses";
 import { EmailIdentity, Identity } from "aws-cdk-lib/aws-ses";
+import { GuAppAwareConstruct } from "../../utils/mixin/app-aware-construct";
 import type { AppIdentity, GuStack } from "../core";
 import { GuCname } from "../dns";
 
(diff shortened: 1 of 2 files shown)


src/constructs/ses/index.ts
```diff
@@ -0,0 +1,28 @@
...
+import { GuCname } from "../dns";
+
+export interface GuEmailIdentityProps extends Omit<EmailIdentityProps, "identity">, AppIdentity {
+  guToolsSubdomain: string;
+}
+
+export class GuEmailIdentity extends EmailIdentity {
+  constructor(scope: GuStack, id: string, props: GuEmailIdentityProps) {
+    super(scope, id, {
+      // Auto verification is only supported for gutools.co.uk domains
```
twrichards, line 14: It supports `code.dev-gutools.co.uk` too - I'd suggest we don't hard code the top level domain, since most applications will have the whole domain as a variable they'd like to pass on which they probably wouldn't want to split just to conform to this inner implementation detail

kenoir: I was trying to encode the requirements we want here in the construct, but agreed it is unnecessarily fiddly.

twrichards: perhaps you could have an `endsWith("gutools.co.uk")` (not DOT gutools, to support CODE too) check and throws if other top level domain is used!

kenoir: Ah, neat - I will do this, less complicated than i'd imagined.

src/constructs/ses/index.ts
```diff
@@ -0,0 +1,47 @@
...
+ *  app: "test",
+ * });
+ **/
+export class GuEmailIdentity extends GuAppAwareConstruct(EmailIdentity) {
+  constructor(scope: GuStack, id: string, props: GuEmailIdentityProps) {
+    const { app, domainName } = props;
+    if (!domainName.endsWith("gutools.co.uk")) {
+      throw new Error(
+        `Auto verification is only supported for gutools.co.uk domains. ${domainName} is not a gutools.co.uk domain.`
+      );
```
akash1810, line 30: Why are we limited to this domain?

twrichards: because we're using GuCname under the hood

akash1810: Ah. We've recently broadened the domains available with this tooling. See https://github.com/guardian/dns-validation-lambda/pull/43/files for a complete list[^1]. It's not massively publicised yet because we still need to properly document use-cases for the domains.

Maybe it's worth adding an inline todo comment along the lines of:

> TODO: expand this as https://github.com/guardian/cfn-private-resource-types/tree/main/dns/guardian-dns-record-set-type does

?

[^1]: The link is to an unrelated repository, however the credentials used between that and the DNS lambdas are the same.

src/constructs/ses/index.ts
```diff
@@ -0,0 +1,47 @@
+import { Duration } from "aws-cdk-lib";
+import type { EmailIdentityProps } from "aws-cdk-lib/aws-ses";
+import { EmailIdentity, Identity } from "aws-cdk-lib/aws-ses";
+import { GuAppAwareConstruct } from "../../utils/mixin/app-aware-construct";
+import type { AppIdentity, GuStack } from "../core";
+import { GuCname } from "../dns";
+
+export interface GuEmailIdentityProps extends Omit<EmailIdentityProps, "identity">, AppIdentity {
+  domainName: string;
```
akash1810, line 9: Could we add some inline documentation here, commenting on the need to be a `gutools.co.uk` domain?

akash1810 (approved):
A couple of documentation requests, but overall LGTM.

PR #1916, opened 2023-07-03
feat!: Add `httpPutResponseHopLimit` to ASG and EC2 props

zekehuntergreen:
Co-authored-by: Philip McMahon <philip.mcmahon@guardian.co.uk>

## What does this change?
Adds an optional parameter to the constructors of `GuAutoScalingGroup` and `GuEc2App`, allowing consumers to specify [`httpPutResponseHopLimit`](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_InstanceMetadataOptionsRequest.html).

## Why
- Our [Aleph server](https://aleph.code.dev-gutools.co.uk/) is running in docker on EC2. 
- We're working on a [change to Aleph](https://github.com/alephdata/aleph/issues/3185) that will enable role-based auth from EC2 to S3 as an alternative to user-based auth. 
- The default `httpPutResponseHopLimit` of 1 prevents the docker container from fetching EC2 instance metadata so the limit needs to be raised to 2.

## How to test

Tested locally by installing this branch in our [Aleph cdk](https://github.com/guardian/investigations-platform/tree/aleph/aleph) package.

## Checklist

### DIFF
--- a/src/constructs/autoscaling/asg.ts
+++ b/src/constructs/autoscaling/asg.ts
@@ -38,6 +38,7 @@ export interface GuAutoScalingGroupProps
   additionalSecurityGroups?: ISecurityGroup[];
   targetGroup?: ApplicationTargetGroup;
   withoutImdsv2?: boolean;
+  httpPutResponseHopLimit?: number;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


akash1810 (commented):
Thanks for this!

The use of `!` in the commit/PR title suggests this is a breaking change, however AFAICT the new property is optional, thus is backwards compatible - do you mean to create a breaking change?

Could we add a test to demonstrate how `httpPutResponseHopLimit` impacts the synthed template when set?

zekehuntergreen:
Thanks @akash1810 !
I've added a test and renamed the PR

akash1810 (approved):
LGTM.

Looking at the commit messages, I'd suggest squashing and merging to ensure a release gets made.

PR #1925, opened 2023-07-10
feat: add Lambda versioning support

jacobwinch:
## What does this change?

This PR adds support for using Lambda [versions](https://docs.aws.amazon.com/lambda/latest/dg/configuration-versions.html) and [aliases](https://docs.aws.amazon.com/lambda/latest/dg/configuration-aliases.html). These features are not currently used at the Guardian, but they are necessary for teams who want to make use of features which rely on versioning (e.g. SnapStart or Provisioned Concurrency).

The primary motivation for this PR is unblocking @tomrf1, who is keen to start using SnapStart after confirming that it [offers very significant performance improvements for his project](https://github.com/guardian/support-frontend/pull/4869). There is some more general background on SnapStart [here](https://docs.google.com/document/d/1ZC9o70gfWoEgfmbd4C0w87SUYqzbMx5CID38EOqf8dA/edit#heading=h.dowx40pffo62).

Because we have never used versions and aliases before, all of our GuCDK patterns currently invoke the [unqualified  function ARN](https://docs.aws.amazon.com/lambda/latest/dg/configuration-versions.html#versioning-versions-using) (i.e. `$LATEST`). `$LATEST` is typically mutated outside of CDK/CloudFormation, via an API call made by Riff-Raff. This PR modifies the Lambda-based patterns so that they invoke an alias (which points to the newest Lambda version) when developers enable versioning via construct/pattern props. 

The new Lambda version (and corresponding alias update) need to be provisioned via CDK/CloudFormation; this can be achieved by modifying the filename used so that AWS understands that the function needs to change.

## How to test

@tomrf1 and I have tested this change for scheduled Lambdas (see https://github.com/guardian/support-frontend/pull/4869 and https://github.com/guardian/elastic-search-monitor/pull/31).

I have also modified all of the other Lambda-based patterns to use the same technique (i.e. invoking the alias instead of `$LATEST`) and have added unit tests for these, but I have not tested these other patterns with real infrastructure.

This should unblock https://github.com/guardian/support-frontend/pull/4869 and allow us to promote this feature more widely.

This is a new approach for the Guardian so any projects which move to this approach should perform thorough testing in `CODE` before deploying to `PROD`.

In addition, this new approach relies on uploading more unique artifacts to S3 (e.g. we upload a new file - i.e. `lambda-v1.jar`, `lambda-v2.jar` etc. - rather than replacing a single artifact - `lambda.jar` - as part of every deployment). In the short-term this should not cause problems (because S3 storage is cheap), but in the longer term we will need to figure out an approach for tidying up old version artifacts.

## Checklist

- [x] I have listed any breaking changes, along with a migration path [^1] - There are no breaking changes so this should not affect existing projects.

### DIFF
--- a/src/constructs/lambda/lambda.test.ts
+++ b/src/constructs/lambda/lambda.test.ts
@@ -192,4 +192,33 @@ describe("The GuLambdaFunction class", () => {
       },
     });
   });
+
+  it("should not create an alias or version if the enableVersioning prop is unset", () => {
+    const stack = simpleGuStackForTesting();
+
+    new GuLambdaFunction(stack, "lambda", {
+      fileName: "my-app.jar",
+      handler: "handler.ts",
+      runtime: Runtime.JAVA_8,
+      app: "testing",
+    });
+
+    Template.fromStack(stack).resourceCountIs("AWS::Lambda::Version", 0);
+    Template.fromStack(stack).resourceCountIs("AWS::Lambda::Alias", 0);
+  });
+
+  it("should create a function version and alias if the enableVersioning prop is set to true", () => {
+    const stack = simpleGuStackForTesting();
+
+    new GuLambdaFunction(stack, "lambda", {
+      enableVersioning: true,
+      fileName: "build123.jar",
+      handler: "handler.ts",
+      runtime: Runtime.JAVA_8,
+      app: "testing",
+    });
+
+    Template.fromStack(stack).resourceCountIs("AWS::Lambda::Version", 1);
+    Template.fromStack(stack).resourceCountIs("AWS::Lambda::Alias", 1);
+  });
 });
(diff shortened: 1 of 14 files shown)


tomrf1 (approved):
Great! 🎉

akash1810 (approved):
👍🏽

jacobwinch:
> In addition, this new approach relies on uploading more unique artifacts to S3 (e.g. we upload a new file - i.e. `lambda-v1.jar`, `lambda-v2.jar` etc. - rather than replacing a single artifact - `lambda.jar` - as part of every deployment). In the short-term this should not cause problems (because S3 storage is cheap), but in the longer term we will need to figure out an approach for tidying up old version artifacts.

There's a potential solution for this here: https://github.com/guardian/riff-raff/pull/1207.

PR #1952, opened 2023-07-25
fix: Don't set prefix settings if using default values in `riff-raff.yml`

bryophyta:
## What does this change?

Only sets an explicit value for `prefixStackToKey`, `prefixAppToKey`, and `prefixStageToKey` if they are set to false (as `true` is the default) in `riff-raff.yml`. RiffRaff complains if you explicitly set these with the default values.

## How to test
I've edited the unit tests, and added a specific test case to check the behaviour.

Less noise when deploying auto-generated `riff-raff.yml`.

## Checklist


akash1810 (approved):
Thanks for this!

PR #1975, opened 2023-08-15
chore: Remove experimental error budget alarm construct

jacobwinch:
## What does this change?

This experimental construct (introduced via https://github.com/guardian/cdk/pull/1576) is currently [only used](https://github.com/search?q=org%3Aguardian%20GuErrorBudgetAlarmExperimental&type=code) in https://github.com/guardian/slo-alerts so the code will be maintained there going forwards (https://github.com/guardian/slo-alerts/pull/22). There was a previous usage in another repo that has been removed via https://github.com/guardian/mobile-n10n/pull/1149.

## How to test

CI should be sufficient here (we're just removing code).

There's less code in this repository; it's easier to modify the construct in its new location.

I don't think there are any particular risks associated with removing an unused experimental construct.

## Checklist

- [x] I have listed any breaking changes, along with a migration path [^1] - N.B. I don't think this is a breaking change because it's always been considered 'experimental'.
- [x] I have updated the documentation as required for the described changes [^2] - This PR removes documentation only.


akash1810 (commented):
This being a `chore` wouldn't trigger a release. Is this desired?

jacobwinch:
> This being a `chore` wouldn't trigger a release. Is this desired?

I wasn't aiming to skip the release[^1], but I think it's fine for this to be bundled up with the next release-worthy change so will merge as-is.

[^1]: I just wasn't sure which type to use - nothing seemed to fit well - I guess this is technically a `!feat` 😀

akash1810 (approved):

PR #2002, opened 2023-09-05
docs: clarify imageId/imageRecipe props

jacobwinch:
## What does this change?

This PR just adds some docs/clarification, as @JamieB-gu and I were a bit confused when looking at this as part of https://github.com/guardian/dotcom-rendering/pull/8754.


src/constructs/autoscaling/asg.ts
```diff
@@ -31,7 +31,18 @@ export interface GuAutoScalingGroupProps
     >,
     AppIdentity,
     GuAsgCapacity {
+  /**
+   * @deprecated
+   * It shouldn't be necessary to specify a value here.
+   * GuCDK will create an AMI parameter by default. Riff-Raff can use this parameter to inject the latest AMI ID for
+   * your AMIgo recipe.
+   */
   imageId?: GuAmiParameter;
```
jacobwinch, line 40: I was trying to think of a use-case for overriding the default parameter when writing this documentation, but I couldn't think of one. 

I was tempted to delete this prop entirely but I don't really want to make a breaking change in this PR, so have marked it as deprecated for now.

What do you think @akash1810 - am I missing something here?!

akash1810: Yeah, this prop should be removed entirely IMO - happy for it to be a separate change too. Looking at the history, it looks like it was added back when GuCDK was being ultra-flexible; we've since made it more opinionated, which is better. Escape hatches can be used to customise things if really needed, but shouldn't be encoded into the library.

akash1810 (approved):
Thanks!

PR #2024, opened 2023-09-15
feat: Allow access on all indexes of a table in GuDynamoDBPolicy

kenoir:
## What does this change?

This change includes all table indexes in the blanket `GuDynamoDBReadPolicy` and `GuDynamoDBWritePolicy` classes.

Without this [consumers will see failures](https://logs.gutools.co.uk/s/editorial-tools/app/discover#/doc/67604290-baa6-11e9-bea2-633437abb232/logstash-ed-tools-2023.09.14?id=RUXwlYoBCdl423wPKe7R) querying or updating items in Global Secondary Indexes.

Adding read/write access to indexes of a table in these policies is intended to make it easier to use these policies by allowing access to indexes as might be expected.

## How to test

- [X] Run the tests. 
- [x] Include this version of the CDK via this branch and test locally the correct CF is provisioned. 

See: https://github.com/guardian/crosswordv2/pull/38

Consumers are able to use the read/write policies in more expected use cases.

This does expand the permissions of these policies, but not in a surprising way (probably not in issue [as currently consumed](https://github.com/search?q=org%3Aguardian%20GuDynamoDBReadPolicy&type=code)).

### DIFF
--- a/src/constructs/iam/policies/dynamodb.ts
+++ b/src/constructs/iam/policies/dynamodb.ts
@@ -15,7 +15,10 @@ abstract class GuDynamoDBPolicy extends GuAllowPolicy {
     super(scope, id, {
       ...props,
       actions: props.actions.map((action) => `dynamodb:${action}`),
-      resources: [`arn:aws:dynamodb:${scope.region}:${scope.account}:table/${props.tableName}`],
+      resources: [
+        `arn:aws:dynamodb:${scope.region}:${scope.account}:table/${props.tableName}`,
+        `arn:aws:dynamodb:${scope.region}:${scope.account}:table/${props.tableName}/index/*`,
+      ],
     });
   }
 }
--- a/src/constructs/iam/policies/dynamodb.ts
+++ b/src/constructs/iam/policies/dynamodb.ts
@@ -15,6 +15,9 @@ abstract class GuDynamoDBPolicy extends GuAllowPolicy {
     super(scope, id, {
       ...props,
       actions: props.actions.map((action) => `dynamodb:${action}`),
+      // Note: although the index resource is not supported for all attached actions
+      // (e.g. BatchWriteItem), it will not cause issues to include it here as it is ignored.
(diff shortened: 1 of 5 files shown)


akash1810 (approved):

src/constructs/iam/policies/dynamodb.ts
```diff
@@ -15,7 +15,10 @@ abstract class GuDynamoDBPolicy extends GuAllowPolicy {
     super(scope, id, {
       ...props,
       actions: props.actions.map((action) => `dynamodb:${action}`),
-      resources: [`arn:aws:dynamodb:${scope.region}:${scope.account}:table/${props.tableName}`],
+      resources: [
+        `arn:aws:dynamodb:${scope.region}:${scope.account}:table/${props.tableName}`,
+        `arn:aws:dynamodb:${scope.region}:${scope.account}:table/${props.tableName}/index/*`,
```
adamnfish, line 20: Are we sure this will work for the actions people are using in practice? There aren't many actions that support indexes as a resource, just the basics like Scan:
https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazondynamodb.html

If people have added other action props, this might break there stacks if CloudFormation then rejects the index resource for that action?

kenoir: Good question. I've tested that the CloudFormation that gets generated will successfully apply by deploying https://github.com/guardian/crosswordv2/pull/38 to our CODE environment (CF updates successfully and the expected behaviour can be observed with regard to permissions). 

I think this is similar to the case where wildcard resources can be applied to sets of actions. See the `AmazonDynamoDBReadOnlyAccess` managed policy attached. 
Screenshot 2023-09-18 at 10 42 29

The [documentation that I can find](https://docs.aws.amazon.com/service-authorization/latest/reference/reference_policies_actions-resources-contextkeys.html) is not explicit on this point, but does state:

> If you specify a resource type in a statement with an action that does not support that resource type, then the statement doesn't allow access.

So I believe we're safe from disrupting existing stacks here. I will add a comment describing this to the change as it does seem like a potential point of confusion.

PR #2034, opened 2023-09-21
chore: Update Node version

akash1810:
## What does this change?
In #2033 we updated the semantic-release dependency.

The version being used requires a newer version of Node, as seen in [the CD logs](https://github.com/guardian/cdk/actions/runs/6257743340/job/16990556733#step:4:376):

```log
[semantic-release]: node version ^18.17 || >=20.6.1 is required. Found v18.16.0.
```

We didn't catch this at PR time as `semantic-release` is only interacted with upon merge to `main`.

## How to test
Given `semantic-release` is only called during CD, a merge to `main` will be the test.

CD is enabled again.

N/A

## Checklist
N/A


.nvmrc
```diff
@@ -1 +1 @@
-v18.16.0
+v18.18.0
```
akash1810, line 1: This is the [current latest LTS version](https://nodejs.org/en/download).

jacobwinch (approved):

PR #2042, opened 2023-09-29
feat(riff-raff.yaml): Support user-added deployment types

akash1810:
> **Note**
> This code change is negligible (changing `private` to `public`), so mainly looking for feedback on the docs and the resulting DX.

## What does this change?
As an alternative to #1657, this change makes the `riffRaffYaml` property of `RiffRaffYamlFileExperimental` public. This means we can supply additional deployment types, if needed. This is more flexible than #1657.

The typical use-case is adding an `aws-s3` deployment type, which `RiffRaffYamlFileExperimental` does not currently support.

No validation is performed[^1] on the parameters you provide, so you might get deployment errors due to an invalid `riff-raff.yaml` file. For example, if an unrecognised parameter or deployment type is provided, an invalid `riff-raff.yaml` file will be created.

Usage of this feature should be uncommon. For this reason, usage is a bit fiddly:

```ts
// bin/cdk.ts
import { App } from "aws-cdk-lib";
import { RiffRaffYamlFileExperimental } from "@guardian/cdk/lib/experimental/riff-raff-yaml-file";

const app = new App();

const { stack, region } = new MyStack(app, "my-stack", {
  stack: "playground",
  stage: "PROD",
  env: { region: "eu-west-1" },
});

const riffRaff = new RiffRaffYamlFileExperimental(app);
const { riffRaffYaml: { deployments } } = riffRaff;

deployments.set("upload-my-static-files", {
  actions: ["uploadStaticFiles"],
  app: "my-static-site",
  contentDirectory: "my-static-site",
  dependencies: [],
  parameters: {
    bucketSsmKey: `/${stack}/my-static-site-origin`,
    publicReadAcl: false,
    cacheControl: "public, max-age=315360000, immutable",
  },
  regions: new Set([region]),
  stacks: new Set([stack]),
  type: "aws-s3",
});

// Write the riff-raff.yaml file to the output directory.
// Must be explicitly called.
riffRaff.synth();
```

## How to test
- Observe CI - the added test should pass.
- Check the docs - do they make sense?

Ultimately, an increase in the number of projects with an auto-generated `riff-raff.yaml` file. For example https://github.com/guardian/pinboard should be able to adopt this now.

The main risk is the lack of validation, and thus no guarentee that a valid `riff-raff.yaml` file will be produced. For example, one could specify an invalid deployment type, or an invalid parameter name.

## Checklist
- [x] I have listed any breaking changes, along with a migration path [^2]
- [x] I have updated the documentation as required for the described changes [^3]

[^1]: There is some [typing](https://github.com/guardian/cdk/blob/b277dda1c5f2e94489716311b32be5f038d07feb/src/experimental/riff-raff-yaml-file/types.ts#L10-L13), but ultimately we're dealing with the `string` type.

### DIFF
--- a/src/experimental/riff-raff-yaml-file/index.test.ts
+++ b/src/experimental/riff-raff-yaml-file/index.test.ts
@@ -1128,4 +1128,66 @@ describe("The RiffRaffYamlFileExperimental class", () => {
       "
     `);
   });
+
+  it("Should support user-added Riff-Raff deployments", () => {
+    const app = new App({ outdir: "/tmp/cdk.out" });
+    class MyApplicationStack extends GuStack {}
+    const { stack, region } = new MyApplicationStack(app, "App-PROD-deploy", {
+      stack: "deploy",
+      stage: "PROD",
+      env: { region: "eu-west-1" },
+    });
+
+    const riffraff = new RiffRaffYamlFileExperimental(app);
+
+    riffraff.riffRaffYaml.deployments.set("upload-my-static-files", {
+      actions: ["uploadStaticFiles"],
+      app: "my-static-site",
+      contentDirectory: "my-static-site",
+      dependencies: [],
+      parameters: {
+        bucketSsmKey: `/${stack}/my-static-site-origin`,
+        publicReadAcl: false,
+        cacheControl: "public, max-age=315360000, immutable",
+      },
+      regions: new Set([region]),
+      stacks: new Set([stack]),
+      type: "aws-s3",
+    });
+
+    const actual = riffraff.toYAML();
+
+    expect(actual).toMatchInlineSnapshot(`
+      "allowedStages:
+        - PROD
+      deployments:
+        cfn-eu-west-1-deploy-my-application-stack:
+          type: cloud-formation
+          regions:
+            - eu-west-1
+          stacks:
+            - deploy
+          app: my-application-stack
+          contentDirectory: /tmp/cdk.out
+          parameters:
+            templateStagePaths:
+              PROD: App-PROD-deploy.template.json
+        upload-my-static-files:
+          actions:
+            - uploadStaticFiles
+          app: my-static-site
+          contentDirectory: my-static-site
+          dependencies: []
+          parameters:
+            bucketSsmKey: /deploy/my-static-site-origin
+            publicReadAcl: false
+            cacheControl: public, max-age=315360000, immutable
+          regions:
+            - eu-west-1
+          stacks:
+            - deploy
+          type: aws-s3
+      "
+    `);
+  });
 });
--- a/src/experimental/riff-raff-yaml-file/index.test.ts
+++ b/src/experimental/riff-raff-yaml-file/index.test.ts
@@ -1141,10 +1141,8 @@ describe("The RiffRaffYamlFileExperimental class", () => {
     const riffraff = new RiffRaffYamlFileExperimental(app);
 
     riffraff.riffRaffYaml.deployments.set("upload-my-static-files", {
-      actions: ["uploadStaticFiles"],
       app: "my-static-site",
       contentDirectory: "my-static-site",
-      dependencies: [],
       parameters: {
         bucketSsmKey: `/${stack}/my-static-site-origin`,
         publicReadAcl: false,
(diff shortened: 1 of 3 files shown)


src/experimental/riff-raff-yaml-file/index.ts
```diff
@@ -69,10 +69,21 @@ export class RiffRaffYamlFileExperimental {
...
+   *
+   * It is useful for specifying additional deployment types that GuCDK does not support.
+   * Consider raising an issue or pull request if you think it should be supported.
+   *
+   * In most cases, you shouldn't need to access this.
+   * No validation is performed on the parameters you provide, so you might get deployment errors.
+   *
+   * @see https://riffraff.gutools.co.uk/docs/magenta-lib/types
+   */
+  public readonly riffRaffYaml: RiffRaffYaml;
```
akash1810, line 85: This is the only meaningful code change in this PR. The other changes are either documentation, or tests.

src/experimental/riff-raff-yaml-file/index.test.ts
```diff
@@ -1128,4 +1128,66 @@ describe("The RiffRaffYamlFileExperimental class", () => {
...
+          contentDirectory: /tmp/cdk.out
+          parameters:
+            templateStagePaths:
+              PROD: App-PROD-deploy.template.json
+        upload-my-static-files:
+          actions:
+            - uploadStaticFiles
+          app: my-static-site
+          contentDirectory: my-static-site
+          dependencies: []
```
bryophyta, line 1180: is `[]` a valid value in YAML?

If you pass a non-empty array for `dependencies` then you get the expected output, e.g. `dependencies: ['a', 'b']` turns into:

```yaml
dependencies:
  - a
  - b
```

If `[]` isn't valid, I'm not sure if this represents a bug in the riffraff generator, or if passing an empty array is just an unexpected use?

bryophyta:
This looks good to me! It works very well for my use-case, at least.

I added an inline question about the suggested config, because it looks to me like the empty array passed to `dependencies` gets converted to YAML incorrectly.

One other point that I'd be inclined to document is that (afaict) the riffraff steps that are added in this way will always come after the rest of the steps in the auto-generated YAML. For static uploads this seems unlikely to matter too much, but I could imagine it causing potential difficulties for other deployment types?

As you mentioned this is not designed to be used frequently, and I don't know how the team would feel if people _did_ start using it more frequently? If that's an acceptable risk though then I'd personally be very pleased to see this change go through to cdk!

src/experimental/riff-raff-yaml-file/index.test.ts
akash1810: Pasting this YAML into the [Riff-Raff validator tool](https://riffraff.gutools.co.uk/configuration/validation) suggests it is valid.

It's a good point though, and an example of the lack of validation. `dependencies` is an optional property, it's only appearing in the YAML because we've set above when calling `riffraff.riffRaffYaml.deployments.set`.

akash1810:
> One other point that I'd be inclined to document is that (afaict) the riffraff steps that are added in this way will always come after the rest of the steps in the auto-generated YAML. For static uploads this seems unlikely to matter too much, but I could imagine it causing potential difficulties for other deployment types?

The order of deployments in the `riff-raff.yaml` file don't have any impact. IIUC Riff-Raff will perform steps in parallel. To control the order, we can describe dependencies between deployments.

src/experimental/riff-raff-yaml-file/index.test.ts
akash1810: From the [YAML spec](https://yaml.org/spec/1.2.2/#sequence):

> Sequence
> The content of a sequence node is an ordered series of zero or more nodes. In particular, a sequence may contain the same node more than once. It could even contain itself.

That last part is confusing!

> It could even contain itself.

WHAT?!

akash1810: > `dependencies` is an optional property, it's only appearing in the YAML because we've set above when calling `riffraff.riffRaffYaml.deployments.set`. A nice example of user error!

I've simplified the test now - https://github.com/guardian/cdk/pull/2042/commits/fc6e23f8b039ed090493a82c2403789ae9f4897a.

bryophyta:
> The order of deployments in the riff-raff.yaml file don't have any impact. IIUC Riff-Raff will perform steps in parallel. To control the order, we can describe dependencies between deployments.

oh great!

twrichards (approved):
This is awesome, thank you @akash1810.

I'll look to use it for https://github.com/guardian/pinboard soon!

PR #2051, opened 2023-10-09
Add 4xx alarms as an optional parameter to ec2 base alarm config

kenoir:
## What does this change?

Being alerted to high volumes of 404 requests is arguably a common pattern that is useful to extend to CDK. We've added this to another project where it would have allowed us to catch a production issue, so contributing upstream to gauge if useful for others.

See: https://github.com/guardian/crosswordv2/pull/80

This change adds an optional `http4xxAlarm` property to the `Alarms` type on the `GuEc2AppProps` interface, as it is perhaps not as critical to consider as the 5xx events, but still hopefully a useful affordance.

## How to test

- [ ] Run the tests
- [ ] Observe use in project consuming the PR branch

Easier to add alarms around high volumes of 4xx requests, teams are able to notice issues surfaced by this symptom faster.

If the `http4xxAlarm` were not optional this would be a breaking change, the addition of this type as `http4xxAlarm?: false | Http4xxAlarmProps;` removes this risk.

## Checklist

   _No breaking changes_
   _Documentation in code hopefully covers this_


jacobwinch:
I am always a bit cautious about configuring alarms based on a high percentage of 4XXs. I agree that there are some scenarios where this symptom is indicative of a 'real' problem e.g.

* Some page / feature _should_ exist and we're serving lots of 404 accidentally (due to a logical error or a problem with another app) - I think this is the scenario described by https://github.com/guardian/crosswordv2/pull/80?
* Our application has authentication and we've broken it such that no users can login / take appropriate actions (serving lots of 401s or 403s)
* We have an API and we've broken the validation logic such that a high number of 400s are being served

However, strictly speaking a 4XX is a _client_ error, which means that we will also receive alarm notifications in the following scenarios which do not require an action:

* Bots/crawlers generating a spike in 404s that we don't care about
* High number of login attempts with bad credentials
* High number of `POST` requests with an invalid body

Overall, I worry that 4XX alarms will generate a lot of noise for the wrong people, because generally speaking a 4XX is a 'good event' from the perspective of the application / load balancer serving the request.

All of that said, I know that at least one other team is doing [something like this](https://github.com/guardian/mobile-apps-api/blob/4847ce5adb3a852dd8aa4b1c479132071c41d1f3/cdk/lib/microservice.ts#L192-L216) and the code looks good, so if you/other reviewers don't share my reservations about configuring alarms based on these metrics then feel free to merge this in.

kenoir:
> Overall, I worry that 4XX alarms will generate a lot of noise (or noise for the wrong team), because generally speaking a 4XX is a 'good event' from the perspective of the application / load balancer serving the request.

@jacobwinch Agree that the need for 4xx alarms is something to consider on a case by case basis. The implementation here doesn't force you to consider them, but lets you reach for them without having to reimplement the pattern if you want them. 

I'll leave it to your discretion as to whether it's worth adding the code complexity / support implications to `guardian/cdk` represented by this PR.

jacobwinch:
> The implementation here doesn't force you to consider them, but lets you reach for them without having to reimplement the pattern if you want them.
> 
> I'll leave it to your discretion as to whether it's worth adding the code complexity / support implications to `guardian/cdk` represented by this PR.

Yeah, this is a fair point - let's merge this 👍

jacobwinch (approved):

PR #2080, opened 2023-10-26
feat(elb)!: Remove support for classic load balancers

akash1810:
## What does this change?
Use of application load balancers (ALBs) is considered best practice, as ALBs are receiving [more capabilities](https://aws.amazon.com/elasticloadbalancing/features/) than elastic (classic) load balancers (ELBs). GuCDK should be encoding best practice, so remove support for ELBs.

BREAKING CHANGE: Classic load balancer support has been removed. Please adopt application load balancers instead, or if necessary, use ELBs directly from AWS CDK.

## How to test
See tests - they'll continue to pass.

A smaller surface area, i.e. less to maintain.

Two public classes are being removed:
1. `GuClassicLoadBalancer`, with [one usage](https://github.com/search?q=org%3Aguardian%20GuClassicLoadBalancer&type=code). This usage is planned for migration to an ALB - https://github.com/guardian/dotcom-rendering/issues/9310.
2. `GuHttpsClassicLoadBalancer`, with [no usages](https://github.com/search?q=org%3Aguardian+GuHttpsClassicLoadBalancer&type=code).

Therefore, I don't think there is any risk here.

## Checklist


jacobwinch:
> 1. `GuClassicLoadBalancer`, with [one usage](https://github.com/search?q=org%3Aguardian%20GuClassicLoadBalancer&type=code). This usage is planned for migration to an ALB - [Migrate from using ELB to ALB Load Balancer Type for DCR dotcom-rendering#9310](https://github.com/guardian/dotcom-rendering/issues/9310).

Thanks for checking the usages of this! 

Do you think we should wait until the migration is done before merging this? Alternatively I guess we could just help the team to swap to the AWS class and use the snapshot tests to confirm that the underlying infrastructure is unchanged?

jacobwinch (approved):

akash1810:
> Do you think we should wait until the migration is done before merging this? Alternatively I guess we could just help the team to swap to the AWS class and use the snapshot tests to confirm that the underlying infrastructure is unchanged?

Another alternative could be to advise the team to stick on a previous version of GuCDK until they've adopted ALBs. Either work IMO; cc @guardian/dotcom-platform for opinions.

PR #2083, opened 2023-10-30
docs(setup): fix link to node versions

mxdvl:
## What does this change?

Link to the correct URL on the Node.js website: https://nodejs.org/en/about/previous-releases

Reminders that “[Cool URIs don’t change](https://www.w3.org/Provider/Style/URI)”

## How to test

Clickity-click the linkity-link!

Prevent linkrot.

N/A

## Checklist


AshCorr (approved):
> Clickity-click the linkity-link! 

😂 

Nice catch

PR #2093, opened 2023-11-02
Ensure non-empty arrays for CIDR ranges

mxdvl:
## What does this change?

Ensure that Internal and Restricted access provide at least one CIDR range.

This is technically a breaking change, as some consumers may currently provide empty arrays.

## How to test

[`type NonEmptyArray<T> = [T, ...T[]]`](https://www.typescriptlang.org/play?#code/C4TwDgpgBAcg9gOwKIFsygIICcsEMRQC8UA2gM7BYCWCA5gDRQB0LF1dJAupwFA8DGiClFoRgAGQh1gACyJQAFLhz4AXLESp0IbHhABKIgD4oyvUwA2U2rL6iJ12Qq76A3FAD0HqAGU4AVyx+aBlcMigABigIKxQpYAUyQwAjf2AoYGV7KCwIAEd-KlzwgEYmRQAmAGYAFgBWfR57SWkZZwByRAh2zjcmsRabNpJOhG7GduAAdzgevqA)

More type safety.

Some consumers might start getting type errors.

## Checklist

### DIFF
--- a/.changeset/cuddly-carrots-lay.md
+++ b/.changeset/cuddly-carrots-lay.md
@@ -1,0 +1,5 @@
+---
+"@guardian/cdk": major
+---
+
+Users must now provide at least one valid CIDR range when using the restricted and internal access scopes
--- a/src/patterns/ec2-app/base.ts
+++ b/src/patterns/ec2-app/base.ts
@@ -135,9 +135,11 @@ export interface GuEc2AppProps extends AppIdentity {
   /**
    * Network access restrictions for your load balancer.
    *
-   * Note, this merely provides defence in depth; you should, for example, limit
-   * access to the VPN and then treat that as sufficient. Instead, use Google
-   * Auth for human access, or a suitable machine auth mechanism.
+   * **Note:** this merely provides defence in depth.
+   *
+   * You should, for example, use Google Auth for human access,
+   * or a suitable machine auth mechanism, rather treating an
+   * access limited to the VPN as sufficient.
    */
   access: AppAccess;
   /**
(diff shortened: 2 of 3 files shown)


chrislomaxjones:
Nice improvement @mxdvl!

This change introduces a type error to [src/patterns/ec2-app/framework.ts](https://github.com/guardian/cdk/blob/38988d7d0045c31d424505628e64cc230948f88f/src/patterns/ec2-app/framework.ts#L28), do you want to fix that as part of this PR? Perhaps the `GuWorkerPlayApp` will need to take a CIDR range as a parameter if the caller must specify one?

chrislomaxjones (commented):
Nice improvement @mxdvl!

This change introduces a type error to [src/patterns/ec2-app/framework.ts](https://github.com/guardian/cdk/blob/38988d7d0045c31d424505628e64cc230948f88f/src/patterns/ec2-app/framework.ts#L28), do you want to fix that as part of this PR? Perhaps the `GuWorkerPlayApp` will need to take _at least one_ CIDR range as a parameter (assuming it's up to the caller to specify one?).

mxdvl:
> This change introduces a type error to [src/patterns/ec2-app/framework.ts](https://github.com/guardian/cdk/blob/38988d7d0045c31d424505628e64cc230948f88f/src/patterns/ec2-app/framework.ts#L28), do you want to fix that as part of this PR? Perhaps the `GuWorkerPlayApp` will need to take _at least one_ CIDR range as a parameter (assuming it's up to the caller to specify one?).

Yes, is that correct? I was wondering whether the example was correct or not and whether my assumption about CIDR ranges made sense in practice.

AshCorr:
Am I mistaken in thinking this change would force Apps to have atleast one Ingress that uses a IP based filter?

I'm not sure thats behaviour that we would want, apps may want to use more sophisticated methods of limiting access, such as only allowing traffic from specific security groups to access the app.

PR #2094, opened 2023-11-03
feat(riff-raff.yaml): Support cross stack dependencies

akash1810:
## What does this change?
Currently the `riff-raff.yaml` generator is not able to create dependencies between `cloud-formation` deployments. This means each `cloud-formation` deployment could happen at the same time.

This does not work in the scenario where we have:
  - Stack A containing a bucket
  - Stack B CODE containing an app that uses A's bucket
  - Stack B PROD containing an app that uses A's bucket

That is, we can't guarantee Stack A is deployed first.

In this change we add support for the scenario where we have a shared resources stack. The generated `riff-raff.yaml` file will describe that Stack B CODE, and Stack B PROD depend on Stack A.

It uses the AWS CDK mechanism to link CloudFormation stacks:

> Add a dependency between this stack and another stack.
> 
> This can be used to define dependencies between any two stacks within an app, and also supports nested stacks.
> – https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.Stack.html#addwbrdependencytarget-reason

## How to test
- See the added unit test.
- Pass the `riff-raff.yaml` to the [validator](https://riffraff.gutools.co.uk/configuration/validation), and observe the desired dependency graph:
   image

Supports another deployment scenario (though admittedly a rather niche one).

N/A

## Checklist

### DIFF
--- a/.changeset/chatty-rabbits-cry.md
+++ b/.changeset/chatty-rabbits-cry.md
@@ -1,0 +1,19 @@
+---
+"@guardian/cdk": minor
+---
+
+feat(riff-raff.yaml): Support cross stack dependencies
+
+Currently the `riff-raff.yaml` generator is not able to create dependencies between `cloud-formation` deployments. This means each `cloud-formation` deployment could happen at the same time.
+
+This does not work in the scenario where we have:
+- Stack A containing a bucket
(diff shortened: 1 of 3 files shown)


twrichards (approved):
🤩

PR #2098, opened 2023-11-07
feat(backup): Support backups provided by DevX

akash1810:
## What does this change?
Adds a new property `withBackup` to `GuStack` to enable [backups provided by DevX](https://github.com/guardian/aws-backup). It is defaulted to `false`.

When `true`, all supported resources in the stack will receive a new tag `devx-backup-enabled`. To opt in/out an individual resource, you can manually apply this tag.

This change uses [Aspects](https://docs.aws.amazon.com/cdk/v2/guide/aspects.html).

## How to test
Observe the added tests.

With this flag, one doesn't need to know the name of the tag. That is, the DX should be a little simpler.

N/A.

## Checklist

### DIFF
--- a/src/aspects/aws-backup.ts
+++ b/src/aspects/aws-backup.ts
@@ -1,0 +1,25 @@
+import type { IAspect } from "aws-cdk-lib";
+import { CfnResource, TagManager } from "aws-cdk-lib";
+import type { IConstruct } from "constructs";
+
+/**
+ * Applies the tags that enable backups for supported resources.
+ *
+ * @see https://github.com/guardian/aws-backup
+ */
+export class AwsBackupTag implements IAspect {
+  /**
+   * These resources are backed up by https://github.com/guardian/aws-backup.
+   */
+  static resourceTypes: string[] = ["AWS::RDS::DBInstance", "AWS::DynamoDB::Table"];
+
+  public visit(node: IConstruct): void {
+    if (node instanceof CfnResource) {
+      const { cfnResourceType } = node;
+
+      if (AwsBackupTag.resourceTypes.includes(cfnResourceType) && TagManager.isTaggable(node)) {
+        node.tags.setTag("devx-backup-enabled", "true");
+      }
+    }
+  }
+}
--- a/src/constructs/core/stack.ts
+++ b/src/constructs/core/stack.ts
@@ -3,6 +3,7 @@ import { Annotations, App, Aspects, CfnParameter, LegacyStackSynthesizer, Stack,
 import { Template } from "aws-cdk-lib/assertions";
 import type { IConstruct } from "constructs";
 import gitUrlParse from "git-url-parse";
+import { AwsBackupTag } from "../../aspects/aws-backup";
 import { CfnIncludeReporter } from "../../aspects/cfn-include-reporter";
 import { CfnParameterReporter } from "../../aspects/cfn-parameter-reporter";
 import { Metadata } from "../../aspects/metadata";
@@ -42,6 +43,15 @@ export interface GuStackProps extends Omit<StackProps, "stackName"> {
    * please do not override this.
    */
   withoutMetadata?: boolean;
+
+  /**
+   * Set to enable all resources in the stack for backup provided by https://github.com/guardian/aws-backup.
+   *
+   * @default false - backups are not enabled
+   *
+   * @see https://github.com/guardian/aws-backup
+   */
+  withBackup?: boolean;
 }
 
 /**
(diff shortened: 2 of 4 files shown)


akash1810:
cc @guardian/devx-reliability for your thoughts

jacobwinch:
Thanks for working on this - it's really cool to see `Aspects` in action!

For a first pass, I wonder if we should use them to enforce that _a_ `devx-backup-enabled` tag is added (i.e. ensure that the user makes a choice about opting-in/out of backups, rather than defaulting to opt-in). It would be interesting to hear what you / others in @guardian/devx-reliability think about that though, as it's not a very strong / well-formed opinion!

src/aspects/aws-backup.ts
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Applies the tags that enable backups for supported resources.
+ *
+ * @see https://github.com/guardian/aws-backup
+ */
+export class AwsBackupTag implements IAspect {
+  /**
+   * These resources are backed up by https://github.com/guardian/aws-backup.
+   */
+  static resources: string[] = [CfnDBInstance.CFN_RESOURCE_TYPE_NAME, CfnTable.CFN_RESOURCE_TYPE_NAME];
```
akash1810, line 16: This is just `[ "AWS::RDS::DBInstance", "AWS::DynamoDB::Table" ]`. There's an argument that using the raw strings is more readable. Thoughts welcomed.

akash1810:
> ensure that the user makes a choice about opting-in/out of backups, rather than defaulting to opt-in

I think defaulting to on fits with the idea of GuCDK being an opinionated library. I'm not sure how mature `aws-backup` is. If we're still developing it, having it be opt-in is sensible.

src/aspects/aws-backup.ts
jacobwinch: Yeah, I find the strings a bit easier to follow.

In terms of `aws-backup` support:

* `AWS::DynamoDB::Table` has been well tested and we're [hoping to roll it out across P&E this quarter](https://docs.google.com/document/d/11EZtuVHCnjavE9AYLroiuDUsMzbp2ulFBtj4VkyCKNc/edit#heading=h.vwt7syo8ng40) (perhaps this will help inform whether defaulting to `true` would be a good choice or not?)
* `AWS::RDS::DBInstance` is being tested with Grafana, but we still have some questions for AWS before deciding on the exact approach / attempting a wider rollout
* `AWS::RDS::DBCluster` could also be supported in theory, but this is completely untested at the moment as we only have a [handful of clusters](https://metrics.gutools.co.uk/goto/ZesH_wVSk?orgId=1).

akash1810: Moved to strings (https://github.com/guardian/cdk/pull/2098/commits/32445bfe17b666f230b7291aece3718f7639e75e). It really was too fancy before 😁.

src/constructs/core/stack.ts
```diff
@@ -127,7 +137,13 @@ export class GuStack extends Stack implements StackStageIdentity {
 
   // eslint-disable-next-line custom-rules/valid-constructors -- GuStack is the exception as it must take an App
   constructor(scope: App, id: string, props: GuStackProps) {
-    const { cloudFormationStackName = process.env.GU_CFN_STACK_NAME, stack, stage, withoutTags } = props;
+    const {
+      cloudFormationStackName = process.env.GU_CFN_STACK_NAME,
+      stack,
+      stage,
+      withoutTags,
+      withoutBackup = false,
```
jorgeazevedo, line 145: While I appreciate that `withoutBackup` is consistent with the existing `withoutTags` property, I found the double negative `withoutBackup = false` hard to parse 😅

akash1810: Totally fair!

Can think of two alternatives:
1. Change the type of `withoutBackup` to `withoutBackup?: true` - when undefined, backups are enabled, when `true` backups are disabled
2. Change the property name to `withBackup`

Though it breaks consistency with the other properties, I think the second option is better. WDYT?

jorgeazevedo:
I wonder if logging is a useful reference here? Is devx-logs/logs.gutools enabled by default in GuCDK? I'm reminded of how reader revenue don't use our logging infrastructure due to their logs containing personal data. If they were to use GuCDK and it's on by default, they would have to disable logging. Would the same argument apply to backups, it's on by default and if your data has special requirements you need to manual disable the feature?

src/constructs/core/stack.ts
jorgeazevedo: Yeah, I agree, I think using `withBackup` would make it easier to read!

akash1810:
> I wonder if logging is a useful reference here?

Ah, this is a brilliant comparison! Integration with DevX logs is [currently opt-in](https://github.com/guardian/cdk/blob/891309e88f9c5ff8303a0215b1ddc21ef274b644/src/patterns/ec2-app/base.ts#L330-L331) for a couple of reasons - other custom logging solutions are still in use, and the overall DX isn't as smooth as it could be - you have to enable a special role in AMIgo.

I think the question of "should this be enabled by default or not yet" is answered. It shouldn't be.

akash1810:
All changes made. This is now ready for a formal review!

jorgeazevedo (approved):
Great work, thanks for adding this Akash! We're currently auditing all the department's dynamodb tables and assessing if they need backups or not. I think once we finish that we'll have another datapoint to help inform if we should evolve this approach and move it to opt-out for example.

PR #2123, opened 2023-11-28
feat: add `enableDistPolicy` prop to `GuEcsTask`

marsavar:
## What does this change?
Adds a new prop to `GuEcsTask` to avoid giving the task IAM role access so the dist bucket in S3.

## Why?
Looking at the CDK template in our project, I noticed that the IAM role that executes our scheduled tasks has access to the dist bucket.

This is undesirable, because our container images contain the application binary and all its dependencies and doesn't need to pull from S3 at all.

This pattern is currently used in Lurch, where the container downloads artifacts from the account's S3 bucket. Disabling this policy by default would break backward compatibility, so the policy needs to be explicitly disabled via the prop.
I think this is a good compromise - keen to hear your thoughts.

## Checklist


src/constructs/ecs/ecs-task.ts
```diff
@@ -114,6 +114,13 @@ export interface GuEcsTaskProps extends AppIdentity {
   customTaskPolicies?: PolicyStatement[];
   environmentOverrides?: TaskEnvironmentVariable[];
   storage?: number;
+  /**
+   * Whether to give the task IAM role access to the account's dist bucket.
+   * This is enabled by default to avoid breaking changes, but consumers of this pattern
+   * that pull a container image that doesn't need extra dependencies from S3
+   * shoud set this value to `false`.
+   */
+  enableDistPolicy?: boolean;
```
philmcmahon, line 123: sorry could you make this enableDistributable policy - realise you've just copied my lazy variable name but might be a bit clearer

PR #2170, opened 2024-01-08
Peer dependencies are never direct

mxdvl:
## What does this change?

Make the following direct dependencies that are already listed as peers, dev: `aws-cdk-lib`, `constructs`. This matches the setup of `aws-cdk`.

## How to test

Install the package somewhere.

Declaring peer dependencies that are also direct ones has no value

Consumers should ensure that they have the correct peer dependencies

## Checklist


akash1810:
> Declaring peer dependencies that are also direct ones has no value

This package uses [`peerDependencies`](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#peerdependencies) as a way to dictate which version of AWS CDK _must_ be used by consumers of GuCDK.

If a different version of `aws-cdk`, `aws-cdk-lib`, or `constructs` is used, some fairly cryptic stack traces are seen[^1], similar to:

```console
    lib/my-app.test.ts:9:29 - error TS2345: Argument of type 'Stack' is not assignable to parameter of type 'Stack'.
      Type 'Stack' is missing the following properties from type 'Stack': _suppressTemplateIndentation, _terminationProtection, toYamlString

    9   expect(Template.fromStack(stack).toJSON()).toMatchSnapshot();
```

Use of `peerDependencies` produces a warning at installation time when mismatched versions of used. The idea is that message provides a way to triage the above issue.

If there's a better way describe this in the `package.json` that'd be amazing! I think the changes in this PR break the above description.

[^1]: The are a few Chat threads sharing confusion on this stack trace.

mxdvl:
> > Declaring peer dependencies that are also direct ones has no value
> 
> This package uses [`peerDependencies`](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#peerdependencies) as a way to dictate which version of AWS CDK _must_ be used by consumers of GuCDK.

Yes, that’s a correct definition of what peer dependencies are meant to do. However, I’m not sure what specifying a dependency as **both** peer and direct achieves. It’s not a defined behaviour in the docs. Peer dependencies are automatically installed by NPM since v7, so there’s no need to specify it as a direct one to install it.

akash1810:
> If a different version of `aws-cdk`, `aws-cdk-lib`, or `constructs` is used, some fairly cryptic stack traces are seen[1](#user-content-fn-1-7ec1244e9fd9fa0005b3a9636c813537), similar to:

https://github.com/guardian/service-catalogue/pull/680/files for a real-life example (see the added annotations).

akash1810 (approved):

akash1810:
Just to note this change follows the advice in https://docs.aws.amazon.com/cdk/v2/guide/manage-dependencies.html:

> If you're developing a construct library, specify its dependencies via a combination of the `peerDependencies` and `devDependencies` sections

PR #2171, opened 2024-01-08
Make `@changeset/cli` development only

### LINKED ISSUE #2148: Stop making `@changesets/cli` a direct dependency
I believe [this should be a `devDependency`](https://github.com/guardian/cdk/blob/15cd297c72a806836087395f1e621f8f06d5520c/package.json#L56), consumers do not need to receive it!

mxdvl:
## What does this change?

Make dependency `@changeset/cli` development only

## How to test

`npx changeset` locally still works

A much smaller package
Fixes #2148 

None!

## Checklist

### DIFF
--- a/.changeset/fuzzy-poets-approve.md
+++ b/.changeset/fuzzy-poets-approve.md
@@ -1,0 +1,5 @@
+---
+"@guardian/cdk": patch
+---
+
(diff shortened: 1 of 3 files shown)


chrislomaxjones (approved):
Thanks @mxdvl, nice spot!

PR #2198, opened 2024-01-29
feat: Support more than one EC2 app with load balancer access logs enabled in a given stack

marsavar:
## What does this change?
Use a singleton to reference the string parameter that contains the name of the S3 bucket used for load balancer access logs.

## Why?
Previously, a new parameter was created per EC2 instance. This meant that only one EC2 in a given stack could have access logs enabled, as attempting to create another one would result in a synth failure due to a parameter with that name already existing.
This change makes it so a new parameter is only created if it doesn't exist in the stack. Subsequent invocations will reference the existing parameter instead of creating a new one.

## How to test
I've added a new test case where 2 EC2 apps exist in the same stack, both with access logs enabled.

## Checklist

### DIFF
--- a/src/constructs/core/parameters/s3.ts
+++ b/src/constructs/core/parameters/s3.ts
@@ -38,3 +38,27 @@ export class GuPrivateConfigBucketParameter extends GuStringParameter {
     });
   }
 }
+
+/**
+ * Creates a CloudFormation parameter which references the bucket used to store load balancer access logs.
+ * By default, the bucket name is stored in an SSM Parameter called `/account/services/access-logging/bucket`.
+ */
+export class GuAccessLoggingBucketParameter extends GuStringParameter {
+  private static instance: GuAccessLoggingBucketParameter | undefined;
+
+  private constructor(scope: GuStack) {
+    super(scope, "AccessLoggingBucket", {
+      description: NAMED_SSM_PARAMETER_PATHS.AccessLoggingBucket.description,
+      default: NAMED_SSM_PARAMETER_PATHS.AccessLoggingBucket.path,
+      fromSSM: true,
+    });
+  }
+
+  public static getInstance(stack: GuStack): GuAccessLoggingBucketParameter {
+    if (!this.instance || !isSingletonPresentInStack(stack, this.instance)) {
+      this.instance = new GuAccessLoggingBucketParameter(stack);
+    }
+
+    return this.instance;
+  }
+}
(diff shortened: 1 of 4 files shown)


src/patterns/ec2-app/base.test.ts
```diff
@@ -934,84 +934,142 @@ describe("the GuEC2App pattern", function () {
...
-      minimumInstances: 1,
-    },
-  });
-  Template.fromStack(stack).hasResourceProperties("AWS::ElasticLoadBalancingV2::TargetGroup", {
-    HealthCheckIntervalSeconds: 10,
-    HealthCheckPath: "/healthcheck",
-    HealthCheckProtocol: "HTTP",
-    HealthCheckTimeoutSeconds: 5,
-    HealthyThresholdCount: 5,
+  it("should provides a default healthcheck", function () {
```
marsavar, line 938: The diff here is due to the fact that this tests and the ones below were incorrectly put outside the `describe` function. I've moved them back into it where they belong.

src/patterns/ec2-app/base.test.ts
```diff
@@ -934,84 +934,142 @@ describe("the GuEC2App pattern", function () {
...
+      },
+    });
   });
-  Template.fromStack(stack).hasResourceProperties("AWS::EC2::LaunchTemplate", {
-    LaunchTemplateData: {
-      MetadataOptions: {
-        HttpPutResponseHopLimit: 2,
-        HttpTokens: "required",
+
+  it("supports more than one EC2 app with load balancer access logs enabled", () => {
```
marsavar, line 1018: This is the test I've added as part of this PR

src/constructs/core/parameters/s3.ts
```diff
@@ -38,3 +38,27 @@ export class GuPrivateConfigBucketParameter extends GuStringParameter {
...
+    super(scope, "AccessLoggingBucket", {
+      description: NAMED_SSM_PARAMETER_PATHS.AccessLoggingBucket.description,
+      default: NAMED_SSM_PARAMETER_PATHS.AccessLoggingBucket.path,
+      fromSSM: true,
+    });
+  }
+
+  public static getInstance(stack: GuStack): GuAccessLoggingBucketParameter {
+    if (!this.instance || !isSingletonPresentInStack(stack, this.instance)) {
+      this.instance = new GuAccessLoggingBucketParameter(stack);
```
mxdvl, line 59: [Nullish coalescing assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment) may better express the intent, here:

```suggestion
    if (!isSingletonPresentInStack(stack, this.instance)) {
      this.instance ??= new GuAccessLoggingBucketParameter(stack);
```

akash1810 (approved):
Thanks for this!

PR #2211, opened 2024-02-08
fix(ec2)!: Tighten default EC2 access to SSM Parameters

akash1810:
## What does this change?
The [SSM+SSH capability](https://github.com/guardian/ssm-scala?tab=readme-ov-file#in-aws) requires an IAM Policy.

This change swaps the [AmazonSSMManagedInstanceCore](https://docs.aws.amazon.com/aws-managed-policy/latest/reference/AmazonSSMManagedInstanceCore.html) AWS Managed Policy with one of our own to more closely follow the principle of least privilege.

In addition to enabling SSM+SSH, AmazonSSMManagedInstanceCore adds permissions to read _all_ SSM Parameters.

<details><summary>AmazonSSMManagedInstanceCore policy statement</summary>
<p>

```json
{
  "Version" : "2012-10-17",
  "Statement" : [
    {
      "Effect" : "Allow",
      "Action" : [
        "ssm:DescribeAssociation",
        "ssm:GetDeployablePatchSnapshotForInstance",
        "ssm:GetDocument",
        "ssm:DescribeDocument",
        "ssm:GetManifest",
        "ssm:GetParameter",
        "ssm:GetParameters",
        "ssm:ListAssociations",
        "ssm:ListInstanceAssociations",
        "ssm:PutInventory",
        "ssm:PutComplianceItems",
        "ssm:PutConfigurePackageResult",
        "ssm:UpdateAssociationStatus",
        "ssm:UpdateInstanceAssociationStatus",
        "ssm:UpdateInstanceInformation"
      ],
      "Resource" : "*"
    },
    {
      "Effect" : "Allow",
      "Action" : [
        "ssmmessages:CreateControlChannel",
        "ssmmessages:CreateDataChannel",
        "ssmmessages:OpenControlChannel",
        "ssmmessages:OpenDataChannel"
      ],
      "Resource" : "*"
    },
    {
      "Effect" : "Allow",
      "Action" : [
        "ec2messages:AcknowledgeMessage",
        "ec2messages:DeleteMessage",
        "ec2messages:FailMessage",
        "ec2messages:GetEndpoint",
        "ec2messages:GetMessages",
        "ec2messages:SendReply"
      ],
      "Resource" : "*"
    }
  ]
}
```

</p>
</details> 

We use SSM Parameters for configuration. With AmazonSSMManagedInstanceCore, every EC2 instance can read the configuration of all apps. This is not least privilege, and counters the effects of the [GuParameterStoreReadPolicy](https://github.com/guardian/cdk/blob/20a4315fb8c7123139abcf735b034b5a5d9924f9/src/constructs/iam/policies/parameter-store-read.ts).

I opted to create a policy, rather than using the [NotResource trick](https://repost.aws/questions/QUhdNOg85CQXqqgYdgoMc24g/limit-scope-of-aws-managed-iam-policies) with the managed policy as it'll be easier to reason about[^1].

https://github.com/guardian/cdk/pull/1694 comments that using AmazonSSMManagedInstanceCore enabled usage of AWS Inspector v2. However, as we haven't enabled AWS Inspector across the estate yet[^2], we can add the permissions in a later PR (cc/@guardian/devx-security).

## How to test
This effectively reverts behaviour to before https://github.com/guardian/cdk/pull/1694. Testing via unit tests should suffice.

Following least privilege more closely. Note however that services will need to update the GuCDK version used.

N/A.

## Checklist
- [ ] I have listed any breaking changes, along with a migration path [^3]
- [ ] I have updated the documentation as required for the described changes [^4]

[^1]: https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_evaluation-logic.html#policy-eval-denyallow shows decision tree of if a request is allowed or denied.
[^2]: It's been (manually) enabled on 3 accounts (out of ~40).
[^3]: Consider whether this is something that will mean changes to projects that have already been migrated, or to the CDK CLI tool. If changes are required, consider adding a checklist here and/or linking to related PRs.
[^4]: If you are adding a new construct or pattern, has new documentation been added? If you are amending defaults or changing behaviour, are the existing docs still valid?

### DIFF
--- a/.changeset/moody-needles-end.md
+++ b/.changeset/moody-needles-end.md
@@ -1,0 +1,82 @@
+---
+"@guardian/cdk": major
+---
+
+This change includes some potentially breaking changes for consumers of:
+- [`GuEc2App`](https://guardian.github.io/cdk/classes/patterns.GuEc2App.html)
+- [`GuPlayApp`](https://guardian.github.io/cdk/classes/patterns.GuPlayApp.html) (a subclass of `GuEc2App`)
+- [`GuPlayWorkerApp`](https://guardian.github.io/cdk/classes/patterns.GuPlayWorkerApp.html) (a subclass of `GuEc2App`)
+- [`GuNodeApp`](https://guardian.github.io/cdk/classes/patterns.GuNodeApp.html) (a subclass of `GuEc2App`)
+
+Since [v49.0.2](https://github.com/guardian/cdk/releases/tag/v49.0.2),
+the EC2 instance profile created in `GuEc2App`, and it's subclasses,
+used the [`AmazonSSMManagedInstanceCore`](https://docs.aws.amazon.com/aws-managed-policy/latest/reference/AmazonSSMManagedInstanceCore.html) AWS Managed Policy
+to enable the [SSM+SSH capability](https://github.com/guardian/ssm-scala?tab=readme-ov-file#in-aws).
+
+In addition to enabling SSM+SSH, this Managed Policy also provided read access to all SSM Parameters.
+This is not least privilege.
+
+In this version, usage of the `AmazonSSMManagedInstanceCore` Managed Policy is swapped for a custom,
+more minimal, policy.
+
+> [!IMPORTANT]
+> Before upgrading to this version,
+> ensure your application is not relying on the IAM Policy behaviour provided by `AmazonSSMManagedInstanceCore`.
+
+If your application is reading SSM Parameters outside the `/STAGE/STACK/APP/*` namespace,
+you will need to add an explicit policy.
+
+An IAM Policy to read SSM Parameters in the `/STAGE/STACK/APP/*` namespace is already provided by the `GuEc2App` construct,
+via [`GuParameterStoreReadPolicy`](https://guardian.github.io/cdk/classes/constructs_iam.GuParameterStoreReadPolicy.html)
+
+To understand if your application is impacted,
+consult [this Service Catalogue query](https://metrics.gutools.co.uk/goto/KZhWJVoIg?orgId=1)
+showing CloudFormation stacks using the above _and_ using GuCDK v49.0.2 or above.
+
+<details><summary>Query ran in Service Catalogue</summary>
+<p>
+
+```sql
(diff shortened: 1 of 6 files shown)


jacobwinch:
The code changes look good to me.

> fix(ec2): Tighten default EC2 access to SSM Parameters

I think we should treat this as a breaking change (and perhaps send comms?) to increase the likelihood of users noticing that we are dropping permissions here.

I am concerned that anyone who started using GuCDK after https://github.com/guardian/cdk/pull/1694 was merged might be (accidentally) relying on the broad permissions that were included via the managed policy. For example, consider the following scenario:

1. An EC2 app relies on reading a parameter called `/my/custom/parameter/path` in order to support some important feature. When this app was defined in standard CloudFormation there was an IAM policy which granted the app permission to read this parameter, allowing this feature to work.
2. The app is migrated to a GuCDK pattern. When the team tests in `CODE` everything 'just works' so they deploy to `PROD` without carefully auditing all of the permissions that their app needs. Unbeknownst to the team, the app 'just works' because the managed policy allows their app to continue reading `/my/custom/parameter/path` as well as parameters under the normal `<stage>/<stack>/<app>` path.
3. The team upgrades GuCDK (perhaps automatically, via a dependabot PR) and the managed policy is removed. The app can no longer read `/my/custom/parameter/path` and the important feature breaks in `PROD`.

akash1810:
> I think we should treat this as a breaking change (and perhaps send comms?) to increase the likelihood of users noticing that we are dropping permissions here.

Agreed, this change will impact a [large number](https://metrics.gutools.co.uk/goto/9PCkpX2SR?orgId=1) of CFN stacks.

akash1810:
> I think we should treat this as a breaking change (and perhaps send comms?) to increase the likelihood of users noticing that we are dropping permissions here.

[Release notes updated](https://github.com/guardian/cdk/blob/aa/ssm-ssh-policy/.changeset/moody-needles-end.md). Do you think it's clear enough to be re-used for comms?

AshCorr (approved):
I think the release notes would do fine as comms!

PR #2218, opened 2024-02-15
Switch from using `INFRA` stage of cognito-lambda to using `PROD` and allow users to optionally use `CODE`

AshCorr:
## What does this change?

- Switch hardcode `INFRA` stage in Cognito lambda config to use `PROD`
- Allow users to optionally use the `CODE` Cognito lambda environment.

## How to test

Still figuring that out.

We've started merging dependabot PRs again for cognito-auth lambda, the problem with switching from INFRA to PROD is that users still on INFRA will start to slowly become more and more out of date. 

## Checklist

### DIFF
--- a/src/patterns/ec2-app/base.ts
+++ b/src/patterns/ec2-app/base.ts
@@ -276,6 +276,15 @@ export interface GuEc2AppProps extends AppIdentity {
      * @defaultValue /:STAGE/:stack/:app/google-auth-credentials
      */
     credentialsSecretsManagerPath?: string;
+
+    /**
+     * When using Auth in the ALB, which stage of cognito-lambda to use.
+     *
+     * For most applications this should always be PROD, even in the CODE environments.
+     *
+     * @defaultValue PROD
+     */
+    cognitoAuthStage?: "PROD" | "CODE";
   };
 
   /**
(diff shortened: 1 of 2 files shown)


src/patterns/ec2-app/base.ts
```diff
@@ -276,6 +276,15 @@ export interface GuEc2AppProps extends AppIdentity {
...
     credentialsSecretsManagerPath?: string;
+
+    /**
+     * When using Auth in the ALB, which stage of cognito-lambda to use.
+     *
+     * For most applications this should always be PROD, even in the CODE environments.
+     *
+     * @defaultValue PROD
+     */
+    cognitoAuthStage?: "PROD" | "CODE";
```
akash1810, line 287: minor: In https://github.com/guardian/cdk/blob/main/docs/architecture-decision-records/006-stage-awareness.md the argument was made to not restrict `stage`, keeping it as a `string`. Does this union type contradict this? Granted the scope of this property is not the same as `stage` for an application. Might possible alternatives be to use a `string` type, or make this a `boolean`, e.g. `usePreProductionCognitoAuth`?

akash1810 (commented):
Brilliant! A minor comment added.

akash1810 (approved):

AshCorr:
Tested in cdk-playground, seems to work.

PR #2223, opened 2024-02-22
feat: Harden load balancers

marsavar:
## What does this change?

Modifies the headers of the requests that the load balancer forwards to the target in the following way:

1) Two headers are added, specifying the TLS version and cipher suite used
2) Invalid headers are dropped

Number 1 makes it easier to spot TLS [downgrade attacks](https://en.wikipedia.org/wiki/Downgrade_attack) when analysing ELB access logs, and number 2 helps mitigate HTTP [smuggling attacks](https://www.virtuesecurity.com/kb/http-request-smuggling-aws/).

These changes improve the overall security posture and we've already applied them to the load balancers in our account; this PR brings the changes to GuCDK so they can be adopted as a default by the department.

## How to test
I've added tests for both features.

## Checklist

- [x] ~I have listed any breaking changes, along with a migration path~ [^1] (N/A)


.changeset/forty-spies-heal.md
```diff
@@ -0,0 +1,6 @@
+---
+"@guardian/cdk": minor
+---
+
+- Load balancers now add headers with information about the TLS version and cipher suite used during negotiation
+- Load balancers now drop invalid headers before forwarding requests to the target
```
AshCorr, line 6: > Load balancers now drop invalid headers before forwarding requests to the target

This sounds like it might be a breaking change if anybody relies on custom headers! Maybe this should be a major version bump?

marsavar: Custom headers (`X-etc-etc`) are considered valid, so they won't be dropped. 

From the AWS docs about this setting:

> Elastic Load Balancing requires that valid HTTP header names conform to the regular expression [-A-Za-z0-9]+, as described in the HTTP Field Name Registry. Each name consists of alphanumeric characters or hyphens. Select true if you want HTTP headers that do not conform to this pattern, to be removed from requests.

Do you still think this should be a major version bump?

AshCorr: Ah gotcha. Highly doubtful that anybody would be using such a header, but I'd still suggest this be a major bump just in case!

AshCorr (approved):

.changeset/forty-spies-heal.md
AshCorr: And could we possibly add the description of what is a valid header to the release notes, just so nobody makes the same mistake i did!

PR #2228, opened 2024-02-23
feat: Various ECS task improvements

marsavar:
## What does this change?

Adds a few improvements to the ECS task pattern to address some Security Hub warnings.

- Add `readonlyRootFilesystem` prop to specify whether the container is given read-only access to its root file system

- Add `containerInsights` prop to enable CloudWatch insights

- Replace deprecated state machine definition. This removes the warning:
```
      [WARNING] aws-cdk-lib.aws_stepfunctions.StateMachineProps#definition is deprecated.
        use definitionBody: DefinitionBody.fromChainable()
        This API will be removed in the next major release.
```

The GuCDK defaults for the new props mirror the CDK defaults to preserve backward compatibility.  

## Checklist

### DIFF
--- a/.changeset/brown-starfishes-grin.md
+++ b/.changeset/brown-starfishes-grin.md
@@ -1,0 +1,9 @@
+---
+"@guardian/cdk": minor
+---
+
+- Add `readonlyRootFilesystem` prop to specify whether the container is given read-only access to its root file system
+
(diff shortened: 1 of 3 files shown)


philmcmahon (approved):
Approving this from the perspective of the other main user of this role!

PR #2246, opened 2024-03-08
fix(GuStack): Set app value 

akash1810:
## What does this change?
Fixes a bug where `this.app` on a `GuStack` is always undefined, as it is never set. See https://github.com/guardian/cdk/pull/1497#issuecomment-1480997050.

Also included, is a refactor to reduce complexity - private properties and public getters are replaced with public properties.


AshCorr (approved):
Noticed this in riffraff-platform too! Wasnt sure if it was intentional!

PR #2247, opened 2024-03-08
fix(ec2-app): Apply the `App` tag to the launch template

akash1810:
## What does this change?
Apply the `App` tag to the launch template created in the EC2 App pattern.

## How to test
See snapshot updates.

More tags!

n/a

## Checklist

### DIFF
--- a/src/constructs/autoscaling/asg.ts
+++ b/src/constructs/autoscaling/asg.ts
@@ -1,4 +1,4 @@
-import { Token } from "aws-cdk-lib";
+import { Tags, Token } from "aws-cdk-lib";
 import { AutoScalingGroup } from "aws-cdk-lib/aws-autoscaling";
 import type { AutoScalingGroupProps, CfnAutoScalingGroup } from "aws-cdk-lib/aws-autoscaling";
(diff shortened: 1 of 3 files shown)


src/constructs/autoscaling/asg.ts
```diff
@@ -162,5 +162,7 @@ export class GuAutoScalingGroup extends GuAppAwareConstruct(AutoScalingGroup) {
     // leaves it to the default value, which is actually false.
     // { UpdatePolicy: { autoScalingScheduledAction: { IgnoreUnmodifiedGroupSizeProperties: true }}
     cfnAsg.addDeletionOverride("UpdatePolicy");
+
+    Tags.of(launchTemplate).add("App", app);
```
AshCorr, line 166: Do we want to make the `app` prop required? I think its a `string | undefined` at the moment, could lead to the App tag being `undefined`

AshCorr (approved):

src/constructs/autoscaling/asg.ts
akash1810: Ha! You've found what I'm doing this 10% day 😄!

It's required on the EC2 App pattern, so it will be `string` here. It's `string | undefined` at the `GuStack` level.

PR #2260, opened 2024-03-21
feat: GuLambdaFunction should use JSON logging by default

NovemberTang:
## What does this change?

I think JSON log formatting is a more sensible default value.

There are two log formats in AWS Lambda, `Text`, and `JSON`. `Text` is the default (possibly for historical reasons?), but has disadvantages, such as not being as easily machine-readable.

Using `Text` means that any changes made to the ApplicationLogLevel will be flat out ignored (defaulting to `DEBUG` as far as I can tell). Application log filtering **only** works when the log format is `JSON`.

I've verified that Lambdas using this configuration are still compatible with Central ELK. [Repocop](https://github.com/guardian/service-catalogue/pull/769) has been using JSON logging for several weeks without issue.

## How to test

- [x] Verify default behaviour of lambdas has changed
- [x] Verify that providing the `Text` value results in the expected snapshot

Teams will be able to filter their application logs by severity level without having to do as much googling/asking around as I had to

This will cause snapshot changes to everyone using a GuLambdaFunction or its descendants. This will be slightly disruptive, but no work is required beyond updating the snapshot tests.

To maintain the previous behaviour, users should add the following to their lambda props

`logFormat: "Text"`

## Checklist

### DIFF
--- a/src/constructs/lambda/lambda.ts
+++ b/src/constructs/lambda/lambda.ts
@@ -12,7 +12,7 @@ import type { GuStack } from "../core";
 import { AppIdentity, GuDistributionBucketParameter } from "../core";
 import { ReadParametersByName, ReadParametersByPath } from "../iam";
 
-export interface GuFunctionProps extends GuDistributable, Omit<FunctionProps, "code">, AppIdentity {
+export interface GuFunctionProps extends GuDistributable, Omit<FunctionProps, "code" | "logFormat">, AppIdentity {
   /**
    * Create a new Lambda version and alias. This is only necessary if you want to use features which rely
    * on versioning (e.g. SnapStart or Provisioned Concurrency).
@@ -60,6 +60,12 @@ export interface GuFunctionProps extends GuDistributable, Omit<FunctionProps, "c
    * an uploadLambda step.
    */
   withoutArtifactUpload?: boolean;
+
+  /**
+   * The format of the Lambda function's logs.
+   * Can be either "JSON" or "Text".
+   */
+  logFormat?: "JSON" | "Text";
 }
 
 function defaultMemorySize(runtime: Runtime, memorySize?: number): number {
@@ -124,6 +130,7 @@ export class GuLambdaFunction extends Function {
       bucketNamePath,
       withoutFilePrefix = false,
       withoutArtifactUpload = false,
+      logFormat,
     } = props;
 
     const bucketName = bucketNamePath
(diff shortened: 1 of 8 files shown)


src/constructs/lambda/lambda.ts
```diff
@@ -12,7 +12,7 @@ import type { GuStack } from "../core";
 import { AppIdentity, GuDistributionBucketParameter } from "../core";
 import { ReadParametersByName, ReadParametersByPath } from "../iam";
 
-export interface GuFunctionProps extends GuDistributable, Omit<FunctionProps, "code">, AppIdentity {
+export interface GuFunctionProps extends GuDistributable, Omit<FunctionProps, "code" | "logFormat">, AppIdentity {
```
akash1810, line 15: Do we mean to remove the `logFormat` property?

akash1810 (commented):
Just to double check - does https://github.com/guardian/cloudwatch-logs-management apply the log markers to JSON format logs?

src/constructs/lambda/lambda.ts
NovemberTang: I did mean to remove it, but I think you're right, we don't actually need to do this.

NovemberTang:
> Just to double check - does https://github.com/guardian/cloudwatch-logs-management apply the log markers to JSON format logs?

It seems like it to me. Here's a link to the [repocop](https://logs.gutools.co.uk/s/devx/app/discover#/?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-15d,to:now))&_a=(columns:!(),filters:!(),index:'626e7830-dd3a-11ea-88b6-37713dba5739',interval:auto,query:(language:kuery,query:'app.keyword%20:%20%22repocop%22%20%20and%20stage.keyword%20:%22PROD%22'),sort:!(!('@timestamp',desc)))) logs, which have been using [JSON formatting for months](https://github.com/guardian/service-catalogue/blame/0cada89a549eee7a7529e40263876f4dbd764341/packages/cdk/lib/repocop.ts#L72).

src/constructs/lambda/lambda.ts
```diff
@@ -141,6 +142,7 @@ export class GuLambdaFunction extends Function {
     const code = Code.fromBucket(bucket, objectKey);
     super(scope, id, {
       ...props,
+      logFormat: !logFormat ? "JSON" : logFormat,
```
akash1810, line 145: [Destructuring with assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) would make this more readable (vs a negated ternary). It also looks like AWS CDK have an [enum](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda.LoggingFormat.html) available:

```ts
import { LoggingFormat } from "aws-cdk-lib/aws-lambda";

const {
  logFormat = LoggingFormat.JSON
} = props;
```

akash1810: These are non-blocking comments.

PR #2276, opened 2024-04-15
feat!: Add DevX Backups support to RDS instance construct

jacobwinch:
## What does this change?

This PR adds support for DevX Backups to our RDS instance construct. Users are now forced to opt-in/out of backups when provisioning an RDS instance via this construct. This makes it less likely that we will forget to add protection to important backups. Users who choose to provision an RDS instance via an AWS construct (instead of the GuCDK one) can still currently forget to make this choice (we will improve this in a follow up PR by modifying [this aspect](https://github.com/guardian/cdk/blob/main/src/aspects/aws-backup.ts) to check for the presence of the correct tag).

Additionally, we help to avoid conflicts by preventing users who opt-in to this feature from setting native RDS backup properties that cannot be managed via CloudFormation once AWS Backup is enabled. Users who opt-out of DevX Backups are still able to set these properties as there is no conflict.

The deletion of automated backups (on instance deletion) is now prevented by default (i.e. we override the [CFN default](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbinstance.html#cfn-rds-dbinstance-deleteautomatedbackups)). This is necessary where DevX Backups are enabled (the lock on these backups will prevent the instance from being deleted otherwise). In cases where users have opted-out of DevX Backups I still consider retaining the automated backups to be a sensible precaution so I have not made this change conditional.

Finally, I've removed some old unit tests that seemed to be testing functionality that is provided via the AWS libraries rather than our own construct: https://github.com/guardian/cdk/pull/2276/commits/c5f348eafbffa03d39ccedc1170eb0fe1f42b0fa.

## How to test

I've added unit tests to cover the new functionality.

Users provisioning new RDS instances via GuCDK will be more likely to remember to opt-in to DevX Backups.

This construct is not widely used and opting-in is an easy change, so I don't think this breaking change poses much of a risk in terms of generating work for teams.

## Checklist

### DIFF
--- a/src/constructs/rds/instance.test.ts
+++ b/src/constructs/rds/instance.test.ts
@@ -1,8 +1,8 @@
 import { Stack } from "aws-cdk-lib";
-import { Match, Template } from "aws-cdk-lib/assertions";
+import { Template } from "aws-cdk-lib/assertions";
 import { Vpc } from "aws-cdk-lib/aws-ec2";
-import { DatabaseInstanceEngine, ParameterGroup, PostgresEngineVersion } from "aws-cdk-lib/aws-rds";
-import { GuTemplate, simpleGuStackForTesting } from "../../utils/test";
+import { DatabaseInstanceEngine, PostgresEngineVersion } from "aws-cdk-lib/aws-rds";
+import { simpleGuStackForTesting } from "../../utils/test";
 import { GuDatabaseInstance } from "./instance";
 
 describe("The GuDatabaseInstance class", () => {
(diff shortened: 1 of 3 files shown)


src/constructs/rds/instance.ts
```diff
@@ -1,12 +1,34 @@
...
+   * We recommend using DevX Backups where possible. If it is not suitable for your use-case please document
+   * this here so that we can understand why this is switched off when performing security audits.
+   */
+  optOutReason: string;
+}
+
+export interface GuDatabaseInstanceProps
+  extends Omit<DatabaseInstanceProps, "instanceType" | "backupRetention" | "preferredBackupWindow">,
+    AppIdentity {
   instanceType: string;
```
jacobwinch, line 25: It seems a shame that we are removing the type from `instanceType` and making it a `string` instead - I'd like to change this but it'd be a breaking change and is not linked to the PR, so have left it as is for now.

src/constructs/rds/instance.test.ts
```diff
@@ -26,86 +26,135 @@ describe("The GuDatabaseInstance class", () => {
...
     });
-
-    const template = Template.fromStack(stack);
-
-    template.hasResourceProperties("AWS::RDS::DBInstance", {
-      DBParameterGroupName: {
-        Ref: Match.stringLikeRegexp("DatabaseInstanceTestingParameterGroup[A-Z0-9]+"),
-      },
+    console.log(JSON.stringify(stack.tags.tagValues()));
+    GuTemplate.fromStack(stack).hasGuTaggedResource("AWS::RDS::DBInstance", {
```
adamnfish, line 83: We should add functionality to check for a single tag, but no worries here! This would encourage good test practice (one assertion per test), and eliminate noise from cases like this one where we are focused on a single tag.

adamnfish (approved):
Looks great, thank you!

src/constructs/rds/instance.ts
michaelwmcnamara: Agree on both counts.

michaelwmcnamara (approved):
Excellent work!

src/constructs/rds/instance.test.ts
jacobwinch: @adamnfish - there's a suggestion for this here: https://github.com/guardian/cdk/pull/2280.

PR #2277, opened 2024-04-17
add changeset that was missed in #2260

NovemberTang:
## What does this change?

Adds a changeset, a step that was missed in #2260


jacobwinch:
I haven't used changesets before but hopefully merging this will lead to a PR similar to https://github.com/guardian/cdk/pull/2279 🤞

jacobwinch (approved):

PR #2280, opened 2024-04-17
test: add simplified tag assertion

jacobwinch:
This PR implements a suggestion from https://github.com/guardian/cdk/pull/2276#discussion_r1568537805.

I don't think AWS provides a way to do this, so I've added a helper function to combine [resource property matching](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.assertions-readme.html#resource-matching--retrieval) and [array matching](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.assertions-readme.html#array-matchers) in order to get this working.  

I think this makes the tests added in #2276 easier to follow and hopefully it will make it easier for others to check tag correctness in the future too.

### DIFF
--- a/src/constructs/rds/instance.test.ts
+++ b/src/constructs/rds/instance.test.ts
@@ -80,14 +80,9 @@ describe("The GuDatabaseInstance class", () => {
       devXBackups: { enabled: true },
     });
     console.log(JSON.stringify(stack.tags.tagValues()));
-    GuTemplate.fromStack(stack).hasGuTaggedResource("AWS::RDS::DBInstance", {
-      appIdentity: { app: "testing" },
-      additionalTags: [
-        {
-          Key: "devx-backup-enabled",
-          Value: "true",
-        },
-      ],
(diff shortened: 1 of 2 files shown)


src/constructs/rds/instance.test.ts
```diff
@@ -80,14 +80,9 @@ describe("The GuDatabaseInstance class", () => {
...
-    GuTemplate.fromStack(stack).hasGuTaggedResource("AWS::RDS::DBInstance", {
-      appIdentity: { app: "testing" },
-      additionalTags: [
-        {
-          Key: "devx-backup-enabled",
-          Value: "true",
-        },
-      ],
+    GuTemplate.fromStack(stack).hasResourceWithTag("AWS::RDS::DBInstance", {
+      Key: "devx-backup-enabled",
```
jacobwinch, line 84: I've manually checked that setting the wrong tag value (or mistyping the key) here causes the test to fail.

akash1810 (approved):

PR #2288, opened 2024-04-29
Add Dynamodb construct with default deletion protection and mandatory opt-in/opt-out setting for DevX-backup

michaelwmcnamara:
## What does this change?
This PR adds a construct for dynamoDB tables
This construct will enable deletion protection by default.
It also adds a mandatory opt-in/opt-out field for aws backups.
This last forces a decision on whether or not to use the AWS backup vault to protect backup data from accidental or malicious deletion.
This decision should almost always be yes, however there are circumstances in which a team might need to use a different form of storage for their backup. The important thing is that the decision is made explicitly.
If users opt-out, they will need to provide a reason for their decision. This is to allow this information to be captured in security audits.

This is a new construct, so will not be a breaking change.

## How to test
Unit tests added to cover main scenarios.

New stacks created in CDK have deletion protection enabled and a clear decision on whether to opt-in or out of AWS Backups.
It is harder to delete DynamoDB tables by accident.
We do not have a build up of tables whose backup status is unknown and un-tracked.
Our data infrastructure is more secure and more resilient

Currently this construct is not in use.
Only risks are:
 - The construct is not adopted and teams continue to use the native AWS dynamoDB construct
 - As an organisation we suddenly want to switch to a multi-region architecture. This dynamoDB construct uses the original Table construct which does not have as good support for multi-region features - as opposed to the new [TableV2 construct](https://aws.amazon.com/blogs/devops/a-new-and-improved-aws-cdk-construct-for-amazon-dynamodb-tables/).
 This risk is minimal for now as a github search revealed we are using the original Table construct across all our current repositories and the TableV2 is not in use on any of our repos.
This PR has been optimised to maximise adoption by existing teams.

That said a new DynamoDB construct that uses the TableV2 classes would be a useful addition should any teams ever need to adopt a multi-regioin AWS infrastructure.

## Checklist

- No breaking changes as this is a new construct

### DIFF
--- a/.changeset/grumpy-socks-divide.md
+++ b/.changeset/grumpy-socks-divide.md
@@ -1,5 +1,5 @@
 ---
-"@guardian/cdk": patch
+"@guardian/cdk": minor
 ---
 
 Add Dynamodb construct with default deletion protection and mandatory opt-in/opt-out setting for DevX-backup.
--- a/.changeset/grumpy-socks-divide.md
+++ b/.changeset/grumpy-socks-divide.md
@@ -1,0 +1,5 @@
+---
+"@guardian/cdk": patch
+---
+
+Add Dynamodb construct with default deletion protection and mandatory opt-in/opt-out setting for DevX-backup.
--- a/src/constructs/dynamodb/dynamodb.ts
+++ b/src/constructs/dynamodb/dynamodb.ts
@@ -25,13 +25,12 @@ export interface GuDynamoTableProps extends TableProps {
   devXBackups: OptIn | OptOut;
 }
 
-  /**
-   * DeletionProtection is enabled by default for this Table.
-   * We recommend enabling this enabled for all active DynamoDB tables.
-   * The default can be overridden in the GuDynamoTable instantiation if needed eg: for table deletion.
-   */
-  export class GuDynamoTable extends Table {
-
+/**
+ * DeletionProtection is enabled by default for this Table.
+ * We recommend enabling this enabled for all active DynamoDB tables.
+ * The default can be overridden in the GuDynamoTable instantiation if needed eg: for table deletion.
+ */
+export class GuDynamoTable extends Table {
   constructor(scope: GuStack, id: string, props: GuDynamoTableProps) {
     super(scope, id, { deletionProtection: true, ...props });
     Tags.of(this).add("devx-backup-enabled", String(props.devXBackups.enabled));
--- a/src/constructs/dynamodb/dynamodb.ts
+++ b/src/constructs/dynamodb/dynamodb.ts
@@ -1,0 +1,41 @@
+import { Tags } from "aws-cdk-lib";
+import { Table } from "aws-cdk-lib/aws-dynamodb";
+import type { TableProps } from "aws-cdk-lib/aws-dynamodb";
+import type { GuStack } from "../core";
+
+export interface OptIn {
+  enabled: true;
+}
+
+export interface OptOut {
+  enabled: false;
+  /**
+   * We recommend using DevX Backups where possible. If it is not suitable for your use-case please document
+   * this here so that we can understand why this is switched off when performing security audits.
+   */
+  optOutReason: string;
+}
+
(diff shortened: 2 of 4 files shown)


src/constructs/dynamodb/dynamodb.ts
```diff
@@ -0,0 +1,41 @@
...
+  /**
+   * We recommend using DevX Backups where possible. If it is not suitable for your use-case please document
+   * this here so that we can understand why this is switched off when performing security audits.
+   */
+  optOutReason: string;
+}
+
+export interface GuDynamoTableProps extends TableProps {
+  /**
+   * We recommend using DevX Backups to protect your RDS instance's backups.
```
jacobwinch, line 21: These docs need to be updated to mention DynamoDB / use the [DynamoDB link](https://docs.google.com/document/d/11EZtuVHCnjavE9AYLroiuDUsMzbp2ulFBtj4VkyCKNc/edit#heading=h.vwt7syo8ng40).

src/constructs/dynamodb/dynamodb.ts
```diff
@@ -0,0 +1,41 @@
...
+}
+
+/*
+ * D
+ */
+export class GuDynamoTable extends Table {
+  /**
+   * DeletionProtection is enabled by default for this Table.
+   * We recommend enabling this enabled for all active DynamoDB tables.
+   * The default can be overridden in the GuDynamoTable instantiation if needed eg: for table deletion.:qq
```
jacobwinch, line 35:
```suggestion
   * The default can be overridden in the GuDynamoTable instantiation if needed eg: for table deletion.
```

src/constructs/dynamodb/dynamodb.ts
michaelwmcnamara: Good catch! Fixed and force-pushed.

src/constructs/dynamodb/dynamodb.ts
michaelwmcnamara: This change has been incorporated

src/constructs/dynamodb/dynamodb.ts
```diff
@@ -0,0 +1,41 @@
...
+   * [documentation](https://docs.google.com/document/d/11EZtuVHCnjavE9AYLroiuDUsMzbp2ulFBtj4VkyCKNc/edit?usp=sharing).
+   */
+  devXBackups: OptIn | OptOut;
+}
+
+/*
+ * D
+ */
+export class GuDynamoTable extends Table {
+  /**
```
jacobwinch, line 32: I think this documentation/comment would typically live above the class rather than above the constructor. There is also a comment fragment above so perhaps this was moved accidentally?

michaelwmcnamara: Not sure where that partial comment was from - deleted it anyway and moved documentation/comment above class declaration.

.changeset/grumpy-socks-divide.md
```diff
@@ -0,0 +1,5 @@
+---
+"@guardian/cdk": patch
```
jacobwinch, line 2:
```suggestion
"@guardian/cdk": minor
```

Looking at the [release history](https://github.com/guardian/cdk/releases), I think we'd normally use `minor` for feature releases (like this) and `patch` for dependency upgrades etc.

jacobwinch (approved):
I think you need to fix formatting to get the CI build to pass (you can do this with `npm run format` ). I've also left one comment on the release type (I think this should be a `minor` release).

Once those small issues have been addressed I think this is good to go 👍

michaelwmcnamara:
> I think you need to reformat to get the CI build to pass (you can do this with `npm run format` ). I've also left one comment on the release type (I think this should be a `minor` release).
> 
> Once those small issues have been addressed I think this is good to go 👍

Thanks for your help with this @jacobwinch!

.changeset/grumpy-socks-divide.md
michaelwmcnamara: updated this to minor release.

PR #2332, opened 2024-06-11
feat(lambda): Explicitly configure the log group 

akash1810:
When creating a `AWS::Lambda::Function` without an explicit [`LoggingConfig.LogGroup`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-function.html#cfn-lambda-function-loggingconfig) property, AWS [implicitly creates a log group](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-lambda-function-loggingconfig.html#cfn-lambda-function-loggingconfig-loggroup). This implicitly created log group is named `/aws/lambda/<function name>`, and does not have any tags applied to it.

Log groups created with CloudFormation can now be tagged. See also https://github.com/guardian/cloudwatch-logs-management/issues/19.

## What does this change?
In this change, an explicit [`AWS::Logs::LogGroup`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-logs-loggroup.html) is created, and the `GuLambdaFunction` is configured to log to it. As we're now creating the log group, we can also apply tags to it 🎉.

> [!CAUTION]
> Two breaking changes are introduced:
> 1. A `GuLambdaFunction` now defaults the `functionName` property to `<stage>-<stack>-<app>`. Previously, it was `undefined` allowing AWS to [auto-generate a name](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-function.html#cfn-lambda-function-functionname). This default can still be overridden.
> 2. Consequently, the `stack`, `stage`, and `app` triple for lambdas must be unique within a region.

This log group's name mimics the naming scheme of AWS: `/aws/lambda/<function name>`. Omitting the name results in AWS auto-generating one. However tools such as [cloudwatch-logs-management](https://github.com/guardian/cloudwatch-logs-management) are looking for groups prefixed `/aws/lambda`, so explicitly naming allows these tools to continue to work.

Looking at [the data](https://metrics.gutools.co.uk/goto/neHpkVUIR?orgId=1), this naming change would impact a large number of lambdas. However, it looks like most of them constitute part of a step function, and haven't been explicitly tagged in their CloudFormation template, for example [elasticsearch-node-rotation](https://github.com/guardian/elasticsearch-node-rotation/blob/f7a1516c89a62627f688e4a9b7ebdce33daae46b/cloudformation.yaml#L200-L217).
## Alternative solutions
Here are some alternative solutions to lessen the breaking changes. I think the last one (generate a random log name) is the most feasible.
### Remove log groups from the tagging obligation
The tagging obligation requires that resources are tagged correctly. That is, an option is to decide log groups do not require tagging.
### Mutate the implicit log group
 https://github.com/guardian/cloudwatch-logs-management already does some mutation, including setting the retention time, and creating a subscription. Could it be updated to also apply tags?
### Omit log group name
Instead of providing an explicit name, we could opt to add an additional tag to the log group. This will result in AWS auto-generating a name, and the log group will not have the `/aws/lambda` prefix.

To allow `cloudwatch-logs-management` to continue to work, it would need updating to look at the tags of log groups.
### Only explicitly name the log group
The log group name needs to start `/aws/lambda` for `cloudwatch-logs-management` to work. Instead of setting the name of the lambda and the log group, we could set the log group's name only: `/aws/lambda/<stage>-<stack>-<app>`.

Theoretically, this has fewer breaking changes. However, it does not remove the need for each lambda in a stack to have a unique `stack`, `stage`, and `app` triple. Else we'd be attempting to create multiple log groups with the same name, which will not deploy.
### Generate a random log name
The breaking changes above are a consequence of needing the `/aws/lambda` prefix.

Theoretically, we could `Ref` the lambda function to make the log group name dynamic. However this creates a cyclical dependency: the lambda requires the log group to be created, and the log group requires the lambda to be created. Further more, this error is first seen at deployment time, which can be confusing.

To meet the above requirement, the log group could be named `/aws/lambda/<guid>`. The GUID cannot be truly random, as that would make snapshot testing tricky. Instead, the randomness should be seeded; [AWS CDK's `Names` class](https://github.com/aws/aws-cdk/blob/665396fa8485ab642c27acf30df85f2b023acde4/packages/aws-cdk-lib/core/lib/names.ts#L69-L91) can be used here.
## How to test

TBD.

More resources defined using infrastructure as code, and more resources with tags.

Yes, there are breaking changes. See above.
## Checklist

### DIFF
--- a/src/constructs/lambda/lambda.ts
+++ b/src/constructs/lambda/lambda.ts
@@ -3,6 +3,7 @@ import { Duration } from "aws-cdk-lib";
 import type { PolicyStatement } from "aws-cdk-lib/aws-iam";
 import type { FunctionProps, Runtime } from "aws-cdk-lib/aws-lambda";
 import { Alias, Code, Function, LoggingFormat, RuntimeFamily } from "aws-cdk-lib/aws-lambda";
+import { LogGroup } from "aws-cdk-lib/aws-logs";
 import { Bucket } from "aws-cdk-lib/aws-s3";
 import { StringParameter } from "aws-cdk-lib/aws-ssm";
 import { GuDistributable } from "../../types";
@@ -115,6 +116,7 @@ export class GuLambdaFunction extends Function {
   public readonly alias?: Alias;
 
   constructor(scope: GuStack, id: string, props: GuFunctionProps) {
+    const { stack, stage } = scope;
     const {
       app,
       fileName,
@@ -125,9 +127,16 @@ export class GuLambdaFunction extends Function {
       withoutFilePrefix = false,
       withoutArtifactUpload = false,
       logFormat = LoggingFormat.JSON,
-    } = props;
 
-    const { stack, stage } = scope;
+      /*
+      This default has a consequence that, within a single CFN stack,
+      each lambda function must have a unique Stage/Stack/App triple.
+      Else we'll create the same lambda twice, which is not possible.
+      This error will also be first seen at runtime, not at CDK synth time,
+      i.e. it has a long feedback loop.
+       */
+      functionName = [stage, stack, app].join("-"),
+    } = props;
 
     const defaultEnvironmentVariables = {
       STACK: stack,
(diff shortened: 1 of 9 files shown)


src/constructs/lambda/lambda.ts
```diff
@@ -125,23 +127,55 @@ export class GuLambdaFunction extends Function {
...
-      ? StringParameter.fromStringParameterName(scope, "bucketoverride", bucketNamePath).stringValue
-      : GuDistributionBucketParameter.getInstance(scope).valueAsString;
+      /*
+      This default has a consequence that, within a single CFN stack,
+      each lambda function must have a unique Stage/Stack/App triple.
+      Else we'll create the same lambda twice, which is not possible.
+      This error will also be first seen at runtime, not at CDK synth time,
+      i.e. it has a long feedback loop.
+       */
+      functionName = [stage, stack, app].join("-"),
```
adamnfish, line 138: What's the reason for needing to do this? I haven't been able to discern what you mean by `and avoid cyclical dependencies`.

As you say, the uniqueness of these isn't a given and I think this is a bit surprising. It also might create problems in the future if we need to do a replacing update for this resource?

akash1810: The log group's name is created using the lambda function's name. An auto-generated lambda name is known at deploy time. Without hard-coding the lambda's name we'll have a cyclical dependency where the lambda requires the log group to exist, and the log group requires the lambda to exist.

The "Generate a random log name" alternative in the PR description would here here...

PR #2343, opened 2024-06-18
Limit length of cognito user pool prefix to 63

philmcmahon:
## What does this change?
I recently ran into an issue when trying to add googleAuth to a project with a rather long app name:

`1 validation error detected: Value 'com-gu-myvery-looong-appname-prod-6ead8aa1447b718dd224f9eab2808b25' at 'domain' failed to satisfy constraint: Member must have length less than or equal to 63 (Service: AWSCognitoIdentityProviderService; Status Code: 400; Error Code: InvalidParameterException; Request ID: f58fe6de-1cf3-44a2-8498-a3afaca5bd3b; Proxy: null)`

This change attempts to resolve the issue by trimming the generated domain prefix to 63 chars. I'm making an assumption that a 5 char hash will be enough to ensure uniqueness (the current md5 hash is 32 chars). 

The important thing here is that we don't want to change the domainPrefix of any existing cognito user pools, which is why I'm only trimming the domainPrefix when the length is greater than 63.

## How to test
I've added a few unit tests - see what you think. I tested against the offending project and was able to succesfully set up google auth:
 - failed deploy: https://riffraff.gutools.co.uk/deployment/view/1bca3a3a-c70c-42b7-be1e-3f75479bde10
 - success: https://riffraff.gutools.co.uk/deployment/view/5accbef5-9532-4a9a-b4f3-3c68bf155434


AshCorr (approved):
Looks good! Don't forget a changeset!

PR #2379, opened 2024-07-22
feat(ec2): Setting UpdatePolicy on our EC2 patterns now configures th…

AshCorr:
## What does this change?

Configures setting up signalling when an UpdatePolicy is passed to our Guardian patterns or Autoscaling group.

This configures the EC2 instance to wait until the ALB Target Group reports the instance as Healthy before telling Cloudformation to delete the old ASG.

It also configured our riff-raff.yaml generator to not configure an autoscaling step for ASGs that use updatePolicies as we don't want RiffRaff to handle rotating the instances for them anymore.

## Checklist

### DIFF
--- a/src/constructs/autoscaling/asg.ts
+++ b/src/constructs/autoscaling/asg.ts
@@ -141,7 +141,7 @@ export class GuAutoScalingGroup extends GuAppAwareConstruct(AutoScalingGroup) {
       maxCapacity: maximumInstances ?? minimumInstances * 2,
       minCapacity: minimumInstances,
       groupMetrics: groupMetrics,
-      signals: updatePolicy ? Signals.waitForAll() : undefined,
+      signals: updatePolicy ? signals ?? Signals.waitForAll() : undefined,
 
       // Omit userData, instanceType, blockDevices & role from asgProps
       // As this are specified by the LaunchTemplate and must not be duplicated
--- a/src/constructs/autoscaling/asg.ts
+++ b/src/constructs/autoscaling/asg.ts
@@ -175,8 +175,8 @@ export class GuAutoScalingGroup extends GuAppAwareConstruct(AutoScalingGroup) {
   }
 
   signalOnHealthyTargetGroup(targetGroup: ApplicationTargetGroup): void {
-    const port = (targetGroup as unknown as CfnTargetGroup).port;
-    const logicalId = (this as unknown as CfnAutoScalingGroup).logicalId;
+    const port = (targetGroup.node.defaultChild as unknown as CfnTargetGroup).port;
+    const logicalId = (this.node.defaultChild as unknown as CfnAutoScalingGroup).logicalId;
 
     // Poll the ALB and wait for the instance to appear as healthy
     this.userData.addCommands(`
--- a/src/constructs/autoscaling/asg.ts
+++ b/src/constructs/autoscaling/asg.ts
@@ -141,6 +141,7 @@ export class GuAutoScalingGroup extends GuAppAwareConstruct(AutoScalingGroup) {
       maxCapacity: maximumInstances ?? minimumInstances * 2,
       minCapacity: minimumInstances,
       groupMetrics: groupMetrics,
+      signals: updatePolicy ? Signals.waitForAll() : undefined,
 
       // Omit userData, instanceType, blockDevices & role from asgProps
       // As this are specified by the LaunchTemplate and must not be duplicated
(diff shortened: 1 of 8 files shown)


src/constructs/autoscaling/asg.ts
```diff
@@ -164,8 +166,56 @@ export class GuAutoScalingGroup extends GuAppAwareConstruct(AutoScalingGroup) {
     // { UpdatePolicy: { autoScalingScheduledAction: { IgnoreUnmodifiedGroupSizeProperties: true }}
     if (!updatePolicy) {
       cfnAsg.addDeletionOverride("UpdatePolicy");
+      Tags.of(this).add("gu:rotated-by", "riff-raff");
+    } else {
+      Tags.of(this).add("gu:rotated-by", "cloudformation");
```
AshCorr, line 171: For monitoring migration progress in Service Catalogue

src/constructs/autoscaling/asg.ts
```diff
@@ -140,6 +141,7 @@ export class GuAutoScalingGroup extends GuAppAwareConstruct(AutoScalingGroup) {
       maxCapacity: maximumInstances ?? minimumInstances * 2,
       minCapacity: minimumInstances,
       groupMetrics: groupMetrics,
+      signals: updatePolicy ? Signals.waitForAll() : undefined,
```
akash1810, line 144: IIUC the default signal is 100% success, within 5 minutes. Might we want to allow customisation of this? Or set it relative to the healthcheck?

src/riff-raff-yaml-file/index.ts
```diff
@@ -200,6 +201,15 @@ export class RiffRaffYamlFile {
     return cdkStack.dependencies.filter((_) => _ instanceof GuStack) as GuStack[];
   }
 
+  private isAutoscalingDeployStack(asg: GuAutoScalingGroup): boolean {
```
akash1810, line 204: Worth adding a doc string?

src/constructs/autoscaling/asg.ts
```diff
@@ -164,8 +166,56 @@ export class GuAutoScalingGroup extends GuAppAwareConstruct(AutoScalingGroup) {
...
       cfnAsg.addDeletionOverride("UpdatePolicy");
+      Tags.of(this).add("gu:rotated-by", "riff-raff");
+    } else {
+      Tags.of(this).add("gu:rotated-by", "cloudformation");
     }
 
     Tags.of(launchTemplate).add("App", app);
   }
+
+  signalOnHealthyTargetGroup(targetGroup: ApplicationTargetGroup): void {
```
akash1810, line 177: Worth a doc string? I imagine we'd want to indicate this function isn't intended for use by consumers.

akash1810 (commented):
I wonder if we want to make this an [experimental feature](https://github.com/guardian/cdk/blob/main/docs/architecture-decision-records/005-x01-package-structure.md) initially whilst we gain more understanding on scaling behaviour, handle edge cases, etc?

src/constructs/autoscaling/asg.ts
AshCorr: Ah I think I made a mistake here, it should be:

```suggestion
      signals: updatePolicy ? Signals.waitForAll() : signals,
```

So that you can pass a `signals` prop to the ASG to override the default. Having it be based on the healthcheck sounds reasonable, but perhaps we can leave it as the default till we understand it a bit more?

src/constructs/autoscaling/asg.ts
AshCorr: I think this would be fine for consumers to use if they're not using our EC2 patterns. Doc string is needed though!

AshCorr:
> I wonder if we want to make this an [experimental feature](https://github.com/guardian/cdk/blob/main/docs/architecture-decision-records/005-x01-package-structure.md) initially whilst we gain more understanding on scaling behaviour, handle edge cases, etc?

Hmm, that would be ideal but I'm not sure how we can do it without making significant refactors to our EC2 pattern to let it accept an experimental ASG?

akash1810:
> > I wonder if we want to make this an [experimental feature](https://github.com/guardian/cdk/blob/main/docs/architecture-decision-records/005-x01-package-structure.md) initially whilst we gain more understanding on scaling behaviour, handle edge cases, etc?
> 
> Hmm, that would be ideal but I'm not sure how we can do it without making significant refactors to our EC2 pattern to let it accept an experimental ASG?

I've made an attempt in https://github.com/guardian/cdk/pull/2395. WDYT?

AshCorr:
Closed in favour of #2395

PR #2395, opened 2024-07-29
feat(experimental-ec2-pattern): Pattern to deploy ASGs updates via CloudFormation (`AutoScalingReplacingUpdate`)

akash1810:
> [!NOTE]
> This is an alternative take of https://github.com/guardian/cdk/pull/2379, differing mostly in the creation of an [experimental](https://github.com/guardian/cdk/blob/main/docs/architecture-decision-records/005-x01-package-structure.md) pattern. The intention is to provide a strong signal that this feature is not yet considered ready for use on high-profile services.
> 
> Another point of difference is usage tracking via the [`Metadata` Aspect](https://github.com/guardian/cdk/blob/main/src/aspects/metadata.ts)  rather than tags. The Aspect updates the stack's `Metadata` section with a list of GuCDK constructs being used.

## What does this change?
This change adds a new [experimental](https://github.com/guardian/cdk/blob/main/docs/architecture-decision-records/005-x01-package-structure.md) pattern `GuEc2AppExperimental` for provisioning an EC2 based service deployed entirely via CloudFormation updates. This is achieved by setting the `UpdatePolicy` attribute of the ASG, specifically [`AutoScalingReplaceUpdate`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatepolicy.html#cfn-attributes-updatepolicy-replacingupdate) is employed.

With an `AutoScalingReplaceUpdate` policy, a CloudFormation update will create a second ASG, and:

> After successfully creating the new Auto Scaling group, CloudFormation deletes the old Auto Scaling group during the cleanup process.
> – https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatepolicy.html#cfn-attributes-updatepolicy-replacingupdate

The CloudFormation update remains `IN_PROGRESS` until the instances in the ASG report healthy. By default this is defined as instance health (did the instance boot correctly), whereas we want an instance to be healthy only once the target group can see it. For this reason, we implement some custom logic in the user-data.

The resulting user-data works like this:
1. Perform the user-data commands provided by the service (i.e. download the artifact, and run it).
2. Start polling the target group for the health of the current instance.
3. Once the instance is healthy, send a [signal](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-signal.html) to indicate success.

On success, the CloudFormation update status moves to `UPDATE_COMPLETE`, and the deployment succeeds. The final state is one ASG, running the new version of the service.

If the target group is unable to report the instance as healthy, then a signal is sent indicating this. The CloudFormation update status moves to `UPDATE_FAILED`, and the changes are rolled back. The final state is one ASG, running the current version of the service.

Updating a service via this mechanism differs from Riff-Raff's `autoscaling` deployment in one key way: the current ASG can continue to scale if needed. In Riff-Raff's [process](https://github.com/guardian/riff-raff/blob/424bff5c2c3e8a37f8546c4f4cee3e31eb3a25b1/magenta-lib/src/main/scala/magenta/deployment_type/AutoScaling.scala#L169-L217), scaling alarms are disabled before doubling the desired capacity of the ASG, and enabled again once the capacity has been halved (technically after the halving _request_ has been made).

The `GuEc2AppExperimental` pattern is not compatible with Riff-Raff's `autoscaling` deployment type too. As a pre-flight check, Riff-Raff checks there is exactly one ASG matching the tagging specification. As we're creating a second, this check will fail. Consequently, the `riff-raff.yaml` generator has been updated to omit the `autoscaling` deployment.

### Why `experimental`?
There are a few requirements for this approach:
- The AWS CLI, and `cfn-signal` binaries need to be available and on the `PATH` (this is added via the [AMIgo role `aws-tools`](https://github.com/guardian/amigo/tree/main/roles/aws-tools))
- The service's artifact should include a build number, as this is the most reliable way to create a difference with the running CloudFormation template, and thus trigger an update.

We're not (yet) validating these are met as its tricky.

There are also a few of unknowns about this approach:
- What is the best signal timeout value?
- If the current ASG scales mid-deployment, how does this impact the new ASG? Is the desired mirrored?

For this reason, the rollout plan looks something like this:
1. Dogfood on some DevX services.
2. Test on a service within the department.
3. Move the pattern to stable, as a breaking change, with communication on the migration path, etc.

## How to test
For a real-world test, I've been using the pattern (and the update to the generated `riff-raff.yaml` file) within `guardian/cdk-playground` - https://github.com/guardian/cdk-playground/pull/496.

Additional testing has been done via unit tests.

ASGs can scale _during_ deployment.

As noted above, there are some unknowns. Any service that starts using this pattern in its `experimental` form implicitly accepts the risk of these unknowns.

## Checklist

### DIFF
--- a/src/experimental/patterns/ec2-app.ts
+++ b/src/experimental/patterns/ec2-app.ts
@@ -49,6 +49,7 @@ export class GuEc2AppExperimental extends GuEc2App {
     See https://github.com/guardian/amigo/tree/main/roles/aws-tools.
      */
     userData.addCommands(
+      `# ${GuEc2AppExperimental.name} UserData Start`,
       `
       INSTANCE_ID=$(ec2metadata --instance-id)
 
@@ -68,6 +69,7 @@ export class GuEc2AppExperimental extends GuEc2App {
           --query "TargetHealthDescriptions[0].TargetHealth.State")
       done
       `,
+      `# ${GuEc2AppExperimental.name} UserData End`,
     );
 
     userData.addOnExitCommands(
--- a/src/experimental/patterns/ec2-app.ts
+++ b/src/experimental/patterns/ec2-app.ts
@@ -1,0 +1,99 @@
+import { Duration } from "aws-cdk-lib";
+import type { CfnAutoScalingGroup } from "aws-cdk-lib/aws-autoscaling";
+import { UpdatePolicy } from "aws-cdk-lib/aws-autoscaling";
+import { Effect, Policy, PolicyStatement } from "aws-cdk-lib/aws-iam";
+import type { GuStack } from "../../constructs/core";
+import type { GuEc2AppProps } from "../../patterns";
+import { GuEc2App } from "../../patterns";
+
+export interface GuEc2AppExperimentalProps extends Omit<GuEc2AppProps, "updatePolicy"> {}
+
+export class GuEc2AppExperimental extends GuEc2App {
+  constructor(scope: GuStack, props: GuEc2AppExperimentalProps) {
+    super(scope, { ...props, updatePolicy: UpdatePolicy.replacingUpdate() });
+
+    const { applicationPort } = props;
+    const { minimumInstances } = props.scaling;
+    const { region, stackId } = scope;
+    const { autoScalingGroup, targetGroup } = this;
+    const { userData, role } = autoScalingGroup;
+    const cfnAutoScalingGroup = autoScalingGroup.node.defaultChild as CfnAutoScalingGroup;
+
+    new Policy(scope, "AsgReplacingUpdatePolicy", {
+      statements: [
+        // Allow usage of command `cfn-signal`.
+        new PolicyStatement({
+          actions: ["cloudformation:SignalResource"],
+          effect: Effect.ALLOW,
+          resources: [stackId],
+        }),
+
+        /*
+        Allow usage of command `aws elbv2 describe-target-health`.
+        AWS Elastic Load Balancing does not support resource based policies, so the resource has to be `*` (any) here.
+        See https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_aws-services-that-work-with-iam.html
+         */
+        new PolicyStatement({
+          actions: ["elasticloadbalancing:DescribeTargetHealth"],
+          effect: Effect.ALLOW,
+          resources: ["*"],
+        }),
+      ],
+    }).attachToRole(role);
+
+    /*
+    `ec2metadata` is available via `cloud-utils` installed on all Canonical Ubuntu AMIs.
+    See https://github.com/canonical/cloud-utils.
+
+    `aws` is available via AMIgo baked AMIs.
+    See https://github.com/guardian/amigo/tree/main/roles/aws-tools.
+     */
+    userData.addCommands(
+      `
+      INSTANCE_ID=$(ec2metadata --instance-id)
+
+      STATE=$(aws elbv2 describe-target-health \
+        --target-group-arn ${targetGroup.targetGroupArn} \
+        --region ${region} \
+        --targets Id=$INSTANCE_ID,Port=${applicationPort} \
+        --query "TargetHealthDescriptions[0].TargetHealth.State")
+
+      until [ "$STATE" == "\\"healthy\\"" ]; do
+        echo "Instance not yet healthy within target group. Current state $STATE. Sleeping..."
+        sleep 5
+        STATE=$(aws elbv2 describe-target-health \
+          --target-group-arn ${targetGroup.targetGroupArn} \
+          --region ${region} \
+          --targets Id=$INSTANCE_ID,Port=${applicationPort} \
+          --query "TargetHealthDescriptions[0].TargetHealth.State")
+      done
+      `,
+    );
+
+    userData.addOnExitCommands(
+      `
+        cfn-signal --stack ${stackId} \
+          --resource ${cfnAutoScalingGroup.logicalId} \
+          --region ${region} \
+          --exit-code $exitCode || echo 'Failed to send Cloudformation Signal'
+        `,
+    );
+
+    // TODO are these sensible values?
+    const signalTimeoutSeconds = Math.max(
+      targetGroup.healthCheck.timeout?.toSeconds() ?? 0,
+      cfnAutoScalingGroup.healthCheckGracePeriod ?? 0,
+      Duration.minutes(5).toSeconds(),
+    );
+
+    cfnAutoScalingGroup.cfnOptions.creationPolicy = {
+      autoScalingCreationPolicy: {
+        minSuccessfulInstancesPercent: 100,
+      },
+      resourceSignal: {
+        count: minimumInstances,
+        timeout: Duration.seconds(signalTimeoutSeconds).toIsoString(),
+      },
+    };
+  }
+}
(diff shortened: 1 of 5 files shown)


src/experimental/patterns/ec2-app.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    /*
+    `ec2metadata` is available via `cloud-utils` installed on all Canonical Ubuntu AMIs.
+    See https://github.com/canonical/cloud-utils.
+
+    `aws` is available via AMIgo baked AMIs.
+    See https://github.com/guardian/amigo/tree/main/roles/aws-tools.
+     */
+    userData.addCommands(
+      `# ${GuEc2AppExperimental.name} UserData Start`,
```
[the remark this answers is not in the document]
akash1810: An practical example of the full diff to the user-data can be seen here https://github.com/guardian/cdk-playground/pull/496#issuecomment-2257682961.

src/experimental/patterns/ec2-app.ts
```diff
@@ -0,0 +1,132 @@
...
+          --exit-code $exitCode || echo 'Failed to send Cloudformation Signal'
+        `,
+    );
+
+    // TODO are these sensible values?
+    const signalTimeoutSeconds = Math.max(
+      targetGroup.healthCheck.timeout?.toSeconds() ?? 0,
+      cfnAutoScalingGroup.healthCheckGracePeriod ?? 0,
+      Duration.minutes(5).toSeconds(),
+    );
```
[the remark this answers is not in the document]
jacobwinch: I think using the health check grace period or 5 minutes (whichever is higher) makes sense as a starting point.

I _think_ `targetGroup.healthCheck.timeout` maps to [`HealthCheckTimeoutSeconds`](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/target-group-health-checks.html#health-check-settings). This is the timeout for _each_ healthcheck request made by the ALB and has a max timeout of 120 seconds, so I think we could safely drop this one.

src/experimental/patterns/ec2-app.ts
```diff
@@ -0,0 +1,132 @@
...
+ * TODO test scaling alarm behaviour.
+ *
+ * To update the application's code, a CloudFormation update must be triggered.
+ * The best way to do this is to include the build number in the application artifact.
+ * TODO test User Data includes a build number.
+ *
+ * NOTE: This pattern:
+ *  - Is NOT compatible with the "autoscaling" Riff-Raff deployment type.
+ *  - Your application should include a build number in its filename.
+ *    This value will change across builds, and therefore create a CloudFormation template difference to be deployed.
```
adamnfish, line 36: If the AMI also gets updated by CloudFormation, would that be a separate CloudFormation change, or would we try and do it together in one update? I think the reason this is important is that CloudFormation will rollback failed changes, and we wouldn't it to rollback halfway to the old ASG with the new AMI.

src/experimental/patterns/ec2-app.ts
```diff
@@ -0,0 +1,132 @@
...
+    }).attachToRole(role);
+
+    /*
+    `ec2metadata` is available via `cloud-utils` installed on all Canonical Ubuntu AMIs.
+    See https://github.com/canonical/cloud-utils.
+
+    `aws` is available via AMIgo baked AMIs.
+    See https://github.com/guardian/amigo/tree/main/roles/aws-tools.
+     */
+    userData.addCommands(
```
adamnfish, line 82: I'd love us to move away from teams having to write the boiler plate of fetching their application from S3. We don't necessarily have to think about this at the same time, but if we're already changing artifacts around to include the build number, is this an opportunity to remove this as well at the same time, with only one breaking change?

I'd say Ideally userdata should be used by teams for appplication specific work if required, and we'd have everything needed for deploying and running a "normal" application in place.

Again, no need to tackle everything at once so if this is non-trivial then for sure let's consider it later.

src/experimental/patterns/ec2-app.ts
```diff
@@ -0,0 +1,132 @@
...
+
+      until [ "$STATE" == "\\"healthy\\"" ]; do
+        echo "Instance not yet healthy within target group. Current state $STATE. Sleeping..."
+        sleep 5
+        STATE=$(aws elbv2 describe-target-health \
+          --target-group-arn ${targetGroup.targetGroupArn} \
+          --region ${region} \
+          --targets Id=$INSTANCE_ID,Port=${applicationPort} \
+          --query "TargetHealthDescriptions[0].TargetHealth.State")
+      done
```
adamnfish, line 101: Should we add a log line here to say words to the effect of "yeah we're all good, I got healthy back from the target health call"?

src/experimental/patterns/ec2-app.ts
akash1810: Currently, Riff-Raff attempts to update the AMI on each deployment. That is, the changes are bundled together into one.

akash1810:
Here is a summary of observations made. I think the negative changes this mechanism creates for incident triaging outweighs the benefits.

## Launch templates
Launch templates are a prerequisite for this solution; we still have a [number of services](https://metrics.gutools.co.uk/goto/BpDztc9SR?orgId=1) using launch configurations.

## Scaling
If, during a deployment, the current ASG scales up, the second ASG will always be under-provisioned.

Let’s say we have this timeline:
1. Within a CloudFormation stack, an ASG has min = 3, max = 10, desired = 3
2. CloudFormation update starts, and a new ASG created. This ASG has capacity 3/10/3 too.
3. The first ASG scales up, doubling capacity; it is now 3/10/6.
4. CloudFormation removes the first ASG, leaving one ASG at 3/10/3.

We’re left with a single ASG with capacity 3/10/3. This is under-provisioned (assuming we’re still in an alarm state).

Indeed, this is confirmed by [AWS support](https://support.console.aws.amazon.com/support/home?region=eu-west-1#/case/?displayId=172251339300370&language=en):

> I think if you want to maintain 6 after update you need to specify the desired capacity to 6 for the update and the old ASG will remove its own 3 and leave you with 6.

AWS recommends doing this by making the scale out alarm more sensitive.

> During the meeting we touched on the bases of Auto scaling behavior during "AutoScalingReplacingUpdate", when an alarm is triggered for the old ASG to add capacity while the new ASG maintains the previous desired capacity. I mentioned to you that the old instances should continue to serve all flights requests as per the connection draining, while new connections will be sent to the new instance. Then, if during this time the aggregated CPUUtilization seem to be spiking to the threshold, the alarm attached to the ASG will react to the change, to either add more instances. However, what determines how quick response of the alarm is the number of datapoints and the period of evaluation.

## Activity history
With new ASGs created, we lose the ability to use the ASG “Activity History” to, for example,  understand why an instance was unexpectedly terminated.

AWS have suggested two alternatives:
1. Use CloudTrail logs. We’d need to know the ARN of the ASG(s), then join the CloudTrail events.
2. Use the CLI. Using the [describe-scaling-activities command](https://docs.aws.amazon.com/cli/latest/reference/autoscaling/describe-scaling-activities.html) and flag --include-deleted-groups, we’re able to view activity history. We’d need to know the ARN of the ASG(s).

## CloudWatch Metrics
Each ASG records metrics to CloudWatch to a namespace based on the ASG name. With multiple ASGs, these metrics will be spread across namespaces. To get a continuous timeline of these metrics, we have to manually join the metrics. How, is yet unanswered.

## Security SLAs
A number of our security SLAs come into effect based on the age of an ASG. With each deployment creating a new ASG, these metrics will become nonsensical.

⚠️ We could create some tooling to improve the DX of observing the metrics, activity, etc. of multiple ASGs for a service. We should consider the cost of doing this beforehand though.

PR #2417, opened 2024-08-20
feat(experimental-ec2-pattern): Pattern to deploy ASG updates w/CFN

akash1810:
## What does this change?
Reduce the “time spent broken” after failed deploys, making [these manual steps](https://github.com/guardian/riff-raff/blob/fc88a4ca485c93423e62724cea4cbc266daa9f00/riff-raff/public/docs/howto/fix-a-failed-deploy.md) redundant, by adding a new [experimental](https://github.com/guardian/cdk/blob/main/docs/architecture-decision-records/005-x01-package-structure.md) pattern `GuEc2AppExperimental` to provision an EC2 based service with an [`AutoScalingRollingUpdate` update policy](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatepolicy.html#cfn-attributes-updatepolicy-rollingupdate). The configuration is based on a number of [tests](https://github.com/guardian/testing-asg-rolling-update) we ran; we've [documented some various observations](https://github.com/guardian/testing-asg-rolling-update/tree/main/observations).

This is an alternative to #2395. It improves on the solution proposed in #2395 by providing continuity of ASG metrics and ASG activity history.

`AutoScalingRollingUpdate` offers an improved DX over the `autoscaling` Riff-Raff deployment type when a deployment fails. On deployment failure, the `autoscaling` Riff-Raff deployment type leaves the ASG over-capacity, requiring [manual intervention to resolve](https://github.com/guardian/riff-raff/blob/fc88a4ca485c93423e62724cea4cbc266daa9f00/riff-raff/public/docs/howto/fix-a-failed-deploy.md), whereas `AutoScalingRollingUpdate` will perform a rollback automatically. This means deployments are more atomic.

> [!NOTE]
> Unfortunately [testing](https://github.com/guardian/testing-asg-rolling-update) has shown unpredictable behaviour if a scale-in event occurs during deployment. Therefore we suspend the `AlarmNotification` process during a rolling update.
> 
> This means scaling events cannot occur during deployment and https://github.com/guardian/riff-raff/issues/1342 is not fixed.

With `AutoScalingRollingUpdate` CloudFormation will deploy application updates to an ASG in a rolling fashion. The rate at which the update rolls out is dependant on how scaled-out an ASG is:
- If an ASG does not have any scaling policies, deployment will behave similar to Riff-Raff; the `DesiredCapacity` will be doubled, once new instances are healthy the capacity will be halved removing the old instances.
- For an ASG that has a scaling policy, the rate will be dynamic:
  - If it is running normally, `DesiredCapacity` will be doubled, once new instances are healthy the capacity will be halved removing the old instances ([testing observations](https://github.com/guardian/testing-asg-rolling-update/blob/main/observations/healthy-to-healthy.md)).
  - If the service is partially scaled, deployment will happen in batches relative to the remaining capacity (`MinCapacity` / (`MaxCapacity` - `DesiredCapacity`)) ([testing observations](https://github.com/guardian/testing-asg-rolling-update/blob/main/observations/healthy-to-healthy-partially-scaled-v2.md)).
  - If the service is fully scaled, deployment will happen in a "one out, one in" fashion ([testing observations](https://github.com/guardian/testing-asg-rolling-update/blob/main/observations/healthy-to-healthy-fully-scaled-v2.md)).

   This will be achieved by Riff-Raff setting [`MinInstancesInService`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatepolicy.html#cfn-attributes-updatepolicy-rollingupdate-mininstancesinservice) at deploy time via a CloudFormation Parameter.
   > 
   This functionality is yet to be added to Riff-Raff. It doesn't block shipping this change, however services that use scaling policies cannot use this experimental pattern until then.

As application updates are performed entirely by CloudFormation, it is not necessary to use the `autoscaling` RIff-Raff deployment type[^1]. If it is used, it shouldn't break anything. However it does mean deployments will take a long time, as the instances will be rotated twice; once by CloudFormation and once by Riff-Raff's custom logic.

### Changes to UserData
With `AutoScalingRollingUpdate` CloudFormation will remain in the `UPDATE_IN_PROGRESS` state whilst waiting for a `SUCCESS` [signal](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-signal.html) to be received from the instance. This signal is sent manually.

The `GuEc2AppExperimental` pattern will add this signal emission to the UserData. A `SUCCESS` signal will be sent once the instance has successfully registered with the Target Group.

### Why `experimental`?
There are a few requirements for this approach:
- The AWS CLI, and `cfn-signal` binaries need to be available and on the `PATH` (this is added via the [AMIgo role `aws-tools`](https://github.com/guardian/amigo/tree/main/roles/aws-tools)).
- The service's artifact should include a build number, as this is the most reliable way to create a difference with the running CloudFormation template, and thus trigger an update.

We're not (yet) validating these are met, as its tricky.

### Proposed rollout
The rollout plan looks something like this:
1. Dogfood on some DevX services.
2. Test on a service within the department.
3. Move the pattern to stable, as a breaking change, with communication on the migration path, etc.

## How to test
These changes have been informed by "real-world" testing that has been documented in https://github.com/guardian/testing-asg-rolling-update.

This branch is being successfully used in [guardian/amiable](https://github.com/guardian/amiable/pull/717), [guardian/cdk-playground](https://github.com/guardian/cdk-playground/pull/522) and [guardian/security-hq](https://github.com/guardian/security-hq/pull/1154).

A DX improvement when a deployment fails.

Once fully adopted, we will be able to remove the `autoscaling` deployment type in Riff-Raff, reducing it's complexity.

As this new pattern is `experimental` there risk levels are fairly small. We should monitor usage and gather feedback before promoting to stable.

## Checklist

- [ - ] I have listed any breaking changes, along with a migration path [^2]
- [x] I have updated the documentation as required for the described changes [^3]

[^1]: We still need the `uploadArtifacts` action of the [`autoscaling` deployment type](https://riffraff.gutools.co.uk/docs/magenta-lib/types#autoscaling). The `cloud-formation` deployment type should depend on this step too.

### DIFF
--- a/.changeset/happy-badgers-compare.md
+++ b/.changeset/happy-badgers-compare.md
@@ -1,0 +1,43 @@
+---
+"@guardian/cdk": minor
+---
+
+feat(experimental-ec2-pattern): Pattern to deploy ASG updates w/CFN
+
+Included in this update is a new experimental pattern `GuEc2AppExperimental`, which can be used in place of a `GuEc2App`:
+
+```ts
+import { GuEc2AppExperimental } from "@guardian/cdk/lib/experimental/patterns/ec2-app";
+```
+
+This pattern will add an [`AutoScalingRollingUpdate` policy](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatepolicy.html#cfn-attributes-updatepolicy-rollingupdate)
+to the autoscaling group.
+This allows application updates to be performed like a standard CloudFormation update,
+and using the custom logic provided by Riff-Raff's `autoscaling` deployment type is unnecessary.
+
+This experimental pattern has few requirements.
+
+## Add the build number to the application artifact
+This change requires versioned artifacts.
+
+The easiest way to achieve this is by adding the build number to the filename of the artifact:
+
+```ts
+import { UserData } from "aws-cdk-lib/aws-ec2";
+// Use a GitHub Actions provided environment variable
+const buildNumber = process.env.GITHUB_RUN_NUMBER ?? "DEV";
+
+const userData = UserData.forLinux();
+userData.addCommands(`aws s3 cp s3://dist-bucket/path/to/artifact-${buildNumber}.deb /tmp/artifact.deb`);
+userData.addCommands(`dpkg -i /tmp/artifact.dep`);
+```
+
+## `riff-raff.yaml`
+The `riff-raff.yaml` file should remove the `deploy` action of the `autoscaling` deployment type.
+Though including it shouldn't break anything, it would result in a longer deployment time as instance will be rotated by both CloudFormation and Riff-Raff's custom logic.
+
+The `uploadArtifacts` step of the `autoscaling` deployment type should still be included, with the `cloud-formation` deployment type depending on it.
+This step uploads the versioned artifact to S3.
+
+> [!TIP]
+> An [auto-generated `riff-raff.yaml` file](https://github.com/guardian/cdk/blob/main/src/riff-raff-yaml-file/README.md) meets this requirement.
--- a/src/experimental/patterns/ec2-app.test.ts
+++ b/src/experimental/patterns/ec2-app.test.ts
@@ -1,17 +1,17 @@
-import { Duration } from "aws-cdk-lib";
-import { Template } from "aws-cdk-lib/assertions";
+import { App, Duration } from "aws-cdk-lib";
+import { Match, Template } from "aws-cdk-lib/assertions";
 import { InstanceClass, InstanceSize, InstanceType, UserData } from "aws-cdk-lib/aws-ec2";
+import { CloudFormationStackArtifact } from "aws-cdk-lib/cx-api";
 import { AccessScope } from "../../constants";
 import { GuUserData } from "../../constructs/autoscaling";
-import type { GuStack } from "../../constructs/core";
+import { GuStack } from "../../constructs/core";
 import { simpleGuStackForTesting } from "../../utils/test";
 import type { GuEc2AppExperimentalProps } from "./ec2-app";
 import { GuEc2AppExperimental } from "./ec2-app";
 
 // TODO test User Data includes a build number
 describe("The GuEc2AppExperimental pattern", () => {
-  function initialProps(scope: GuStack): GuEc2AppExperimentalProps {
-    const app = "test-gu-ec2-app";
+  function initialProps(scope: GuStack, app: string = "test-gu-ec2-app"): GuEc2AppExperimentalProps {
     const buildNumber = 123;
 
     const { userData } = new GuUserData(scope, {
@@ -139,5 +139,68 @@ describe("The GuEc2AppExperimental pattern", () => {
 
     // The target group healthcheck polling should be the last thing in the user data.
     expect(endMarkerPosition).toEqual(totalLines - 1);
+  });
+
+  it("should adjust properties of a horizontally scaling service", () => {
+    const cdkApp = new App();
+    const stack = new GuStack(cdkApp, "test", {
+      stack: "test-stack",
+      stage: "TEST",
+    });
+
+    const scalingApp = "my-scaling-app";
+    const { autoScalingGroup } = new GuEc2AppExperimental(stack, {
+      ...initialProps(stack, scalingApp),
+      scaling: {
+        minimumInstances: 5,
+      },
+    });
+    autoScalingGroup.scaleOnRequestCount("ScaleOnRequests", {
+      targetRequestsPerMinute: 100,
+    });
+
+    /*
+    We're ultimately testing an `Aspect`, which appear to run only at synth time.
+    As a work-around, synth the `App`, then perform assertions on the resulting template.
+
+    See also: https://github.com/aws/aws-cdk/issues/29047.
+     */
+    const { artifacts } = cdkApp.synth();
+    const cfnStack = artifacts.find((_): _ is CloudFormationStackArtifact => _ instanceof CloudFormationStackArtifact);
+
+    if (!cfnStack) {
+      throw new Error("Unable to locate a CloudFormationStackArtifact");
+    }
+
+    const template = Template.fromJSON(cfnStack.template as Record<string, unknown>);
+
+    const parameterName = `MinInstancesInServiceFor${scalingApp.replaceAll("-", "")}`;
+
+    template.hasParameter(parameterName, {
+      Type: "Number",
+      Default: 5,
+      MaxValue: 9, // (min * 2) - 1
+    });
+
+    template.hasResource("AWS::AutoScaling::AutoScalingGroup", {
+      Properties: {
+        MinSize: "5",
+        MaxSize: "10",
+        DesiredCapacity: Match.absent(),
+        Tags: Match.arrayWith([{ Key: "App", Value: scalingApp, PropagateAtLaunch: true }]),
+      },
+      UpdatePolicy: {
+        AutoScalingRollingUpdate: {
+          MaxBatchSize: 10,
+          SuspendProcesses: ["AlarmNotification"],
+          MinSuccessfulInstancesPercent: 100,
+          WaitOnResourceSignals: true,
+          PauseTime: "PT5M",
+          MinInstancesInService: {
+            Ref: parameterName,
+          },
+        },
+      },
+    });
   });
 });
--- a/src/experimental/patterns/ec2-app.test.ts
+++ b/src/experimental/patterns/ec2-app.test.ts
@@ -10,6 +10,31 @@ import { simpleGuStackForTesting } from "../../utils/test";
 import type { GuEc2AppExperimentalProps } from "./ec2-app";
 import { GuEc2AppExperimental } from "./ec2-app";
 
+/**
+ * `Aspects` appear to run only at synth time.
+ * This means we must synth the stack to see the results of the `Aspect`.
+ *
+ * @see https://github.com/aws/aws-cdk/issues/29047
+ *
+ * @param stack the stack to synthesise
+ */
+function getTemplateAfterAspectInvocation(stack: GuStack): Template {
+  const app = App.of(stack);
+
+  if (!app) {
+    throw new Error(`Unable to locate the enclosing App from GuStack ${stack.node.id}`);
+  }
+
+  const { artifacts } = app.synth();
+  const cfnStack = artifacts.find((_): _ is CloudFormationStackArtifact => _ instanceof CloudFormationStackArtifact);
+
+  if (!cfnStack) {
+    throw new Error("Unable to locate a CloudFormationStackArtifact");
+  }
+
+  return Template.fromJSON(cfnStack.template as Record<string, unknown>);
+}
+
 // TODO test User Data includes a build number
 describe("The GuEc2AppExperimental pattern", () => {
   function initialProps(scope: GuStack, app: string = "test-gu-ec2-app"): GuEc2AppExperimentalProps {
@@ -143,11 +168,7 @@ describe("The GuEc2AppExperimental pattern", () => {
   });
 
   it("should only adjust properties of a horizontally scaling service", () => {
-    const cdkApp = new App();
-    const stack = new GuStack(cdkApp, "test", {
-      stack: "test-stack",
-      stage: "TEST",
-    });
+    const stack = simpleGuStackForTesting();
 
     const scalingApp = "my-scaling-app";
     const { autoScalingGroup } = new GuEc2AppExperimental(stack, {
@@ -163,20 +184,7 @@ describe("The GuEc2AppExperimental pattern", () => {
     const staticApp = "my-static-app";
     new GuEc2AppExperimental(stack, initialProps(stack, staticApp));
 
-    /*
-    We're ultimately testing an `Aspect`, which appear to run only at synth time.
-    As a work-around, synth the `App`, then perform assertions on the resulting template.
-
-    See also: https://github.com/aws/aws-cdk/issues/29047.
-     */
-    const { artifacts } = cdkApp.synth();
-    const cfnStack = artifacts.find((_): _ is CloudFormationStackArtifact => _ instanceof CloudFormationStackArtifact);
-
-    if (!cfnStack) {
-      throw new Error("Unable to locate a CloudFormationStackArtifact");
-    }
-
-    const template = Template.fromJSON(cfnStack.template as Record<string, unknown>);
+    const template = getTemplateAfterAspectInvocation(stack);
 
     // The scaling ASG should NOT have `DesiredCapacity` set, and `MinInstancesInService` set via a CFN Parameter
     const parameterName = `MinInstancesInServiceFor${scalingApp.replaceAll("-", "")}`;
@@ -214,11 +222,7 @@ describe("The GuEc2AppExperimental pattern", () => {
   });
 
   it("should add a single CFN Parameter per ASG regardless of how many scaling policies are attached to it", () => {
-    const cdkApp = new App();
-    const stack = new GuStack(cdkApp, "test", {
-      stack: "test-stack",
-      stage: "TEST",
-    });
+    const stack = simpleGuStackForTesting();
 
     const scalingApp = "my-scaling-app";
     const { autoScalingGroup } = new GuEc2AppExperimental(stack, {
@@ -245,20 +249,7 @@ describe("The GuEc2AppExperimental pattern", () => {
       scalingAdjustment: -1,
     });
 
-    /*
-    We're ultimately testing an `Aspect`, which appear to run only at synth time.
-    As a work-around, synth the `App`, then perform assertions on the resulting template.
-
-    See also: https://github.com/aws/aws-cdk/issues/29047.
-     */
-    const { artifacts } = cdkApp.synth();
-    const cfnStack = artifacts.find((_): _ is CloudFormationStackArtifact => _ instanceof CloudFormationStackArtifact);
-
-    if (!cfnStack) {
-      throw new Error("Unable to locate a CloudFormationStackArtifact");
-    }
-
-    const template = Template.fromJSON(cfnStack.template as Record<string, unknown>);
+    const template = getTemplateAfterAspectInvocation(stack);
 
     const parameterName = `MinInstancesInServiceFor${scalingApp.replaceAll("-", "")}`;
 
--- a/src/experimental/patterns/ec2-app.test.ts
+++ b/src/experimental/patterns/ec2-app.test.ts
@@ -1,0 +1,105 @@
+import { Duration } from "aws-cdk-lib";
+import { Template } from "aws-cdk-lib/assertions";
+import { InstanceClass, InstanceSize, InstanceType } from "aws-cdk-lib/aws-ec2";
+import { AccessScope } from "../../constants";
+import { GuUserData } from "../../constructs/autoscaling";
+import type { GuStack } from "../../constructs/core";
+import { simpleGuStackForTesting } from "../../utils/test";
+import type { GuEc2AppExperimentalProps } from "./ec2-app";
+import { GuEc2AppExperimental } from "./ec2-app";
+
+// TODO test User Data includes a build number
+describe("The GuEc2AppExperimental pattern", () => {
+  function initialProps(scope: GuStack): GuEc2AppExperimentalProps {
+    const app = "test-gu-ec2-app";
+    const buildNumber = 123;
+
+    const { userData } = new GuUserData(scope, {
+      app,
+      distributable: {
+        fileName: `${app}-${buildNumber}.deb`,
+        executionStatement: `dpkg -i /${app}/${app}-${buildNumber}.deb`,
+      },
+    });
+
+    return {
+      applicationPort: 9000,
+      app,
+      access: { scope: AccessScope.PUBLIC },
+      instanceType: InstanceType.of(InstanceClass.T4G, InstanceSize.MEDIUM),
+      monitoringConfiguration: { noMonitoring: true },
+      userData,
+      certificateProps: {
+        domainName: "domain-name-for-your-application.example",
+      },
+      scaling: {
+        minimumInstances: 1,
+      },
+    };
+  }
+
+  it("matches the snapshot", () => {
+    const stack = simpleGuStackForTesting();
+    new GuEc2AppExperimental(stack, initialProps(stack));
+    expect(Template.fromStack(stack).toJSON()).toMatchSnapshot();
+  });
+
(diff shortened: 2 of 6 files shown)


src/experimental/patterns/ec2-app.ts
```diff
@@ -0,0 +1,282 @@
...
+ *  - If the service is fully scaled, it'll be set to (at least) `MaxSize` - 1
+ *
+ * @privateRemarks
+ * - Temporarily implemented as a singleton to ensure only one instance is added to a stack,
+ *   else multiple attempts to add the same CFN Parameter will be made and fail.
+ * - Once out of experimental, instantiate this `Aspect` directly in {@link GuStack}.
+ *
+ * @see https://github.com/guardian/testing-asg-rolling-update
+ */
+class HorizontallyScalingDeploymentProperties implements IAspect {
```
akash1810, line 33: This is implemented as an [`Aspect`](https://docs.aws.amazon.com/cdk/v2/guide/aspects.html) to handle cases when a scaling policy is added after the call to `super` of an `GuStack` implementer's constructor.

src/experimental/patterns/ec2-app.ts
```diff
@@ -0,0 +1,282 @@
...
+        as a rolling update will set the current `Desired` back to the template version,
+        undoing any changes that a scale-out event may have done.
+         */
+        cfnAutoScalingGroup.desiredCapacity = undefined;
+
+        const minInstancesInService = new CfnParameter(guStack, `MinInstancesInServiceFor${autoScalingGroup.app}`, {
+          type: "Number",
+          default: parseInt(cfnAutoScalingGroup.minSize),
+          maxValue: parseInt(cfnAutoScalingGroup.maxSize) - 1,
+        });
```
akash1810, line 76: If the `app` contains hyphens (`-`), or other illegal characters for a CFN Parameter, they'll get stripped out. This means Riff-Raff cannot reliably just find CFN Parameters that start with `MinInstancesInServiceFor`. Therefore Riff-Raff's logic would be something like this:
1. Find all `AWS::AutoScaling::ScalingPolicy`s in the CloudFormation template that has `MinInstancesInService` set via a CloudFormation Parameter
2. Find current `DesiredCapacity` of ASG that relates to the Scaling Policy
3. Set the CFN Parameter identified in step 1 to the value identified in step 2

src/experimental/patterns/ec2-app.ts
akash1810, line 1: This file includes three classes, only one of which is exported. This is for a couple of reasons:
- It doesn't make sense to use the two other classes in any context other than in `GuEc2AppExperimental`
- It's less work 😅, we can split into multiple files if this feature gets promoted to stable

src/experimental/patterns/ec2-app.ts
```diff
@@ -0,0 +1,310 @@
...
+        const asgNodeId = autoScalingGroup.node.id;
+
+        if (!this.asgToParamMap.has(asgNodeId)) {
+          this.asgToParamMap.set(
+            asgNodeId,
+            new CfnParameter(guStack, `MinInstancesInServiceFor${autoScalingGroup.app}`, {
+              type: "Number",
+              default: parseInt(cfnAutoScalingGroup.minSize),
+              maxValue: parseInt(cfnAutoScalingGroup.maxSize) - 1,
+            }),
```
akash1810, line 99: If the `app` contains hyphens (`-`), or other illegal characters for a CFN Parameter, they'll get stripped out. This means Riff-Raff cannot reliably just find CFN Parameters that start with `MinInstancesInServiceFor`. Therefore Riff-Raff's logic would be something like this:
1. Find all `AWS::AutoScaling::ScalingPolicy`s in the CloudFormation template that has `MinInstancesInService` set via a CloudFormation Parameter
2. Find current `DesiredCapacity` of ASG that relates to the Scaling Policy
3. Set the CFN Parameter identified in step 1 to the value identified in step 2

src/experimental/patterns/ec2-app.ts
```diff
@@ -0,0 +1,309 @@
...
+        const asgNodeId = autoScalingGroup.node.id;
+
+        if (!this.asgToParamMap.has(asgNodeId)) {
+          this.asgToParamMap.set(
+            asgNodeId,
+            new CfnParameter(this.stack, `MinInstancesInServiceFor${autoScalingGroup.app}`, {
+              type: "Number",
+              default: parseInt(cfnAutoScalingGroup.minSize),
+              maxValue: parseInt(cfnAutoScalingGroup.maxSize) - 1,
+            }),
```
akash1810, line 98: If the `app` contains hyphens (`-`), or other illegal characters for a CFN Parameter, they'll get stripped out. This means Riff-Raff cannot reliably just find CFN Parameters that start with `MinInstancesInServiceFor`. Therefore Riff-Raff's logic would be something like this:
1. Find all `AWS::AutoScaling::ScalingPolicy`s in the CloudFormation template that has `MinInstancesInService` set via a CloudFormation Parameter
2. Find current `DesiredCapacity` of ASG that relates to the Scaling Policy
3. Set the CFN Parameter identified in step 1 to the value identified in step 2

src/experimental/patterns/ec2-app.ts
```diff
@@ -0,0 +1,307 @@
...
+
+    /*
+    `aws` is available via AMIgo baked AMIs.
+    See https://github.com/guardian/amigo/tree/main/roles/aws-tools.
+     */
+    userData.addCommands(
+      `# ${GuEc2AppExperimental.name} UserData Start`,
+      `
+      TOKEN=$(curl -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 21600")
+      INSTANCE_ID=$(curl -H "X-aws-ec2-metadata-token: $TOKEN" "http://169.254.169.254/latest/meta-data/instance-id")
```
akash1810, line 272: Curiously the [`ec2metadata` command](https://github.com/canonical/cloud-utils/blob/main/bin/ec2metadata) was failing with a 401 in AMIable CODE in deployTools account. IIUC a 401 is returned when an IMDSv2 token is not provided. I'm not sure how this is possible, as [`ec2metadata` only works with IMDSv2](https://github.com/canonical/cloud-utils/blob/49e5dd7849ee3c662f3db35e857148d02e72694b/bin/ec2metadata#L132).

Using the [AWS way](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html) of getting the instance ID seems to be more reliable. Not sure why. I can only conclude it's something with the AWS account, as [guardian/cdk-playground](https://github.com/guardian/cdk-playground/pull/522) and [guardian/testing-asg-rolling-update](https://github.com/guardian/testing-asg-rolling-update/pull/15) are fine in the developerPlayground account.

<details><summary>stack trace</summary>
<p>

```console
root@ip-10-248-51-213:/var/lib/cloud/instance# ec2metadata --instance-id
Traceback (most recent call last):
  File "/usr/bin/ec2metadata", line 249, in <module>
    main()
  File "/usr/bin/ec2metadata", line 245, in main
    display(metaopts, burl, prefix)
  File "/usr/bin/ec2metadata", line 192, in display
    value = m.get(metaopt)
  File "/usr/bin/ec2metadata", line 177, in get
    return self._get('meta-data/' + metaopt)
  File "/usr/bin/ec2metadata", line 137, in _get
    resp = urllib_request.urlopen(urllib_request.Request(url))
  File "/usr/lib/python3.8/urllib/request.py", line 222, in urlopen
    return opener.open(url, data, timeout)
  File "/usr/lib/python3.8/urllib/request.py", line 531, in open
    response = meth(req, response)
  File "/usr/lib/python3.8/urllib/request.py", line 640, in http_response
    response = self.parent.error(
  File "/usr/lib/python3.8/urllib/request.py", line 569, in error
    return self._call_chain(*args)
  File "/usr/lib/python3.8/urllib/request.py", line 502, in _call_chain
    result = func(*args)
  File "/usr/lib/python3.8/urllib/request.py", line 649, in http_error_default
    raise HTTPError(req.full_url, code, msg, hdrs, fp)
urllib.error.HTTPError: HTTP Error 401: Unautho
```

</p>
</details>

src/experimental/patterns/ec2-app.ts
```diff
@@ -0,0 +1,307 @@
...
+ * This improves on the behaviour of Riff-Raff's `autoscaling` deployment which requires manual intervention on deployment failure.
+ *
+ * To update the application's code, a CloudFormation update must be triggered.
+ * The best way to do this is to include the build number in the application artifact.
+ * TODO test User Data includes a build number.
+ *
+ * @remarks
+ * - The rate at which instances are replaced are conditional based on how scaled up the service is.
+ *   The more an application is scaled, the longer it'll take to deploy.
+ * - It is not necessary to use the `autoscaling` Riff-Raff deployment type in conjunction with this.
```
akash1810, line 182: https://github.com/guardian/security-hq/pull/1154 shows the `riff-raff.yaml` changes required if creating the file manually:
1. Use the `autoscaling` deployment type, but only the `uploadArtifacts` action
2. Have the `cloud-formation` deployment type depend on step 1

Obviously, it's not required, but a reduced risk migration path would look like this:
1. Update to latest version of GuCDK
2. Migrate to using the auto-generated `riff-raff.yaml` file
3. Use the experimental pattern

Where step 1 and 2 can go to `main` independent of 3.

akash1810: I'll add this nuanced detail to the release notes.

akash1810: Done - https://github.com/guardian/cdk/pull/2417/commits/f3072e101d71bba24ba987b416b6753d1c1c4834.

src/experimental/patterns/ec2-app.test.ts
```diff
@@ -0,0 +1,407 @@
...
+    const { artifacts } = cdkApp.synth();
+    const cfnStack = artifacts.find((_): _ is CloudFormationStackArtifact => _ instanceof CloudFormationStackArtifact);
+
+    if (!cfnStack) {
+      throw new Error("Unable to locate a CloudFormationStackArtifact");
+    }
+
+    const template = Template.fromJSON(cfnStack.template as Record<string, unknown>);
+
+    /*
```
jacobwinch, line 244: IIUC we have already checked a lot of this in the test above (`should adjust properties of a horizontally scaling service`), so I think we could simplify by only making assertions about the static ASG here?

src/experimental/patterns/ec2-app.test.ts
```diff
@@ -0,0 +1,407 @@
...
+          PauseTime: "PT5M",
+          MinInstancesInService: {
+            Ref: parameterName,
+          },
+        },
+      },
+    });
+
+    /*
+    The static ASG should:
```
jacobwinch, line 277: I wonder if it would be clearer to have a separate test for each of these assertions rather than including this comment?

src/experimental/patterns/ec2-app.test.ts
```diff
@@ -0,0 +1,407 @@
...
+
+    const parameterName = `MinInstancesInServiceFor${scalingApp.replaceAll("-", "")}`;
+
+    template.hasParameter(parameterName, {
+      Type: "Number",
+      Default: 5,
+      MaxValue: 9, // (min * 2) - 1
+    });
+
+    template.hasResource("AWS::AutoScaling::AutoScalingGroup", {
```
jacobwinch, line 186: As your comment below mentions, I think we are really interested in the values of two properties here. Perhaps it would be clearer to have a separate test/test name for each?

src/experimental/patterns/ec2-app.test.ts
```diff
@@ -0,0 +1,407 @@
...
+            Ref: parameterName,
+          },
+        },
+      },
+    });
+
+    template.resourceCountIs("AWS::AutoScaling::ScalingPolicy", 3);
+  });
+
+  it("should throw an error when a scaling policy is not created with aa GuAutoScalingGroup scope", () => {
```
jacobwinch, line 380: Minor: 

```suggestion
  it("should throw an error when a scaling policy is not created with a GuAutoScalingGroup scope", () => {
```

src/experimental/patterns/ec2-app.ts
```diff
@@ -0,0 +1,282 @@
...
+    /*
+    Enable all ASG metrics for improved observability.
+    Specifying `granularity` without specifying `metrics` results in all metrics being enabled.
+    See https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-autoscaling-autoscalinggroup-metricscollection.html
+     */
+    cfnAutoScalingGroup.metricsCollection = [
+      {
+        granularity: "1Minute",
+      },
+    ];
```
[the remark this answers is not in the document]
jacobwinch: I've just approved https://github.com/guardian/cdk/pull/2455.

src/experimental/patterns/ec2-app.ts
```diff
@@ -0,0 +1,307 @@
...
+    const currentRollingUpdate = cfnAutoScalingGroup.cfnOptions.updatePolicy?.autoScalingRollingUpdate;
+
+    cfnAutoScalingGroup.cfnOptions.updatePolicy = {
+      autoScalingRollingUpdate: {
+        ...currentRollingUpdate,
+        pauseTime: Duration.seconds(signalTimeoutSeconds).toIsoString(),
+      },
+    };
+
+    cfnAutoScalingGroup.cfnOptions.creationPolicy = {
```
jacobwinch, line 252: Minor: do we need to add the creation policy as part of this change? I think it could be added as a standalone feature without being coupled to this experimental pattern?

src/riff-raff-yaml-file/index.ts
```diff
@@ -253,7 +254,18 @@ export class RiffRaffYamlFile {
...
           });
 
-          autoscalingGroups.forEach((asg) => {
+          // ASGs without an UpdatePolicy can be deployed via Riff-Raff's (legacy) `autoscaling` deployment type.
+          // ASGs with an UpdatePolicy are updated via Riff-Raff's `cloud-formation` deployment type.
+          const legacyAutoscalingGroups = autoscalingGroups.filter((asg) => {
+            const { cfnOptions } = asg.node.defaultChild as CfnAutoScalingGroup;
+            const { updatePolicy } = cfnOptions;
+            return (
+              updatePolicy?.autoScalingReplacingUpdate === undefined &&
```
jacobwinch, line 263: Minor: I think we've decided against using replacing update, so perhaps we can remove this line so it's clearer that we don't support this?

src/experimental/patterns/ec2-app.ts
jacobwinch: I wonder if our `Aspect` could add [`Outputs`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/outputs-section-structure.html) for Riff-Raff to read. 

This would mean that Riff-Raff does not need to parse the full template or lookup the associated ASG name (which will not be included in the template unless [this property](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-autoscaling-autoscalinggroup.html#cfn-autoscaling-autoscalinggroup-autoscalinggroupname) is set).

For example:

```
Outputs:
  MinInstancesInServiceForAppA:
    Description: The AutoScaling group name associated with the MinInstancesInServiceForAppA parameter.    
    Value: !Ref AsgForAppA 
  MinInstancesInServiceForAppB:
    Description: The AutoScaling group name associated with the MinInstancesInServiceForAppB parameter.    
    Value: !Ref AsgForAppB
```

I think we can obtain these outputs pretty easily via the existing [`describeStack` function](https://github.com/guardian/riff-raff/blob/d02c0e5835a20d8389a97a19df896ebc4bed2f15/magenta-lib/src/main/scala/magenta/tasks/AWS.scala#L854-L868) and IIUC [using `!Ref` here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-autoscaling-autoscalinggroup.html#aws-resource-autoscaling-autoscalinggroup-return-values) will mean that we can avoid looking up the ASG name ourselves.

src/experimental/patterns/ec2-app.test.ts
akash1810: I've stripped this back to testing only the properties that we're interested in. Hopefully that addresses this comment?

src/experimental/patterns/ec2-app.ts
akash1810: Technically it can be its own change. However, this creation policy is configured to keep the ASG resource in `CREATE_IN_PROGRESS` until the correct number of `SUCCESS` signals are received. The signals need to be explicitly sent, which we've starting doing in this branch, so I think coupling to the experimental pattern is simplest. WDYT?

src/experimental/patterns/ec2-app.ts
akash1810: Oh interesting idea! Tempted to make these changes in a separate PR once we've know more about what the changes in Riff-Raff would look like. WDYT?

src/experimental/patterns/ec2-app.ts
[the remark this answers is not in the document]
akash1810: Merged that PR and rebased against `main` now.

src/riff-raff-yaml-file/index.ts
akash1810: Done.

.changeset/happy-badgers-compare.md
```diff
@@ -0,0 +1,43 @@
...
+userData.addCommands(`aws s3 cp s3://dist-bucket/path/to/artifact-${buildNumber}.deb /tmp/artifact.deb`);
+userData.addCommands(`dpkg -i /tmp/artifact.dep`);
+```
+
+## `riff-raff.yaml`
+The `riff-raff.yaml` file should remove the `deploy` action of the `autoscaling` deployment type.
+Though including it shouldn't break anything, it would result in a longer deployment time as instance will be rotated by both CloudFormation and Riff-Raff's custom logic.
+
+The `uploadArtifacts` step of the `autoscaling` deployment type should still be included, with the `cloud-formation` deployment type depending on it.
+This step uploads the versioned artifact to S3.
```
akash1810, line 40: An example of this can be seen in https://github.com/guardian/security-hq/pull/1154.

src/experimental/patterns/ec2-app.ts
jacobwinch: Ah yes, sorry, I now see that it's not a useful feature without the `SUCCESS` signals being sent - let's leave it in!

src/experimental/patterns/ec2-app.ts
```diff
@@ -0,0 +1,307 @@
...
+     */
+    cfnAutoScalingGroup.metricsCollection = [
+      {
+        granularity: "1Minute",
+      },
+    ];
+
+    // TODO are these sensible values?
+    const signalTimeoutSeconds = Math.max(
+      targetGroup.healthCheck.timeout?.toSeconds() ?? 0,
```
jacobwinch, line 238: I think we could drop this one for simplicity. IIUC it is the timeout _per healthcheck request_, so it will typically be very short. For example, I think we're [defaulting to 5 seconds here](https://github.com/guardian/cdk/blob/12817108d384a561c85f7713a3e943c26fa3a6f3/src/constructs/loadbalancing/alb/application-target-group.ts#L33-L44).

src/experimental/patterns/ec2-app.ts
```diff
@@ -0,0 +1,307 @@
...
+      cfnAutoScalingGroup.healthCheckGracePeriod ?? 0,
+      Duration.minutes(5).toSeconds(),
```
jacobwinch, line 240: I think it makes sense to explicitly couple this timeout to the healthcheck grace period (and I think this will be [2-minutes for most services](https://github.com/guardian/cdk/blob/12817108d384a561c85f7713a3e943c26fa3a6f3/src/patterns/ec2-app/base.ts#L428)).

Users can already tune this property based on how long their instances take to start passing healthchecks and adding this (potentially) longer and non-configurable default will cause the deployment of 'bad' builds to fail more slowly than necessary in many cases.

jacobwinch: If the above sounds sensible I think we could also drop a bit of logic and the `should create an ASG with the maximum resource signal timeout` unit test?

akash1810: This variable is used for the [`PauseTime` in the update policy](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatepolicy.html#cfn-attributes-updatepolicy-rollingupdate-pausetime), which is described as:

> It determines the maximum time that CloudFormation waits to receive the required number of valid signals from the instances being replaced during a rolling update and from new instances being added by increasing the [DesiredCapacity](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-autoscaling-autoscalinggroup.html#cfn-autoscaling-autoscalinggroup-desiredcapacity) property of the AWS::AutoScaling::AutoScalingGroup resource.

My reading of this is the timeout starts when `DesiredCapacity` is changed, and thus includes the time spent [pending](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-lifecycle.html). Do we know if the healthcheck grace period covers pending too? I thought it doesn't, but can't find any docs to suggest either way.

jacobwinch (approved):
Great work 💯

src/experimental/patterns/ec2-app.ts
akash1810: Done - https://github.com/guardian/cdk/pull/2417/commits/7a12f608297ced73a44c50b73c7312e22ffeafef.

This was a bit more involved than anticipated as we don't currently expose the ASG healthcheck grace period as a configurable property, it is hard-coded. Any [customisation is done via escape hatches](https://github.com/search?q=org%3Aguardian+.healthCheckGracePeriod+language%3ATypeScript&type=code&l=TypeScript). This means we need to use an `Aspect`.

akash1810:
[These services](https://metrics.gutools.co.uk/goto/FO8fT7RHg?orgId=1) can adopt this pattern today:

```psql
select  asg.tags ->> 'Stack' as stack
        , asg.tags ->> 'Stage' as stage
        , asg.tags ->> 'App' as app
        , asg.tags ->> 'gu:repo' as repo
        , asg.tags ->> 'gu:cdk:version' as gu_cdk_version
from    aws_autoscaling_groups asg
where   asg.arn not in (select group_arn from aws_autoscaling_group_scaling_policies)
        and asg.tags ->> 'gu:cdk:version' is not null
order by repo asc, stage asc

```

PR #2423, opened 2024-08-23
enable InstanceMetadataTags on EC2 patterns

andrew-nowak:
## What does this change?

https://github.com/guardian/instance-tag-discovery recommends enabling tags on instance metadata, but that option isn't available to users of the EC2 patterns in gucdk. This doesn't have any real problems, [aside from an error log on startup](https://logs.gutools.co.uk/goto/61f12e30-613d-11ef-9d32-0b00a291b047) that currently can't be resolved and can cause confusion.

This PR enables the option for all users of the EC2 patterns. I could have made it a configurable option, but I've neither thought up nor discovered a reason _not_ to have it enabled, so I opted to hard-code it to true. I'd be happy to relax that back to a configurable option if there's any objection though! 

## How to test

See the updated snapshots.

Error logs from `instance-tag-discovery` should fall as users adopt the new version.

Could this cause a breaking change to a gucdk consumer? I can't see how...

## Checklist

### DIFF
--- a/.changeset/cold-coins-flow.md
+++ b/.changeset/cold-coins-flow.md
@@ -1,0 +1,5 @@
+---
+"@guardian/cdk": minor
+---
+
+enable InstanceMetadataTags on EC2 patterns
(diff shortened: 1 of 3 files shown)


andrew-nowak:
^ I'd appreciate a steer on what sort of version bump this change deserves from someone more familiar with changes in gucdk

akash1810 (approved):
I spotted this this week too! Thanks for raising this!!

akash1810:
> ^ I'd appreciate a steer on what sort of version bump this change deserves from someone more familiar with changes in gucdk

I think a minor version would be best here.

PR #2438, opened 2024-09-02
vpc construct: only try setting context if unset

andrew-nowak:
## What does this change?

currently a single stack cannot create multiple vpcs, as each will try to set the context itself, and the subsequent creations will throw errors like

```
Cannot set context after children have been added: MyVpc
```

Instead, only attempt to set the context if it is unset

## How to test

Run the provided unit test

Stacks in the account setup repo can succeed even if one account attempts to create two "new VPCs".

I don't **really** understand this error, I may have missed a more correct fix for this situation.

## Checklist

### DIFF
--- a/src/constructs/vpc/vpc.ts
+++ b/src/constructs/vpc/vpc.ts
@@ -61,11 +61,11 @@ export class GuVpc extends Vpc {
       );
     }
 
-    node.setContext(`availability-zones:account=${account}:region=eu-west-1`, [
-      "eu-west-1a",
-      "eu-west-1b",
-      "eu-west-1c",
-    ]);
+    const contextKey = `availability-zones:account=${account}:region=eu-west-1`;
+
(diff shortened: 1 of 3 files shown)


src/constructs/vpc/vpc.ts
```diff
@@ -61,11 +61,11 @@ export class GuVpc extends Vpc {
       );
     }
 
-    node.setContext(`availability-zones:account=${account}:region=eu-west-1`, [
-      "eu-west-1a",
-      "eu-west-1b",
-      "eu-west-1c",
-    ]);
+    const contextKey = `availability-zones:account=${account}:region=eu-west-1`;
```
akash1810, line 64: Not for this PR necessarily, but I wonder why the region has been hard-coded here, rather than `this.region`?

akash1810 (approved):

PR #2440, opened 2024-09-04
Allow inputs in GuScheduledLambda

tomrf1:
Currently the `GuScheduledLambda` pattern lets you pass in schedules, but you cannot associate an input with each schedule.

This PR changes the props to optionally take a `RuleTargetInput` value with each scheduled rule.

### DIFF
--- a/src/patterns/scheduled-lambda.test.ts
+++ b/src/patterns/scheduled-lambda.test.ts
@@ -1,6 +1,6 @@
 import { Duration } from "aws-cdk-lib";
 import { Template } from "aws-cdk-lib/assertions";
-import { Schedule } from "aws-cdk-lib/aws-events";
+import { RuleTargetInput, Schedule } from "aws-cdk-lib/aws-events";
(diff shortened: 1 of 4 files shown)


src/patterns/scheduled-lambda.test.ts
```diff
@@ -15,7 +15,7 @@ describe("The GuScheduledLambda pattern", () => {
       functionName: "my-lambda-function",
       handler: "my-lambda/handler",
       runtime: Runtime.NODEJS_12_X,
-      rules: [{ schedule: Schedule.rate(Duration.minutes(1)) }],
+      rules: [{ schedule: Schedule.rate(Duration.minutes(1)), input: RuleTargetInput.fromText("Testing") }],
```
jamesgorrie, line 18: Would this be better in it's own test given that this test is `should create the correct resources with minimal config`?

jamesgorrie (commented):
Tiny question on if we should have a separate test?

jacobwinch (approved):
🎉

PR #2465, opened 2024-09-20
feat(experimental-ec2-pattern): Add `buildIdentifier` prop

akash1810:
## What does this change?
A requirement of using `GuEc2AppExperimental` is the filename of the application artifact should include the build identifier. Adding a required `buildIdentifier` prop to `GuEc2AppExperimental` uses the type-system to enforce the requirement.

This change also adds a `gu:build-identifier` tag to the launch template, meaning launched instances receive this tag. Furthermore, if instance logs are shipped with [devx-logs](https://github.com/guardian/devx-logs) then this tag becomes a marker on the log lines in Central ELK.

## How to test
See updated snapshot.

Improved observability.

N/A.

## Checklist

### DIFF
--- a/src/constants/metadata-keys.ts
+++ b/src/constants/metadata-keys.ts
@@ -4,4 +4,5 @@ export const MetadataKeys = {
   REPOSITORY_NAME: "gu:repo",
   LOG_KINESIS_STREAM_NAME: "LogKinesisStreamName",
   SYSTEMD_UNIT: "SystemdUnit",
+  BUILD_IDENTIFIER: "gu:build-identifier",
 };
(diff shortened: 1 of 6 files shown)


jorgeazevedo (approved):

.changeset/sixty-eggs-cheat.md
```diff
@@ -0,0 +1,5 @@
+---
+"@guardian/cdk": patch
```
akash1810, line 2: Argh, this should have been minor, as it's adding new behaviour rather than fixing existing. This change is to the experimental pattern, so hopefully this doesn't cause too much confusion!

PR #2477, opened 2024-10-15
feat: Require subnet property for ecs task, new subnetsFromParameter property

philmcmahon:
## What does this change?
I ran into an issue when working with the ecs-task role where I found it only worked if you provided a cdk context file. This was because, by default an EcsRunTask will try to assign itself to the 'PRIVATE' subnets identified by the cdk context file. This works for lurch (which was the project the Ecs task role was created for) because that project is a private repo, with a [checked in context file](https://github.com/guardian/lurch/blob/main/packages/infra/cdk.context.json#L8).

Relying on a context file isn't ideal. This PR removes the need for a context file by requiring consumers to either specify which subnets to use directly, or falling back to using `GuVpc.subnetsFromParameterFixedNumber`

But what is `GuVpc.subnetsFromParameterFixedNumber`?

Unfortunately, when I tried to use just the normal `GuVpc.subnetsFromParameter` it resulted in CDK creating a custom resource, presumably to deal with some assumed circular dependency. You can see an example of the custom resource mess [here](https://github.com/guardian/cdk/actions/runs/11348935560/job/31563876756?pr=2477).

The issue is caused by this line in GuVpc: https://github.com/guardian/cdk/blob/b61bbc6f06d05f3cebfab057e310d9b3323041f8/src/constructs/ec2/vpc.ts#L45

With some experimentation, I discovered that if you specify the subnets using .bySubnetId, and use a standard cloudformation list operation (Fn.select), then this custom resource is avoided. The thing is, I was a bit reluctant (scared) to change the existing approach to subnets, as it would affect all GuEc2Apps. The alternative approach of using Fn.select also makes for slightly messier cloudformation output - instead of looking like this:

        "Subnets": {
          "Ref": "testguec2appPublicSubnets",
        },

the subnets definition ends up looking like this:

```
{"Subnets":["",
              {
                "Fn::Select": [
                  0,
                  {
                    "Ref": "ecstestPrivateSubnets",
                  },
                ],
              },
              "","",
              {
                "Fn::Select": [
                  1,
                  {
                    "Ref": "ecstestPrivateSubnets",
                  },
                ],
              },
              "","",
              {
                "Fn::Select": [
                  2,
                  {
                    "Ref": "ecstestPrivateSubnets",
                  },
                ],
              },
              ""]
```

The new approach also assumes knowledge of how many subnets there are - whilst we usually have 3, in the tests we sometimes just have one, so more effort would be needed there.

I think this approach is a reasonable compromise, but very happy to talk it through/try to pair on a better solution.

Whilst I try and resolve that, this is a placeholder PR so I can share the work with others

### DIFF
--- a/src/constructs/ecs/ecs-task.test.ts
+++ b/src/constructs/ecs/ecs-task.test.ts
@@ -1,6 +1,6 @@
 import { Template } from "aws-cdk-lib/assertions";
-import type { ISubnet, IVpc } from "aws-cdk-lib/aws-ec2";
-import { SecurityGroup, Vpc } from "aws-cdk-lib/aws-ec2";
+import type { ISubnet, IVpc} from "aws-cdk-lib/aws-ec2";
+import { SecurityGroup, Subnet, Vpc } from "aws-cdk-lib/aws-ec2";
 import { Effect, PolicyStatement } from "aws-cdk-lib/aws-iam";
 import { GuTemplate, simpleGuStackForTesting } from "../../utils/test";
 import type { GuStack } from "../core";
@@ -81,13 +81,11 @@ describe("The GuEcsTask pattern", () => {
   it("should support overriding the subnets used by the task", () => {
     const stack = simpleGuStackForTesting();
 
-    const vpc2 = Vpc.fromVpcAttributes(stack, "VPC2", {
-      vpcId: "test",
-      availabilityZones: [""],
-      publicSubnetIds: [""],
-      privateSubnetIds: ["abc-123", "def-456"],
-    });
-    generateComplexStack(stack, "ecs-test", makeVpc(stack), vpc2.privateSubnets);
+    const subnets = [
+      Subnet.fromSubnetId(stack, "subnet-abc-123", "abc-123"),
+      Subnet.fromSubnetId(stack, "subnet-def-456", "def-456"),
+    ];
+    generateComplexStack(stack, "ecs-test", makeVpc(stack), subnets);
 
     expect(Template.fromStack(stack).toJSON()).toMatchSnapshot();
   });
--- a/src/constructs/ecs/ecs-task.test.ts
+++ b/src/constructs/ecs/ecs-task.test.ts
@@ -1,5 +1,5 @@
 import { Template } from "aws-cdk-lib/assertions";
-import type {ISubnet, IVpc} from "aws-cdk-lib/aws-ec2";
+import type { ISubnet, IVpc } from "aws-cdk-lib/aws-ec2";
 import { SecurityGroup, Vpc } from "aws-cdk-lib/aws-ec2";
 import { Effect, PolicyStatement } from "aws-cdk-lib/aws-iam";
 import { GuTemplate, simpleGuStackForTesting } from "../../utils/test";
--- a/src/constructs/ecs/ecs-task.test.ts
+++ b/src/constructs/ecs/ecs-task.test.ts
@@ -1,5 +1,5 @@
 import { Template } from "aws-cdk-lib/assertions";
-import type { ISubnet, IVpc} from "aws-cdk-lib/aws-ec2";
+import type { ISubnet, IVpc } from "aws-cdk-lib/aws-ec2";
 import { SecurityGroup, Subnet, Vpc } from "aws-cdk-lib/aws-ec2";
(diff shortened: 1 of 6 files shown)


philmcmahon:
I found that if I specify the subnets like this: https://github.com/guardian/transcription-service/blob/pm-media-download-infra/packages/cdk/lib/transcription-service.ts#L525-L557 then the extra custom resource doesn't get added

src/constructs/ecs/ecs-task.ts
```diff
@@ -188,6 +197,12 @@ export class GuEcsTask extends Construct {
       );
     }
 
+    const taskSubnets = subnets
+      ? subnets
+      : assignPublicIp
+        ? GuVpc.subnetsFromParameterFixedNumber(scope, { type: SubnetType.PUBLIC, app }, 3)
+        : GuVpc.subnetsFromParameterFixedNumber(scope, { type: SubnetType.PRIVATE, app }, 3);
+
```
akash1810, line 205: I think this nested ternary is unnecessary complexity; could we remove the `assignPublicIp` prop and pass in the correct `subnets` when instantiating this construct? The consequence is the `subnets` prop becomes mandatory, but I think it'll result in a simpler interface. For example:

```ts
new GuEcsTask(scope, "my-task", {
  subnets: GuVpc.subnetsFromParameterFixedNumber(scope, { type: SubnetType.PUBLIC, app }
});
```

philmcmahon: Good idea, thanks

src/patterns/scheduled-ecs-task.test.ts
```diff
@@ -17,13 +17,14 @@ const makeVpc = (stack: GuStack) =>
...
-
+    const vpc = makeVpc(stack);
     new GuScheduledEcsTask(stack, "test", {
       schedule: Schedule.rate(Duration.minutes(1)),
       containerConfiguration: { id: "node:10", type: "registry" },
       monitoringConfiguration: { noMonitoring: true },
-      vpc: makeVpc(stack),
+      vpc,
       app: "ecs-test",
+      subnets: vpc.privateSubnets,
```
akash1810, line 27: IIUC the snapshot tests haven't changed, and are still passing. I can't quite work out why 🤔. Does it indicate that we don't have enough test coverage? Or that we don't need this change after all?

philmcmahon: Good point. I've added a test here which I think might help a bit https://github.com/guardian/cdk/pull/2477/commits/116280e45f94163015ab34f1f8c96ef12632d838

The ecs task itself hasn't changed much - it's just that `subnets` is now a required prop - whereas previously it would be inferred from context. I think in the tests the 'inferring from context' maybe just worked and picked the private subnets, which we're now explicitly setting as the subnets

src/constructs/ecs/ecs-task.test.ts
```diff
@@ -75,4 +77,18 @@ describe("The GuEcsTask pattern", () => {
...
+  it("should support overriding the subnets used by the task", () => {
+    const stack = simpleGuStackForTesting();
+
+    const vpc2 = Vpc.fromVpcAttributes(stack, "VPC2", {
+      vpcId: "test",
+      availabilityZones: [""],
+      publicSubnetIds: [""],
+      privateSubnetIds: ["abc-123", "def-456"],
+    });
+    generateComplexStack(stack, "ecs-test", makeVpc(stack), vpc2.privateSubnets);
```
akash1810, line 90: It looks like we've references to two VPCs here: the return of `makeVpc` and `vpc2`. Is this typical usage? If not, should this test reflect real usage more closely?

akash1810 (commented):
LGTM, with a small comment on the test.

src/constructs/ecs/ecs-task.test.ts
philmcmahon: this was me being lazy and creating a whole vpc in order to get a list if ISubnet I could use to test overriding the subnets. In https://github.com/guardian/cdk/pull/2477/commits/1cfb5ac77a9c83b436fb333390746be78497e42f I do it properly, just creating some subnets to override rather than a whole new vpc

akash1810 (approved):

PR #2483, opened 2024-10-18
feat: Add exports for constructs

AshCorr:
## What does this change?

Add explicit exports to our `package.json`

I always seem to have issues with Intellisense and `@guardian/cdk`, Intellisense seems to work fine with regular `aws-cdk-lib` constructs but it can almost never find @guardian/cdk exports. From asking about it sounds like I'm not the only user affected by this.

Furthermore Deno is able to compile and synth regular `aws-cdk` projects, but the moment we include `@guardian/cdk` it starts complaining about being unable to find imports for a lot of @guardian/cdk classes.

My suspicion is because `@guardian/cdk` exports each construct as its own module instead of exporting everything through the root `index.ts` file. `aws-cdk-lib` does this too but it explicitly [specifies all its exports in its package.json](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/package.json#L209-L509)

## How to test

Copied the `package.json` to my node_modules folder.

Before:

image

After:

image

## Checklist


package.json
```diff
@@ -71,5 +71,37 @@
...
+  "exports": {
+    ".": "./lib/index.js",
+    "./lib/constants/access": "./lib/constants/access.js",
+    "./lib/constants/context-keys": "./lib/constants/context-keys.js",
+    "./lib/constants/fastly-aws-account-ids": "./lib/constants/fastly-aws-account-ids.js",
+    "./lib/constants/library-info": "./lib/constants/library-info.js",
+    "./lib/constants/metadata-keys": "./lib/constants/metadata-keys.js",
+    "./lib/constants/regex-pattern": "./lib/constants/regex-pattern.js",
+    "./lib/constants/ssm-parameter-paths": "./lib/constants/ssm-parameter-paths.js",
+    "./lib/constants/tracking-tag": "./lib/constants/tracking-tag.js",
```
AshCorr, line 84: I'm not supeerrr happy with the lack of consistency in our modules, sometimes a single file is a module like our constants, or `root.ts` and sometimes we treat an entire folder as a module.

I'd like to avoid a big refactor for the time being so I'll just make it work, but might be worth exploring in a future cleanup.

PR #2484, opened 2024-10-18
fix: Publish new version reverting 59.5.3

AshCorr:
## What does this change?

v59.5.3 was published by mistake whilst trying to publish a beta build from the `beta` branch.

Instead of unpublishing 59.5.3 which would leave a hole in our NPM versions as per https://docs.npmjs.com/policies/unpublish instead publish a new 59.5.4 version which reverts the changes that were made in 59.5.3.

Also include the latest version bumps.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # @guardian/cdk
 
+## 59.5.3
+
+**DO NOT USE**: Published by mistake.
+
 ## 59.5.2
 
 ### Patch Changes
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@guardian/cdk",
   "description": "Generic Guardian flavoured AWS CDK components",
-  "version": "59.5.2",
+  "version": "59.5.3",
   "main": "lib/index.js",
   "types": "lib/index.d.ts",
   "files": [
--- a/.changeset/wise-starfish-prove.md
+++ b/.changeset/wise-starfish-prove.md
@@ -1,0 +1,5 @@
+  ---
+  "@guardian/cdk": patch
+  ---
+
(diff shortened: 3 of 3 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@guardian/cdk",
   "description": "Generic Guardian flavoured AWS CDK components",
-  "version": "59.5.2",
+  "version": "59.5.3",
```
AshCorr, line 4: Don't let it try to publish to 59.5.3 when it generates a release.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # @guardian/cdk
 
+## 59.5.3
+
+**DO NOT USE**: Published by mistake.
+
```
AshCorr, line 6: Add a changelog entry for 59.5.3 to let people know why it exists and to not use it.

.changeset/tall-eagles-heal.md
```diff
@@ -0,0 +1,10 @@
+---
+"@guardian/cdk": patch
+---
+
+Bump dependencies"
+  - e905b31: bump typedoc from  0.26.7 to 0.26.8
+  - a334ea9: bump @types/node from 22.7.4 to 22.7.5
+  - 504fe94: bump @changesets/cli from 2.27.8 to 2.27.9
+  - 5a5498d: bump @types/node from 22.7.0 to 22.7.4
+  - 7dbd8ac: bump @types/node from 22.5.5 to 22.7.0
```
akash1810, line 10: These are all dev dependencies; I'm surprised to see them causing a version change.

AshCorr: Ah yea, good point! Removed!

akash1810 (approved):

PR #2485, opened 2024-10-18
fix: Remove contributing docs referencing beta branch

AshCorr:
## What does this change?

Removes the docs referencing the beta branch, it looks like this only worked before we updated to changesets as using different release channels is a [semantic-release](https://github.com/semantic-release/semantic-release/blob/master/docs/recipes/release-workflow/distribution-channels.md#publishing-on-distribution-channels) feature


akash1810 (approved):
Thanks! Shame we can't replicate this behaviour with changesets.

AshCorr:
> Thanks! Shame we can't replicate this behaviour with changesets.

With a bit of work we could probably achieve a similar result, just not as straightforward anymore.

PR #2486, opened 2024-10-21
Parameterise cpuArchitecture for GuEcsTask

philmcmahon:
## What does this change?
Previously we could only launch x86 ecs tasks. This resolves that.

Note - @akash1810 suggested pulling the ecs task out of GuCDK since it's only used by investigations. V happy to do this but in the short term it would be helpful to have this extra configuration property

## How to test
I've tested this by deploying to a project in the investigations account.

### DIFF
--- a/.changeset/brave-pots-act.md
+++ b/.changeset/brave-pots-act.md
@@ -1,0 +1,5 @@
+---
+"@guardian/cdk": major
+---
+
+Add cpuArchitecture property to guecstask, default it to ARM64
--- a/.changeset/brave-pots-act.md
+++ b/.changeset/brave-pots-act.md
@@ -2,4 +2,4 @@
 "@guardian/cdk": major
 ---
 
-Add cpuArchitecture property to guecstask, default it to ARM64
+Set CPU architecture of ECS tasks to ARM64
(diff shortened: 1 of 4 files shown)


.changeset/brave-pots-act.md
```diff
@@ -0,0 +1,5 @@
+---
+"@guardian/cdk": major
```
akash1810, line 2: Is this really a breaking change? Or is it a minor?

philmcmahon: It's breaking because the previous default was to use X86 rather than ARM

akash1810: Ah! Hadn't spotted that. Are there any scenarios where we'd ever use x86? I wonder if a "paved road" solution would be to not expose `cpuArchitecture` at all. Where a service not be be ARM compatible, it'll use an [escape hatch](https://docs.aws.amazon.com/cdk/v2/guide/cfn_layer.html) to change the architecture.

philmcmahon: Sorry I missed this! Sounds good, have updated

akash1810 (approved):

PR #2531, opened 2024-12-05
fix(GitHubOidcProvider): Ensure `AWS::IAM::OIDCProvider` resource is tagged

akash1810:
## What does this change?
Apply the standard `Stack`, `Stage`, `App` and `gu:repo` tags to the `AWS::IAM::OIDCProvider` resource created by the `GitHubOidcProvider` construct.

This was previously absent as we're creating `AWS::IAM::OIDCProvider` via a level 1 construct; AWS CDK only automatically tags level 2 constructs.

## How to test
See the added unit test. The test uses the [`hasGuTaggedResource`](https://github.com/guardian/cdk/blob/5f483a0abe04e4f962729f5fa43ba1a7f3afb664/src/utils/test/assertions.ts#L67) assertion helper.

Tagging resources makes it easier to identify their origin, and to allocate costs, etc.

N/A.

## Checklist

### DIFF
--- a/.changeset/nasty-parrots-mix.md
+++ b/.changeset/nasty-parrots-mix.md
@@ -1,0 +1,6 @@
+---
+"@guardian/cdk": patch
+---
+
+Apply the standard `Stack`, `Stage`, `App` and `gu:repo` tags to the `AWS::IAM::OIDCProvider` resource
+created via the `GitHubOidcProvider` construct.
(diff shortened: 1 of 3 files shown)


rtyley (approved):
Yes! Very nice!

rtyley:
> I shouldn't put my own tags on the cloudformation stack itself?

Ah yes, if I don't apply my own `stage` & `stack` tags to the Cloudformation stack when creating it in UI, everything works perfectly.

akash1810:
Confirming this works as expected in `guardian/aws-account-setup` (https://github.com/guardian/aws-account-setup/pull/265):

image

PR #2544, opened 2024-12-19
feat: mount a volume and enforce readonlyRootFilesystem in ecs-task

philmcmahon:
## What does this change?
I'm trying to resolve [[ECS.5] ECS containers should be limited to read-only access to root filesystems
](https://docs.aws.amazon.com/securityhub/latest/userguide/ecs-controls.html#ecs-5) for a number of ECS tasks in the investigations account. To do this we need to switch those tasks over to writing to a non-root volume.

Given that this is best practice, in this PR I am enforcing readonlyRootFilesystem. Of the 3 services that use this pattern, coverdrop already has readonlyRootFilesystem set to true. I'll update lurch and transcription service once this feature is out. 

This PR adds an extra parameter to ecs-task `volumeMountPath` which, when set, will mount a volume for use by the container. I opted to just support a single volume rather than making this an array or similar to keep the task simple - of the 3 services which use this pattern, all of them will be fine with just the one volume.

## How to test
I tested this change using one of the lurch CODE tasks definitions, and verified that it was able to run succesfully (writing to disk along the way) without any problems

### DIFF
--- a/.changeset/fluffy-cameras-enjoy.md
+++ b/.changeset/fluffy-cameras-enjoy.md
@@ -1,0 +1,5 @@
+---
+"@guardian/cdk": major
+---
+
+Enforce readonlyRootFilesystem in ecs-task. Support mounting a volume in ecs-task.
--- a/src/constructs/ecs/ecs-task.ts
+++ b/src/constructs/ecs/ecs-task.ts
@@ -120,6 +120,13 @@ export interface GuEcsTaskProps extends AppIdentity {
   securityGroups?: ISecurityGroup[];
   customTaskPolicies?: PolicyStatement[];
   environmentOverrides?: TaskEnvironmentVariable[];
+  /**
+   * If your container needs to write to disk whilst running, you will need to mount a non-root volume to use. Setting
+   * volumeMountPath will ensure a volume is mounted at that location, making use of Faragate ephemeral storage (set by
+   * 'storage' param)
+   *
+   */
+  volumeMountPath?: string;
   storage?: number;
   /**
    * Whether to give the task IAM role access to the account's dist bucket.
@@ -128,11 +135,6 @@ export interface GuEcsTaskProps extends AppIdentity {
    * shoud set this value to `false`.
    */
   enableDistributablePolicy?: boolean;
-  /**
-   * When this parameter is true, the container is given read-only access to its root file system.
-   * @default false
-   */
-  readonlyRootFilesystem?: boolean;
   /**
    * If `true`, CloudWatch Container Insights will be enabled for the cluster
    * @default false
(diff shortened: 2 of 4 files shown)


src/constructs/ecs/ecs-task.ts
```diff
@@ -227,9 +229,21 @@ export class GuEcsTask extends Construct {
...
+      readonlyRootFilesystem: true,
     });
 
+    if (volumeMountPath) {
+      taskDefinition.addVolume({
+        name: `${app}-volume`,
+      });
+
+      containerDefinition.addMountPoints({
+        sourceVolume: `${app}-volume`,
```
akash1810, line 241: Does this work only because the string matches `name` in `taskDefinition.addVolume` above? If so, I think we could make this more explicit via:

```ts
import type { Volume } from 'aws-cdk-lib/aws-ecs';

const volume: Volume = {
  name: `${app}-volume`,
};

taskDefinition.addVolume(volume);

containerDefinition.addMountPoints({
  sourceVolume: volume.name,
  containerPath: volumeMountPath,
  readOnly: false
});
```

src/constructs/ecs/ecs-task.ts
```diff
@@ -120,6 +120,13 @@ export interface GuEcsTaskProps extends AppIdentity {
   securityGroups?: ISecurityGroup[];
   customTaskPolicies?: PolicyStatement[];
   environmentOverrides?: TaskEnvironmentVariable[];
+  /**
+   * If your container needs to write to disk whilst running, you will need to mount a non-root volume to use. Setting
+   * volumeMountPath will ensure a volume is mounted at that location, making use of Faragate ephemeral storage (set by
+   * 'storage' param)
+   *
+   */
+  volumeMountPath?: string;
```
akash1810, line 129: I wonder if a cleaner API might be to expose the created `taskDefinition` and `containerDefinition` as properties from `GuEcsTask`:

```ts
export class GuEcsTask extends Construct {
  public readonly task: TaskDefinition; // Been lazy and made up the type here!
  public readonly container: ContainerDefinition;

  constructor() {
    super();
    const taskDefinition = new TaskDefinition(...);
    const containerDefinition = new ContainerDefinition(...);

    this.task = taskDefinition;
    this.container = containerDefinition;
  }
}
```

This approach avoids adding a new optional prop and increases flexibility as we can now call any function available on `TaskDefinition` and `ContainerDefinition`:

```ts
const { task, container } = new GuEcsTask(...);

task.addVolume({
  name: 'asd'
});

task.someOtherUsefulFunction(...);
```

.changeset/fluffy-cameras-enjoy.md
```diff
@@ -0,0 +1,5 @@
+---
+"@guardian/cdk": major
+---
+
+Enforce readonlyRootFilesystem in ecs-task. Support mounting a volume in ecs-task.
```
akash1810, line 5: Should we more descriptive here and offer an example implementation?

src/constructs/ecs/ecs-task.ts
philmcmahon: Oh that's a much better approach!

philmcmahon:
Closing in favour of https://github.com/guardian/cdk/pull/2545

PR #2545, opened 2024-12-20
feat: Expose components of ecs-task, enforce read only root filesystem

philmcmahon:
## What does this change?
Replaces https://github.com/guardian/cdk/pull/2544 following a great suggestion from @akash1810. 

This PR enforces readonlyRootFilesystem. This change may require apps that use this pattern to mount a volume onto the container to use for writing application data to disk. Exposing the container definition and task definition makes this possible. It should also make it easier to work with this pattern without having to keep making changes to the core GuCDK library.

## How to test
I've tested this in Lurch by referencing this branch


.changeset/wet-pillows-judge.md
```diff
@@ -0,0 +1,5 @@
+---
+"@guardian/cdk": minor
+---
+
+Expose taskDefinition, containerDefinition and task in ecsruntask
```
akash1810, line 5:
```suggestion
Expose taskDefinition, containerDefinition and task in `GuEcsTask`
```

akash1810 (approved):

PR #2567, opened 2025-01-24
Aspect to enforce unique app/region/stack on lambdas

Jakeii:
## What does this change?
Introduce an `Aspect` to enforce unique combinations of app/region/stack on `GuLambdaFunction`.

A few projects at the Guardian seem to be mistakenly doing this, and when used in combination with a generated `riff-raff.yaml` can cause unexpected behaviour where only the last lambda defined will make it into the the deployments as the app/region/stack combination is used for the key.

I ran into some circular dependency issues, so had to make some imports more specific in some other modules to resolve.


src/patterns/api-multiple-lambdas.test.ts
```diff
@@ -58,6 +61,61 @@ describe("The GuApiGatewayWithLambdaByPath pattern", () => {
...
+        {
+          path: "/test/api-key",
+          httpMethod: "GET",
+          lambda: lambdaFour,
+          apiKeyRequired: true,
+        },
+      ],
+    });
+    expect(() => Template.fromStack(stack)).toThrow(
+      "GuLambdaFunction must have a unique combination of app, region and stack",
```
akash1810, line 115: The message above also includes "Found duplicate: ${node.app}". Any reason to not capture that here?

Jakeii: I guess not!

PR #2574, opened 2025-02-03
feat!(vpc-import): Reimplement reference to a pre-existing VPC

akash1810:
> [!NOTE]
> Need to consider the impact on these projects:
> - [Instances of passing `publicSubnets` to `GuEc2App`](https://github.com/search?q=org%3Aguardian+publicSubnets+language%3ATypeScript++NOT+is%3Aarchived+NOT+repo%3Aguardian%2Fcdk+NOT+repo%3Aguardian%2Faws-account-setup+&type=code)
> - [Instances of passing `privateSubnets` to `GuEc2App`](https://github.com/search?q=org%3Aguardian+privateSubnets+language%3ATypeScript++NOT+is%3Aarchived+NOT+repo%3Aguardian%2Fcdk+NOT+repo%3Aguardian%2Faws-account-setup+&type=code)
> - [Uses of `GuVpc.fromIdParameter`](https://github.com/search?q=org%3Aguardian+fromIdParameter+language%3ATypeScript++NOT+is%3Aarchived+NOT+repo%3Aguardian%2Fcdk+NOT+repo%3Aguardian%2Faws-account-setup+&type=code)
> - [Uses of `GuVpc.subnetsFromParameterFixedNumber`](https://github.com/search?q=org%3Aguardian+subnetsFromParameterFixedNumber+language%3ATypeScript++NOT+is%3Aarchived+NOT+repo%3Aguardian%2Fcdk+NOT+repo%3Aguardian%2Faws-account-setup+&type=code)

## What does this change?
The current implementation of importing a VPC, via SSM Parameters, for use in an application makes reference to the [AWS CDK method](https://github.com/aws/aws-cdk/blob/86ce1550abf9809ca9c16a4842b7445f0cce34e3/packages/aws-cdk-lib/core/lib/stack.ts#L829-L866) `scope.availabilityZones`. AWS CDK comments about this:

>  Returns the list of AZs that are available in the AWS environment (account/region) associated with this stack.
>  
>  If the stack is environment-agnostic (either account and/or region are tokens), this property will return an array with 2 tokens that will resolve at deploy-time to the first two availability zones returned from CloudFormation's `Fn::GetAZs` intrinsic function.
>  
>  If they are not available in the context, returns a set of dummy values and reports them as missing, and let the CLI resolve them by calling EC2 `DescribeAvailabilityZones` on the target environment.

That is, AWS CDK attempts to dynamically resolve the AZs by interrogating the account and storing values in the context file. If we don't configure AWS CDK with AWS credentials, attempts to reference a subnet results in the error:

> Found an encoded list token string in a scalar string context. Use 'Fn.select(0, list)' (not 'list[0]') to extract elements from token lists.

This is very confusing!

We don't need to call `scope.availabilityZones` as we usually do not explicitly specify the AZ to place our resources. These changes reimplement how we reference a pre-existing VPC. The following have been replaced with `GuVpcImport.fromSsmParameters`:
- `GuVpc.fromIdParameter`
- `GuVpc.subnetsFromParameter`

For the most part, this should not impact any stacks as GuCDK's patterns have been updated. It is a breaking change, however, as this change includes a change to the CloudFormation parameter names.
## How to test
See added and updated unit tests.

I've also tested these changes in https://github.com/guardian/cdk-playground/pull/629.

More intuitive DX, with a reduction in confusing errors.

We'd need to check the compatibility of these changes with those in https://github.com/guardian/cdk/pull/1566 and https://github.com/guardian/cdk/pull/2477.

## Checklist


.changeset/kind-dots-dress.md
```diff
@@ -0,0 +1,80 @@
...
+2. Reduce the need to add values to `cdk.context.json`
+
+For most applications, these changes will be minimal with a change to the name of CloudFormation (CFN) parameters.
+
+### CloudFormation Parameter changes
+
+Previously, the CFN parameters were prefixed with the `app` being deployed.
+For example, with an `app` called `api`, GuCDK added the following CFN parameters:
+- `apiPrivateSubnets` - referencing an SSM Parameter holding the private subnets
+- `apiPublicSubnets` - referencing an SSM Parameter holding the public subnets
```
akash1810, line 23: This behaviour was introduced in https://github.com/guardian/cdk/pull/399. I don't think it makes sense to have these parameters be `app` aware. The `VpcId` parameter is not `app` aware, that is, every app in a CFN stack uses the same VPC. We also don't have a subnet per `app`.

PR #2578, opened 2025-02-07
Add `GuKCLPolicy`, a IAM-policy construct for the Kinesis Client Library (KCL)

rtyley:
The new `GuKCLPolicy` construct creates an `AWS::IAM::Policy` to grant all the required permissions for the [Kinesis Client Library](https://github.com/awslabs/amazon-kinesis-client) - specifically for KCL **v3**, which requires [many additional permissions](https://docs.aws.amazon.com/streams/latest/dev/kcl-iam-permissions.html) over KCL v2 (largely because KCL 3.x introduced **two** new DynamoDB tables for metadata: [worker metrics](https://docs.aws.amazon.com/streams/latest/dev/kcl-dynamoDB.html#kcl-worker-metrics-table) & [coordinator state](https://docs.aws.amazon.com/streams/latest/dev/kcl-dynamoDB.html#kcl-coordinator-state-table)).

## Example of use

This PR is an example of swapping out existing KCL v2 policy code to use the new `GuKCLPolicy` construct:

* https://github.com/guardian/apple-news/pull/426

## Effective IAM permissions

The key code in `GuKCLPolicy` _defining_ the required policy statements is this: 

https://github.com/guardian/cdk/blob/ea3e23f87f817307549ed2169833f1473794c35f/src/constructs/iam/policies/kcl.ts#L56-L61

...we can compare this with the full cloudformation example taken from [AWS's documentation](https://docs.aws.amazon.com/streams/latest/dev/kcl-iam-permissions.html) (converted to YAML):

```yaml
- Effect: Allow
  Action:
    - kinesis:DescribeStream
    - kinesis:DescribeStreamSummary
    - kinesis:RegisterStreamConsumer
    - kinesis:GetRecords
    - kinesis:GetShardIterator
    - kinesis:ListShards
  Resource: arn:aws:kinesis:REGION:ACCOUNT_ID:stream/STREAM_NAME
- Effect: Allow
  Action:
    - kinesis:SubscribeToShard
    - kinesis:DescribeStreamConsumer
  Resource: arn:aws:kinesis:REGION:ACCOUNT_ID:stream/STREAM_NAME/consumer/*
- Effect: Allow
  Action:
    - dynamodb:CreateTable
    - dynamodb:DescribeTable
    - dynamodb:UpdateTable
    - dynamodb:GetItem
    - dynamodb:UpdateItem
    - dynamodb:PutItem
    - dynamodb:DeleteItem
    - dynamodb:Scan
  Resource:
    - arn:aws:dynamodb:REGION:ACCOUNT_ID:table/KCL_APPLICATION_NAME
- Effect: Allow
  Action:
    - dynamodb:CreateTable
    - dynamodb:DescribeTable
    - dynamodb:GetItem
    - dynamodb:UpdateItem
    - dynamodb:PutItem
    - dynamodb:DeleteItem
    - dynamodb:Scan
  Resource:
    - arn:aws:dynamodb:REGION:ACCOUNT_ID:table/KCL_APPLICATION_NAME-WorkerMetricStats
    - arn:aws:dynamodb:REGION:ACCOUNT_ID:table/KCL_APPLICATION_NAME-CoordinatorState
- Effect: Allow
  Action:
    - dynamodb:Query
  Resource:
    - arn:aws:dynamodb:REGION:ACCOUNT_ID:table/KCL_APPLICATION_NAME/index/*
- Effect: Allow
  Action:
    - cloudwatch:PutMetricData
  Resource: '*'
```

## See also

* https://github.com/guardian/content-api-firehose-client/pull/56
* Examples of previous PRs adding an IAM-policy construct:
  * https://github.com/guardian/cdk/pull/139
  * https://github.com/guardian/cdk/pull/278

### DIFF
--- a/.changeset/neat-roses-smell.md
+++ b/.changeset/neat-roses-smell.md
@@ -1,0 +1,5 @@
+---
+"@guardian/cdk": major
+---
+
+Add GuKCLPolicy, a IAM-policy construct for the Kinesis Client Library
--- a/.changeset/neat-roses-smell.md
+++ b/.changeset/neat-roses-smell.md
@@ -1,5 +1,5 @@
 ---
-"@guardian/cdk": major
+"@guardian/cdk": minor
 ---
 
 Add GuKCLPolicy, a IAM-policy construct for the Kinesis Client Library
--- a/src/constructs/iam/policies/kcl.ts
+++ b/src/constructs/iam/policies/kcl.ts
@@ -18,7 +18,13 @@ const kinesisActions: string[] = [
 
 const kinesisEnhancedFanOutActions: string[] = ["SubscribeToShard", "DescribeStreamConsumer"];
 
-const actionsOnAllTables: string[] = ["Scan", "CreateTable", "DescribeTable", "GetItem", "PutItem"];
+const actionsOnAllTables: string[] = [
+  "Scan",
+  "CreateTable", // Allow KCL creating the lease, metrics, and coordinator tables - see https://github.com/guardian/cdk/pull/2578#discussion_r1954061397
+  "DescribeTable",
+  "GetItem",
+  "PutItem",
+];
 
 const additionalLeaseTableActions: string[] = ["UpdateTable", "UpdateItem", "DeleteItem"];
 
--- a/src/constructs/iam/policies/kcl.ts
+++ b/src/constructs/iam/policies/kcl.ts
@@ -1,0 +1,65 @@
+import { Effect, PolicyStatement } from "aws-cdk-lib/aws-iam";
+import type { GuStack } from "../../core";
+import { GuPolicy } from "./base-policy";
+
+export interface KCLApplication {
+  streamName: string;
+  applicationName: string;
+}
+
+const kinesisActions: string[] = [
+  "DescribeStream",
+  "DescribeStreamSummary",
+  "RegisterStreamConsumer",
+  "GetRecords",
+  "GetShardIterator",
+  "ListShards",
+];
+
+const kinesisEnhancedFanOutActions: string[] = ["SubscribeToShard", "DescribeStreamConsumer"];
+
+const actionsOnAllTables: string[] = ["Scan", "CreateTable", "DescribeTable", "GetItem", "PutItem"];
+
+const additionalLeaseTableActions: string[] = ["UpdateTable", "UpdateItem", "DeleteItem"];
+
+/**
+ * Creates an `AWS::IAM::Policy` to grant all the required permissions for the Kinesis Client Library
+ * (https://github.com/awslabs/amazon-kinesis-client) - specifically for KCL v3, which requires many
+ * additional permissions over KCL v2.
+ *
+ * @see https://docs.aws.amazon.com/streams/latest/dev/kcl-iam-permissions.html
+ */
+export class GuKCLPolicy extends GuPolicy {
+  constructor(scope: GuStack, props: KCLApplication) {
+    function allow(actionType: string, actions: string[], resources: string[]): PolicyStatement {
+      return new PolicyStatement({
+        effect: Effect.ALLOW,
+        actions: actions.map((a) => `${actionType}:${a}`),
+        resources: resources.map((r) => `arn:aws:${actionType}:${scope.region}:${scope.account}:${r}`),
+      });
+    }
+
+    function allowDynamoDB(actions: string[], tableOrIndexNames: string[]): PolicyStatement {
+      return allow(
+        "dynamodb",
+        actions,
+        tableOrIndexNames.map((name) => `table/${name}`),
+      );
+    }
+
+    const leaseTable = props.applicationName;
+    const metricsTable = `${props.applicationName}-WorkerMetricStats`;
+    const coordinatorTable = `${props.applicationName}-CoordinatorState`;
+
+    super(scope, "GuKCLPolicy", {
+      statements: [
+        allow("kinesis", kinesisActions, [`stream/${props.streamName}`]),
+        allow("kinesis", kinesisEnhancedFanOutActions, [`stream/${props.streamName}/consumer/*`]),
+        allowDynamoDB(actionsOnAllTables, [leaseTable, metricsTable, coordinatorTable]),
+        allowDynamoDB(additionalLeaseTableActions, [leaseTable]),
+        allowDynamoDB(["Query"], [`${leaseTable}/index/*`]),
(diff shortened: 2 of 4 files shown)


.changeset/neat-roses-smell.md
```diff
@@ -0,0 +1,5 @@
+---
+"@guardian/cdk": major
```
akash1810, line 2: Not sure if this is a breaking change?

src/constructs/iam/policies/kcl.ts
```diff
@@ -0,0 +1,65 @@
...
+  "DescribeStreamSummary",
+  "RegisterStreamConsumer",
+  "GetRecords",
+  "GetShardIterator",
+  "ListShards",
+];
+
+const kinesisEnhancedFanOutActions: string[] = ["SubscribeToShard", "DescribeStreamConsumer"];
+
+const actionsOnAllTables: string[] = ["Scan", "CreateTable", "DescribeTable", "GetItem", "PutItem"];
```
akash1810, line 21: IIRC v2 of the KCL would create the dynamodb table by default, but it was also happy to "find" it, allowing us to create it via IaC. It looks like we know the names of the tables; should we create them ourselves and by extension apply tags to them?

.changeset/neat-roses-smell.md
rtyley: Ah, fair play! I guess all this is doing is adding a new class, and this won't break any existing code. So I guess this is `minor`, according to the rules of https://semver.org/

src/constructs/iam/policies/kcl.ts
rtyley: > should we create them ourselves and by extension apply tags to them?

Hmm, to create the tables we'd also need to know the DynamoDB primary key & global secondary index structures - by contrast with the IAM permissions and table names, which are explicitly documented, the primary keys & secondary index structure are _not_ - they feel like internal details which AWS are not encouraging us to depend on.

Obviously we could work them out if we wanted to, allowing us to create-&-tag them, but if so, I think it would be reasonable to do that in a subsequent PR!

src/constructs/iam/policies/kcl.ts
```diff
@@ -0,0 +1,65 @@
...
+        actions,
+        tableOrIndexNames.map((name) => `table/${name}`),
+      );
+    }
+
+    const leaseTable = props.applicationName;
+    const metricsTable = `${props.applicationName}-WorkerMetricStats`;
+    const coordinatorTable = `${props.applicationName}-CoordinatorState`;
+
+    super(scope, "GuKCLPolicy", {
```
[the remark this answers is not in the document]
rtyley: Ah, great spot - addressed in 5ab19bc6e49af3413b84f3a3b1da4cf51eb81dfc.

src/constructs/iam/policies/kcl.ts
akash1810: Fair. Wonder if its worth having an inline comment here to explain the `CreateTable` permission?

src/constructs/iam/policies/kcl.ts
```diff
@@ -0,0 +1,65 @@
+import { Effect, PolicyStatement } from "aws-cdk-lib/aws-iam";
+import type { GuStack } from "../../core";
+import { GuPolicy } from "./base-policy";
+
+export interface KCLApplication {
+  streamName: string;
```
akash1810, line 6: minor: Could this be a [`Stream`](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_kinesis.Stream.html) for stricter typing? If we're referencing a pre-existing stream, we could use the [`fromStreamArn` static function](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_kinesis.Stream.html#static-fromwbrstreamwbrarnscope-id-streamarn) to create this type.

akash1810 (approved):
LGTM.

Some additional, but non-blocking, comments.

src/constructs/iam/policies/kcl.ts
rtyley: I do love things being strongly-typed!.. Does it work with the magic around SSM parameters though? In situations like `apple-news` consuming the CAPI `crier` stream, [the ARN of the Crier stream is read from SSM](https://github.com/guardian/apple-news/blob/77c416ef32aa41dafe1beaa2d4288a4461e222ae/cdk/lib/apple-news.ts#L75-L102) - could we pass the _SSM parameter_ to [fromStreamArn()](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_kinesis.Stream.html#static-fromwbrstreamwbrarnscope-id-streamarn) - I'm not sure if that would work?!

src/constructs/iam/policies/kcl.ts
rtyley: > Wonder if its worth having an inline comment here to explain the `CreateTable` permission?

👍  - done with 2bd57a563e22e9caa118b4171ed9194dae57bbcf!

src/constructs/iam/policies/kcl.ts
akash1810: It can work, but is a bit longer and maybe less intuitive?

```ts
const streamArnParam = new CfnParameter(scope, "Stream", {
  type: "AWS::SSM:;Parameter::Value<string>",
  default: "/a/path/to/ssm/parameter",
});

const streamArn = streamArnParam.valueAsString; // This is a `Token` at compile time. The compiler treats it as a `string`. See also https://docs.aws.amazon.com/cdk/v2/guide/tokens.html.

const stream = Stream.fromStreamArn(streamArnParam.valueAsString); // Strong type!
```

rtyley: I did try this out, with a couple of test branches:

* https://github.com/guardian/cdk/pull/2585
* https://github.com/guardian/apple-news/pull/428 - couldn't get it to work here, for some reason

> a bit longer and maybe less intuitive?

Yeah, that's a shame - the `Stream.fromStreamArn()` method ends up taking 3 arguments too (scope, id, streamArn), so it ends up being quite verbose to just create the `IStream`:

```ts
    const streamName: string = "streamFoo";
    const policy = new GuKCLPolicy(stack, {
      stream: Stream.fromStreamArn(
        stack,
        streamName,
        `arn:aws:kinesis:${stack.region}:${stack.account}:stream/${streamName}`,
      ),
      applicationName: "appBar",
    });
```

..in the end, given it adds a fair bit of lines-of-code overhead, I think I'll probably leave it 😞

PR #2623, opened 2025-04-07
chore: Specify minimum `aws-cdk` version in peerDependencies instead of a specific version

AshCorr:
## What does this change?

We often get Dependabot PRs on our repos that try to update our version of `aws-cdk` before a new version of `@guardian/cdk` has been released, its quite annoying to have to wait till a new version of `@guardian/cdk` is released just to make a minor version update. Particularly as we group Dependabot PRs together causing us to have to manually remove the `aws-cdk` bumps from the group PR.

---

This pull request specifies a minimum `aws-cdk` version in `peerDependencies` instead of tagging a specific version. This is as per AWS's own recommendation [here](https://www.npmjs.com/package/aws-cdk-lib#installation):

> To make sure your library is compatible with the widest range of CDK versions: pick the minimum aws-cdk-lib version that your library requires; declare a range dependency with a caret on that version in peerDependencies, and declare a point version dependency on that version in devDependencies.

In practice we use Snapshot testing in most of our CDK projects meaning any compatability issues should be flagged by either failing snapshots or failing CI.

### DIFF
--- a/script/ci-project-generation
+++ b/script/ci-project-generation
@@ -22,12 +22,10 @@ npm pack --pack-destination $TMP_DIR
     exit 1
   fi
 
-  npm install guardian-cdk-*.tgz
-
   # Ensure the CLI runs - the exit code should be 0.
-  npx gu-cdk --version
+  npx --yes guardian-cdk-*.tgz --version
 
-  NODE_ENV=test npx gu-cdk new \
+  NODE_ENV=test npx --yes guardian-cdk-*.tgz new \
     --app integration-test \
     --stack cdk \
     --stage CODE \
--- a/.changeset/gold-owls-give.md
+++ b/.changeset/gold-owls-give.md
@@ -1,0 +1,5 @@
+---
+"@guardian/cdk": patch
+---
+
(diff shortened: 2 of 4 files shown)


script/ci-project-generation
```diff
@@ -22,12 +22,10 @@ npm pack --pack-destination $TMP_DIR
     exit 1
   fi
 
-  npm install guardian-cdk-*.tgz
```
AshCorr, line 25: When we previously ran `npm install` the test project folder structure would end up looking like:

```bash
- node_modules
  - @guardian/cdk
  - aws-cdk v2.1007
  ...
- cdk
 - node_modules
  - aws-cdk v2.1005
```

This made Node's import resolution a bit funky as even though the test project didn't include `@guardian/cdk` as a dependency it would be able to resolve it thanks to it being in the parent `node_modules` directory. 

This was helpful for testing as it meant that we didn't have to `npm install` the `@guardian/cdk` dependency in the test project, but it did have the side effect of making the parent `node_modules` folder include different set of dependencies than the test project. As we no longer enforce identical aws-sdk versions in our peer deps it meant that whenever `@guardian/cdk` imported `aws-cdk` it might get a different version of `aws-cdk` then what would be imported by the test project

PR #2624, opened 2025-04-08
feat: Bump `eslint` to `9.19.0`

AshCorr:
## Checklist


.eslintrc.js
AshCorr, line 1: `.eslintrc.js` has been replaced with `eslint.config.mjs` in ESLint 9 with a considerably different format. 

https://eslint.org/docs/latest/use/configure/migration-guide

package.json
```diff
@@ -13,7 +13,7 @@
   "repository": "github:guardian/cdk",
   "scripts": {
     "build": "tsc",
-    "lint": "eslint src --ext .ts",
```
AshCorr, line 16: Linting `.ts` files is enabled by `@guardian/eslint-config` now.

.changeset/clever-apes-hammer.md
```diff
@@ -0,0 +1,49 @@
+---
+"@guardian/cdk": major
```
akash1810, line 2: I'm not sure this is a breaking change to the GuCDK library as its mostly focused on the internal workings/configuration of it. Might a patch release suffice?

Regarding the upgrade guide - these notes are great! I wonder they're generic enough to form the release notes of `@guardian/eslint-config`?

akash1810 (commented):
One blocking comment on the changeset, the other changes look great!

.changeset/clever-apes-hammer.md
AshCorr: > Might a patch release suffice?

I suppose I was thinking that if you wanted to upgrade your version of `@guardian/cdk` you would have to also upgrade `eslint`, but I don't think thats quite true as `eslint` is a dev dependency... Yea this could probably be a patch version.

> I wonder they're generic enough to form the release notes of @guardian/eslint-config?

Most CDK projects have the same if not similar eslint configs making them somewhat easy to automate, don't think this script would work well in other places as they might be using [one of 6 different config locations](https://eslint.org/docs/latest/use/configure/configuration-files-deprecated), or have non-default ESLint rules or plugins enabled!

I'd probably suggest using ESLints own guide for migrating other projects which includes a helper script for migrating config files to the new format: https://eslint.org/docs/latest/use/configure/configuration-files-deprecated

akash1810 (approved):

PR #2629, opened 2025-04-09
feat: add rolling update support for MAPI

jacobwinch:
Refactor new deployment functionality (introduced in https://github.com/guardian/cdk/pull/2417) to allow it to be used outside of the experimental pattern.

This is useful for MAPI, which uses GuCDK constructs but cannot use the `GuEC2App` pattern (or the closely related `GuEC2AppExperimental` pattern) because it has multiple target groups[^1].

There's a demonstration of how this could be utilised for `mobile-sport` [here](https://github.com/guardian/mobile-apps-api/pull/3610/files#diff-4d018da4e363cfcecdb3fbe02e21f5818cf4826c27c6df7d0269377981071503). Note that most of the changes in https://github.com/guardian/mobile-apps-api/pull/3610 would be required regardless of whether we take this approach or implement a new experimental pattern for the multi-target group architecture, as there is quite a bit of wiring involved in moving MAPI to versioned deployments.

[^1]: The aforementioned patterns support the common case i.e. a single target group.

### DIFF
--- a/src/experimental/patterns/ec2-app.ts
+++ b/src/experimental/patterns/ec2-app.ts
@@ -7,6 +7,7 @@ import type { IConstruct } from "constructs";
 import { MetadataKeys } from "../../constants";
 import { GuAutoScalingGroup } from "../../constructs/autoscaling";
 import type { GuStack } from "../../constructs/core";
+import type { GuApplicationTargetGroup } from "../../constructs/loadbalancing";
 import type { GuEc2AppProps } from "../../patterns";
 import { GuEc2App } from "../../patterns";
 import { isSingletonPresentInStack } from "../../utils/singleton";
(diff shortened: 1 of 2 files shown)


src/experimental/patterns/ec2-app.ts
```diff
@@ -253,6 +253,90 @@ export interface GuEc2AppExperimentalProps extends Omit<GuEc2AppProps, "updatePo
...
+    For this reason, we suspend the `AlarmNotification` process, else availability of a service cannot be guaranteed.
+    Consequently, services cannot scale-out during deployments.
+    If AWS ever supports suspending scale-out and scale-in independently, we should allow scale-out.
+     */
+    suspendProcesses: [ScalingProcess.ALARM_NOTIFICATION],
+    // Note: there is also an important property called pauseTime, but this is set via an Aspect.
+  });
+}
+
+export function modifyRoleForRollingUpdate(scope: GuStack, autoScalingGroup: GuAutoScalingGroup) {
```
akash1810, line 273: Could we convert this to a class so that it's picked up by the [metadata aspect](https://github.com/guardian/cdk/blob/main/src/aspects/metadata.ts), allowing us to track usage in case we need to make breaking changes in future?

Similar to `modifyUserDataForRollingUpdate` below too.

.changeset/slow-bats-slide.md
```diff
@@ -0,0 +1,5 @@
+---
+"@guardian/cdk": minor
+---
+
+Addition of new deployment mechanism support for MAPI.
```
akash1810, line 5:
```suggestion
Addition of new experimental deployment mechanism support for MAPI.
```

Just trying to note that usage brings risk (more from it being unused in production thus far).

src/experimental/patterns/ec2-app.ts
```diff
@@ -41,17 +42,17 @@ export const RollingUpdateDurations: AutoScalingRollingUpdateDurations = {
  *
  * TODO Expose the healthcheck grace period as a property on {@link GuEc2App} and remove this `Aspect`.
  */
-class AutoScalingRollingUpdateTimeout implements IAspect {
+export class AutoScalingRollingUpdateTimeoutExperimental implements IAspect {
```
akash1810, line 45: I think this, and other exported classes would need to be [prefixed `Gu` to be picked up by the metadata aspect](https://github.com/guardian/cdk/blob/e416a34f2dbe454169e8351622640a5074286c11/src/aspects/metadata.ts#L27).

PR #2638, opened 2025-04-16
feat: Remove `withoutImdsv2` property from `GuEc2App` and `GuAutoScalingGroup`

akash1810:
## What does this change?
When this property was set to `true`, launched instances would not meet the [FSBP EC2.8 control](https://docs.aws.amazon.com/securityhub/latest/userguide/ec2-controls.html#ec2-8). Removing this property reinforces that GuCDK provides a paved path, following best practice by default.

If for whatever reason one needs to disable IMDSv2, the escape hatch patten can be used:

```typescript
import { CfnLaunchTemplate } from "aws-cdk-lib/aws-ec2";

declare const asg: GuAutoScalingGroup;

// Set the value to "optional", allowing IMDSv1 and IMDSv2.
// See https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-metadataoptions.html#cfn-ec2-launchtemplate-metadataoptions-httptokens.
launchTemplate.addPropertyOverride("LaunchTemplateData.MetadataOptions.HttpTokens", "optional");

// Or remove the property entirely.
launchTemplate.addPropertyDeletionOverride("LaunchTemplateData.MetadataOptions.HttpTokens");
```

## How to test
N/A.

Reinforcement of the idea that GuCDK provides a paved path.

As this change removes an optional property, and doesn't change the default, I don't think it is a breaking change. That said, the [services that make use of the `withoutImdsv2` property](https://github.com/search?q=org%3Aguardian+withoutImdsv2++NOT+is%3Aarchived+NOT+repo%3Aguardian%2Fcdk&type=code) would need to upgrade carefully, ensuring to implement the escape hatch.

## Checklist

### DIFF
--- a/.changeset/rare-plums-fix.md
+++ b/.changeset/rare-plums-fix.md
@@ -1,0 +1,25 @@
+---
+"@guardian/cdk": minor
+---
+
+Removal of the `withoutImdsv2` property from `GuEc2App` and `GuAutoScalingGroup`.
+When this property was set to `true`, launched instances would not meet the [FSBP EC2.8 control](https://docs.aws.amazon.com/securityhub/latest/userguide/ec2-controls.html#ec2-8).
+
+Removing this property as a signal that GuCDK will follow FSBP controls by default.
+
+If for whatever reason you need to disable IMDSv2, you can do so via an [escape hatch](https://docs.aws.amazon.com/cdk/v2/guide/cfn_layer.html):
+
+```typescript
+import { CfnLaunchTemplate } from "aws-cdk-lib/aws-ec2";
+
+declare const asg: GuAutoScalingGroup;
+
+const launchTemplate = asg.instanceLaunchTemplate.node.defaultChild as CfnLaunchTemplate;
+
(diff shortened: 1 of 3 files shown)


jacobwinch:
> As this change removes an optional property, and doesn't change the default, I don't think it is a breaking change. That said, the [services that make use of the `withoutImdsv2` property](https://github.com/search?q=org%3Aguardian+withoutImdsv2++NOT+is%3Aarchived+NOT+repo%3Aguardian%2Fcdk&type=code) would need to upgrade carefully, ensuring to implement the escape hatch.

I think we should treat this as a breaking change because it's possible that things will break if teams update to IMDSv2 without noticing (for `editorial-wires` it [looks like a 3rd party might be relying on the current setting](https://github.com/guardian/editorial-wires/pull/22)).

That said, there is only one service affected and this aligns GuCDK with FSBP, so I think we should go ahead with this.

jacobwinch (approved):

PR #2644, opened 2025-04-30
feat(GuEc2App): Replace `enabledDetailedInstanceMonitoring` optional property with mandatory `instanceMetricGranularity` property

akash1810:
## What does this change?
Replaces the `enabledDetailedInstanceMonitoring` optional property from the `GuEc2App` pattern with a mandatory `instanceMetricGranularity` property. There are two values: `1Minute`, `5Minute`. By default, AWS produces metrics every 5 minutes.

Adding a mandatory property aims to encourage teams to enable granular metrics in production as it allows for earlier horizontal scaling and provides more detail during incident triage.

> [!NOTE]
> - This change will cost roughly $3 per instance per month. We might want to consider disabling it for non-production environments.
> - These are not the same as metrics from the [CloudWatch Agent](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/metrics-collected-by-CloudWatch-agent.html), which services might also be using.

See also:
- https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/manage-detailed-monitoring.html
- https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/viewing_metrics_with_cloudwatch.html

## How to test
See updated tests.

Our services can horizontally scale earlier.

This has a direct cost implication. At the time of writing, we have [~630 running instances](https://metrics.gutools.co.uk/goto/zuzCf7bHg?orgId=1) across all accounts. At ~$3 per instance per month, this'll be a total of ~$23,000[^1]. If we only enable this on [production instances](https://metrics.gutools.co.uk/goto/Pojuf7bNR?orgId=1), this reduces to ~$15k.

## Checklist

- [ ] I have listed any breaking changes, along with a migration path [^2]
- [ ] I have updated the documentation as required for the described changes [^3]

[^1]: 630 * 3 * 12

### DIFF
--- a/src/constructs/autoscaling/asg.ts
+++ b/src/constructs/autoscaling/asg.ts
@@ -13,6 +13,8 @@ import type { AppIdentity, GuStack } from "../core";
 import { GuHttpsEgressSecurityGroup } from "../ec2";
 import { GuInstanceRole } from "../iam";
 
+export type InstanceMetricGranularity = "1Minute" | "5Minute";
+
 // Since we want to override the types of what gets passed in for the below props,
 // we need to use Omit<T, U> to remove them from the interface this extends
 // https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys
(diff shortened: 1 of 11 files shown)


mchv:
I am supportive of the additional cost for `PROD` instance, and so would like to be added as default on `PROD` if that possible.

markjamesbutler:
@akash1810 - I took this out of review. Best of luck with the deployment.

akash1810:
> I am supportive of the additional cost for `PROD` instance, and so would like to be added as default on `PROD` if that possible.

GuCDK is intentionally stage agnostic ([ADR here](https://github.com/guardian/cdk/blob/main/docs/architecture-decision-records/006-stage-awareness.md)), therefore we cannot implement the "default on PROD" behaviour unfortunately. In an attempt to provide an accessible API and DX, I've created a new mandatory property. Usage could look something like this:

```ts
const stage = props.stage;
const instanceMetricGranularity = stage === "PROD" ? "1Minute" ? "5Minute";

const myApp = new GuEc2App(this, "my-app", {
  instanceMetricGranularity,
});
```

src/constructs/autoscaling/asg.ts
```diff
@@ -47,10 +49,21 @@ export interface GuAutoScalingGroupProps
   imageRecipe?: string | AmigoProps;
   instanceType: InstanceType;
   userData: UserData;
+
+  /**
+   * How often to send produce EC2 metrics, such as CPU usage.
```
jacobwinch, line 54:
```suggestion
   * How often to send EC2 metrics, such as CPU usage.
```

PR #2561, opened 2025-01-15
feat: Remove `GuWazuhAccess` class

akash1810:
## What does this change?
Remove the `GuWazuhAccess` class which adds a security group of logical ID `WazuhSecurityGroup` to any stack using `GuAutoScalingGroup` either directly or via a pattern. This follows https://github.com/guardian/amigo/pull/1601 which removed Wazuh from AMIgo.

## How to test
- See the updated snapshots
- See https://github.com/guardian/cdk-playground/pull/688 for a practical test

## How can we measure success?
Closing of unused ports.

## Have we considered potential risks?
N/A.

## Checklist
- [ ] I have listed any breaking changes, along with a migration path [^1]
- [ ] I have updated the documentation as required for the described changes [^2]

[^1]: Consider whether this is something that will mean changes to projects that have already been migrated, or to the CDK CLI tool. If changes are required, consider adding a checklist here and/or linking to related PRs.
[^2]: If you are adding a new construct or pattern, has new documentation been added? If you are amending defaults or changing behaviour, are the existing docs still valid?

PR #2647, opened 2025-04-30
docs: Add detailed notes for v61.5.0

akash1810:
## What does this change?
Updates the release notes for v61.5.0 and the changes in https://github.com/guardian/cdk/pull/2561. When using [Riff-Raff's `autoscaling` deployment type](https://riffraff.gutools.co.uk/docs/magenta-lib/types#autoscaling), upgrading needs to be performed in two stages.

## How to test
Do the docs make sense?

A simpler upgrade path.

N/A.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,39 +12,78 @@
 
 ### Minor Changes
 
-- 2a75b9f: Removes `GuWazuhAccess` security group as Wazuh has been deprecated.
+Removes `GuWazuhAccess` security group as Wazuh has been deprecated (#2561).
 
-  This change will remove a resource of logical ID `WazuhSecurityGroup` from stacks that use a `GuAutoScalingGroup`.
-  The snapshot diff will include the removal of the following resource:
+This change will remove a resource of logical ID `WazuhSecurityGroup` from stacks that use a `GuAutoScalingGroup`.
+The snapshot diff will include the removal of the following resource:
 
-  ```json
-  {
-    "Resources": {
-      "WazuhSecurityGroup": {
-        "Properties": {
-          "GroupDescription": "Allow outbound traffic from wazuh agent to manager",
-          "SecurityGroupEgress": [
-            {
-              "CidrIp": "0.0.0.0/0",
-              "Description": "Wazuh event logging",
-              "FromPort": 1514,
-              "IpProtocol": "tcp",
-              "ToPort": 1514
-            },
-            {
-              "CidrIp": "0.0.0.0/0",
-              "Description": "Wazuh agent registration",
-              "FromPort": 1515,
-              "IpProtocol": "tcp",
-              "ToPort": 1515
-            }
-          ],
-          "Type": "AWS::EC2::SecurityGroup"
-        }
+```json
+{
+  "Resources": {
+    "WazuhSecurityGroup": {
+      "Properties": {
+        "GroupDescription": "Allow outbound traffic from wazuh agent to manager",
+        "SecurityGroupEgress": [
+          {
+            "CidrIp": "0.0.0.0/0",
+            "Description": "Wazuh event logging",
+            "FromPort": 1514,
+            "IpProtocol": "tcp",
+            "ToPort": 1514
+          },
+          {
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -12,39 +12,79 @@
...
+     // See https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-securitygroup.html#cfn-ec2-securitygroup-groupdescription.
+     description: "Allow outbound traffic from wazuh agent to manager",
+   });
+   this.overrideLogicalId(tempSecurityGroup, {
+     logicalId: "WazuhSecurityGroup",
+     reason:
+       "Part one of updating to GuCDK 61.5.0+ whilst using Riff-Raff's ASG deployment type",
+   });
+   ```
+   This should be a PR of its own.
```
jorgeazevedo, line 79: When you say this should be a PR of its own, do you mean that clients should update to 6.1.5.0 in one PR, have that merged, and then put in this fix in its own PR?

CHANGELOG.md
```diff
@@ -12,39 +12,79 @@
 
 ### Minor Changes
```
jorgeazevedo, line 13: I think the dual step nature of this makes it a major change, but that ship has sailed, right? We can't make this version 62?

CHANGELOG.md
```diff
@@ -12,39 +12,79 @@
...
+#### How to update to this version
+This version of `@guardian/cdk` detaches the `WazuhSecurityGroup` security group from any autoscaling group and deletes it in one step.
+When using [Riff-Raff's autoscaling deployment type](https://riffraff.gutools.co.uk/docs/magenta-lib/types#autoscaling), upgrading needs to be performed in two steps.
+If we do not, Riff-Raff will fail with an error similar to:
+
+> WazuhSecurityGroup(AWS::EC2::SecurityGroup}: DELETE_FAILED resource sg-1234 has a dependent object (Service: Ec2, Status Code: 400) (SDK Attempt Count: 1)
+
+1. Detach the `WazuhSecurityGroup` security group from the autoscaling group
+
+   In this step, we detach the `WazuhSecurityGroup` security group from the autoscaling group by upgrading to v61.5.0
```
jorgeazevedo, line 58: Can you help me understand what you mean by detaching the security group from the ASG? Like you would detach an instance from an ASG? Or do you mean removing it from the lists of SGs in the launch template?

CHANGELOG.md
```diff
@@ -12,39 +12,79 @@
...
+   and temporarily recreate `WazuhSecurityGroup` as a resource in the CloudFormation stack:
+
+   ```ts
+   declare const myApp: GuEc2App;
+
+   const { vpc } = myApp;
+
+   // A temporary security group with a fixed logical ID, replicating the one removed from GuCDK v61.5.0.
```
jorgeazevedo, line 66: Is the idea that by doing this move, from a YAML perspective the WazuhSecurityGroup will remain there but just with its rules removed? Therefore, the cfn deployment will update but not replace the wazuh security group? Though in reality it's no longer part of the normal cdk stack, so all resources that used to reference it like the launch template, will stop doing so?

CHANGELOG.md
jorgeazevedo: Hmm personally I'd find it surprising to have to go through this process for just a minor version bump, but if you think in the context of this library it doesn't quite make the cut for a major I'm happy to leave it. I think in the end it doesn't really make much practical difference, given clients tend to update to the latest version anyway.

CHANGELOG.md
jorgeazevedo: Ok, got it!

jorgeazevedo (approved):
I left some suggestions that would've made me understand this a bit better, but I'm happy for you to merge as is and see how it works for the teams. I think Tom Wey will immediately want to give it a try.

CHANGELOG.md
akash1810: Good point! I've updated each step now, prefixing the description with "For the first/second pull request, ...".

PR #2660, opened 2025-05-16
feat: add slow start support to new deployment mechanism

jacobwinch:
## What does this change?

Adds [slow start mode](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/edit-target-group-attributes.html#slow-start-mode) support to the new/experimental deployment mechanism (which was added in https://github.com/guardian/cdk/pull/2417).

## How to test

I've used `cdk-playground` to test this.

There is a successful deployment example [here](https://github.com/guardian/cdk-playground/pull/703).

I've also created a special build that starts failing the healthcheck _whilst the instance is in slow start mode_, to demonstrate the new failure case - an example for that can be found [here](https://github.com/guardian/cdk-playground/pull/704).

This allows high-traffic services to warm-up instances before sending them a full share of requests. The Ophan team have recently found this to be a successful strategy: https://github.com/guardian/ophan/pull/6273.

Teams ([e.g. WebX](https://github.com/guardian/frontend/blob/95f0ffe8857843ca557700f8d1d62294f96078b9/riff-raff.yaml#L13)) who were previously relying on Riff-Raff's ['warm-up grace' feature](https://riffraff.gutools.co.uk/docs/magenta-lib/types#autoscaling), can now use slow-start mode if they migrate to the new deployment mechanism, as this should offer similar benefits.

I don't think there are any particular risks associated with this PR. We are adding an optional prop to an experimental pattern.

## Checklist

### DIFF
--- a/src/experimental/patterns/ec2-app.ts
+++ b/src/experimental/patterns/ec2-app.ts
@@ -276,7 +276,11 @@ export interface GuEc2AppExperimentalProps extends Omit<GuEc2AppProps, "updatePo
  * If you need to use this class for a service other than MAPI, please speak to DevX about your use-case first.
  */
 export class GuRollingUpdatePolicyExperimental {
-  static getPolicy(maximumInstances: number, minimumInstances?: number, slowStartDuration?: Duration) {
+  static getPolicy(
+    maximumInstances: number,
+    minimumInstances?: number,
+    slowStartDuration: Duration = Duration.seconds(0),
+  ) {
     return UpdatePolicy.rollingUpdate({
       maxBatchSize: maximumInstances,
       minInstancesInService: minimumInstances,
@@ -289,8 +293,19 @@ export class GuRollingUpdatePolicyExperimental {
       If AWS ever supports suspending scale-out and scale-in independently, we should allow scale-out.
        */
       suspendProcesses: [ScalingProcess.ALARM_NOTIFICATION],
-      // Note: this is increased via an Aspect which also takes healthcheck grace period into account.
-      pauseTime: slowStartDuration ?? Duration.seconds(0),
+      /*
+      Note: this is increased via an Aspect which also takes healthcheck grace period into account.
+
+      It was easier to pass the slow start duration through here rather than trying to do this via the
+      GuAutoScalingRollingUpdateTimeoutExperimental aspect.
+
+      If we do this via the aspect then we need to find all mappings between ASGs and target groups and then
+      get the relevant slow start duration for each. This is probably possible but would be reasonably complex.
+      However, it's worth mentioning that by taking this approach we open up a small risk that things will not work
+      as expected if users enable slow start via a different mechanism (e.g. by modifying the target group outside
+      the pattern).
+       */
+      pauseTime: slowStartDuration,
     });
   }
 }
--- a/src/experimental/patterns/ec2-app.ts
+++ b/src/experimental/patterns/ec2-app.ts
@@ -276,7 +276,7 @@ export interface GuEc2AppExperimentalProps extends Omit<GuEc2AppProps, "updatePo
  * If you need to use this class for a service other than MAPI, please speak to DevX about your use-case first.
  */
 export class GuRollingUpdatePolicyExperimental {
-  static getPolicy(slowStartDuration: Duration, maximumInstances: number, minimumInstances?: number) {
+  static getPolicy(maximumInstances: number, minimumInstances?: number, slowStartDuration?: Duration) {
     return UpdatePolicy.rollingUpdate({
       maxBatchSize: maximumInstances,
       minInstancesInService: minimumInstances,
@@ -290,7 +290,7 @@ export class GuRollingUpdatePolicyExperimental {
        */
       suspendProcesses: [ScalingProcess.ALARM_NOTIFICATION],
       // Note: this is increased via an Aspect which also takes healthcheck grace period into account.
-      pauseTime: slowStartDuration,
+      pauseTime: slowStartDuration ?? Duration.seconds(0),
     });
   }
 }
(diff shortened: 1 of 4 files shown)


src/experimental/patterns/ec2-app.ts
```diff
@@ -355,9 +372,36 @@ export class GuUserDataForRollingUpdateExperimental {
...
+        sleep ${slowStartInSeconds}s
+        echo "Instance running build ${buildIdentifier} should have warmed up by now..."
+
+        STATE=$(aws elbv2 describe-target-health \
+        --target-group-arn ${targetGroup.targetGroupArn} \
+        --region ${region} \
+        --targets Id=$INSTANCE_ID,Port=${applicationPort} \
+        --query "TargetHealthDescriptions[0].TargetHealth.State")
+
+        if [ "$STATE" == "\\"healthy\\"" ]; then
```
jacobwinch, line 394: Note that we do a one-off check here rather than polling.

If an instance is not healthy after it's warm-up period[^1] then this means that it has:

1) Initially launched correctly and become healthy
2) Fallen over despite only being asked to deal with a lower-than-average share of traffic

Consequently we just assume that something is wrong with the new build at this point and fail the deployment.

[^1]: In this scenario the instance will have [exited slow-start mode](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/edit-target-group-attributes.html#slow-start-mode)[^1]; in theory it could become healthy again and re-enter slow start mode, but we are not waiting to find out if this happens.

src/experimental/patterns/ec2-app.ts
```diff
@@ -261,7 +276,7 @@ export interface GuEc2AppExperimentalProps extends Omit<GuEc2AppProps, "updatePo
  * If you need to use this class for a service other than MAPI, please speak to DevX about your use-case first.
  */
 export class GuRollingUpdatePolicyExperimental {
-  static getPolicy(maximumInstances: number, minimumInstances?: number) {
+  static getPolicy(maximumInstances: number, minimumInstances?: number, slowStartDuration?: Duration) {
```
jacobwinch, line 279: To get this working correctly for MAPI we'll need to pass the slow start duration in here...

src/experimental/patterns/ec2-app.ts
```diff
@@ -324,15 +341,15 @@ export interface GuUserDataForRollingUpdateExperimentalProps {
 export class GuUserDataForRollingUpdateExperimental {
   constructor(scope: GuStack, props: GuUserDataForRollingUpdateExperimentalProps) {
```
jacobwinch, line 342: And we'll also need to pass the slow start duration here if we enable this for MAPI.

src/experimental/patterns/ec2-app.ts
```diff
@@ -72,7 +72,9 @@ export class GuAutoScalingRollingUpdateTimeoutExperimental implements IAspect {
         construct.cfnOptions.updatePolicy = {
           autoScalingRollingUpdate: {
             ...currentRollingUpdate,
-            pauseTime: Duration.seconds(signalTimeoutSeconds).toIsoString(),
+            pauseTime: Duration.seconds(
```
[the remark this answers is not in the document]
akash1810: Worth adding a note/TODO here, to prompt us to look at this when promoting this experimental pattern to stable?

src/experimental/patterns/ec2-app.ts
```diff
@@ -274,7 +289,8 @@ export class GuRollingUpdatePolicyExperimental {
       If AWS ever supports suspending scale-out and scale-in independently, we should allow scale-out.
        */
       suspendProcesses: [ScalingProcess.ALARM_NOTIFICATION],
-      // Note: there is also an important property called pauseTime, but this is set via an Aspect.
+      // Note: this is increased via an Aspect which also takes healthcheck grace period into account.
+      pauseTime: slowStartDuration ?? Duration.seconds(0),
```
akash1810, line 293: minor: Would having a default value in the function signature highlight it more/be more maintainable?

```ts
getPolicy(maximumInstances: number, minimumInstances?: number, slowStarDuration: Duration = Duration.seconds(0)) {
  ...
}
```

src/experimental/patterns/ec2-app.ts
```diff
@@ -251,6 +253,19 @@ export interface GuEc2AppExperimentalProps extends Omit<GuEc2AppProps, "updatePo
...
   buildIdentifier: string;
+  /**
+   * The amount of time that newly launched instances will spend warming-up / serving fewer requests than other instances
+   * (defaults to 0 seconds).
+   *
+   * See https://docs.aws.amazon.com/elasticloadbalancing/latest/application/edit-target-group-attributes.html#slow-start-mode
+   * for more details.
+   *
+   * We recommend enabling this setting if you run a high-traffic service, particularly if it is JVM-based.
+   *
```
akash1810, line 264: Are we able to recommend a value? Or how to calculate a value? For example, would it be always be greater than one of the configurable healthcheck values?

Alternatively, could we log a warning/error if the duration is something unadvisable, like 1 hour?

akash1810 (approved):
A couple of comments, nothing blocking.

src/experimental/patterns/ec2-app.ts
jacobwinch: Unfortunately I think it will depend on the app and will likely need performance testing to optimise this effectively. 

For now I've just copied over some of the [AWS docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-elasticloadbalancingv2-targetgroup-targetgroupattribute.html) and added some validation that was missing before: https://github.com/guardian/cdk/pull/2660/commits/228d7dafda503e2c6b7c2b19a0967ec677eebdc5.

PR #2677, opened 2025-06-04
refactor(update-aws-cdk): Replace PAT with a GitHub App

akash1810:
## Checklist

- [ ] I have listed any breaking changes, along with a migration path [^1]
- [ ] I have updated the documentation as required for the described changes [^2]

[^1]: Consider whether this is something that will mean changes to projects that have already been migrated, or to the CDK CLI tool. If changes are required, consider adding a checklist here and/or linking to related PRs.
[^2]: If you are adding a new construct or pattern, has new documentation been added? If you are amending defaults or changing behaviour, are the existing docs still valid?

PR #2681, opened 2025-06-04
refactor(cd): Replace PAT with a GitHub App

akash1810:
## What does this change?
Follows on from #2677, replacing the PAT used by the [CD workflow](https://github.com/guardian/cdk/blob/main/.github/workflows/cd.yaml).

## How to test
I've [added](https://github.com/guardian/cdk/commit/7b347c642c7fa2e83b3851cf08acb08f5c31028d) (and then removed) a blank changeset file. [Running the CD workflow from this branch](https://github.com/guardian/cdk/actions/runs/15446764215/job/43478474750) with this change created https://github.com/guardian/cdk/pull/2680. This demonstrates the changes for the majority of the workflow; the ["release docs" section](https://github.com/guardian/cdk/blob/ccc5b1df9f4ca3ac5bde44473d2e962864710cee/.github/workflows/cd.yaml#L52-L56) can only be tested on `main`.

Once merged, we can remove the `GU_GUARDIAN_CI_TOKEN` repository secret and revoke the PAT.

N/A.

## Checklist

### DIFF
--- a/.github/workflows/cd.yaml
+++ b/.github/workflows/cd.yaml
@@ -33,6 +33,23 @@ jobs:
       - name: Test
         run: npm run test
 
+      - uses: actions/create-github-app-token@v2
+        id: app-token
+        with:
+          app-id: ${{ vars.APP_ID }}
+          private-key: ${{ secrets.PRIVATE_KEY }}
+
+      - name: Get GitHub App User ID
+        id: get-user-id
+        run: echo "user-id=$(gh api "/users/${{ steps.app-token.outputs.app-slug }}[bot]" --jq .id)" >> "$GITHUB_OUTPUT"
+        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/cd.yaml
```diff
@@ -33,24 +33,37 @@ jobs:
...
+
       - name: Create Release Pull Request or Publish to npm
         id: changesets
         uses: changesets/action@v1
         with:
           publish: npx changeset publish
           title: "🦋 Release package updates"
           commit: "Bump package version"
-
+          setupGitUser: false
```
akash1810, line 60: This ensures the [changesets action](https://github.com/changesets/action/blob/04d574e831923498156e0b2b93152878063203a3/action.yml#L23-L26) uses the git user we've configured within the "Configure git" step.

PR #2688, opened 2025-06-16
fix(GuEcsTask): Update import path of `ContainerInsights`

akash1810:
## What does this change?
Updates the import path to resolve the following in clients CI builds:

```
error TS2307: Cannot find module 'aws-cdk-lib/aws-ecs/lib/cluster' or its corresponding type declarations.
```

## How to test
Observe CI, it should pass.

N/A.

N/A.

## Checklist

### DIFF
--- a/.changeset/lovely-onions-relate.md
+++ b/.changeset/lovely-onions-relate.md
@@ -1,0 +1,5 @@
+---
+"@guardian/cdk": patch
+---
+
+Update import path of `ContainerInsights`.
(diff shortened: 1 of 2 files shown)


emdash-ie:
Did you work out why this bug didn’t cause a failure in CI originally? I don’t think the reason was explained in chat. (If I understood right it sounded like https://github.com/guardian/cdk/pull/2643 should have failed CI?)

PR #2703, opened 2025-07-17
Add necessary permissions for guard duty runtime monitoring of ecs tasks

philmcmahon:
## What does this change?
Recently we've spotted these 'guard duty runtime monitoring' containers being spun up alongside our ECS tasks. Currently they're all failing because of missing the permissions detailed here https://docs.aws.amazon.com/guardduty/latest/ug/prereq-runtime-monitoring-ecs-support.html#before-enable-runtime-monitoring-ecs

This PR adds those permissions to the base ECS task execution role. 

## How to test
I've deployed this branch to our app lurch on CODE and verified that it resolves the guard duty container failure


dominickendrick (approved):
Looks in line with the docs : )

src/constructs/ecs/ecs-task.ts
```diff
@@ -236,6 +236,20 @@ export class GuEcsTask extends Construct {
...
+    // See https://docs.aws.amazon.com/guardduty/latest/ug/prereq-runtime-monitoring-ecs-support.html#before-enable-runtime-monitoring-ecs
+    const guardDutyPolicy = new PolicyStatement({
+      effect: Effect.ALLOW,
+      actions: [
+        "ecr:GetAuthorizationToken",
+        "ecr:BatchCheckLayerAvailability",
+        "ecr:GetDownloadUrlForLayer",
+        "ecr:BatchGetImage",
+      ],
+      resources: ["*"],
```
dominickendrick, line 248: Does this need to be so permissive ?

philmcmahon: No it doesn't! I left it as * because there's a different guard duty fargate url for each region. I've updated it now to only allow fetching images from the eu-west-1 guard duty fargate container location, as referenced here https://docs.aws.amazon.com/guardduty/latest/ug/runtime-monitoring-ecr-repository-gdu-agent.html

jacobwinch (approved):

PR #2704, opened 2025-07-23
Fix guard duty runtime monitoring policy

philmcmahon:
## What does this change?
I made a mistake in https://github.com/guardian/cdk/pull/2703 - specifying the repository url rather than the ARN when trying to limit access to the fargate repository in eu-west-1. In addition, `ecr:GetAuthorizationToken` is a global policy so shouldn't be scoped to a single resource.

Currently, using this pattern results in a cloudformation error `Resource handler returned message: "Resource 694911143906.dkr.ecr.eu-west-1.amazonaws.com/aws-guardduty-agent-fargate must be in ARN format or "*". `

## How to test
I've tested this using lurch (I know I said I'd done that for https://github.com/guardian/cdk/pull/2703!) I'm not certain what happened there but this time I verified that the iam execution role had been updated as expected.

### DIFF
--- a/.changeset/quiet-states-pump.md
+++ b/.changeset/quiet-states-pump.md
@@ -1,0 +1,5 @@
+---
+"@guardian/cdk": patch
+---
+
+Fix resource property of guard duty IAM role for ECS task pattern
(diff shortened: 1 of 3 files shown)


akash1810 (approved):
Surprised AWS CDK accepts non-ARN strings when creating an IAM Policy 🤔.

PR #2714, opened 2025-08-12
fix(update-aws-cdk): Ensure a valid git branch name

akash1810:
## What does this change?
Remove invalid `^` character from branch name as described in the spec - https://git-scm.com/docs/git-check-ref-format.

See also https://github.com/guardian/cdk/pull/2698.

## How to test
I've run the `update-aws-cdk` script locally from this branch and it created https://github.com/guardian/cdk/pull/2713.

The `update-aws-cdk` script runs on the 10th of every month. The invocation for August 2025 failed at the point of creating a branch:

image

This change should allow the script to run to completion.

N/A.

## Checklist


AshCorr (approved):
Ah nice catch, thank you!

PR #2742, opened 2025-09-26
feat!: Enable access logging by default on Application Load Balancers 

akash1810:
> [!NOTE]
> - This is an alternative to https://github.com/guardian/cdk/pull/2739. This change does not default the region as there were a bit too many edge-cases to this at the moment.
> - Easiest to review [commit by commit](https://github.com/guardian/cdk/pull/2742/commits).

## What does this change?
The `access_logs.s3.enabled` Application Load Balancer attribute is now explicitly set, with a default value of `true`, enabling access logs by default. Previously this was opt-in behaviour. These logs are written to an S3 bucket in the same AWS account as the ALB and are queryable with Athena.

Consequently, the optional `accessLogging` object property for a `GuEc2` app (and it subclasses) has been removed. In its place is a `withAccessLogging` boolean property with a default value of `true`.

### Why drop `accessLogging`?
The `accessLogging` property had the following type:

```ts
export interface AccessLoggingProps {
  /**
   * Enable (load balancer) access logs.
   *
   * Note, you will need to specify a region in your stack declaration to use
   * this.
   * See`https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-elasticloadbalancingv2.ApplicationLoadBalancer.html#logwbraccesswbrlogsbucket-prefix`
   */
  enabled: boolean;
  /**
   * S3 prefix for the logs.
   *
   * @defaultValue no prefix
   */
  prefix?: string;
}
```

It allowed the `prefix` to be set. DevX RelOps have recently updated https://github.com/guardian/aws-account-setup to provision an Athena database, table and saved query for this data. These Athena resources assume a specific prefix which is now hardcoded in GuCDK.

If a service doesn't want to use the RelOps solution, they still can by setting `withAccessLogging: false` and manually configuring the load balancer.

## How to test
See the updated unit tests.

Access logs are another datasource that can be used during incident triage. Success is that incident triage becomes easier, though this is hard to measure!

This change should only impact clients that use `GuEc2App` or its sub-classes. The changes require a snapshot update before passing CI and being deployed. The risk (if it can be considered a risk) is that team's won't understand why their snapshot is failing; the release notes should help there.

## Checklist

### DIFF
--- a/.changeset/great-eyes-tan.md
+++ b/.changeset/great-eyes-tan.md
@@ -1,0 +1,49 @@
+---
+"@guardian/cdk": major
+---
+
+Access logging for Application Load Balancers (ALBs) is now enabled by default.
+
+[Application Load Balancer (ALB) access logs](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-access-logs.html) describe, in detail, each request processed by a load balancer, including request paths and status codes.
+They are helpful during incident response and are now enabled by default.
+
+Previously users of the `GuEc2App`, `GuNodeApp`, `GuPlayApp `, `GuPlayWorkerApp` and `GuEc2AppExperimental` patterns could opt-in to this logging via the `accessLogging` property and configure the S3 prefix.
+
+This property is now removed and replaced with a new optional boolean property `withAccessLogging` which defaults to `true`.
+- When `true` the ALB will have access logs enabled, configured to write to the account's S3 bucket using a specific prefix for compatibility with the `gucdk_access_logs` database created in Athena via https://github.com/guardian/aws-account-setup.
+- When `false` the [`access_logs.s3.enabled` attribute](https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-properties-elasticloadbalancingv2-loadbalancer-loadbalancerattribute.html) is now explicitly set to `false`.
+
+A `withAccessLogging` property is also added to the `GuApplicationLoadBalancer` construct, with the same behaviour.
+
+NOTE: This feature requires a region to be set at the `GuStack` level, else the following error will be thrown:
+
+> ValidationError: Region is required to enable ELBv2 access logging
+
+Here's an example of how to set the region:
+
+```typescript
+class MyStack extends GuStack {
+  constructor(scope: App, id: string, props: GuStackProps) {
+    super(scope, id, props);
+  }
+}
+
+const stackInstance = new MyStack(app, 'MyStack', {
+  env: {
+    region: 'eu-west-1'
+  },
+});
+```
+
+There are three cost areas to this feature:
+- Writing to S3.
+
+  AWS absorbs these costs.
+
+- S3 data storage.
+
+  This cost will vary depending on the volume of traffic received; more traffic, more logs. To somewhat mitigate this, the target S3 bucket has already been configured to retain logs for 14 days.
+
+- Reading from S3 using Athena.
+
+  This cost will vary depending on the volume of logs queried.
--- a/src/constructs/loadbalancing/alb/application-load-balancer.ts
+++ b/src/constructs/loadbalancing/alb/application-load-balancer.ts
@@ -1,8 +1,10 @@
-import { CfnOutput } from "aws-cdk-lib";
+import { Annotations, CfnOutput } from "aws-cdk-lib";
 import { ApplicationLoadBalancer } from "aws-cdk-lib/aws-elasticloadbalancingv2";
 import type { ApplicationLoadBalancerProps, CfnLoadBalancer } from "aws-cdk-lib/aws-elasticloadbalancingv2";
+import { Bucket } from "aws-cdk-lib/aws-s3";
 import { GuAppAwareConstruct } from "../../../utils/mixin/app-aware-construct";
-import type { AppIdentity, GuStack } from "../../core";
+import type { GuStack } from "../../core";
+import { AppIdentity, GuAccessLoggingBucketParameter } from "../../core";
 
 /**
  * Adds the following headers to each request before forwarding it to the target:
@@ -26,6 +28,16 @@ interface GuApplicationLoadBalancerProps extends ApplicationLoadBalancerProps, A
    * If a Load Balancer is replaced it is likely to lead to downtime.
    */
   removeType?: boolean;
+
+  /**
+   * Enable access logging for this load balancer.
+   * Access logs are written to an S3 bucket within your AWS account.
+   * The bucket is created by {@link https://github.com/guardian/aws-account-setup}.
+   * The logs are queryable via the `gucdk_access_logs` Athena database.
+   *
+   * @default true
+   */
+  withAccessLogging?: boolean;
 }
 
 /**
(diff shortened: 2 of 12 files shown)


src/constructs/loadbalancing/alb/application-load-balancer.ts
```diff
@@ -40,16 +52,32 @@ interface GuApplicationLoadBalancerProps extends ApplicationLoadBalancerProps, A
...
+    if (withAccessLogging) {
+      const bucket = Bucket.fromBucketName(
+        scope,
+        AppIdentity.suffixText(props, "AccessLoggingBucket"),
+        GuAccessLoggingBucketParameter.getInstance(scope).valueAsString,
+      );
+      const prefix = `application-load-balancer/${stage}/${stack}/${app}`;
+      this.logAccessLogs(bucket, prefix);
+    } else {
+      this.setAttribute("access_logs.s3.enabled", "false");
```
akash1810, line 77: From the [AWS docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-properties-elasticloadbalancingv2-loadbalancer-loadbalancerattribute.html) `access_logs.s3.enabled` is implicitly `false`. Setting the value explicitly is a lot easier to reason about (and test!).

.changeset/great-eyes-tan.md
```diff
@@ -0,0 +1,42 @@
+---
+"@guardian/cdk": major
+---
+
+The `access_logs.s3.enabled` Application Load Balancer attribute is explicitly set, with a default value of `true`.
+
+Access logging for Application Load Balancers (ALBs) is now enabled by default.
+These logs are helpful during incident triage; therefore, they are now enabled by default.
+
+Previously, access logging was opt-in via the `accessLogging` object property on the `GuEc2App` pattern and its subclasses.
```
jacobwinch, line 10: Minor point, but I think we could make life easier for users here and just name the relevant patterns (as knowing about the subclasses requires knowledge of this library code IIUC):

```suggestion
Previously users of the `GuEc2App`, `GuNodeApp`, `GuPlayApp ` and `GuEc2AppExperimental` patterns could opt-in to this logging via the `accessLogging` property.
```

.changeset/great-eyes-tan.md
```diff
@@ -0,0 +1,42 @@
+---
+"@guardian/cdk": major
+---
+
+The `access_logs.s3.enabled` Application Load Balancer attribute is explicitly set, with a default value of `true`.
```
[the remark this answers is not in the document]
akash1810: I'd like to keep this as it indicates that, regardless of the value of `withAccessLogs`, a snapshot update is required. Agreed, it's too implementation detail-y for a subject line! Will move it down.

.changeset/great-eyes-tan.md
```diff
@@ -0,0 +1,42 @@
...
+
+See also https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-properties-elasticloadbalancingv2-loadbalancer-loadbalancerattribute.html.
+
+A `withAccessLogging` property is also added to the `GuApplicationLoadBalancer` construct.
+
+NOTE: This feature requires a region to be set at the `GuStack` level, else the following error will be thrown:
+
+> ValidationError: Region is required to enable ELBv2 access logging
+
+Here's an example of how to set the region:
```
jacobwinch, line 25: 🎉

.changeset/great-eyes-tan.md
```diff
@@ -0,0 +1,42 @@
...
+
+const stackInstance = new MyStack(app, 'MyStack', {
+  env: {
+    region: 'eu-west-1'
+  },
+});
+```
+
+The main cost of this feature is incurred at query time (via Athena).
+There are no storage costs, as they are absorbed by AWS.
```
jacobwinch, line 42: I think there is some storage cost associated with this (i.e. GBs stored in S3), even though AWS covers the cost of write requests from the ELB service?

If that's correct, we might want to mention it just in case it has an impact on very high traffic services like Ophan? But we could definitely mention that for most services the storage costs are negligible?

akash1810: Ah, good point! From https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-access-logs.html:

> You are charged storage costs for Amazon S3, but not charged for the bandwidth used by Elastic Load Balancing to send log files to Amazon S3. For more information about storage costs

src/patterns/ec2-app/framework.test.ts
```diff
@@ -52,7 +52,7 @@ describe("Framework level EC2 app patterns", () => {
   });
 
   it("GuPlayWorkerApp has no internal or external access", function () {
```
jacobwinch, line 54: I wonder if access logging should be disabled by default for the `GuPlayWorkerApp` because the load balancer is inaccessible / does not serve requests?

I'd actually made the assumption that it would be disabled when suggesting the list of patterns that we should mention in the release notes.

src/patterns/ec2-app/base.test.ts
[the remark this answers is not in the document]
akash1810: I think having a discrete `describe("Access logging behaviour")` test block would be most explicit here. I'll leave this test alone in this change, and move things around in the next PR.

src/patterns/ec2-app/framework.test.ts
akash1810: Whilst this ALB doesn't receive external traffic, I think it can technically receive traffic from within the VPC. I think for simplicity it's easier to have access logs enabled as standard. For a `GuPlayWorkerApp` there is almost no financial cost here. Any cost would be less than $1.00. WDYT?

jacobwinch: I'd assumed it was completely inaccessible due to the [empty array being passed for `cidrRanges`](https://github.com/guardian/cdk/blob/5569c749211b518001666cffb558fe403ff0539c/src/patterns/ec2-app/framework.ts#L28), but maybe I've misunderstood that.

I initially suggested switching the feature off to avoid confusion (for example, I remember us briefly thinking the access logging feature was broken for `crier` due to the absence of traffic). But on reflection I just find the overall architecture pattern (having a load balancer when there is no load to balance!) confusing[^1] and I don't think that enabling access logging makes this much worse. 

You're right that it won't cost us any money (I actually think the cost will be $0 if it really is receiving 0 requests?!), so feel free to leave it as is.

[^1]: I know we do this to take advantage of ELB healthchecks, which is a fair reason, I just still get surprised whenever I see a `PROD` load balancer with 0 requests!

jacobwinch (approved):

src/patterns/ec2-app/framework.test.ts
akash1810: > I just still get surprised whenever I see a PROD load balancer with 0 requests!

💯 !

.changeset/great-eyes-tan.md
```diff
@@ -0,0 +1,54 @@
...
+Access logging for Application Load Balancers (ALBs) is now enabled by default.
+
+The Application Load Balancer (ALB) access logs are helpful during incident triage.
+Therefore, they are now enabled by default.
+
+Previously users of the `GuEc2App`, `GuNodeApp`, `GuPlayApp `, `GuPlayWorkerApp` and `GuEc2AppExperimental` patterns could opt-in to this logging via the `accessLogging` property and configure the S3 prefix.
+
+This property is now removed and replaced with a new optional boolean property `withAccessLogging` which defaults to `true`.
+- When `true` the ALB will have access logs enabled, configured to write to the account's S3 bucket
+using a specific prefix for compatibility with the Athena database created in https://github.com/guardian/aws-account-setup.
```
jorgeazevedo, line 14: Maybe this isn't the right place to mention it, but I think it'd be useful to write the name of the database somewhere.

```suggestion
using a specific prefix for compatibility with the `gucdk_access_logs` database created in Athena via https://github.com/guardian/aws-account-setup.
```

akash1810: Updated. I've also added this to the docstrings too.

PR #2743, opened 2025-09-29
docs: attempt to improve GuCDK migration docs

jacobwinch:
## What does this change?

This PR updates our migration docs following a recent review.

### DIFF
--- a/docs/migration-guide.md
+++ b/docs/migration-guide.md
@@ -9,7 +9,6 @@ If you are starting from scratch, see the [new project guide](setting-up-a-gucdk
 ## Overall migration process
 
 1. Identify the [GuCDK pattern](https://guardian.github.io/cdk/modules/patterns.html) which works for your app
-1. Check account readiness
 1. Add GuCDK to your repository and CI/CD process
 1. Instantiate a GuCDK pattern alongside your legacy infrastructure (dual-stack), for stateless resources
 1. Switch your production workload (e.g. serving HTTP traffic or processing data) to the GuCDK infrastructure
@@ -25,16 +24,6 @@ If you are starting from scratch, see the [new project guide](setting-up-a-gucdk
 | `AWS::Lambda::Function` (single) with `AWS::ApiGateway::RestApi`                                                                           | Serving HTTP traffic                | `GuApiLambda`                   |
 | `AWS::Lambda::Function` (multiple) with `AWS::ApiGateway::RestApi`                                                                         | Serving HTTP traffic                | `GuApiGatewayWithLambdaByPath`  |
 | `AWS::Lambda::Function` with `AWS::Events::Rule`                                                                                           | Polling/running tasks on a schedule | `GuScheduledLambda`             |
-
-## Check account readiness
-
-`@guardian/cdk` patterns and constructs rely on parameter store values for things like VPC identifiers and dist buckets.
-
-To ensure AWS your account is set up correctly, run:
-
-    npx @guardian/cdk@latest account-readiness --profile [profile]
-
-Then follow the instructions for any errors.
 
 ## Add GuCDK to your repository and CI/CD process
 
--- a/docs/migration-guide.md
+++ b/docs/migration-guide.md
@@ -18,15 +18,13 @@ If you are starting from scratch, see the [new project guide](setting-up-a-gucdk
 
 ## Identify the right GuCDK pattern
 
-| YAML Infrastructure                                                                   | Application type                    | GuCDK pattern to use            |
-|---------------------------------------------------------------------------------------|-------------------------------------|---------------------------------|
-| `AWS::AutoScaling::AutoScalingGroup` with `AWS::ElasticLoadBalancing::LoadBalancer`   | Serving HTTP traffic                | `GuEc2App`                      |
-| `AWS::AutoScaling::AutoScalingGroup` with `AWS::ElasticLoadBalancingV2::LoadBalancer` | Serving HTTP traffic                | `GuEc2App`                      |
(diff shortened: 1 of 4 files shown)


docs/migration-guide.md
```diff
@@ -6,7 +6,29 @@ If you are starting from scratch, see the [new project guide](setting-up-a-gucdk
 
 ---
 
-## Account readiness
+## Overall migration process
+
+1. Identify the [GuCDK pattern](https://guardian.github.io/cdk/modules/patterns.html) which works for your app
+1. Check account readiness
```
jacobwinch, line 12: Hopefully most accounts are ready by now, so I'm not sure if we still need this step?

jorgeazevedo: That's a great point, if we could confirm that (and maybe set up the one's missing?) it'd be one less thing in the guide!

docs/migration-guide.md
```diff
@@ -6,7 +6,29 @@ If you are starting from scratch, see the [new project guide](setting-up-a-gucdk
...
+1. Instantiate a GuCDK pattern alongside your legacy infrastructure (dual-stack), for stateless resources
+1. Switch your production workload (e.g. serving HTTP traffic or processing data) to the GuCDK infrastructure
+1. Remove legacy infrastructure for stateless resources
+1. Where necessary, define stateful resources (e.g. DBs) via cdk components instead of via the original template
+
+## Identify the right GuCDK pattern
+
+| YAML Infrastructure                                                                   | Application type                    | GuCDK pattern to use            |
+|---------------------------------------------------------------------------------------|-------------------------------------|---------------------------------|
+| `AWS::AutoScaling::AutoScalingGroup` with `AWS::ElasticLoadBalancing::LoadBalancer`   | Serving HTTP traffic                | `GuEc2App`                      |
```
jorgeazevedo, line 23: Did you experiment with the readability of merging these two? I had to do a double take between the rows to clock that the difference is just the version, but maybe the merged version is a bit too busy?
```suggestion
| `AWS::AutoScaling::AutoScalingGroup` with either `AWS::ElasticLoadBalancing::LoadBalancer` or `AWS::ElasticLoadBalancingV2::LoadBalancer`   | Serving HTTP traffic                | `GuEc2App`                      |
```

docs/migration-guide.md
```diff
@@ -77,28 +96,10 @@ The end goal is for CDK to do all the work, and remove the YAML template:
...
-- [for EC2 apps](./migration-guide-ec2.md)
-- [for Lambdas which run on a schedule/cron](./migration-guide-scheduled-lambda.md)
-- [for an API which serves requests using AWS Lambda](./migration-guide-api-with-lambda.md)
+- [for `GuEc2App`](./migration-guide-ec2.md)
+- [for `GuApiLambda` or `GuApiGatewayWithLambdaByPath`](./migration-guide-api-with-lambda.md)
+- [for `GuScheduledLambda`](./migration-guide-scheduled-lambda.md)
 
-:information_source: Migration guides will be coming soon for other architectures.
-Please contact DevX if you'd like to discuss the migration of an application which uses an architecture that is not listed above.
+If your pattern is not listed here then please contact the DevX Reliability & Operations team for support.
```
jorgeazevedo, line 105: I'm a bit confused by what "your pattern" means here. Is the idea here to point out that if you intend to use a pattern for which we don't have a guide (i.e. GuPlayWorkerApp) you should reach out for support? Or is it more if the app you're migrating uses an architecture for which we don't have a pattern?

docs/migration-guide.md
jacobwinch: I hadn't experimented with this before but I think the merged version is better so I've updated this. Thanks for the suggestion!

docs/migration-guide.md
jorgeazevedo: Yeah, from what I can tell `account-readiness` checks for the existence of the parameters defined here https://github.com/guardian/cdk/blob/6caaf42ab406de16b990612a551fdfe1741cbf82/src/constants/ssm-parameter-paths.ts . I'm not sure if this is now enforced for all accounts defined in aws-account-setup, or if there's an easy way to check which accounts are missing some parameters.

akash1810: Yeah `guardian/aws-account-setup` is aimed at getting (and keeping) AWS accounts GuCDK-ready. I think we can remove this step from the GuCDK docs. Indeed, we might be able to remove this from the GuCDK CLI (this might need a bit more thinking about though).

Maybe the [AWS Account Information dashboard](https://metrics.gutools.co.uk/d/de7dtpiflscn4a/aws-account-information) could be expanded with a "GuCDK-ready" box?

jacobwinch: Thanks both; I've removed this in https://github.com/guardian/cdk/pull/2743/commits/c8288fe6dd4d8166e5abd4fb017504b8974735b9.

docs/migration-guide.md
```diff
@@ -77,28 +96,10 @@ The end goal is for CDK to do all the work, and remove the YAML template:
...
-When migrating stateful resources:
-1. Identify a resource(s) to move from YAML to CDK.
-2. Write the CDK and comment out/delete the resource(s) from the YAML template.
-3. Run the tests:
-   - If the changes are acceptable, update the snapshot.
-   - If the changes are not acceptable, refactor CDK and re-run the tests.
-4. With the smallest change made, we can now raise a PR, merge it, and move to the next resource(s).
-
-That is, we can use the snapshot in a [red/green refactor cycle](https://blog.cleancoder.com/uncle-bob/2014/12/17/TheCyclesOfTDD.html).
-We'll repeat this process until the YAML template is empty and can be removed.
```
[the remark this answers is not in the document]
akash1810: Oh interesting, this is still how I think about a migration. Maybe the wording isn't great here as it reads like we're migrating every resource individually.

akash1810 (approved):

PR #2874, opened 2026-03-30
Convert GuDeveloperPolicy to allow withoutPolicyChecks marker

aug24:
## What does this change?

After discussion, the restriction on both resource and action to not use '*' is too strong.  We have therefore provided an interface to simply turn it off, with an optional marker property of "withoutPolicyChecks=true".

This prevents the creation of the checker class altogether, which provides visibility of the choice in the metadata of the output cloudformation, so that we can analyse how often this is required and finesse the check.

## Checklist


akash1810 (commented):
I think we'd want to have this in a minor release? If so, let's add a changeset (`npx changeset add`).

src/experimental/constructs/iam/policies/developer-policy.ts
```diff
@@ -69,8 +73,10 @@ export class GuDeveloperPolicyExperimental extends ManagedPolicy {
       throw new Error("Empty statements array passed to GuDeveloperPolicyExperimental");
     }
 
-    // Later, apply to the stack and check for specific errors
-    Aspects.of(this).add(new GuDeveloperPolicyExperimentalChecker());
+    if (!props.withoutPolicyChecks) {
```
kelvin-chappell, line 76: Does this check cover both absent and present but false?
Also I don't think there's a test for when it's false.

src/experimental/constructs/iam/policies/developer-policy.test.ts
```diff
@@ -103,7 +116,29 @@ describe("GuDeveloperPolicyExperimental", () => {
...
       permission: "test123",
     });
-    Annotations.fromStack(stack).hasError("*", "Statement Action is too broad: *");
+    Annotations.fromStack(stack).hasError(
+      "*",
+      "Statement Action is too broad: *. If this is necessary and intended, use withoutPolicyChecks: true in properties to turn off this check",
+    );
+  });
+
+  test("does not add an error if a wide-open action/allow policy statement is present but the isBroad marker is used", () => {
```
kelvin-chappell:
```suggestion
  test("does not add an error if a wide-open action/allow policy statement is present but the withoutPolicyChecks marker is used", () => {
```

src/experimental/constructs/iam/policies/developer-policy.ts
aug24: Absent is "all the other tests" :)

I've just added a test for present but false.

kelvin-chappell (approved):
👍

PR #2887, opened 2026-04-14
Pass shell: false to childProcess.spawn

tjmw:
## What does this change?

Pass `shell: false` to `childProcess.spawn` when running commands.

This is more secure. After upgrading a project to Node 24 I started seeing deprecation warnings like this, coming from @guardian/cdk:

`DeprecationWarning: Passing args to a child process with shell option true can lead to security vulnerabilities, as the arguments are not escaped, only concatenated.`

## How to test

Verify that build checks/CI pass.

If it's possible to release a BETA version to npm then checking that against a real project would be a good way to verify.

The deprecation warning mentioned above should disappear after updating to a version of `@guardian/cdk` with this change.

## Checklist

- [x] I have listed any breaking changes, along with a migration path [^1] (No breaking changes)
- [x] I have updated the documentation as required for the described changes [^2] (This is an internal change, I don't think documentation changes are required)

### DIFF
--- a/src/bin/commands/new-project/utils/utils.ts
+++ b/src/bin/commands/new-project/utils/utils.ts
@@ -6,8 +6,8 @@ export interface Name {
   pascal: string;
 }
 
-async function runTask(packageManager: PackageManager, cwd: string, task: string): Promise<string> {
-  return await execute(`${packageManager} ${task}`, [], { cwd });
+async function runTask(packageManager: PackageManager, cwd: string, args: string[]): Promise<string> {
+  return await execute(packageManager, args, { cwd });
 }
 
 export function getCommands(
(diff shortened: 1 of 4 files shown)


src/bin/commands/new-project/utils/utils.ts
```diff
@@ -6,8 +6,8 @@ export interface Name {
   pascal: string;
 }
 
-async function runTask(packageManager: PackageManager, cwd: string, task: string): Promise<string> {
-  return await execute(`${packageManager} ${task}`, [], { cwd });
+async function runTask(packageManager: PackageManager, cwd: string, args: string[]): Promise<string> {
```
tjmw, line 9: It's possible this function is now no longer needed since it's signature is so similar to `execute`. I think the main thing we're getting is the stronger typing of the first arg.

jorgeazevedo:
I'm confused that the origin of the warning seems to be the `new-project` utility, but you were seeing issues on an existing project? Which commands did you run that triggered the deprecation notice?

tjmw:
> I'm confused that the origin of the warning seems to be the new-project utility, but you were seeing issues on an existing project? Which commands did you run that triggered the deprecation notice?

Hi @jorgeazevedo good spot. You're right `execute` is only ever called from the new-project utility. But `executeSync` (which I just pushed up a missing change for) is triggered indirectly via the constructor for `GuStack` which is where I was seeing the warning. This is happening when I run the cdk tests after upgrading to Node 24.

akash1810 (commented):
Amazing, thanks for this! I think we'd want to release a new minor version for this. Could we add a changeset?

akash1810 (approved):

PR #2903, opened 2026-05-11
ecs-app as experimental pattern

JuliaBrigitte:
## What does this change?

This is a very rough draft of what a new ECS pattern might look like.

There's an example of how we might migrate existing services to this pattern here: https://github.com/guardian/cdk-playground/pull/1100

---
  - https://app.asana.com/0/0/1214114211358839


src/experimental/patterns/ecs-app.ts
```diff
@@ -0,0 +1,355 @@
...
+
+interface Migration {
+  loadBalancer: IApplicationLoadBalancer;
+  listener: IApplicationListener;
+  targetGroup: IApplicationTargetGroup;
+  // Must be between 0 and 999 (inclusive)
+  weightForEcsTargetGroup: number;
+}
+
+// We still need to port (or consider porting) the following functionality over from GuEc2App:
```
jacobwinch, line 51: I guess we might need to refactor some of this code to make it easier to share it between patterns?

src/experimental/patterns/ecs-app.ts
```diff
@@ -0,0 +1,355 @@
...
+      sourceVolume: logVolume.name,
+      readOnly: false,
+    });
+
+    // We need a new target group even if we share the other load balancer components with the EC2 infrastructure
+    const ecsTargetGroup = new GuApplicationTargetGroup(scope, "EcsTargetGroup", {
+      vpc,
+      app,
+      port: applicationPort,
+      targetGroupName: `${app}-${scope.stage}`, // Add the name here to make it more easily identifiable in the console etc.
```
jacobwinch, line 306: We might need to include the stack here to ensure that it's unique.

src/experimental/patterns/ecs-app.ts
```diff
@@ -0,0 +1,355 @@
...
+import type { GuDomainName } from "../../types";
+import { GuRiffRaffDeploymentIdParameterExperimental } from "../constructs/riff-raff-deployment-id";
+
+interface Scaling {
+  minimumTasks: number;
+  maximumTasks: number;
+}
+
+interface Migration {
+  loadBalancer: IApplicationLoadBalancer;
```
jacobwinch, line 44: I've used these types as I hope that they might leave the door open for less common migration scenarios.

For example, re-using an ALB provisioned via AWS CDK constructs or via YAML. 

Unfortunately we can't easily support migrations if the app still uses a classic load balancer because they don't work with target groups. I think we'd have to move to an ALB first or migrate via DNS in these cases.

src/experimental/patterns/ecs-app.ts
```diff
@@ -0,0 +1,355 @@
...
+    // We need a new target group even if we share the other load balancer components with the EC2 infrastructure
+    const ecsTargetGroup = new GuApplicationTargetGroup(scope, "EcsTargetGroup", {
+      vpc,
+      app,
+      port: applicationPort,
+      targetGroupName: `${app}-${scope.stage}`, // Add the name here to make it more easily identifiable in the console etc.
+      targets: [ecsService],
+    });
+
+    if (!migrationProps) {
```
jacobwinch, line 310: Ideally we'd drop this eventually, but I think we'll need it for quite a while. I don't think it actually makes the code too much harder to work with though.

jacobwinch:
Closing this as we've decided to take the approach that we explored in https://github.com/guardian/cdk/pull/2905 forwards instead.

PR #2904, opened 2026-05-14
Add new GuLoadBalancedAppExperimental pattern (which supports EC2 and ECS)

jacobwinch:
## What does this change?

This PR introduces a new experimental pattern: `GuLoadBalancedAppExperimental`.

It supports all of the functionality that `GuEc2App` offers (['EC2-only mode'](https://github.com/guardian/cdk/blob/22efb9e110e0c5617b661aaea0861023f7d7826f/src/experimental/patterns/gu-load-balanced-app.test.ts#L211-L227)). Additionally, users can set up an ECS service and ECS tasks instead of an Autoscaling Group and EC2 instances (['ECS-only mode'](https://github.com/guardian/cdk/blob/22efb9e110e0c5617b661aaea0861023f7d7826f/src/experimental/patterns/gu-load-balanced-app.test.ts#L16-L31)).

The new pattern can also work in ['hybrid mode'](https://github.com/guardian/cdk/blob/22efb9e110e0c5617b661aaea0861023f7d7826f/src/experimental/patterns/gu-load-balanced-app.test.ts#L36-L63), where we have a single load balancer and 2 target groups (one for EC2 and one for ECS). This allows us to use [weighted forwarding](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/rule-action-types.html#forward-actions) to split traffic between the target groups. This is how we currently expect most migrations from EC2 to ECS to work.

The fact that the pattern can operate in all 3 modes gives us quite a lot of flexibility to experiment without negatively impacting readers whilst its under development. For example, we could run some testing in hybrid mode, fallback to EC2-only mode in order to make drastic/breaking changes to the ECS setup that the pattern provides and then provision it again without incurring any downtime.

Note that there are a few things that are not yet supported in this PR:
* Currently the permissions for pulling images assumes that the ECR repository is in the same account as the app being deployed ([Asana task](https://app.asana.com/1/1210045093164357/project/1214113788426225/task/1214975013349595)).
* I've added support for versioned/CloudFormation only EC2 deployments via a separate PR: https://github.com/guardian/cdk/pull/2908 ([Asana task](https://app.asana.com/1/1210045093164357/project/1214113788426225/task/1215167013385621)).
* Users cannot currently use the Google Auth feature unless they are using the pattern in EC2-only mode (we deliberately throw an error in other scenarios). The CFN builds fine in these scenarios but I think we should do some dedicated testing (particularly in 'hybrid mode') before we open up this feature, as getting this wrong would be really problematic. We don't actually need this feature for `dotcom-rendering`, so I think it's OK to deprioritise this for now. I've created an [Asana task](https://app.asana.com/1/1210045093164357/project/1214113788426225/task/1215167013385624) so that we don't forget about this.

## How to test

The existing suite of tests for `GuEc2App` is now also running against `GuLoadBalancedAppExperimental`. This gives us confidence that `GuEc2App` users could move to `GuLoadBalancedAppExperimental` without changing the underlying infrastructure. Here's an example of how that will look in a project: https://github.com/guardian/cdk-playground/pull/1104.

I've also got a draft here which sets this pattern up in 'ECS-only mode' - https://github.com/guardian/cdk-playground/pull/1114. This is the state that we want most apps to end up in. I've deployed that to `CODE` and it works as expected.

Finally, I've tested the full migration path from 'EC2-only mode' to 'ECS-only mode' using an early version of this pattern:
https://github.com/guardian/cdk-playground/pull/1104
https://github.com/guardian/cdk-playground/pull/1105
https://github.com/guardian/cdk-playground/pull/1106
https://github.com/guardian/cdk-playground/pull/1107
https://github.com/guardian/cdk-playground/pull/1108

This demonstrates how we expect the migration to work for existing GuCDK users.

Moving our ECS infrastructure definition into GuCDK and adding a migration path for existing GuCDK EC2 services allows us to start trialling this architecture with real applications.

The main risk here is the fact that we're duplicating quite a lot of code. If we add features to `GuEc2App` (or modify existing features) then we'll need to port those changes to `GuLoadBalancedAppExperimental` as well.

Personally I'm comfortable with that risk because I don't think there are any plans to make improvements to `GuEc2App` at this stage, so I expect the rate of change in this area to be low. Furthermore, if we ever forgot to do this then we'd notice it when moving services from `GuEc2App` to `GuLoadBalancedAppExperimental` (snapshot tests would start failing).

I think we should aim to move `GuLoadBalancedAppExperimental` out of `experimental` as soon as possible (perhaps once we have it working with `dotcom-rendering`). Once we do that I think `GuEc2App` could just instantiate `GuLoadBalancedApp` behind the scenes ([example](https://github.com/guardian/cdk/pull/2909/changes)), which would allow us to remove the duplication that this PR introduces.

## Checklist

---
  - https://app.asana.com/0/0/1214114211358839


src/constructs/loadbalancing/alb/application-listener.ts
```diff
@@ -31,9 +31,8 @@ export class GuApplicationListener extends GuAppAwareConstruct(ApplicationListen
   }
 }
 
-export interface GuHttpsApplicationListenerProps
-  extends Omit<GuApplicationListenerProps, "defaultAction" | "certificates">, AppIdentity {
-  targetGroup: GuApplicationTargetGroup;
+export interface GuHttpsApplicationListenerProps extends Omit<GuApplicationListenerProps, "certificates">, AppIdentity {
+  targetGroup?: GuApplicationTargetGroup;
```
jacobwinch, line 35: I think this change is OK because:

1) Most users should not be interacting with this construct directly
2) This won't be a breaking change for any users who are

Let me know if you want me to look into an alternative to this.

src/experimental/constructs/riff-raff-deployment-id.ts
```diff
@@ -0,0 +1,22 @@
+import { CfnParameter } from "aws-cdk-lib";
+import type { GuStack } from "../../constructs/core";
+import { isSingletonPresentInStack } from "../../utils/singleton";
+
+export class GuRiffRaffDeploymentIdParameterExperimental extends CfnParameter {
```
jacobwinch, line 5: This has been moved from `src/experimental/patterns/ec2-app.ts` because `GuEc2AppExperimental` and `GuLoadBalancedAppExperimental` both rely on it now.

That also means that we can remove it from `cdk-playground`: https://github.com/guardian/cdk-playground/pull/1114/changes#diff-54a9fd10fa40c2c980236632b17076aeed58180d6b5140b151a5eccd97c3fa1c

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -0,0 +1,839 @@
...
+    ec2: number;
+  };
+}
+
+interface TargetGroups {
+  ec2?: GuApplicationTargetGroup;
+  ecs?: GuApplicationTargetGroup;
+}
+
+export class GuLoadBalancedAppExperimental extends Construct {
```
jacobwinch, line 354: GitHub is not rendering this large diff by default, but this is actually the most interesting part of this PR!

src/experimental/patterns/gu-load-balanced-app.test.ts
```diff
@@ -0,0 +1,1541 @@
...
+import { InstanceClass, InstanceSize, InstanceType, Peer, Port, UserData, Vpc } from "aws-cdk-lib/aws-ec2";
+import type { CfnLoadBalancer } from "aws-cdk-lib/aws-elasticloadbalancingv2";
+import { AccessScope, MetadataKeys } from "../../constants";
+import { GuPrivateConfigBucketParameter } from "../../constructs/core";
+import { GuSecurityGroup } from "../../constructs/ec2";
+import { GuDynamoDBWritePolicy } from "../../constructs/iam";
+import { GuTemplate, simpleGuStackForTesting } from "../../utils/test";
+import { GuLoadBalancedAppExperimental } from "./gu-load-balanced-app";
+
+describe("the GuLoadBalancedAppExperimental pattern should support new ECS and hybrid functionality", function () {
```
jacobwinch, line 13: These test cases cover all of the new functionality that this pattern is adding.

src/experimental/patterns/gu-load-balanced-app.test.ts
```diff
@@ -0,0 +1,1541 @@
...
+          googleAuth: {
+            enabled: true,
+            domain,
+          },
+        }),
+    ).toThrow("Using Google Auth with ECS is currently unsupported");
+  });
+});
+
+describe("the GuLoadBalancedAppExperimental pattern should support all existing GuEc2App functionality", function () {
```
jacobwinch, line 251: All of the remaining test cases have been copied over from https://github.com/guardian/cdk/blob/main/src/patterns/ec2-app/base.test.ts. I've modified them slightly so they are checking the template produced by `GuLoadBalancedAppExperimental` instead of `GuEc2App`.

I did this because I wanted to check that anyone who migrates from `GuEc2App` to `GuLoadBalancedAppExperimental` can do so without changing their infrastructure (the only snapshot test changes in these  cases are to the `Metadata` section of the template).

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -0,0 +1,839 @@
...
+import type { Alarms, ApplicationLoggingProps } from "../../patterns";
+import { restrictedCidrRanges } from "../../patterns";
+import { AppAccess } from "../../types";
+import type { GuAsgCapacity, GuDomainName } from "../../types";
+import type { AmigoProps } from "../../types/amigo";
+import { getUserPoolDomainPrefix } from "../../utils/cognito/cognito";
+import { GuRiffRaffDeploymentIdParameterExperimental } from "../constructs/riff-raff-deployment-id";
+
+export interface GuLoadBalancedAppExperimentalProps extends AppIdentity {
+  /**
```
jacobwinch, line 69: All of these top-level props have been copied across from https://github.com/guardian/cdk/blob/main/src/patterns/ec2-app/base.ts

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -0,0 +1,839 @@
...
+   *
+   * Remove the old param and immediately redeploy with the new param (this does not affect
+   * protection unless and until the WAF configuration is redeployed)
+   *
+   *   OR
+   *
+   * Create an escape hatch by overriding the logical id (see "ssm param escape hatch" test for example)
+   **/
+  waf?: WafProps;
+  ec2Props?: {
```
jacobwinch, line 231: Any props in https://github.com/guardian/cdk/blob/main/src/patterns/ec2-app/base.ts which were specific to the EC2 infrastructure have now been nested here.

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -0,0 +1,839 @@
...
+     * By default, AWS will produce `5Minute` granular metrics.
+     *
+     * It is recommended to produce `1Minute` granular metrics in production,
+     * especially when using ASG metrics to trigger horizontal scaling as it allows for earlier scaling.
+     *
+     * @see https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/viewing_metrics_with_cloudwatch.html
+     */
+    instanceMetricGranularity: "1Minute" | "5Minute";
+  };
+  ecsProps?: {
```
jacobwinch, line 304: And these are all of the ECS-specific props that we've needed so far in `cdk-playground`. I expect this list to grow over time (for example, we don't currently have support for passing custom IAM permissions to tasks).

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -0,0 +1,879 @@
...
+        roleConfiguration = { withoutLogShipping: false, additionalPolicies: [] },
+        scaling: { minimumInstances, maximumInstances = minimumInstances * 2 },
+        userData: userDataLike,
+        imageRecipe,
+        instanceMetadataHopLimit,
+        updatePolicy,
+        defaultInstanceWarmup,
+        instanceMetricGranularity,
+      } = ec2Props;
+
```
jacobwinch, line 426: The code here is copied across from https://github.com/guardian/cdk/blob/main/src/patterns/ec2-app/base.ts.

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -0,0 +1,879 @@
...
+          applicationLogging.systemdUnitName ? `${applicationLogging.systemdUnitName}.service` : `${app}.service`,
+        );
+      }
+    }
+
+    // Setup ECS-specific infrastructure
+    if (ecsProps) {
+      const { cpu, memoryLimitMiB, repositoryName, imageIdentifier, scaling } = ecsProps;
+
+      const cluster = new Cluster(this, "EcsCluster", { vpc });
```
jacobwinch, line 508: This code is all based on https://github.com/guardian/cdk-playground/blob/main/cdk/lib/cdk-playground-ecs.ts

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -0,0 +1,879 @@
...
+        targets: [ecsService],
+      });
+
+      targetGroups = {
+        ...targetGroups,
+        ecs: ecsTargetGroup,
+      };
+    }
+
+    // Set up the load balancer and listener components
```
jacobwinch, line 670: Most of this code for setting up the load balancer is also taken from https://github.com/guardian/cdk/blob/main/src/patterns/ec2-app/base.ts.

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -0,0 +1,879 @@
...
+      vpcSubnets: {
+        subnets: access.scope === AccessScope.INTERNAL ? privateSubnets : publicSubnets,
+      },
+      withAccessLogging,
+      waf,
+    });
+
+    let defaultAction: ListenerAction | undefined = undefined;
+    if (targetGroups.ec2 && !targetGroups.ecs) {
+      defaultAction = ListenerAction.forward([targetGroups.ec2]);
```
jacobwinch: This is the current behaviour that `GuEc2App` users get via: https://github.com/guardian/cdk/blob/50ceac8c3f475eda3748b7dea3f1680725786a6f/src/constructs/loadbalancing/alb/application-listener.ts#L63

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -0,0 +1,879 @@
...
+     */
+    instanceType: InstanceType;
+    /**
+     * Enable and configures application logs.
+     */
+    applicationLogging?: ApplicationLoggingProps;
+    /**
+     * Configure IAM roles for autoscaling group EC2 instances.
+     */
+    roleConfiguration?: GuInstanceRoleProps;
```
jacobwinch, line 248: Some of these props (e.g. `roleConfiguration` and `scaling`) could be made more generic and moved to the top-level so that they apply to ECS and EC2. 

I've deliberately avoided that for now because I want it to be as easy as possible for `GuEc2App` users to start using this new pattern.

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -0,0 +1,879 @@
...
+      taskDefinition.addVolume(logVolume);
+
+      logRouter.addMountPoints({
+        containerPath: "/init",
+        sourceVolume: logVolume.name,
+        readOnly: false,
+      });
+
+      // We need a new target group even if we share the other load balancer components with the EC2 infrastructure
+      const ecsTargetGroup = new GuApplicationTargetGroup(scope, "EcsTargetGroup", {
```
jacobwinch, line 657: In our prototype we are setting the target group name: https://github.com/guardian/cdk-playground/blob/320bf032091636c83770141ee304d712613ef6ba/cdk/lib/cdk-playground-ecs.ts#L263

When I tried to include `stack`, `stage` and 'ecs' in the name I went over a 32-character limit, so I decided to remove that.

It's tricky to ensure that we generate something unique for the account whilst staying under the character limit.

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -0,0 +1,879 @@
...
+          resources: [
+            // See https://docs.aws.amazon.com/guardduty/latest/ug/runtime-monitoring-ecr-repository-gdu-agent.html
+            "arn:aws:ecr:eu-west-1:694911143906:repository/aws-guardduty-agent-fargate",
+          ],
+        }),
+      ];
+
+      guardDutyPolicies.forEach((policy) => taskDefinition.addToExecutionRolePolicy(policy));
+
+      const ecsService = new FargateService(scope, "EcsService", {
```
jacobwinch, line 582: We're not adding a [service name](https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-resource-ecs-service.html#cfn-ecs-service-servicename) at the moment - perhaps we should?

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -0,0 +1,879 @@
...
+        cluster,
+        taskDefinition,
+        vpcSubnets: { subnets: privateSubnets },
```
jacobwinch, line 585: If we pass the private subnets explicitly here then we can avoid all of this [annoying VPC wiring](https://github.com/guardian/cdk-playground/blob/320bf032091636c83770141ee304d712613ef6ba/cdk/lib/cdk-playground-ecs.ts#L98-L114).

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -0,0 +1,879 @@
...
+       * Note that this max can be exceeded when a deployment runs (unlike the ASG max size). E.g. if maximumTasks is 6,
+       * the service is running 6 tasks and a deployment starts, the ECS service will briefly run with 12 tasks to get
+       * the deployment through.
+       */
+      maximumTasks: number;
+    };
+  };
+  /**
+   * If you are specifying `ec2Props` and `ecsProps` use these weights to distribute traffic across the different compute
+   * types. The weights must sum to 999.
```
jacobwinch, line 341: I've just realised that AWS are [actually more flexible than this](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/rule-action-types.html#forward-actions):

> For example, if you specify two target groups, each with a weight of 10, each target group receives half the requests. If you specify two target groups, one with a weight of 10 and the other with a weight of 20, the target group with a weight of 20 receives twice as many requests as the other target group.

So we can support a split of 99.9%/0.01% like this, but it looks like we could also accept:

```typescript
targetGroupWeights: {
  ecs: 1;
  ec2: 1;
};
```

For a 50%/50% split.

Perhaps we want to change this interface?

JuliaBrigitte:
The PR description is very thorough, thank you!

src/experimental/constructs/riff-raff-deployment-id.ts
akash1810: Would it be worth making this "stable" in its own release? Maybe even adding it to `GuStack` by default as Riff-Raff would always fill it in (https://github.com/guardian/riff-raff/pull/1469).

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -0,0 +1,879 @@
...
+import type { GuAsgCapacity, GuDomainName } from "../../types";
+import type { AmigoProps } from "../../types/amigo";
+import { getUserPoolDomainPrefix } from "../../utils/cognito/cognito";
+import { GuRiffRaffDeploymentIdParameterExperimental } from "../constructs/riff-raff-deployment-id";
+
+export interface GuLoadBalancedAppExperimentalProps extends AppIdentity {
+  /**
+   * Network access restrictions for your load balancer.
+   *
+   * Note, this merely provides defence in depth; you should, for example, limit
```
akash1810: minor: I think this doc string is incorrect. It's also difficult to read IMO as it suggests VPN restriction is sufficient, then states what should be done instead 😵.

We'd want to fix the original docstring too.

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -0,0 +1,879 @@
...
+   * Specify private subnets if using a non-default VPC or (generally
+   * discouraged) to limit to a subset of the available subnets.
+   */
+  privateSubnets?: ISubnet[];
+
+  /**
+   * Specify private subnets if using a non-default VPC or (generally
+   * discouraged) to limit to a subset of the available subnets.
+   */
+  publicSubnets?: ISubnet[];
```
akash1810, line 113: Should we take this opportunity to simplify these VPN properties? One possibility could be to remove the props and if a custom VPC is required, we escape hatch by changing the `default` of the CFN parameters?

src/experimental/patterns/gu-load-balanced-app.ts
akash1810: I wonder if "ratio" rather than "weight" is more intuitive phrasing? Alas, AWS use "weight", so it'll be simpler to use the same.

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -0,0 +1,879 @@
...
+}
+
+export class GuLoadBalancedAppExperimental extends Construct {
+  public readonly vpc: IVpc;
+  public readonly certificate?: GuCertificate;
+  public readonly loadBalancer: GuApplicationLoadBalancer;
+  public readonly autoScalingGroup?: GuAutoScalingGroup;
+  public readonly ecsService?: FargateService;
+  public readonly listener: GuHttpsApplicationListener;
+  public readonly targetGroups: TargetGroups;
```
akash1810, line 388: Could we add doc strings to these public properties? (Maybe we should tweak [this eslint rule](https://github.com/guardian/cdk/blob/50ceac8c3f475eda3748b7dea3f1680725786a6f/eslint.config.mjs#L37-L42)?)

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -0,0 +1,879 @@
...
+            app: `${app}-ecs`,
+            vpc,
+          }),
+        ],
+      });
+
+      ecsService.autoScaleTaskCount({
+        minCapacity: scaling.minimumTasks,
+        maxCapacity: scaling.maximumTasks,
+      });
```
akash1810, line 606: Should we validate max >= min? And/or default `max` to be relative to `min`, [as we do for EC2](https://github.com/guardian/cdk/blob/50ceac8c3f475eda3748b7dea3f1680725786a6f/src/patterns/ec2-app/base.ts#L389)?

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -0,0 +1,879 @@
...
+
+      this.ecsService = ecsService;
+
+      // It's possible to opt-out of log shipping to ELK in the EC2 patterns; should we mirror that here
+      const logRouter = taskDefinition.addFirelensLogRouter("LogShipping", {
+        // See https://github.com/guardian/devx-logs
+        image: ContainerImage.fromRegistry("ghcr.io/guardian/devx-logs:2.1.0"),
```
akash1810: Shall we add a TODO here to reference the image by its digest?

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -0,0 +1,879 @@
...
+     *
+     * @see https://docs.docker.com/dhi/core-concepts/digests
+     */
+    imageIdentifier: string;
+    cpu: number;
+    memoryLimitMiB: number;
+    /**
+     * ECR repository name which contains your images.
+     */
+    repositoryName: string;
```
akash1810: Could we default this to the [`GuStack`'s repository](https://github.com/guardian/cdk/blob/50ceac8c3f475eda3748b7dea3f1680725786a6f/src/constructs/core/stack.ts#L178-L197)?

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -0,0 +1,879 @@
...
+      const logRouter = taskDefinition.addFirelensLogRouter("LogShipping", {
+        // See https://github.com/guardian/devx-logs
+        image: ContainerImage.fromRegistry("ghcr.io/guardian/devx-logs:2.1.0"),
+
+        // Required by https://github.com/guardian/devx-logs
+        environment: {
+          STACK: scope.stack,
+          STAGE: scope.stage,
+          APP: app,
+          GU_REPO: repositoryName,
```
akash1810: minor: Should this be read from `scope`, to match the [`gu:repo` tag](https://github.com/guardian/cdk/blob/50ceac8c3f475eda3748b7dea3f1680725786a6f/src/constructs/core/stack.ts#L165-L168) that'll be applied?

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -0,0 +1,879 @@
...
+      if (targetGroupWeights.ec2 + targetGroupWeights.ecs !== 999) {
+        throw new Error("Combined target group weights for EC2 and ECS must be equal to 999");
+      }
+      defaultAction = ListenerAction.weightedForward([
+        { targetGroup: targetGroups.ec2, weight: targetGroupWeights.ec2 },
+        { targetGroup: targetGroups.ecs, weight: targetGroupWeights.ecs },
+      ]);
+    } else {
+      throw new Error("At least one of 'ec2Props' or 'ecsProps' must be specified");
+    }
```
akash1810: Could this be extracted to a function of its own for the benefit of terse unit tests?

src/experimental/patterns/gu-load-balanced-app.test.ts
```diff
@@ -0,0 +1,1541 @@
...
+                  Ref: "EcsTargetGroupTestguFCB7574C",
+                },
+                Weight: 500,
+              },
+            ],
+          },
+        },
+      ],
+    });
+  });
```
akash1810, line 80: Is there a way to relate the target group ARN to it's origin (EC2/ECS)? We're currently testing that the 499/500 input is observed, but without checking the origin of the target group we can't assert the weighting is correctly applied.

src/experimental/patterns/gu-load-balanced-app.ts
jacobwinch: Thanks, yes we need to include the changes in https://github.com/guardian/cdk-playground/pull/1120 once it's merged and I've added a reminder for that now.

src/experimental/constructs/riff-raff-deployment-id.ts
jacobwinch: > Would it be worth making this "stable" in its own release?

I've kept this in experimental for now because it's only needed by experimental patterns at the moment; I don't think advertising its existence to users is helpful at this stage.

> Maybe even adding it to GuStack by default as Riff-Raff would always fill it in (https://github.com/guardian/riff-raff/pull/1469).

I'm reluctant to do this because IIUC it will cause snapshot tests to fail for _all_ projects when they upgrade? I think we should avoid changing the 'stable' parts of the library unless there is a clear benefit to users, whereas I think the [small metadata change](https://github.com/guardian/cdk/pull/2904/changes#diff-1f091e2da91e736513871b044acf9638802f0e6eef880785026510821c3b3ff4) for users who've intentionally used experimental features is acceptable.

src/experimental/patterns/gu-load-balanced-app.ts
jacobwinch: Good catch, we seem to be missing a very important 'not' here! I'll update this.

jacobwinch: https://github.com/guardian/cdk/pull/2904/changes/4eb24fbcdf4bbbf2ca1882dba193b478832ac3f5 & https://github.com/guardian/cdk/pull/2910

src/experimental/patterns/gu-load-balanced-app.ts
jacobwinch: For now I think we should focus on making it as easy as possible for any existing user to move from `GuEc2App` to `GuLoadBalancedApp` (once it's stable). At the moment all users will just need to [move some existing props down a level](https://github.com/guardian/cdk-playground/pull/1104/changes#diff-5ab716a433d4ed4816c22ae39141b7146b4d1c1a8117eefa8ce7dab4cb4dbf89) and I'm reluctant to add any further caveats that would add friction to an ECS migration.

I think we could look to drop these props in a breaking change further down the line as I agree that using escape hatches would have been fine for this scenario.

src/experimental/patterns/gu-load-balanced-app.ts
jacobwinch: Good question. I think that relative default is needed to ensure that ASGs have enough 'headroom' for Riff-Raff's `autoscaling` deployment algorithm (where we always double capacity and max is a hard limit that cannot be exceeded).

In this scenario max seems to be a soft limit:

https://github.com/guardian/cdk/blob/4eb24fbcdf4bbbf2ca1882dba193b478832ac3f5/src/experimental/patterns/gu-load-balanced-app.ts#L331-L335

So I actually think if you want to run a single task (and you never want to automatically scale) then setting both of `minimumTasks` and `maximumTasks` to 1 is fine.

src/experimental/patterns/gu-load-balanced-app.test.ts
jacobwinch: I needed some help from Copilot to get the right syntax, but perhaps this is better: https://github.com/guardian/cdk/pull/2904/changes/327a5c869080b2163ccac63bb037a6f7b5ab3085

jacobwinch: Ideally this would be clear from the logical ID names alone, but unfortunately we can't add the `EC2` prefix without replacing existing target groups.

src/experimental/patterns/gu-load-balanced-app.ts
jacobwinch: https://github.com/guardian/cdk/pull/2904/changes/b2687b05db7f1ee6a5cb7a5971bb0436a9ae6171

src/experimental/patterns/gu-load-balanced-app.ts
jacobwinch: https://github.com/guardian/cdk/pull/2904/changes/b2687b05db7f1ee6a5cb7a5971bb0436a9ae6171

jacobwinch: Thanks - the fact that we now have a GitHub repo _and_ an ECR repo keeps catching me out!

src/experimental/patterns/gu-load-balanced-app.ts
jacobwinch: https://github.com/guardian/cdk/pull/2904/commits/ed98024b8e64cd3d36d9fbb9f038a478d0f2cbd9

Personally I find testing the overall pattern behaviour the most helpful (as it also serves as documentation), but I agree that moving this into a separate function is an improvement.

src/experimental/patterns/gu-load-balanced-app.ts
jacobwinch: https://github.com/guardian/cdk/pull/2904/changes/21ef9d96a4b82371ae76ece030b7ab15c606766a

src/experimental/patterns/gu-load-balanced-app.ts
jacobwinch: These logging changes have been tested now so I've updated this: https://github.com/guardian/cdk/pull/2904/changes/72fd729589fb56a796b74eb8dffe73a82251f9f7

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -0,0 +1,927 @@
...
+          userPool: userPool,
+          userPoolClient: userPoolClient,
+          userPoolDomain: userPoolDomain,
+          next: defaultAction,
+          sessionTimeout: Duration.minutes(sessionTimeoutInMinutes),
+        }),
+      });
+
+      // Need to give the ALB outbound access on 443 for the IdP endpoints.
+      const idpEgressSecurityGroup = new GuHttpsEgressSecurityGroup(scope, "ldp-access", {
```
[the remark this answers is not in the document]
jacobwinch: Fixing this typo (which originates [here](https://github.com/guardian/cdk/blob/50ceac8c3f475eda3748b7dea3f1680725786a6f/src/patterns/ec2-app/base.ts#L677-L681)) is going to change a logical ID and force the security group to be replaced as part of a migration. 

I think it's better to live with the typo.

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -0,0 +1,927 @@
...
+
+      userPoolClient.node.addDependency(userPoolIdp);
+
+      if (props.ecsProps) {
+        throw new Error("Using Google Auth with ECS is currently unsupported");
+        // I think that the code here will probably work fine but it needs some dedicated testing.
+        // For Google auth to work we need to 'authenticate-cognito' and then forward to a target group.
+        // If we are operating with EC2 and ECS backends then this forwarded request could be sent to either backend.
+        // Let's deploy this to a DevX service and check it's valid/works as expected before opening this up to other teams.
+      }
```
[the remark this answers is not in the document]
jacobwinch: This is a temporary safety measure and locating it here makes it easier for people working on this code to understand why the restriction currently exists.

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -0,0 +1,927 @@
...
+    return ListenerAction.forward([targetGroups.ec2]);
+  } else if (targetGroups.ecs && !targetGroups.ec2) {
+    return ListenerAction.forward([targetGroups.ecs]);
+  } else if (targetGroups.ec2 && targetGroups.ecs) {
+    if (!targetGroupWeights) {
+      throw new Error("EC2 and ECS are both enabled but no target group weights were provided");
+    }
+    if (targetGroupWeights.ec2 + targetGroupWeights.ecs !== 999) {
+      throw new Error("Combined target group weights for EC2 and ECS must be equal to 999");
+    }
```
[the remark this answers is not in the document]
jacobwinch: This needs more thought but at least restricts errors to synth-time (fast feedback loop). The suggestion here allows for deploy time errors, which are much more time consuming (if these weights sum to more than 999 the deployment will fail; AWS CDK doesn't document this but I hit this when testing).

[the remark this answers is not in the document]
jacobwinch: Not sure why Copilot says this is undocumented and then refers to the docs, or why it refers to a passing test as an example of getting the weights wrong 🙃

src/experimental/patterns/gu-load-balanced-app.ts
jacobwinch: I'm still thinking about the least-bad way to present this given the AWS model; I don't want that thinking to block this PR so will try to address it in a follow-up.

akash1810 (approved):

PR #2908, opened 2026-05-21
Add support for EC2 versioned deployments to GuLoadBalancedAppExperimental

jacobwinch:
## What does this change?

This PR is a follow-up to https://github.com/guardian/cdk/pull/2904. It unblocks the EC2 to ECS migration for users who currently rely on `GuEc2AppExperimental` (i.e. the [versioned deployment mechanism](https://docs.google.com/document/d/18M6nW6bknvFYjICMaSmDwkIisAuHsn4y_HSTdKu7Z-o/edit?tab=t.0#heading=h.fg4plmhts30o)).

## How to test

Similarly to https://github.com/guardian/cdk/pull/2904, I've brought across the existing test suite (with the slight amendments needed) to ensure that this functionality will be identical for users of the new pattern.

This has also been tested with a real app [here](https://github.com/guardian/cdk-playground/pull/1115) - only the [Metadata section](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/metadata-section-structure.html) is changing (no resources are modified).

We can support migrations for users who currently rely on versioned EC2 deployments. Users who want to adopt this feature _before_ moving to ECS (in order to keep deployments as fast as possible during the migration) are now also free to do this.

As with https://github.com/guardian/cdk/pull/2904, this PR is introducing a bit of duplication. To mitigate this risk, I've run the existing test suite against this new pattern. We should also pick up any edge cases / accidental changes via snapshot test failures when users migrate to this new pattern.

It's less clear to me how we should remove this duplication again in the future - we could consider removing `GuEc2AppExperimental` entirely at this stage so that we don't have to maintain this feature for 2 patterns? However, I think that would make it more challenging to advertise the deployment mechanism as stable without also marking this new pattern (including ECS functionality) as stable.

## Checklist

### DIFF
--- a/src/experimental/patterns/gu-load-balanced-app.test.ts
+++ b/src/experimental/patterns/gu-load-balanced-app.test.ts
@@ -1,13 +1,23 @@
-import { Duration } from "aws-cdk-lib";
+import { App, Duration } from "aws-cdk-lib";
 import { Match, Template } from "aws-cdk-lib/assertions";
-import { BlockDeviceVolume, EbsDeviceVolumeType, UpdatePolicy } from "aws-cdk-lib/aws-autoscaling";
+import {
+  BlockDeviceVolume,
+  type CfnAutoScalingGroup,
+  CfnScalingPolicy,
+  EbsDeviceVolumeType,
+  UpdatePolicy,
+} from "aws-cdk-lib/aws-autoscaling";
 import { InstanceClass, InstanceSize, InstanceType, Peer, Port, UserData, Vpc } from "aws-cdk-lib/aws-ec2";
 import type { CfnLoadBalancer } from "aws-cdk-lib/aws-elasticloadbalancingv2";
 import { AccessScope, MetadataKeys } from "../../constants";
-import { GuPrivateConfigBucketParameter } from "../../constructs/core";
+import { GuUserData } from "../../constructs/autoscaling";
+import type { NoMonitoring } from "../../constructs/cloudwatch";
+import { GuPrivateConfigBucketParameter, GuStack, type GuStackProps } from "../../constructs/core";
 import { GuSecurityGroup } from "../../constructs/ec2";
 import { GuDynamoDBWritePolicy } from "../../constructs/iam";
-import { GuTemplate, simpleGuStackForTesting } from "../../utils/test";
+import type { AppAccess, GuAsgCapacity } from "../../types";
+import { getTemplateAfterAspectInvocation, GuTemplate, simpleGuStackForTesting } from "../../utils/test";
+import { RollingUpdateDurations } from "./ec2-app";
 import { GuLoadBalancedAppExperimental } from "./gu-load-balanced-app";
 
 describe("the GuLoadBalancedAppExperimental pattern should support new ECS and hybrid functionality", function () {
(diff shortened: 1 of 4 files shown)


src/experimental/patterns/gu-load-balanced-app.test.ts
```diff
@@ -1582,3 +1592,554 @@ UserData from accessed construct`);
...
+  it("should throw an error if the user passes an updatePolicy whilst using versioned deployments", function () {
+    const stack = simpleGuStackForTesting({ env: { region: "eu-west-1" } });
+    expect(
+      () =>
+        new GuLoadBalancedAppExperimental(stack, {
+          ...topLevelPropsForVersionedDeployment(),
+          ec2Props: { updatePolicy: UpdatePolicy.replacingUpdate(), ...ec2PropsForVersionedDeployment(stack) },
+        }),
+    ).toThrow("If versionedDeployments are enabled then updatePolicy should not be set via ec2Props");
+  });
```
jacobwinch, line 1638: All tests/assertions after this point have come from [`experimental/patterns/ec2-app.test.ts`](https://github.com/guardian/cdk/blob/1e889541a85dcca73cb74a33a202df45cdb9d4f0/src/experimental/patterns/ec2-app.test.ts#L54) (I've just modified the test setup so that it works with `GuLoadBalancedAppExperimental` instead of `GuEc2AppExperimental`).

I wanted to make sure that I hadn't missed anything and this helped me to catch a few bugs (https://github.com/guardian/cdk/pull/2908/changes/b6312da26ec58e7f5f91016834ebd1861cdce1e2).

jorgeazevedo (approved):

jorgeazevedo:
> It's less clear to me how we should remove this duplication again in the future - we could consider removing GuEc2AppExperimental entirely at this stage so that we don't have to maintain this feature for 2 patterns? However, I think that would make it more challenging to advertise the deployment mechanism as stable without also marking this new pattern (including ECS functionality) as stable.

Apologies if I'm misremembering this, but I think the new deployment mechanism feature is going to be deprecated as we won't be able to make it work on a non-riff-raff world, right? I think that means the path forward is keeping both for now, but once we have a non experimental `GuLoadBalancedApp`  we deprecate `GuEc2AppExperimental` and move existing clients to `GuLoadBalancedApp` with the new `versionedDeployments` option . When we deprecate riff-raff we can then deprecate the `versionedDeployments` option. Clients will then go have the option to either go back to a basic deployment strategy, or upgrade to ECR to get identical behaviour.

jacobwinch:
> I think that means the path forward is keeping both for now, but once we have a non experimental `GuLoadBalancedApp` we deprecate `GuEc2AppExperimental` and move existing clients to `GuLoadBalancedApp` with the new `versionedDeployments` option .

Yes, that makes sense to me. Let's keep `GuEc2AppExperimental` separate for now so it's not coupled to the ECS work. It's still possible that we'll need to change our approach if the ECS trial doesn't work as well as we're hoping (e.g. due to cost).

If the ECS trial _does_ goes well, then I think we can release a stable `GuLoadBalancedApp` and migrate all current users of `GuEc2AppExperimental` to it before deleting the experimental Ec2App pattern.

> I think the new deployment mechanism feature is going to be deprecated as we won't be able to make it work on a non-riff-raff world, right? 

@jorgeazevedo and I also discussed this a bit in person - we might still need to support versioned EC2 deployments without Riff-Raff, but perhaps this would be simplified version compared to what we support today (e.g. Riff-Raff wouldn't be able to inject any parameter values, so users might need to do this themselves, or just lose that functionality).

PR #2921, opened 2026-06-29
fix(GuLoadBalancedAppExperimental): Validate each target group weight individually

akash1810:
## What does this change?
The [AWS docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-properties-elasticloadbalancingv2-listener-targetgrouptuple.html) and the AWS console suggests each target group weight must be between 0 and 999. Previously we incorrectly asserted the sum total weight was 999. Now we check each weight independently. This format is easier to convert to a percentage too. For example:

```ts
// Route 0.1% of traffic to ECS
targetGroupWeights: {
  ecs: 1,
  ec2: 999
}

// Route 1% of traffic to ECS
targetGroupWeights: {
  ecs: 1,
  ec2: 99
}

// Evenly route traffic
targetGroupWeights: {
  ecs: 1,
  ec2: 1
}
```

## How to test
See updated unit test.

Providing a simpler API.

N/A.

## Checklist

### DIFF
--- a/src/experimental/patterns/gu-load-balanced-app.ts
+++ b/src/experimental/patterns/gu-load-balanced-app.ts
@@ -357,8 +357,17 @@ export interface GuLoadBalancedAppExperimentalProps extends AppIdentity {
     };
   };
   /**
-   * If you are specifying `ec2Props` and `ecsProps` use these weights to distribute traffic across the different compute
-   * types. The weights must sum to 999.
+   * If you are specifying `ec2Props` and `ecsProps` use these weights to distribute traffic across the different compute types.
+   * Each weight must be between 0 and 999. AWS will work out the relative percentage.
+   *
+   * @example - Send 10% of traffic to ECS:
+   *
+   * ```ts
+   * targetGroupWeights: {
+   *   ecs: 1,
+   *   ec2: 9
+   * }
+   * ```
    */
   targetGroupWeights?: {
     ecs: number;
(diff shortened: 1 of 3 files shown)


src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -992,17 +1001,30 @@ function configureListenerActions(
...
     }
-    if (targetGroupWeights.ec2 + targetGroupWeights.ecs !== 999) {
-      throw new Error("Combined target group weights for EC2 and ECS must be equal to 999");
+
+    // An individual target can have a weight between 0 and 999. AWS will work out the relative percentage.
+    const isBetween = (n: number, min: number, max: number): boolean => n >= min && n <= max;
+    if (!isBetween(targetGroupWeights.ecs, 0, 999)) {
+      throw new Error("targetGroupWeights.ecs must be between 0 and 999");
+    }
+    if (!isBetween(targetGroupWeights.ec2, 0, 999)) {
```
akash1810, line 1024: AWS CDK doesn't throw when out of bounds here, so we'll do it. Interestingly, they [accept a range of 0 to 1000](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_elasticloadbalancingv2.WeightedTargetGroup.html), which is different from the [CloudFormation docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-properties-elasticloadbalancingv2-listener-targetgrouptuple.html) of 0-999.

jorgeazevedo: Yeah this is worth doing I think!

jorgeazevedo (approved):

PR #2929, opened 2026-07-02
feat(GuLoadBalancedAppExperimental): Apply `additionalPolicies` to ECS task role

akash1810:
> [!NOTE]
> Recommended to review [commit by commit](https://github.com/guardian/cdk/pull/2929/commits).

## What does this change?
Updates `GuLoadBalancedAppExperimental` to add the IAM Policies declared in the `additionalPolicies` prop to the ECS task role.

To achieve this, a couple of major changes are needed:
- The `props` interface has changed, removing the `roleConfiguration` object option, placing `additionalPolicies` at the root of the `props` and removing `withoutLogShipping`. As noted, a [GitHub search](https://github.com/search?q=org%3Aguardian+withoutLogShipping+NOT+repo%3Aguardian%2Fcdk++NOT+is%3Aarchived+NOT+repo%3Aguardian%2Faws-account-setup&type=code)
shows the `withoutLogShipping` property is never set by clients when instantiating a `GuEc2App` or `GuLoadBalancedAppExperimental`. That is, I don't expect clients to be directly impacted here.
- The `GuParameterStoreReadPolicy` construct, which provides permissions to read config from SSM Parameter Store, is implemented as a singleton. Previously, the ECS version erroneously changed the `app` path being granted; see also https://github.com/guardian/cdk-playground/pull/1145#pullrequestreview-4598913165. As noted, a [GitHub search](https://github.com/search?q=org%3Aguardian+GuParameterStoreReadPolicy+NOT+repo%3Aguardian%2Fcdk++NOT+is%3Aarchived&type=code) shows `GuParameterStoreReadPolicy` is never directly instantiated by clients. That is, I don't expect clients to be directly impacted here.

## How to test
See updated unit tests.

The snapshot changes are isolated to the ECS resources added in `GuLoadBalancedAppExperimental`. This demonstrates a no-op for all other infrastructure.

The ECS flavour of an application can obtain the necessary AWS permissions to run.

There are breaking changes to the public API. However I think the impact is isolated to one scenario: within the constructor of `GuEc2App` and `GuLoadBalancedAppExperimental`, the `additionalPolicies` are no longer nested under `roleConfiguration`.

### Before

```ts
new GuEc2App(this, {
  // other props
  roleConfiguration: {
    additionalPolicies: [
      new GuAllowPolicy(this, 'AllowPolicyCloudwatchLogs', {
        actions: ['cloudwatch:*', 'logs:*'],
        resources: ['*'],
      }),
      new GuAllowPolicy(
        this,
        'AllowPolicyDescribeDecryptKms',
        {
          actions: ['kms:Decrypt', 'kms:DescribeKey'],
          resources: [
            `arn:aws:kms:${region}:${account}:FrontendConfigKey`,
          ],
        },
      ),
    ],
  },
});
```

### After

```ts
new GuEc2App(this, {
  // other props
  additionalPolicies: [
    new GuAllowPolicy(this, 'AllowPolicyCloudwatchLogs', {
      actions: ['cloudwatch:*', 'logs:*'],
      resources: ['*'],
    }),
    new GuAllowPolicy(
      this,
      'AllowPolicyDescribeDecryptKms',
      {
        actions: ['kms:Decrypt', 'kms:DescribeKey'],
        resources: [
          `arn:aws:kms:${region}:${account}:FrontendConfigKey`,
        ],
      },
    ),
  ],
});
```

## Checklist

### DIFF
--- a/.changeset/sweet-chicken-travel.md
+++ b/.changeset/sweet-chicken-travel.md
@@ -1,0 +1,8 @@
+---
+"@guardian/cdk": major
+---
+
+Implement `GuParameterStoreReadPolicy` as a singleton to support `GuLoadBalancedAppExperimental` instantiating it once for the EC2 app and once for the ECS app.
+The result is a single `AWS::IAM::Policy` resource in the CloudFormation template, which is attached to both the EC2 and ECS roles, keeping the diff small.
+
+A [GitHub search](https://github.com/search?q=org%3Aguardian+GuParameterStoreReadPolicy+NOT+repo%3Aguardian%2Fcdk++NOT+is%3Aarchived&type=code) shows `GuParameterStoreReadPolicy` is never directly instantiated by clients.
--- a/src/constructs/iam/policies/parameter-store-read.ts
+++ b/src/constructs/iam/policies/parameter-store-read.ts
@@ -38,11 +38,43 @@ export class ReadParametersByName extends PolicyStatement {
  * configuration. See [[`ReadParametersByPath`]] and [[`ReadParametersByName`]] for more details.
  */
 export class GuParameterStoreReadPolicy extends GuAppAwareConstruct(GuPolicy) {
-  constructor(scope: GuStack, props: AppIdentity) {
+  /**
+   * A record of existing instances of this construct, per {@link GuStack}.
+   * This allows us to implement singleton like behaviour.
+   * @private
+   */
+  private static instances: WeakMap<GuStack, Record<string, GuParameterStoreReadPolicy>> = new WeakMap();
+
+  private constructor(scope: GuStack, props: AppIdentity) {
     super(scope, "ParameterStoreRead", {
       policyName: "parameter-store-read-policy",
       statements: [new ReadParametersByPath(scope, props), new ReadParametersByName(scope, props)],
       ...props,
     });
+  }
+
+  public static getInstance(stack: GuStack, props: AppIdentity): GuParameterStoreReadPolicy {
+    const maybeStackInstances = this.instances.get(stack);
+
+    if (!maybeStackInstances) {
+      const instance = new GuParameterStoreReadPolicy(stack, props);
+      this.instances.set(stack, { [props.app]: instance });
+      return instance;
+    }
+
+    const maybeInstance = maybeStackInstances[props.app];
+
+    if (!maybeInstance) {
+      const instance = new GuParameterStoreReadPolicy(stack, props);
+
+      this.instances.set(stack, {
+        ...maybeStackInstances,
+        [props.app]: instance,
+      });
+
+      return instance;
+    }
+
+    return maybeInstance;
   }
 }
--- a/src/constructs/iam/policies/parameter-store-read.ts
+++ b/src/constructs/iam/policies/parameter-store-read.ts
@@ -43,7 +43,7 @@ export class GuParameterStoreReadPolicy extends GuAppAwareConstruct(GuPolicy) {
    * This allows us to implement singleton like behaviour.
    * @private
    */
-  private static instances: WeakMap<GuStack, Record<string, GuParameterStoreReadPolicy>> = new WeakMap();
+  private static instances: Map<GuStack, Map<string, GuParameterStoreReadPolicy>> = new Map();
 
   private constructor(scope: GuStack, props: AppIdentity) {
     super(scope, "ParameterStoreRead", {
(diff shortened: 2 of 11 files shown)


.changeset/sweet-chicken-travel.md
```diff
@@ -0,0 +1,8 @@
+---
+"@guardian/cdk": major
```
akash1810, line 2: Practically speaking, no clients will be impacted by this change. However as its a change to the public API, semver says create a new major version.

src/patterns/ec2-app/base.ts
akash1810, line 1: The `GuEc2App` pattern is changed to maintain a similar prop API between it and `GuLoadBalancedAppExperimental`.

src/constructs/iam/policies/parameter-store-read.ts
```diff
@@ -38,11 +38,43 @@ export class ReadParametersByName extends PolicyStatement {
...
+
+  private constructor(scope: GuStack, props: AppIdentity) {
     super(scope, "ParameterStoreRead", {
       policyName: "parameter-store-read-policy",
       statements: [new ReadParametersByPath(scope, props), new ReadParametersByName(scope, props)],
       ...props,
     });
   }
+
+  public static getInstance(stack: GuStack, props: AppIdentity): GuParameterStoreReadPolicy {
```
akash1810, line 56: We can't use the [existing singleton helper function](https://github.com/guardian/cdk/blob/main/src/utils/singleton.ts) here as we need to instantiate across two axes: the `GuStack` and the `app`.

.changeset/sweet-chicken-travel.md
jorgeazevedo: When you say that no clients will be impacted by this change, you're referring to the removal of `withoutLogShipping`, which no client depends on, right? But am I right in thinking that every client of GuEc2App that relies on `roleConfiguration` needs to be patched? e.g. https://github.com/guardian/amigo/blob/18acf04b17992018b516a695533cd6e0747ab6dc/cdk/lib/amigo.ts#L319-L321

jorgeazevedo: Ah hold on I didn't realise there were 3 changesets! I thought I was commenting on the removal of `roleConfiguration`, but this one is about making `GuParameterStoreReadPolicy` a singleton, right? How does it work, will changeset release 2 major versions and a minor?

akash1810: > But am I right in thinking that every client of GuEc2App that relies on roleConfiguration needs to be patched?

Correct.

> How does it work, will changeset release 2 major versions and a minor?

A new library is only created once we merge the PR changeset will [create](https://github.com/guardian/cdk/pulls?q=is%3Apr+%22Release+package+updates%22+is%3Aclosed) when we merge to main.

The changeset PR will accumulate all the changes merged to main into a single new version of the library.

src/constructs/iam/policies/parameter-store-read.ts
```diff
@@ -38,11 +38,43 @@ export class ReadParametersByName extends PolicyStatement {
  * configuration. See [[`ReadParametersByPath`]] and [[`ReadParametersByName`]] for more details.
  */
 export class GuParameterStoreReadPolicy extends GuAppAwareConstruct(GuPolicy) {
-  constructor(scope: GuStack, props: AppIdentity) {
+  /**
+   * A record of existing instances of this construct, per {@link GuStack}.
+   * This allows us to implement singleton like behaviour.
+   * @private
+   */
+  private static instances: WeakMap<GuStack, Record<string, GuParameterStoreReadPolicy>> = new WeakMap();
```
jorgeazevedo: I found this a bit counterintuive. I'd not heard of WeakMap before, but in the documentation it says

> an object's presence as a key in a WeakMap does not prevent the object from being garbage collected

So I think this can be useful in long running applications, but in a cdk context where we call `synth`, can we get away with a regular `Map`?

Similarly, could we use a `Map` instead of `Record`? So something like

```ts
private static instances: Map<GuStack, Map<string, GuParameterStoreReadPolicy>> = new Map();
```

The extra `.get()`, `.set()`, `.has()` methods being available allow the implementation of `getInstance()` below to be simplified I think.

akash1810: > can we get away with a regular `Map`?

For sure! I used a `WeakMap` after Copilot's suggestion; I didn't really think about the context 😬.

> Similarly, could we use a `Map` instead of `Record`? The extra `.get()`, `.set()`, `.has()` methods being available allow the implementation of `getInstance()` below to be simplified I think.

Done in https://github.com/guardian/cdk/pull/2929/changes/af2012c121f2e40af1c2752c20d0ba68f9270fef. I don't think its any more or less complex.

jorgeazevedo (approved):

PR #2943, opened 2026-07-31
Add log link to alarms

aug24:
## What does this change?

This adds a useful 3am feature to an alarm: a direct deep link to the relevant logging.

It either takes a complete link, or (and this should be sufficient for most purposes) the ELK space in which to search for app/stack/stage.

## Checklist

### DIFF
--- a/src/constructs/cloudwatch/alarm.ts
+++ b/src/constructs/cloudwatch/alarm.ts
@@ -1,13 +1,19 @@
-import { Alarm } from "aws-cdk-lib/aws-cloudwatch";
 import type { AlarmProps } from "aws-cdk-lib/aws-cloudwatch";
+import { Alarm } from "aws-cdk-lib/aws-cloudwatch";
 import { SnsAction } from "aws-cdk-lib/aws-cloudwatch-actions";
-import { Topic } from "aws-cdk-lib/aws-sns";
 import type { ITopic } from "aws-cdk-lib/aws-sns";
+import { Topic } from "aws-cdk-lib/aws-sns";
 import type { AppIdentity, GuStack } from "../core";
 
 export interface GuAlarmProps extends AlarmProps, AppIdentity {
   snsTopicName: string;
   okAction?: boolean;
+  cta?: GuAlarmCtaProps;
+}
+
+export interface GuAlarmCtaProps {
+  link?: string;
+  elkSpace?: string;
 }
 
 export interface Http4xxAlarmProps extends Omit<
@@ -24,6 +30,29 @@ export interface Http5xxAlarmProps extends Omit<
 > {
   tolerated5xxPercentage: number;
   numberOfMinutesAboveThresholdBeforeAlarm?: number;
+}
+
+export class GuAlarmCta {
+  ctaLinks: string[];
+  constructor(scope: GuStack, props: GuAlarmCtaProps) {
+    const providedLink = props.link ? [props.link] : [];
+    const generatedLink =
+      props.elkSpace && scope.app
+        ? [
+            [
+              `https://logs.gutools.co.uk/s/${props.elkSpace}/app/discover#/?`,
+              "_g=(filters:!(),refreshInterval:(pause:!t,value:60000),time:(from:now-1d,to:now))",
+              "&",
+              "_a=(columns:!(stack,stage,message,app),filters:!(",
+              `('$state':(store:appState),meta:(alias:!n,disabled:!f,key:stack.keyword,negate:!f,params:(query:${scope.stack}),type:phrase),query:(match_phrase:(stack.keyword:${scope.stack}))),`,
+              `('$state':(store:appState),meta:(alias:!n,disabled:!f,key:app.keyword,negate:!f,params:(query:${scope.app}),type:phrase),query:(match_phrase:(app.keyword:${scope.app}))),`,
+              `('$state':(store:appState),meta:(alias:!n,disabled:!f,key:stage.keyword,negate:!f,params:(query:${scope.stage}),type:phrase),query:(match_phrase:(stage.keyword:${scope.stage}))))`,
+              ",hideChart:!t,interval:auto,query:(language:kuery,query:exception),sort:!(!('@timestamp',desc)))",
+            ].join(""),
+          ]
+        : [];
+    this.ctaLinks = [...providedLink, ...generatedLink];
+  }
 }
 
 /**
--- a/src/constructs/cloudwatch/alarm.ts
+++ b/src/constructs/cloudwatch/alarm.ts
@@ -12,7 +12,7 @@ export interface GuAlarmProps extends AlarmProps, AppIdentity {
 }
 
 export interface GuAlarmCtaProps {
-  link?: string;
+  runbook?: string;
   elkSpace?: string;
 }
 
@@ -36,7 +36,7 @@ export class GuAlarmCta {
   ctaLinks: string[];
   markdown: string;
   constructor(scope: GuStack, props: GuAlarmCtaProps) {
-    const providedLink = props.link;
+    const providedRunbook = props.runbook;
     const generatedLink =
       props.elkSpace && scope.app
         ? [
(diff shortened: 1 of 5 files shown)


aug24:
I think it might be better if this is moved to a separate class, with a properties object requiring space, app, stack, stage.  Not sure.

src/constructs/cloudwatch/alarm.ts
```diff
@@ -8,6 +8,12 @@ import type { AppIdentity, GuStack } from "../core";
 export interface GuAlarmProps extends AlarmProps, AppIdentity {
   snsTopicName: string;
   okAction?: boolean;
+  logLink?: GuLogLinkProps;
```
akash1810: Our Central ELK stack does not have role based access. As a result, it is not used by some teams who, for example, cannot guarantee the absence of PII in application logs.

For this reason, I'd be tempted to adjust this to be more generic, for example:

```ts
export interface GuAlarmProps extends AlarmProps, AppIdentity {
  /**
   * The name (not arn) of the topic to report alarm status on.
   * This topic should be subscribed by necessary recipients.
   */
  snsTopicName: string;

  /**
   * Should an action be triggered if the alarm returns from breaching state into ok state?
   */   
  okAction?: boolean;

  /**
   * A call to action to help responders.
   * For example, this could be a link to a runbook, or a {@link createLogsUrl | link to logs in Central ELK}.
   * This will be added to the alarm's description field.
   */
  cta?: string;
}

export function createLogsUrl(): string {
   return "https://logs.gutools.co.uk ...";
}
```

src/constructs/cloudwatch/lambda-alarms.ts
```diff
@@ -41,7 +49,7 @@ export class GuLambdaErrorPercentageAlarm extends GuAlarm {
       comparisonOperator: ComparisonOperator.GREATER_THAN_THRESHOLD,
       evaluationPeriods: props.numberOfEvaluationPeriodsAboveThresholdBeforeAlarm ?? 1,
       alarmName: props.alarmName ?? defaultAlarmName,
-      alarmDescription: props.alarmDescription ?? defaultDescription,
+      alarmDescription: `${alarmDescription}${logLink}`,
```
akash1810: minor: we can simplify this a little by removing the leading `\n\n` on `logLink` and doing:

```ts
alarmDescription: [alarmDescription, logLink].join('\n\n'),
```

src/constructs/cloudwatch/alarm.ts
aug24: I've done this for an initial version of `GuAlarmCta`.  We can expand it as needed once its basic version works.

src/constructs/cloudwatch/alarm.ts
```diff
@@ -26,6 +32,29 @@ export interface Http5xxAlarmProps extends Omit<
   numberOfMinutesAboveThresholdBeforeAlarm?: number;
 }
 
+export class GuAlarmCta {
+  ctaLinks: string[];
```
aug24, line 36: Makes more sense if it creates zero or more links.

akash1810 (commented):
After noticing that [guardian/anghammarad](https://github.com/guardian/anghammarad/blob/06aefe123cf98cbc90f27b8aeb5ea857c76dab62/cloudformation/cfn.yaml#L108-L113) does something similar, I wonder if the current implementation of `GuAlarm` already supports this idea? That is, if clients want to add CTAs to their alarm, they'd set the `alarmDescription` property appropriately.

What if this change only provided a function to generate a deep-link to Central ELK, which clients could call when building `alarmDescription`?

aug24:
> After noticing that [guardian/anghammarad](https://github.com/guardian/anghammarad/blob/06aefe123cf98cbc90f27b8aeb5ea857c76dab62/cloudformation/cfn.yaml#L108-L113) does something similar, I wonder if the current implementation of `GuAlarm` already supports this idea? That is, if clients want to add CTAs to their alarm, they'd set the `alarmDescription` property appropriately.

I had in mind that this would be a first pass which made it easy to link to logs, and perhaps turn it into markdown links or similar to make it prettier on a second pass.  Might be better to do this immediately.

> 
> What if this change only provided a function to generate a deep-link to Central ELK, which clients could call when building `alarmDescription`?

Honestly I prefer this as it's structured, but I'll take direction.  I'd not do what Anghammarad does, because it won't support a CODE/PROD distinction from the same template.  Anghammarad could just have `cta: {elkSpace: "devx"}` which I prefer.

akash1810:
> I'd not do what Anghammarad does, because it won't support a CODE/PROD distinction from the same template.

Is that correct? I'd have thought we'd do something like this:

```ts
export class Anghammarad extends GuStack {
    super(scope, id, props);
    const { app, stack, stage } = this;
  
    const alarm = new GuAlarm({
      alarmDescription: [
        `Triggers if Anghammarad failed to process some messages in ${stage}.`,
        'For suggested actions, see: https://docs.google.com/document/d/1cDd9mVAKFAYibuBPiCw4HkSCl16dG5zSu0qdSXDPuSE/edit#',
        `To check if there are any failed messages in the logs, see: ${getDeepLinkToLogs(app, stack, stage)}`,
      ].join('\n');
    });
  }
}


const codeStack = new Anghammarad(scope, 'anghammarad-CODE', { 
  app: 'anghammarad',
  stack: 'deploy',
  stage: 'CODE'
});

const prodStack = new Anghammarad(scope, 'anghammarad-PROD', { 
  app: 'anghammarad',
  stack: 'deploy',
  stage: 'PROD'
});
```

aug24:
I see what you mean.  But isn't this even nicer for people to use generically?
```
export class Anghammarad extends GuStack {
    super(scope, id, props);
    const { app, stack, stage } = this;
  
    const alarm = new GuAlarm({
      alarmDescription: `Triggers if Anghammarad failed to process some messages in ${stage}.`,
      cta: {
        "link": "https://docs.google.com/document/d/1cDd9mVAKFAYibuBPiCw4HkSCl16dG5zSu0qdSXDPuSE/edit#",
        "elkSpace": "whatever",
    });
  }
}
```

Which would generate the following:
```
Triggers if Anghammarad failed to process some messages in PROD.

---
[link](https://docs.google.com/document/d/1cDd9mVAKFAYibuBPiCw4HkSCl16dG5zSu0qdSXDPuSE/edit#)
[logs](http://logs.gutools.co.uk/?space=whatever&stage=PROD&blahblahblah)
```

We could/should rename `[link]` to `[runbook]`, I think.  It encourages people to use it correctly.  Ideally, each alarm would link to a specific section.

src/constructs/cloudwatch/alarm.ts
```diff
@@ -26,6 +32,34 @@ export interface Http5xxAlarmProps extends Omit<
   numberOfMinutesAboveThresholdBeforeAlarm?: number;
 }
 
+export class GuAlarmCta {
```
akash1810, line 35: Do we need a class here? I think this could be simplified as a function.

src/constructs/cloudwatch/alarm.ts
```diff
@@ -26,6 +32,34 @@ export interface Http5xxAlarmProps extends Omit<
...
+            ",hideChart:!t,interval:auto,query:(language:kuery,query:exception),sort:!(!('@timestamp',desc)))",
+          ].join("")
+        : undefined;
+    this.ctaLinks = [providedRunbook, generatedLink].filter((link): link is string => !!link);
+
+    const markdownLinks = [
+      providedRunbook ? `[runbook](${providedRunbook})` : undefined,
+      generatedLink ? `[logs](${generatedLink})` : undefined,
+    ].filter((link): link is string => !!link);
+    this.markdown = markdownLinks.length > 0 ? ["---", ...markdownLinks].join("\n") : "";
```
akash1810, line 64: Alert messages go to SNS and that SNS topic has an email subscription. Can email clients render this markdown?

akash1810 (commented):
Let's release a canary version and test on a real service? This can be done by adding the `🐥 Canaries` label to the PR.

src/constructs/cloudwatch/alarm.ts
aug24: That's the whole point of markdown, right?  It's readable if not rendered.

akash1810: True. However, if it can never be rendered, should we be using plain text?

PR #2944, opened 2026-08-03
fix(GuLoadBalancedAppExperimental): Ensure all taggable resources use standard Guardian tags

akash1810:
## What does this change?
Updates all resources created by `GuLoadBalancedAppExperimental` with standard Guardian tags. Namely `App` is added to:
- `AWS::ECS::Cluster`
- `AWS::ECS::TaskDefinition`
- `AWS::ECS::Service`

Other resources, for example [`AWS::ElasticLoadBalancingV2::LoadBalancer`](https://github.com/guardian/cdk/blob/6cd190b1304da630e2b533ae300dc00d0840fa51/src/constructs/loadbalancing/alb/application-load-balancer.ts#L59), already have Guardian tags via the [`GuAppAwareConstruct` mixin](https://github.com/guardian/cdk/blob/6cd190b1304da630e2b533ae300dc00d0840fa51/src/utils/mixin/app-aware-construct.ts).

> [!NOTE]
> Previously, these resources implicitly inherited the `App` tag of the parent CloudFormation stack.
The CloudFormation's `App` tag is implicitly set by Riff-Raff via the contents of the `riff-raff.yaml` configuration. For example, in [this CloudFormation deployment of a DCR service](https://github.com/guardian/dotcom-rendering/blob/792024d40354ae2dbaa90d5ef8ae5a34eda7ea1f/dotcom-rendering/scripts/deploy/riff-raff.yaml#L70-L78), Riff-Raff sets `App` to `tag-page-rendering-cfn`. In this instance `tag-page-rendering` is the desired `App` value.

To support this, the [`hasGuTaggedResource` assertion](https://github.com/guardian/cdk/blob/6cd190b1304da630e2b533ae300dc00d0840fa51/src/utils/test/assertions.ts#L67) needed tweaking to perform an "includes" [match](https://github.com/aws/aws-cdk/blob/a5148255673fe7d4f37d229ba2839f3ab7a3b707/packages/aws-cdk-lib/assertions/README.md) (previously it was an "exact" match). This is because AWS CDK adds tags to some resources, for example [`Name` is added to a `AWS::CertificateManager::Certificate` resource](https://github.com/aws/aws-cdk/blob/a5148255673fe7d4f37d229ba2839f3ab7a3b707/packages/aws-cdk-lib/aws-certificatemanager/lib/certificate.ts#L363).

## How to test
See added unit test.

Tags of a resource are explicitly created and are reflected in the snapshot test.

N/A.

## Checklist

### DIFF
--- a/src/experimental/patterns/gu-load-balanced-app.test.ts
+++ b/src/experimental/patterns/gu-load-balanced-app.test.ts
@@ -1,4 +1,4 @@
-import { App, Duration } from "aws-cdk-lib";
+import { App, CfnResource, Duration, TagManager } from "aws-cdk-lib";
 import { Match, Template } from "aws-cdk-lib/assertions";
 import {
   BlockDeviceVolume,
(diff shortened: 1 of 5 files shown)


src/experimental/patterns/gu-load-balanced-app.test.ts
```diff
@@ -41,6 +41,35 @@ describe("the GuLoadBalancedAppExperimental pattern should support new ECS and h
     expect(Template.fromStack(stack).toJSON()).toMatchSnapshot();
   });
 
+  it("should apply standard tags to all taggable resources", function () {
```
AshCorr, line 44: Nice!

AshCorr (approved):

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -755,6 +755,12 @@ export class GuLoadBalancedAppExperimental extends Construct {
         ...targetGroups,
         ecs: ecsTargetGroup,
       };
+
+      /**
+       * Specifically apply App tag to resources.
+       * Other resources obtain this tag by extending {@link GuAppAwareConstruct}
+       */
```
connoromalleyatwork: Trivial comment from me but here it goes: Since this lies within a method i don't think it should use the documentation comment format (even though it is a multiline comment)
```suggestion
      // Specifically apply App tag to resources.
      // Other resources obtain this tag by extending {@link GuAppAwareConstruct}
```

connoromalleyatwork (approved):

src/experimental/patterns/gu-load-balanced-app.ts
akash1810: Fair. I used it to make `{@link GuAppAwareConstruct}` clickable in the IDE. I only tested in IntelliJ. Will switch to traditional comments 👍🏽 .

PR #2949, opened 2026-08-05
feat(GuLoadBalancedAppExperimental): Add App to start of logical ID to influence target group name

akash1810:
## What does this change?
Changes the stem of the logical ID fo the target group created for ECS in the `GuLoadBalancedAppExperimental` pattern.

A target group has the following restrictions on its [name](https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-resource-elasticloadbalancingv2-targetgroup.html#cfn-elasticloadbalancingv2-targetgroup-name):
- Max 32 characters
- Must be regionally unique in an account

The former can be checked at compile time. The latter is only checked at deploy time. For this reason, we let CloudFormation auto-generate a name. The algorithm used to generate a name appears to be: first 6 alphanumeric characters of CloudFormation stack name, first 6 alphanumeric characters of logical ID, 12 character GUID.

Our CloudFormation stack names typically take the form `<stack>-<stage><app>`, for example `deploy-CODE-cdk-playground` or `deploy-PROD-prism`. For these stacks, the ECS target groups have names like `deploy-EcsTar-123456123456` and `deploy-EcsTar-123456123457`; it isn't possible to know which app these relate to at a glance.

With this change, the target group will have a logical ID stem of `MyAppEcsTargetGroup`, resulting in a name such as `deploy-MyAppE-123456123456`.

### Why not change all target groups?
We should consider target groups as a [stateful resource](https://github.com/guardian/cdk/blob/main/docs/stateful-resources.md) as they're tied to metrics used to understand the performance of an application. Changing all target groups (e.g. the ones created for EC2 apps via the `GuEc2App` pattern) would be a significant breaking change.

The ECS infrastructure created by the `GuLoadBalancedAppExperimental` pattern is not yet used in production, so we can afford to make this destructive change in this context.

> [!NOTE]
> This is not considered to be a breaking change as its an update to an experimental pattern.

## How to test
See updated tests.

Target group resources are more glanceable in the AWS console.

As we're changing the logical ID, this is a replacement operation. Brief testing suggests AWS will provision the new target group first and then delete the existing one. That is, there should not be impact to service availability.

## Checklist

### DIFF
--- a/.changeset/pink-rockets-sink.md
+++ b/.changeset/pink-rockets-sink.md
@@ -1,0 +1,20 @@
+---
+"@guardian/cdk": minor
+---
+
+Changes the stem of the logical ID fo the target group created for ECS in the `GuLoadBalancedAppExperimental` pattern.
+
+A target group has the following restrictions on its [name](https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-resource-elasticloadbalancingv2-targetgroup.html#cfn-elasticloadbalancingv2-targetgroup-name):
+- Max 32 characters
+- Must be regionally unique in an account
+
+The former can be checked at compile time. The latter is only checked at deploy time. For this reason, we let CloudFormation auto-generate a name. The algorithm used to generate a name appears to be: first 6 alphanumeric characters of CloudFormation stack name, first 6 alphanumeric characters of logical ID, 12 character GUID.
+Our CloudFormation stack names typically take the form `<stack>-<stage><app>`, for example `deploy-CODE-cdk-playground` or `deploy-PROD-prism`. For these stacks, the ECS target groups have names like `deploy-EcsTar-123456123456` and `deploy-EcsTar-123456123457`; it isn't possible to know which app these relate to at a glance.
+
+With this change, the target group will have a logical ID stem of `MyAppEcsTargetGroup`, resulting in a name such as `deploy-MyAppE-123456123456`.
+
+## Why not change all target groups?
+We should consider target groups as a [stateful resource](../docs/stateful-resources.md) as they're tied to metrics used to understand the performance of an application.
+Changing all target groups (e.g. the ones created for EC2 apps via the `GuEc2App` pattern) would be a significant breaking change.
+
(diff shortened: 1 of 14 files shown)


src/constructs/core/identity.ts
```diff
@@ -18,12 +18,26 @@ export const AppIdentity = {
...
+
+    return `${alphanumericTitleCaseApp}${text}`;
+  },
+
+  addAppToStringEnd(appIdentity: AppIdentity, text: string): string {
+    return AppIdentity.addAppToString(appIdentity, text, "end");
   },
+
+  addAppToStringStart(appIdentity: AppIdentity, text: string): string {
+    return AppIdentity.addAppToString(appIdentity, text, "start");
```
jorgeazevedo: Would you consider changing the way this is structured so that instead of having a share addAppToString with a string based switch to control behaviour, we implement the behaviour we want here but encapsulate building `alphanumericTitleCaseApp` behind a function?

So something like
```ts
addAppToStringStart(appIdentity: AppIdentity, text: string): string {
    const alphanumericTitleCaseApp = new_method(appIdentity)
    return `${text}${alphanumericTitleCaseApp}`
```

That way we avoid the conditional whilst avoiding duplication at the same time. What do you think?

.changeset/pink-rockets-sink.md
```diff
@@ -0,0 +1,20 @@
+---
```
jorgeazevedo, line 1: This is more a question of convention that it being right/wrong, but to me this feels like too much detail for a release changelog. From a client's perspective it's useful to understand what is going to change from installing this version. The reasoning for why the maintainers pushed this changed I think might be best homed in a PR description or an ADR.

src/constructs/core/identity.ts
akash1810: Good shout! Done!

.changeset/pink-rockets-sink.md
akash1810: Agreed. Updated now.

jorgeazevedo (approved):

jorgeazevedo:
Thanks for the changes!

PR #2950, opened 2026-08-05
feat: Add support for publishing canaries

AshCorr:
## What does this change?

Adds support to our release workflow to release Canaries, by labelling a PR with `🐥 Canaries`.

This publishes the PR branch to NPM under the `canary` tag (not latest, to avoid automated tooling picking it up as a valid version to upgrade to), and under a specific canary version such as `0.0.0-canary-20260805130549`.

After publishing the canary the Github action then removes the `🐥 Canaries` so that it may be re-applied again to trigger another release, and comments on the PR to let the author know the artifact version to use.

This allows us to more easily test CDK changes as we can now use these canary versions to test how the library perform in CODE in other repositories such as DCR.

I've shamelessly stolen most of the Canary behaviour from [guardian/csnx](https://github.com/guardian/csnx/blob/main/.github/workflows/npm-publish.yml) and [guardian/stand](https://github.com/guardian/stand)

## How do Canaries work?

Essentially 2 commands:

 - `npx changeset version --snapshot canary`: Prepares the project for a canary release by setting the version field in `package.json` to a canary version, eg `0.0.0-canary-20260805133757`
 - `npx changeset publish --tag canary`: Publish the canary version to NPM. Tagging the version as `canary` instead of the default `latest` to avoid confusion with automated tooling.

## How to test

Labelled this PR with `🐥 Canaries` and then used the subsequent artifact in another repository.

Example from DCR:

image

### DIFF
--- a/.github/workflows/cd.yaml
+++ b/.github/workflows/cd.yaml
@@ -1,12 +1,19 @@
 name: CD
 on:
-  push:
-    branches:
-      - main
-  workflow_dispatch:
+  # Trigger for canary release on PR label
+  pull_request:
+    types: [labeled, synchronize]
+
+  # Trigger for standard (changesets) release after CI on main
+  # If a changeset PR is merged to main, this will publish a new version of the package to npm
+  workflow_run:
+    workflows: [CI]
+    branches: [main]
+    types: [completed]
 
 jobs:
-  CD:
+  canaries:
+    if: ${{ github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, '🐥 Canaries') }}
     runs-on: ubuntu-latest
 
     permissions:
@@ -15,16 +22,11 @@ jobs:
       pull-requests: write
 
     steps:
-      - uses: actions/checkout@v7
-      - uses: actions/setup-node@v6
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      - uses: actions/setup-node@249970729cb0ef3589644e2896645e5dc5ba9c38 # v6.5.0
         with:
           node-version-file: ".tool-versions"
-          cache: npm
-
-      # See https://docs.npmjs.com/trusted-publishers
-      # Find the latest version with `npm info npm@latest version`
-      - name: Install suitable NPM version for trusted publishing
-        run: npm install -g npm@11.6.2
+          package-manager-cache: false # never use caching in release builds
 
       - name: Install
         run: npm ci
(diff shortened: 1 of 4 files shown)


.github/workflows/cd.yaml
```diff
@@ -32,11 +39,71 @@ jobs:
       - name: Build
         run: npm run build
 
-      - name: Lint
```
AshCorr, line 35: Since the workflow now only runs when CI passes (or when the Canaries label is applied) the check for lint and tests has already ran.

We should probably avoid running unnecessary commands in our pre-publish environment to keep it as clean as possible, so I've opted to remove these for now. [This is similar to how other projects have configured their canary releases](https://github.com/guardian/csnx/blob/main/.github/workflows/npm-publish.yml#L9-L12).

.github/workflows/cd.yaml
```diff
@@ -15,30 +22,81 @@ jobs:
...
-      - uses: actions/setup-node@v6
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      - uses: actions/setup-node@249970729cb0ef3589644e2896645e5dc5ba9c38 # v6.5.0
         with:
           node-version-file: ".tool-versions"
-          cache: npm
-
-      # See https://docs.npmjs.com/trusted-publishers
-      # Find the latest version with `npm info npm@latest version`
-      - name: Install suitable NPM version for trusted publishing
```
AshCorr, line 26: Our node version is high enough now that it comes with a version of NPM that supports trusted publishing.

.github/workflows/cd.yaml
```diff
@@ -1,12 +1,19 @@
 name: CD
 on:
-  push:
-    branches:
-      - main
```
AshCorr, line 5: Workflow is now triggered by either a label being applied to a PR, or CI finishing successfully on main.

.github/workflows/cd.yaml
```diff
@@ -15,30 +22,81 @@ jobs:
...
+      - uses: actions/setup-node@249970729cb0ef3589644e2896645e5dc5ba9c38 # v6.5.0
         with:
           node-version-file: ".tool-versions"
-          cache: npm
-
-      # See https://docs.npmjs.com/trusted-publishers
-      # Find the latest version with `npm info npm@latest version`
-      - name: Install suitable NPM version for trusted publishing
-        run: npm install -g npm@11.6.2
+          package-manager-cache: false # never use caching in release builds
```
AshCorr, line 29: https://docs.npmjs.com/trusted-publishers#step-2-configure-your-cicd-workflow

.github/workflows/cd.yaml
```diff
@@ -15,30 +22,81 @@ jobs:
...
       - name: Install
         run: npm ci
 
       - name: Build
         run: npm run build
 
-      - name: Lint
-        run: npm run lint
+      - name: Create Snapshot Version
+        run: npx changeset version --snapshot canary
```
AshCorr, line 38: Sets the `version` field of `package.json` to a canary version such as `0.0.0-canary-20260805133731`

akash1810 (approved):

PR #2954, opened 2026-08-12
Set containerInsights to enhanced and add Monitoring properties for ecsService

JuliaBrigitte:
## What does this change?
Sets containerInsights to enhanced and add Monitoring properties for ecsService as a default

We need to update [dotcom-rendering](https://github.com/guardian/dotcom-rendering/pull/16557), where it is currently set manually.

---
  - https://app.asana.com/0/0/1217199727794975

### DIFF
--- a/src/experimental/patterns/gu-load-balanced-app.ts
+++ b/src/experimental/patterns/gu-load-balanced-app.ts
@@ -10,7 +10,7 @@ import {
 import type { InstanceType, ISubnet, IVpc } from "aws-cdk-lib/aws-ec2";
 import { UserData } from "aws-cdk-lib/aws-ec2";
 import { Repository } from "aws-cdk-lib/aws-ecr";
-import { OperatingSystemFamily } from "aws-cdk-lib/aws-ecs";
+import { ContainerInsights, OperatingSystemFamily } from "aws-cdk-lib/aws-ecs";
 import { CpuArchitecture } from "aws-cdk-lib/aws-ecs";
 import { PropagatedTagSource } from "aws-cdk-lib/aws-ecs";
 import {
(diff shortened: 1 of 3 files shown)


src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -708,6 +710,24 @@ export class GuLoadBalancedAppExperimental extends Construct {
         ],
       });
 
+      const cfnCluster = cluster.node.defaultChild as CfnCluster;
+      cfnCluster.clusterSettings = [
+        {
+          name: "containerInsights",
+          value: "enhanced",
+        },
+      ];
```
akash1810: I think we could set [`containerInsightsV2`](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ecs.Cluster.html#containerinsightsv2) when instantiating a `Cluster`, in place of using the L1 construct?

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -708,6 +710,24 @@ export class GuLoadBalancedAppExperimental extends Construct {
...
+
+      const cfnService = ecsService.node.defaultChild as CfnService;
```
connoromalleyatwork, line 715: Looks good! I had a look and its annoying that aws-cdk doesn't let us pass this configuration to the `FargateService`. It's not even available as a mixin, even though this matches their description of a mixin use case. 

No action needed from this comment, just a comment:)

package.json
```diff
@@ -69,5 +69,6 @@
     "aws-cdk": "^2.1135.0",
     "aws-cdk-lib": "^2.263.0",
     "constructs": "^10.8.1"
-  }
+  },
+  "packageManager": "yarn@4.12.0+sha512.f45ab632439a67f8bc759bf32ead036a1f413287b9042726b7cc4818b7b49e14e9423ba49b18f9e06ea4941c1ad062385b1d8760a8d5091a1a31e5f6219afca8"
```
akash1810: I don't think this change is needed.

akash1810 (commented):
Other than an accidental edit of `package.json`, LGTM!

package.json
JuliaBrigitte: Oh, good catch, fixing that now, thanks.

akash1810 (approved):

PR #2958, opened 2026-09-01
Add artifacts account id instead of deploy tools

costasmantzoukas:
## What does this change?

In this PR the objective is the allow the `cdk` to pull the image digest in `GuLoadBalancedAppExperimental`, using the Artifacts AWS account. As part of this piece of work we had to introduce a new SSM property, namely `/organisation/accounts/artifacts`, to parameterize the name of the Artifacts AWS account.

## Checklist

---
  - https://app.asana.com/0/0/1217416594685413

### DIFF
--- a/.changeset/every-flies-battle.md
+++ b/.changeset/every-flies-battle.md
@@ -1,0 +1,5 @@
+---
+"@guardian/cdk": minor
+---
+
+Use Artifact account to publish an image to ECR
--- a/.changeset/every-flies-battle.md
+++ b/.changeset/every-flies-battle.md
@@ -2,4 +2,4 @@
 "@guardian/cdk": minor
 ---
 
-Use Artifact account to publish an image to ECR
+Pull images from the Artifacts AWS account within `GuLoadBalancedAppExperimental`.
--- a/src/constants/ssm-parameter-paths.ts
+++ b/src/constants/ssm-parameter-paths.ts
@@ -20,6 +20,7 @@ interface NamedSsmParameterPaths {
   FastlyCustomerId: SsmParameterPath;
   OrganisationDistributionBucket: SsmParameterPath;
   DeployToolsAccountId: SsmParameterPath;
+  ArtifactsAccountId: SsmParameterPath;
 }
 
 export const VPC_SSM_PARAMETER_PREFIX = "/account/vpc";
@@ -73,6 +74,10 @@ export const NAMED_SSM_PARAMETER_PATHS: NamedSsmParameterPaths = {
   DeployToolsAccountId: {
     path: "/organisation/accounts/deployTools",
     description: "Account ID of the Deploy Tools account. Useful as some shared resources live here.",
+  },
+  ArtifactsAccountId: {
+    path: "/organisation/accounts/artifacts",
+    description: "Account ID of the Artifacts account. Useful as some shared resources live here.",
   },
 };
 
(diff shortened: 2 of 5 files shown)


src/constants/ssm-parameter-paths.ts
```diff
@@ -74,6 +75,10 @@ export const NAMED_SSM_PARAMETER_PATHS: NamedSsmParameterPaths = {
     path: "/organisation/accounts/deployTools",
     description: "Account ID of the Deploy Tools account. Useful as some shared resources live here.",
   },
+  ArtifactsAccountId: {
+    path: "/organisation/accounts/artifacts",
```
costasmantzoukas: According to [this](https://github.com/guardian/aws-account-setup/blob/main/packages/cdk/lib/config/artifacts.ts) I was expecting to see a parameter with name `/organisation/accounts/artifacts` in SSM but was unable to find one. Not sure why that is. 

Image

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -593,13 +593,13 @@ export class GuLoadBalancedAppExperimental extends Construct {
...
       const image = ContainerImage.fromEcrRepository(
-        // Images are published to the ECR registry in the DeployTools account, so reference that here
+        // Images are published to the ECR registry in the Artifacts account, so reference that here
         Repository.fromRepositoryAttributes(this, "Repo", {
           repositoryArn: scope.formatArn({
             account: StringParameter.fromStringParameterName(
               scope,
-              "DeployToolsAccountId",
-              NAMED_SSM_PARAMETER_PATHS.DeployToolsAccountId.path,
+              "Artifact",
```
akash1810: minor: This string is used for the name of the CloudFormation parameter, could it be more explict?

```suggestion
              "ArtifactsAccountId",
```

src/constants/ssm-parameter-paths.ts
akash1810, line 79: Unfortunately this PR alone will not provision this parameter. We'd need to:
1. [Create a new release of this library](https://github.com/guardian/cdk/blob/main/docs/contributing.md#releasing)
2. Update the [version of the library being used in `guardian/aws-account-setup`](https://github.com/guardian/aws-account-setup/blob/cb1a6a241c4619f90b4f15ca53d0d9663c146e74/packages/cdk/package.json#L15)
3. Provision the SSM parameter in `guardian/aws-account-setup`; we can [copy how the deployTools parameter is provisioned](https://github.com/guardian/aws-account-setup/blob/cb1a6a241c4619f90b4f15ca53d0d9663c146e74/packages/cdk/lib/aws-account-setup.ts#L172-L177)

src/experimental/patterns/gu-load-balanced-app.ts
costasmantzoukas: Done!

.changeset/every-flies-battle.md
```diff
@@ -0,0 +1,5 @@
+---
+"@guardian/cdk": minor
+---
+
+Use Artifact account to publish an image to ECR
```
akash1810: This change relates to where an image is pulled from, rather than published to.


```suggestion
Pull images from the Artifacts AWS account within `GuLoadBalancedAppExperimental`.
```

akash1810 (commented):
Let's add some detail to the PR description to help any future travellers.

.changeset/every-flies-battle.md
costasmantzoukas: Done!

akash1810 (approved):

PR #2961, opened 2026-09-07
feat(`GuLoadBalancedAppExperimental`): Make the FireLens log driver accessible

akash1810:
## What does this change?
Updates `GuLoadBalancedAppExperimental` to make the FireLens log driver accessible. This allows clients to use it to ship the logs of their additional sidecars to Central ELK.

## How to test
N/A.

Reducing the amount of code in clients, for example [guardian/dotcom-rendering](https://github.com/guardian/dotcom-rendering/blob/cbd424c81ffbc090332d0a2063d4572eb3e6a470/dotcom-rendering/cdk/lib/renderingStack.ts#L323-L330).

N/A.

## Checklist

### DIFF
--- a/src/experimental/patterns/gu-load-balanced-app.ts
+++ b/src/experimental/patterns/gu-load-balanced-app.ts
@@ -421,6 +421,13 @@ export class GuLoadBalancedAppExperimental extends Construct {
    */
   public readonly targetGroups: TargetGroups;
 
+  /**
+   * A log driver used to ship logs to Central ELK (via Kinesis).
+   *
+   * @note Only available when using ECS, i.e. when {@link GuLoadBalancedAppExperimental.ecsService} is defined.
+   */
+  public readonly fireLensLogDriver?: FireLensLogDriver;
+
   constructor(scope: GuStack, props: GuLoadBalancedAppExperimentalProps) {
     const {
       access,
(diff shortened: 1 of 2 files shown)


src/experimental/patterns/gu-load-balanced-app.ts
aug24, line 624:
```suggestion

this.fireLensLogDriver = new FireLensLogDriver({
options: {
Name: `kinesis_streams`,
region: scope.region,
stream: loggingStreamName,
retry_limit: "2",
},
});

```

akash1810: Hmm, not too keen on this, as it means current usage https://github.com/guardian/cdk/blob/ba9cad3d54ac863a3fa33a6625165fd62f2c67cd/src/experimental/patterns/gu-load-balanced-app.ts#L652 would need updating to handle the possibility that `this.fireLensLogDriver` is `undefined`. Obviously _we_ know it is defined, by TypeScript doesn't. We'd either need an `if` statement, or use a non-null assertion (`this.fireLensLogDriver!`).

AshCorr: I would have imagined Typescript to know that `this.fireLensLogDriver` is not `undefined` considering we're setting it in the same scope that we're using it?

PR #2962, opened 2026-09-07
ECS CPU Target Scaling

connoromalleyatwork:
Add support for ECS scaling using CPU Utilisation as a target metric to `GuLoadBalancedAppExperimental`.

A mandatory field is added to `GuLoadBalancedAppExperimentalProps` called `cpuAvgUtilisationTarget`. This is a non backwards compatible change, however `GuLoadBalancedAppExperimental` is only being used by RelOps team to move tag page rendering over to ecs, and so there should be no need to make any other updates. The change is also in the `experimental` directory, and so we make no promises 🤭


src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -722,11 +734,22 @@ export class GuLoadBalancedAppExperimental extends Construct {
...
         maxCapacity: scaling.maximumTasks,
       });
 
+      // The high resolution predefined metric evaluates every 20s rather than 60s,
+      // so scaling reacts far faster. It isn't in the CDK `PredefinedMetric` enum yet.
+      // https://docs.aws.amazon.com/AmazonECS/latest/developerguide/target-tracking-faster-auto-scaling.html
+      new TargetTrackingScalingPolicy(this, "CpuScaling", {
+        scalingTarget: ecsScalableTarget,
+        targetValue: scaling.cpuAvgUtilisationTarget,
+        predefinedMetric: "ECSServiceAverageCPUUtilizationHighResolution" as unknown as PredefinedMetric,
```
akash1810: Could this metric be obtained via the [AWS CDK helper function](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ecs.FargateService.html#metricwbrcpuwbrutilizationprops)? Just as a way to remove a static string?

akash1810 (approved):
LGTM! Added a non-blocking comment.

src/experimental/patterns/gu-load-balanced-app.ts
connoromalleyatwork: I'm not sure. If we switch to cdk helper function we will for now have to set this up as a scaling policy using a custom metric, rather than a predefined metric. 

Ideally, the enum `PredefinedMetric` will get updated with this valid value soon.

akash1810: > Ideally, the enum `PredefinedMetric` will get updated with this valid value soon.

Related https://github.com/aws/aws-cdk/issues/38207.

AshCorr: Before we move this out of experimental we might want to make this an optional `scaling` prop so users can scale on other metrics.

AshCorr (approved):

src/experimental/patterns/gu-load-balanced-app.ts
connoromalleyatwork: That sounds great! We certainly want to make this more flexible in future

akash1810: Is it worth adding an inline TODO comment here so we remember to check on it in the future?

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -359,6 +361,32 @@ export interface GuLoadBalancedAppExperimentalProps extends AppIdentity {
        * the deployment through.
        */
       maximumTasks: number;
+      cpuScaling?: {
```
AshCorr, line 364: I would be tempted to make this a [`Pick<TargetTrackingScalingPolicyProps, 'targetValue' | 'scaleInCooldown' | 'scaleOutCooldown'>` type](https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys) so we end up with a similar API to the base CDK `TargetTrackingScalingPolicy` construct and automatically get the documentation for the parameters too.

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -722,11 +750,25 @@ export class GuLoadBalancedAppExperimental extends Construct {
...
       });
 
-      ecsService.autoScaleTaskCount({
+      const ecsScalableTarget = ecsService.autoScaleTaskCount({
         minCapacity: scaling.minimumTasks,
         maxCapacity: scaling.maximumTasks,
       });
 
+      if (scaling.cpuScaling) {
+        const cpuScaling = scaling.cpuScaling;
```
AshCorr, line 759: Maybe we can do some destructuring here?

```ts
const { scaleOutCooldown, scaleInCooldown } = scaling;
```

src/experimental/patterns/gu-load-balanced-app.ts
connoromalleyatwork: Great suggestion! Will switch to this as it achieves exactly what i was aiming for, i didn't know this existed:)

src/experimental/patterns/gu-load-balanced-app.ts
connoromalleyatwork: Nice have added it like this:
```ts
const { targetValue, scaleOutCooldown, scaleInCooldown } = scaling.cpuScaling;
```

src/experimental/patterns/gu-load-balanced-app.ts
connoromalleyatwork: Added a todo!

AshCorr (approved):
LGTM!

PR #2937, opened 2026-07-15
Include CHANGELOG.md in published npm package

jorgeazevedo:
## What does this change?

I've seen an AI agent look for the changelog in the root of the package for information about the currently installed version and why it's breaking. We include migration instructions in the release notes in the changelog, so I think this inclusion will improve agents ability to perform CDK upgrades without steering.

## How to test

On `main`
```bash
$ npm pack --dry-run --ignore-scripts 2>&1 | wc -l 
     864
$ npm pack --dry-run --ignore-scripts 2>&1 | grep CHANGELOG
$
```

On this branch
```bash
$ npm pack --dry-run --ignore-scripts 2>&1 | wc -l         
     865
$ npm pack --dry-run --ignore-scripts 2>&1 | grep CHANGELOG
npm notice 50.7kB CHANGELOG.md
$
```

## Checklist

- [x] I have listed any breaking changes, along with a migration path [^1]
- [x] I have updated the documentation as required for the described changes [^2]

[^1]: Consider whether this is something that will mean changes to projects that have already been migrated, or to the CDK CLI tool. If changes are required, consider adding a checklist here and/or linking to related PRs.
[^2]: If you are adding a new construct or pattern, has new documentation been added? If you are amending defaults or changing behaviour, are the existing docs still valid?

PR #2964, opened 2026-09-09
First step moving developer-policy out of experimental

JuliaBrigitte:
## What does this change?
Move the code out of the experimental directory
Next step will be to issue a new major version with release notes that detailed the migration path

## Checklist

---
  - https://app.asana.com/0/0/1217646438082602


.changeset/stale-readers-try.md
```diff
@@ -0,0 +1,8 @@
+---
+"@guardian/cdk": major
+---
+
+
+BREAKING CHANGE: GuDeveloperPolicyExperimental has been removed. Use GuDeveloperPolicy instead.
+
+Promote developer policy CDK construct from experimental to stable.
```
akash1810: I think we could be a bit terser here:

```suggestion
BREAKING CHANGE: `GuDeveloperPolicyExperimental` has been promoted to stable and renamed `GuDeveloperPolicy`
```

akash1810: This text is used to update the [CHANGELOG](https://github.com/guardian/cdk/blob/main/CHANGELOG.md). Since https://github.com/guardian/cdk/pull/2937, the CHANGELOG file is included in the NPM module to enable AI agents to perform updates. With that said, could we include an example of this migration? For example:

```ts
// Before
import { GuDeveloperPolicyExperimental } from '@guardian/cdk/lib/experimental/constructs/iam/policies';
new GuDeveloperPolicyExperimental();

// After
import { GuDeveloperPolicy } from '@guardian/cdk/iam/policies';
new GuDeveloperPolicy();
```

JuliaBrigitte: Will add an example, thanks for the suggestion.

PR #2965, opened 2026-09-11
Possible mount approach for s3 config

aug24:
## What does this change?

This PR adds capability to specify mount points in S3 as part of a task.

## Checklist

### DIFF
--- a/src/experimental/patterns/gu-load-balanced-app.test.ts
+++ b/src/experimental/patterns/gu-load-balanced-app.test.ts
@@ -41,6 +41,57 @@ describe("the GuLoadBalancedAppExperimental pattern should support new ECS and h
     expect(Template.fromStack(stack).toJSON()).toMatchSnapshot();
   });
 
+  it("should pass S3 Files volume configuration through to the ECS task definition", function () {
+    const stack = simpleGuStackForTesting({ env: { region: "eu-west-1" } });
+    new GuLoadBalancedAppExperimental(stack, {
+      monitoringConfiguration: { noMonitoring: true },
+      applicationPort: 3000,
+      access: { scope: AccessScope.PUBLIC },
+      app: "test-gu",
+      certificateProps: {
+        domainName: "domain-name-for-your-application.example",
+      },
+      ecsProps: {
+        cpu: 1024,
+        memoryLimitMiB: 2048,
+        scaling: { minimumTasks: 3, maximumTasks: 6 },
+        imageIdentifier: "sha256:12345",
+        s3FilesMounts: [
+          {
+            containerPath: "/amiable",
+            fileSystemArn: "arn:aws:s3files:eu-west-1:123456789012:file-system/fs-0123456789abcdef0",
+            rootDirectory: "/amiable",
+            accessPointArn: "arn:aws:s3files:eu-west-1:123456789012:file-system/fs-0123456789abcdef0/access-point/fsap-0123456789abcdef0",
+          },
+        ],
+      },
+    });
+
+    Template.fromStack(stack).hasResourceProperties("AWS::ECS::TaskDefinition", {
+      Volumes: Match.arrayWith([
+        Match.objectLike({
+          Name: "s3files-volume-0",
+          S3FilesVolumeConfiguration: {
+            FileSystemArn: "arn:aws:s3files:eu-west-1:123456789012:file-system/fs-0123456789abcdef0",
+            RootDirectory: "/amiable",
+            AccessPointArn: "arn:aws:s3files:eu-west-1:123456789012:file-system/fs-0123456789abcdef0/access-point/fsap-0123456789abcdef0",
+          },
+        }),
+      ]),
+      ContainerDefinitions: Match.arrayWith([
+        Match.objectLike({
+          MountPoints: Match.arrayWith([
+            Match.objectLike({
+              ContainerPath: "/amiable",
+              SourceVolume: "s3files-volume-0",
+              ReadOnly: true,
+            }),
+          ]),
+        }),
+      ]),
+    });
+  });
+
   it("should apply standard tags to all taggable resources", function () {
     const stack = simpleGuStackForTesting({ env: { region: "eu-west-1" } });
     new GuLoadBalancedAppExperimental(stack, {
--- a/src/experimental/patterns/gu-load-balanced-app.ts
+++ b/src/experimental/patterns/gu-load-balanced-app.ts
@@ -10,7 +10,7 @@ import {
 import type { InstanceType, ISubnet, IVpc } from "aws-cdk-lib/aws-ec2";
 import { UserData } from "aws-cdk-lib/aws-ec2";
 import { Repository } from "aws-cdk-lib/aws-ecr";
-import { ContainerInsights, OperatingSystemFamily } from "aws-cdk-lib/aws-ecs";
+import { CfnTaskDefinition, ContainerInsights, OperatingSystemFamily } from "aws-cdk-lib/aws-ecs";
 import { CpuArchitecture } from "aws-cdk-lib/aws-ecs";
 import { PropagatedTagSource } from "aws-cdk-lib/aws-ecs";
 import {
@@ -74,6 +74,50 @@ import {
   GuRollingUpdatePolicyExperimental,
   GuUserDataForRollingUpdateExperimental,
 } from "./ec2-app";
+
+export interface GuS3FilesVolumeConfiguration {
+  /**
+   * Full ARN of the S3 Files file system to mount.
+   */
+  fileSystemArn: string;
+  /**
+   * Path within the mounted file system to use as the root.
+   *
+   * @defaultValue "/"
+   */
+  rootDirectory?: string;
+  /**
+   * Optional S3 Files access point ARN.
+   */
+  accessPointArn?: string;
+}
+
+export interface GuS3FileMount {
+  /**
+   * The mount path inside the container.
+   */
+  containerPath: string;
+  /**
+   * Full ARN of the S3 Files file system to mount.
+   */
+  fileSystemArn: string;
+  /**
+   * Path within the mounted file system to use as the root.
+   *
+   * @defaultValue "/"
+   */
+  rootDirectory?: string;
+  /**
+   * Optional S3 Files access point ARN.
+   */
+  accessPointArn?: string;
+  /**
+   * Whether the mount should be read-only.
+   *
+   * @defaultValue true
+   */
+  readOnly?: boolean;
+}
 
 export interface GuLoadBalancedAppExperimentalProps extends AppIdentity {
   /**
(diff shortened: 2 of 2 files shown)


src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -640,7 +707,56 @@ export class GuLoadBalancedAppExperimental extends Construct {
         runtimePlatform: { cpuArchitecture: CpuArchitecture.ARM64, operatingSystemFamily: OperatingSystemFamily.LINUX },
       });
 
-      taskDefinition.addContainer(app, {
+      const defaultS3FilesSource = {
+        bucket: GuDistributionBucketParameter.getInstance(scope).valueAsString,
+        path: `${stack}/${stage}/${app}/`,
+      };
```
AshCorr: Is it possible to instantiate this only if necessary to avoid creating the `GuDistributionBucketParameter` construct in stacks that wont use it?

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -75,6 +78,64 @@ import {
   GuUserDataForRollingUpdateExperimental,
 } from "./ec2-app";
 
+export interface GuS3FilesVolumeConfiguration {
```
adamnfish-gu: Do we need this? This could be an internal implementation detail for the S3 config mounts, which would simplify this change and keep our surface area small. Teams are free to add other volume configurations themselves of course, but I'm not sure we need to offer that as part of this change?

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -75,6 +78,64 @@ import {
...
+  /**
+   * Path within the mounted file system to use as the root.
+   *
+   * @defaultValue "/"
+   */
+  rootDirectory?: string;
+}
+
+/**
+ * This interface defines the configuration for mounting an S3 Files file system into a container.
```
adamnfish-gu, line 81: Here again, the fact that we're using an S3 Files filesystem is relevant, but maybe not the headline point - especially if we remove the above! This will mount a specified (explicitly or by convention) S3 location as a container volume, and for app configuration.

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -75,6 +78,64 @@ import {
...
+ *
+ * The implementation comes with defaults for every value:
+ *
+ *   s3ConfigMounts: [ {} ]
```
adamnfish-gu, line 85: This seems super confusing to me! Passing a list that only includes an empty object isn't clearly "truthy".

Firstly, I'm not convinced this needs to be a list - can we have this be an optional object?

The secondly, maybe we can think about how to convey that this key will opt you into the mount behaviour if it is present, even if we're then using the default convention-based settings.

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -640,7 +707,56 @@ export class GuLoadBalancedAppExperimental extends Construct {
...
+            physicalResourceId: PhysicalResourceId.of(`s3files-linked-role-${index}`),
+          },
+          onUpdate: {
+            service: "IAM",
+            action: "getRole",
+            parameters: { RoleName: "AWSServiceRoleForAmazonS3Vectors" },
+            physicalResourceId: PhysicalResourceId.of(`s3files-linked-role-${index}`),
+          },
+          policy: AwsCustomResourcePolicy.fromSdkCalls({ resources: AwsCustomResourcePolicy.ANY_RESOURCE }),
+        });
```
AshCorr: Not entirely sure I understand what this is doing, we're creating a custom resource which goes and finds some `s3files-linked-role`, what's creating these roles?

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -75,6 +78,64 @@ import {
...
+ *
+ *   /etc/gu/app/ -> s3://dist-bucket/app/stack/stage/conf/
+ *
+ * Any files which need to present in a container in a directory that isn't owned by
+ * the container (eg /etc/nginx/nginx.d/appconfig) should be symlinked.
+ *
+ */
+export interface GuS3ConfigMount {
+  /**
+   * The mount path inside the container.
```
adamnfish-gu, line 97: I think this is missing some pointers about being careful to make sure you won't mount over the top of other files?

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -75,6 +78,64 @@ import {
...
+  source?: {
+    bucket: string;
+    path: string;
+  };
+  /**
+   * Sub-path within the file system source to mount into the container.
+   *
+   * @defaultValue `/conf/`
+   */
+  subPath?: string;
```
adamnfish-gu, line 116: We spoke about this separately, but I'm relatively unconvinced by the presence of this key because we already have a path. Offering more levers to teams sometimes looks easier as a library author because we're close to the problem and are imagining all the ways they can be pulled, but typically teams really want fewer simpler levers!

As you document above we have the $stack/$stage/$app here, so it's no huge challenge to just write the path if it isn't the default. If they don't want to do that, then it is nice to provide an incentive for them to move onto the default, too!

Before proposing this extra complecity, I'd want to see evidence that there is a lot of divergence specifically on the path suffix that this helps with. I'd also want to see that there isn't also similar divergence elsewhere in the path (e.g. a prefix, or other changes to how the Stack/Stage/App are used).

src/experimental/patterns/gu-load-balanced-app.ts
AshCorr: Ah sorry, is it trying to get the role ARN for the `AWSServiceRoleForAmazonS3Vectors` role?

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -75,6 +78,64 @@ import {
...
+ *
+ * will result in the following S3 location being mounted into the container as a file system:
+ *
+ *   /etc/gu/app/ -> s3://dist-bucket/app/stack/stage/conf/
+ *
+ * Any files which need to present in a container in a directory that isn't owned by
+ * the container (eg /etc/nginx/nginx.d/appconfig) should be symlinked.
+ *
+ */
+export interface GuS3ConfigMount {
```
akash1810, line 95: I think all the properties in this object can be required for simplicity.

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -75,6 +78,64 @@ import {
...
+  /**
+   * The mount path inside the container.
+   *
+   * @defaultValue `/etc/gu/${app}/`
+   */
+  containerPath?: string;
```
akash1810, line 101: I'd suggest hard-coding this to `/etc/${app}` as part of the paved-road. Clients can escape-hatch an alternative path. This is what we do for EC2 user-data - https://github.com/guardian/cdk/blob/16129c931b5d611892228218d1231ac3fd9a677e/src/constructs/autoscaling/user-data.ts#L45-L64:

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -640,7 +707,56 @@ export class GuLoadBalancedAppExperimental extends Construct {
...
+        bucket: GuDistributionBucketParameter.getInstance(scope).valueAsString,
+        path: `${stack}/${stage}/${app}/`,
+      };
+
+      const resolvedS3FilesMounts = s3ConfigMounts.map((mount) => ({
+        ...mount,
+        containerPath: mount.containerPath ?? `/etc/gu/${app}`,
+        subPath: mount.subPath ?? "/conf",
+        readOnly: mount.readOnly ?? true,
+      }));
```
akash1810: I think we could simplify here:

```ts
const resolvedS3FilesMounts: Required<GuS3ConfigMount> = s3ConfigMounts.map((mount) => {
  const defaults: Required<GuS3ConfigMount> = {
    containerPath: `/etc/gu/${app}`,
    subPath: '/conf',
    readOnly: true,
    source: {
        bucket: GuDistributionBucketParameter.getInstance(scope).valueAsString,
        path: `${stack}/${stage}/${app}/`,
      }
  };

  return { ...defaults, ...mount };
});
```

This also removes `const source = mount.source ?? defaultS3FilesSource;` in the `map` below too.

akash1810: [This comment](https://github.com/guardian/cdk/pull/2965/changes#r4005162077) would simplify even further, however.

src/experimental/patterns/gu-load-balanced-app.ts
akash1810: Same, I'm not sure what this does. If its explicitly required, lets add some inline comments to help future travellers?

src/experimental/patterns/gu-load-balanced-app.ts
```diff
@@ -756,6 +895,21 @@ export class GuLoadBalancedAppExperimental extends Construct {
...
+          Name: name,
+          ...(configuredAtLaunch !== undefined && { ConfiguredAtLaunch: configuredAtLaunch }),
+          S3FilesVolumeConfiguration: {
+            FileSystemArn: s3FilesVolumeConfiguration.fileSystemArn,
+            ...(s3FilesVolumeConfiguration.rootDirectory !== undefined && {
+              RootDirectory: s3FilesVolumeConfiguration.rootDirectory,
+            }),
+          },
+        })),
+      ]);
```
akash1810: I'm finding this quite difficult to read. Could we refactor a little? Some suggestions:
- Add type annotations
- Destructure explicit variables, rather than function arguments
- Add an explicit `return` statement in the `map`
- Add an inline comment to explain what we're doing and why

src/experimental/patterns/gu-load-balanced-app.test.ts
akash1810, line 1: Could we add a test for the [Grafana](https://github.com/guardian/grafana/blob/e8f4e9ec5040b009f86da3a9b218369769ce36b0/cdk/lib/grafana-stack.ts#L122-L127) use-case too, where the S3 file includes place holder values that are filled in later? Just to get an idea on the API being provided.

src/experimental/patterns/gu-load-balanced-app.ts
AshCorr: If it is for getting the service role, could we provision a role ourselves? If not, I think the Role ARN should be deterministic and we can just hardcode it:

`arn:aws:iam::${AWS::AcountId}:role/aws-service-role/s3files.amazonaws.com/AWSServiceRoleForAmazonS3Vectors`

(not sure if this is the exact ARN, best guess, couldn't actually find it in the AWS Dashboard)

src/experimental/patterns/gu-load-balanced-app.ts
aug24: Good idea. Done.

src/experimental/patterns/gu-load-balanced-app.ts
aug24: Agreed.  Removed.

PR #2969, opened 2026-09-18
include the hd parameter in google auth redirects

andrew-nowak:
## What does this change?

This hints to the signin page what email domain users should use when signing into this service. If the user is signed into any other Google accounts on their device, they will not be offered as alternative options. This will usually reduce the number of options to only their guardian.co.uk account, at which point the signin screen will be automatically bypassed and not require any interaction at all, saving a click or two.

## How to test

Deploy to a service. Attempt to log in to that service from a tab not currently signed into a Google account (eg. incognito tab). The signin screen should default to showing `@guardian.co.uk` in the email input.

ie. before (does not include the hd parameter):

image

and after (using another tool which don't use Cognito auth, and does send the hd parameter):
image

Slightly clearer interaction for users signing in for the first time (use a guardian.co.uk account and not another), and simpler & faster reauthentication flow for any users signed into more than one account.

## Checklist


andrew-nowak:
This doesn't work. This configuration sends the `hd` parameter in the authentication request to Cognito, which does not forward the parameter in the subsequent authentication request to Google. I don't believe this can be done with our current setup

